PR #1427, opened 2014-04-08
Check in schema.rb as recommended

garethrees:
Not sure why this is currently ignored?

> It's strongly recommended to check this file into your version
> control system.

### DIFF
--- a/db/.gitignore
+++ b/db/.gitignore
@@ -1,1 +1,0 @@
-schema.rb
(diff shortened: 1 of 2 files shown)


stevenday:
Not sure if this is worthy of a re-open, but here goes (actually I can't re-open anyway since the branch has gone). Since we're all using the same vagrant box to do development on now (aren't we?) I wonder if this issue of slight variation has gone away?

lizconlan:
We also support multiple Linux versions on vagrant so we could still end up with different Postgres versions (passing thought, not tested)

PR #1957, opened 2014-11-11
Add widget

crowbot:


garethrees:
I'm just going to review the final diff as it looks like there's a bunch that's changed over the commit history

app/controllers/widgets_controller.rb
```diff
@@ -0,0 +1,56 @@
+# app/controllers/widget_controller.rb:
+# Handle widgets, if enabled
+#
+# Copyright (c) 2014 UK Citizens Online Democracy. All rights reserved.
+# Email: hello@mysociety.org; WWW: http://www.mysociety.org/
+
+class WidgetsController < ApplicationController
+
+    require 'securerandom'
```
garethrees: `require` before the class definition for readability

app/controllers/widgets_controller.rb
```diff
@@ -0,0 +1,56 @@
...
+
+    before_filter :find_info_request, :check_widget_config, :check_prominence
```
garethrees: Check that the widget is enabled first so that we avoid db queries if the feature is not enabled?

garethrees: Might be worth an explicit spec for this too:

```ruby
AlaveteliConfiguration.stub!(:widget_page).and_return(false)
expect(InfoRequest.find).to_not be_called # can't remember the syntax off hand!
get :show
```

app/controllers/widgets_controller.rb
```diff
@@ -0,0 +1,56 @@
...
+    end
+
+    def new
+        long_cache
+    end
+
+    # Track interest in a request from a non-logged in user
+    def update
+        if not @user and cookies[:widget_vote]
+            wv = @info_request.widget_votes.where(:cookie => cookies[:widget_vote]).first_or_create
```
garethrees: use a clearer variable name and split over a few lines:

```ruby
widget_vote = @info_request.widget_votes.
                  where(:cookie => cookies[:widget_vote]).
                      first_or_create
```

`wv` isn't currently used anywhere

app/helpers/widget_helper.rb
```diff
@@ -0,0 +1,47 @@
+module WidgetHelper
+    def status_description(info_request)
```
garethrees: This might be useful in a class of its own?

```ruby
status = InfoRequestStatusPresenter.new(@info_request).description
```

app/models/widget_vote.rb
```diff
@@ -0,0 +1,19 @@
+# == Schema Information
+#
+# Table name: widget_votes
+#
+#  id              :integer          not null, primary key
+#  cookie          :string(255)
+#  info_request_id :integer
```
garethrees: Could add a constraint here to match the validation?

app/controllers/widgets_controller.rb
```diff
@@ -0,0 +1,56 @@
...
+        end
+
+        track_thing = TrackThing.create_track_for_request(@info_request)
+        redirect_to do_track_path(track_thing), status => :temporary_redirect
+    end
+
+    private
+
+    def find_info_request
+        @info_request = InfoRequest.find(params[:request_id])
```
garethrees: Would be nicer to use `/request/:url_name/widget/new` for the URL

app/views/widgets/show.html.erb
```diff
@@ -0,0 +1,54 @@
...
+<body>
+  <div class="alaveteli-widget">
+    <div class="alaveteli-widget__top">
+      <div class="alaveteli-widget__left">
+        <div class="alaveteli-widget__title">
+          <%= link_to @info_request.title, request_path(@info_request), :target => "_top" %>
+        </div>
+        <div class="alaveteli-widget__status <%= @info_request.calculate_status %>">
+          <p class="alaveteli-widget__status__status-label">Status</p>
+          <%= status_description(@info_request) %>
```
garethrees: We're calling `@info_request.calculate_status` twice (L11, L13).

If we used an [`InfoRequestStatusPresenter`](https://github.com/mysociety/alaveteli/pull/1957/files#r28501316) we could cache the result of this.

app/views/widgets/show.html.erb
```diff
@@ -0,0 +1,54 @@
...
+        </div>
+      </div>
+      <div class="alaveteli-widget__people-count">
+      <% count = TrackThing.count(:all, :conditions => ["info_request_id = ?", @info_request.id]) + 1 + WidgetVote.count(:all, :conditions => ["info_request_id = ?", @info_request.id]) %>
```
garethrees: Looks like something that the controller should be doing. Might also be worth caching the value in memcached?

app/views/widgets/show.html.erb
```diff
@@ -0,0 +1,54 @@
...
+         <%= n_('<div class="alaveteli-widget__count">{{count}}</div> person wants to know', '<div class="alaveteli-widget__count">{{count}}</div> people want to know', count, :count => count) %>
+      </div>
+    </div>
+
+
+    <%
+      if @user
+        existing_track = TrackThing.find_existing(@user, @track_thing)
```
garethrees: controller should be doing this

lib/configuration.rb
```diff
@@ -85,6 +85,7 @@ module AlaveteliConfiguration
             :USE_MAILCATCHER_IN_DEVELOPMENT => true,
             :UTILITY_SEARCH_PATH => ["/usr/bin", "/usr/local/bin"],
             :VARNISH_HOST => '',
+            :WIDGET_PAGE => false,
```
garethrees: As mentioned in https://github.com/mysociety/alaveteli/pull/2328#discussion_r28497880, I think `ENABLE_WIDGETS` would be a better name for this key.

app/views/widgets/show.html.erb
garethrees: Just noticed we actually assign the status in the controller https://github.com/mysociety/alaveteli/blob/add-widget/app/controllers/widgets_controller.rb#L17

spec/controllers/widgets_controller_spec.rb
```diff
@@ -0,0 +1,181 @@
...
+                    raise_error(ActiveRecord::RecordNotFound)
+            end
+
+        end
+
+        context "when the request's prominence is not 'normal'" do
+
+            it 'should return a 403' do
+                @info_request.prominence = 'hidden'
+                @info_request.save!
```
garethrees: Could shorten `L67-68` to `@info_request.update_attributes!(:prominence => 'hidden')`

spec/controllers/widgets_controller_spec.rb
```diff
@@ -0,0 +1,181 @@
...
+        end
+
+        it 'should find the info request' do
+            get :update, :request_id => @info_request.id
+            assigns[:info_request].should == @info_request
+        end
+
+        it 'should redirect to the track path for the info request' do
+            get :update, :request_id => @info_request.id
+            track_thing = TrackThing.create_track_for_request(@info_request)
```
garethrees: Should this be using the `TrackThing` created in the controller?

```ruby
get :update, :request_id => @info_request.id
expect(response).to redirect_to(do_track_path(assigns[:track_thing]))
```

app/models/widget_vote.rb
```diff
@@ -0,0 +1,19 @@
...
+#  created_at      :datetime         not null
+#  updated_at      :datetime         not null
+#
+
+class WidgetVote < ActiveRecord::Base
+  belongs_to :info_request
+  validates :info_request, :presence => true
```
garethrees: Probably need to define the messages for translation purposes

spec/models/widget_vote_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+        end
+
+        it 'validates the cookie length' do
+            widget_vote = WidgetVote.new
+            widget_vote.should_not be_valid
+            widget_vote.errors[:cookie].should == ["is the wrong length (should be 20 characters)"]
+        end
+
+        it 'is valid with a cookie and info request' do
+            widget_vote = FactoryGirl.create(:info_request).widget_votes.build(:cookie => 'x' * 20)
```
garethrees: Better to add a `:widget_vote` factory?

```ruby
require 'securerandom'
FactoryGirl.define do
  factory :widget_vote do
    info_request
    cookie { SecureRandom.hex(10) }
  end
end
```

app/assets/stylesheets/widget.scss
```diff
@@ -0,0 +1,113 @@
...
+    font-family: "Helvetica Neue", Arial, Helvetica, Helmet, Freesans, sans-serif;
+    font-weight: normal;
+    font-style: normal;
+    line-height: 1.5em;
+    position: relative;
+    cursor: default;
+    font-size: 1em;
+}
+
+a {
```
garethrees: Don't need this if its empty?

app/controllers/widgets_controller.rb
```diff
@@ -0,0 +1,56 @@
...
+    require 'securerandom'
+
+    before_filter :find_info_request, :check_widget_config, :check_prominence
+    skip_before_filter :set_x_frame_options_header, :only => [:show]
+
+    def show
+        medium_cache
+        @track_thing = TrackThing.create_track_for_request(@info_request)
+        @status = @info_request.calculate_status
+        unless @user or cookies[:widget_vote]
```
garethrees: Improve consistency between the user/cookie conditional:

```ruby
if !@user || cookies[:widget_vote]
```

app/controllers/widgets_controller.rb
```diff
@@ -0,0 +1,56 @@
...
+        render :action => 'show', :layout => false
+    end
+
+    def new
+        long_cache
+    end
+
+    # Track interest in a request from a non-logged in user
+    def update
+        if not @user and cookies[:widget_vote]
```
garethrees: Improve consistency between the user/cookie conditional:

```ruby
if !@user && cookies[:widget_vote]
```

garethrees:
Looks good; mostly just some tidying up and a few tweaks

garethrees:
Also noticed this PR commits `spec/fixtures/files/tnf_mail.email` but isn't used in any of the final diff?

app/views/widgets/show.html.erb
crowbot: Memcached seems like overengineering at this point. If it turns out to be a performance problem we can come back to it.

spec/controllers/widgets_controller_spec.rb
crowbot: I think I'll stick with the current code, given that it's more robust to changes in our approach to `attr_accessible`.

spec/controllers/widgets_controller_spec.rb
crowbot: I think this approach is mildly preferable as it tests that the track thing that gets assigned is also the track thing for the info request.

app/models/widget_vote.rb
crowbot: There's no place where these appear in the interface, so I don't think we should add to the translation burden until they do.

app/controllers/widgets_controller.rb
crowbot: That would change the logic flow. I've changed it to `unless @user || cookies[:widget_vote]`

app/views/widgets/show.html.erb
crowbot: I've used that in the other calls for the sake of simplicity.

app/controllers/widgets_controller.rb
crowbot: Yeah, but this is already in production on AskTheEU. So we'd need to redirect the id based routes. The use of the id is also a pattern that's already in use for [other routes](https://github.com/mysociety/alaveteli/blob/rails-3-develop/config/routes.rb#L51-L56) so I'd rather clear up all those routes in a separate pull request that puts in appropriate redirects.

PR #2108, opened 2015-01-20
Hotfix/0.20.0.5

garethrees:
Fixes https://github.com/mysociety/sobanukirwa-theme/issues/60

**TODO**

- [x] Fix intermittent test failure
- [x] Use `@public_body.translations.build` to assign translations for the form
- [x] ~~Use form fields associated to the form builder rather than plain form tags~~ [Doesn't look like](http://api.rubyonrails.org/v3.2.21/classes/ActionView/Helpers/FormHelper.html#method-i-fields_for) this is doable anyway
- [x] ~~Remove~~ Deprecate `Array` handling from `PublicBody#translations_attributes=`


garethrees:
```sh

Failures:
  1) AdminPublicBodyController when updating a public body updates an existing translation and adds a third translation
     Failure/Error: expect(pb.name).to eq('Renamed Department for Humpadinking')
       
       expected: "Renamed Department for Humpadinking"
            got: "El Department for Humpadinking"
       
       (compared using ==)
     # ./spec/controllers/admin_public_body_controller_spec.rb:297
     # ./spec/controllers/admin_public_body_controller_spec.rb:296
Finished in 5 minutes 24.68 seconds
1627 examples, 1 failure, 8 pending
Failed examples:
rspec ./spec/controllers/admin_public_body_controller_spec.rb:265 # AdminPublicBodyController when updating a public body updates an existing translation and adds a third translation
Randomized with seed 27553
```

garethrees:
Looks like the same issue affects editing a PublicBodyCategory too :crying_cat_face:

app/models/public_body.rb
```diff
@@ -151,12 +152,11 @@ def translated_versions
         translations
     end
 
-    def translated_versions=(translation_attrs)
+    def translations_attributes=(translation_attrs)
         def empty_translation?(attrs)
-            attrs_with_values = attrs.select{ |key, value| value != '' and key != 'locale' }
+            attrs_with_values = attrs.select{ |key, value| value != '' and key.to_s != 'locale' }
```
garethrees: I have a feeling there was an issue with this. Just making a note to check IRC history.

garethrees:
now in https://github.com/mysociety/alaveteli/tree/fix-add-and-edit-body-translations

app/models/public_body.rb
garethrees: Ah nope it was caused by doing `id.to_sym`.

PR #2114, opened 2015-01-23
Add make request button to search results

crowbot:

### DIFF
--- a/app/views/public_body/show.html.erb
+++ b/app/views/public_body/show.html.erb
@@ -39,30 +39,22 @@
         <% end %>
         </p>
 
-        <% if !@public_body.special_not_requestable_reason? %>
-            <% if @public_body.has_notes? %>
-                <p><%= @public_body.notes_as_html.html_safe %></p>
-            <% end %>
-            <% if @public_body.eir_only? %>
-                <p><%= _('You can only request information about the environment from this authority.')%></p>
-            <% end %>
-        <% else %>
-            <% if @public_body.not_requestable_reason == 'not_apply' %>
-                <p><%= _('Freedom of Information law does not apply to this authority, so you cannot make
-                a request to it.')%></p>
-            <% elsif @public_body.not_requestable_reason == 'defunct' %>
-                <p><%= _('This authority no longer exists, so you cannot make a request to it.')%></p>
-            <% end %>
-        <% end %>
-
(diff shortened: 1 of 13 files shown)


garethrees:
I rebuilt this yesterday but still failing with the same issue

app/views/public_body/show.html.erb
```diff
@@ -27,7 +27,7 @@
         <% end %>
         </p>
 
-        <% if @public_body.is_requestable? || @public_body.not_requestable_reason == 'bad_contact' %>
+        <% if !@public_body.special_not_requestable_reason? %>
```
garethrees: This section is pretty confusing. It would be good to refactor so that its clear when to show the button, and what to do if the button can't be shown. Maybe something like:

```erb
<%# app/views/public_body_show.html.erb %>
<div id="stepwise_make_request">
  <% if @public_body.requestable? %>
    <%= link_to _("Make a request to this authority"),
                new_request_to_body_path(:url_name => @public_body.url_name),
                :class => "link_button_green" %>
  <% else %>
    <% public_body_not_requestable_reasons(@public_body).each do |reason| %>
      <p class="not-requestable"><%= reason %></p>
    <% end %>
  <% end %>

  <% if @public_body.has_notes? %>
      <%= @public_body.notes_as_html.html_safe %>
  <% end %>
</div>
```

```ruby
# app/helpers/public_body_helper.rb
module PublicBodyHelper

  def public_body_not_requestable_reasons(public_body)

    reasons = []

    if public_body.eir_only?
      reasons.push _('You can only request information about the environment from this authority.')
    end

    # ideally @public_body.not_apply?
    if @public_body.not_requestable_reason == 'not_apply'
      reasons.push _('Freedom of Information law does not apply to this authority, so you cannot make a request to it.')
    end

    # ideally @public_body.defunct?
    if public_body.not_requestable_reason == 'defunct'
      reasons.push _('This authority no longer exists, so you cannot make a request to it.')
    end

  end

end
```

garethrees:
Looks great other than a few inline comments

garethrees:
> Looks like the specs are failing.

Sorry, missed a comma. Fixed in https://github.com/mysociety/alaveteli/commit/d21aaa2fcbd6b699f817494e2edfd0c55d15fd56

PR #2117, opened 2015-01-26
docs for tags and categories (new page, and links/glossary to it)

### LINKED ISSUE #2086: document the use of tags on authorities
They're useful for allowing specific searches, and matching to categories. But there are also at least four tags which are hardwired with magic.

Possibly this can be largely covered by a glossary entry, and links to it where needed. 

   * `defunct` and `not_apply` 
   * `site_administration` won't be included in CSV downloads
   * `eir_only` (UK only) instead of saying FOI in reference to the request, we say EIR

Example of some magic happening here based on tag value:
https://github.com/mysociety/alaveteli/blob/rails-3-develop/app/models/public_body.rb#L217

Noticed when documenting the CSV upload feature (in the admin guide), which talks about using tags without really explaining what they are for.

(thanks to @crowbot for clarifying some of this)

davewhiteland:
Covers the tag/categories that are mentioned but not explained in the new
http://alaveteli.org/docs/running/admin_manual/#creating-changing-and-uploading-public-authority-data
section... so this sort of span out from that. Issue creep :-( sorry -- this closes #2086 (which was not in the current sprint).

Also "Alaveteli" in the glossary was not in the correct position if the glossary is supposed to be ordered in the traditional alphabetical order, whoops.

### DIFF
--- a/_layouts/page.html
+++ b/_layouts/page.html
@@ -66,6 +66,7 @@ layout: default
           <li><a href="{{ site.baseurl }}docs/running/">Running</a>
             <ul>
               <li><a href="{{ site.baseurl }}docs/running/admin_manual/">Admin manual</a></li>
+              <li><a href="{{ site.baseurl }}docs/running/categories_and_tags/">Categories &amp; tags</a></li>
               <li><a href="{{ site.baseurl }}docs/running/redaction">Redaction</a></li>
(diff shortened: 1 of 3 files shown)


davewhiteland:
Hmm I think I didn't --no-ff the merge when I brought this in, whoops

Issue #2119, opened 2015-01-27
Change wording of label in incoming mail display in admin: "Mark as spam address"


davewhiteland:
This is just a label, don't make it look like an action. The real action required here is to open the raw email view, copy the `To:` address, and _then_ click the link to the right of _Mark as spam address_ and then paste it into the form (ugh).

So better maybe to have label:

```
Spam?
```

and expland link to 

```
"Copy To: address and paste it into the 
 <a href="/admin/spam_addresses">spam address list</a>"
```

Possibly with a link to the glossary entry if/when that's gone live ;-)

Relates to #1409 <-- would be much nicer to implement @garethrees's workflow, but until we do, clearer label+link would be kinder.

Yes this isn't something admins probably need to do very often so it's never going to be high priority, I know.

davewhiteland:
OK so I've now realised that *all* the labels are calls to action. I don't like this: the buttons are! So it would be nice to tidy this up, and change the labels (if they are really needed at all). E.g., label "Destroy message" has button "Destroy message" (and so on) -- the button alone is clear there.

PR #2120, opened 2015-01-27
Issues/2064 document spam addresses

davewhiteland:
Just describes the spam address list, and pasting emails into it.

It's a fairly horrible process though: it would be lovely if there was an orange "Mark as spam" button as in the mockup on #1409 . Currently the label is worded as an action which is confusing/cruel, so I've raised an issue for it #2119.


davewhiteland:
Hmm I think I didn't --no-ff the merge when I brought this in, whoops

PR #2140, opened 2015-02-03
[#2113] Fix admin form translations

garethrees:
Fixes #2113 
~~Requires #2127 (for the `config/test.yml` addition 5b776d1daf8f6bc42f95dbfc86c20409e03ba4a2)~~

PR #2142, opened 2015-02-04
[WIP][#2112] Validate PublicBodyCategory::Translation

### LINKED ISSUE #2112: Require a description for a PublicBodyCategory translation
Given a `PublicBody` is marked with a tag that relates to a `PublicBodyCategory`,
When you visit the "show" page of that `PublicBody` where the `PublicBodyCategory` does not have a `description` in the current locale
Then there should be no error.

```
A TypeError occurred in public_body#show:

can't convert nil into String
app/models/public_body.rb:347:in `+'
```

Need to add a validation on `PublicBodyCategory` to ensure all locales have a description if the primary locale contains a description.

Alternatively, modify `PublicBody#type_of_authority` to only try to render the description if it's available in the current locale.

https://github.com/mysociety/alaveteli/blob/rails-3-develop/app/models/public_body.rb#L347

garethrees:
Fixes #2112 

Validate the translated attributes of a PublicBodyCategory.

Validations (and counterpart specs) are defined in the same file as
PublicBodyCategory since the validations must be added after
PublicBodyCategory is loaded.

TODO: Render errors in the form. Waiting on
https://github.com/mysociety/alaveteli/pull/2140

### DIFF
--- a/app/models/public_body_category.rb
+++ b/app/models/public_body_category.rb
@@ -85,4 +85,7 @@ class PublicBodyCategory < ActiveRecord::Base
     end
 end
 
-
+PublicBodyCategory::Translation.class_eval do
+  validates_presence_of :title, :message => "Title can't be blank"
+  validates_presence_of :description, :message => "Description can't be blank"
+end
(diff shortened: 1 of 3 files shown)


garethrees:
Done in https://github.com/mysociety/alaveteli/pull/2166

PR #2152, opened 2015-02-10
Add page explaining holding pen, update links in glossary & admin_guide

### LINKED ISSUE #2063: explain how things come to end up in the "holding pen"
*Misaddressed messages*: explanation of how messages become misaddressed and end up
in the "holding pen" could be useful (e.g., one is errors made when email addresses are copied (perhaps by hand within public bodies!) but the other is due to unusual things such as a response being sent to multiple request addresses, perhaps using BCC) (re: http://alaveteli.org/docs/running/admin_manual/#removing-a-message-from-the-holding-pen )

Broken out from #1075

davewhiteland:
Closes #2063

Promotes the Holding Pen into a concept with its own page -- but (obviously) I am happy with this because the admin guide seems like the place for instructions, and I'm slowly making it more navigable section by section.

The "how Alaveteli works" at the top of this page probably deserves a diagram -- but actually such a diagram would belong in a more overiew-y place. Will think about this separately.

### DIFF
--- a/_layouts/page.html
+++ b/_layouts/page.html
@@ -66,6 +66,7 @@ layout: default
           <li><a href="{{ site.baseurl }}docs/running/">Running</a>
             <ul>
               <li><a href="{{ site.baseurl }}docs/running/admin_manual/">Admin manual</a></li>
+              <li><a href="{{ site.baseurl }}docs/running/holding_pen/">The holding pen</a></li>
               <li><a href="{{ site.baseurl }}docs/running/categories_and_tags/">Categories &amp; tags</a></li>
               <li><a href="{{ site.baseurl }}docs/running/redaction">Redaction</a></li>
               <li><a href="{{ site.baseurl }}docs/running/security/">Security &amp; Maintenance</a></li>
--- a/docs/running/holding_pen.md
+++ b/docs/running/holding_pen.md
@@ -1,0 +1,92 @@
+---
+layout: page
+title: The holding pen
+---
+
+#  The holding pen
+
+<p class="lead">
+  
+  The <em>holding pen</em> is where Alaveteli puts any incoming
+  <a href="{{ site.baseurl }}docs/glossary/#response" class="glossary__link">responses</a>
+  that can't be matched to a
+  <a href="{{ site.baseurl }}docs/glossary/#request" class="glossary__link">request</a>
+  automatically.
+</p>
+
+
+Alaveteli works by emailing requests to the correct target
+<a href="{{ site.baseurl }}docs/glossary/#authority" class="glossary__link">authority</a>.
+That email message is sent from a unique email address &mdash; that is, an
+email address that is associated with that single request (technically,
+Alaveteli hashes the request ID to generate a unique address and uses this as
+the `Reply-to:` address).
+
+So whenever an authority replies (by email) to a request that Alaveteli has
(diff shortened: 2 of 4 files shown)


_layouts/page.html
```diff
@@ -66,6 +66,7 @@
           <li><a href="{{ site.baseurl }}docs/running/">Running</a>
             <ul>
               <li><a href="{{ site.baseurl }}docs/running/admin_manual/">Admin manual</a></li>
+              <li><a href="{{ site.baseurl }}docs/running/holding_pen/">The holding pen</a></li>
```
garethrees: Should this page be included in this PR?

davewhiteland:
added the page (an oversight, obviously) and push --forced it back up for you to inspect

docs/running/holding_pen.md
```diff
@@ -0,0 +1,92 @@
...
+
+* **the authority "broke" the reply-to email**<br>
+  This can happen if the authority replies "by hand" to the incoming email &mdash;
+  for example if the person at the authority accidentally loses the first
+  letter of the email address when they copy-and-paste it. Or if they copy
+  it manually and simply get it wrong.
+  
+* **the message is sent to multiple (valid) email addresses**<br>
+  Sometimes an officer working for an authority may attempt to send email using
+  a collection of email addresses for you site. When this happens &mdash;
```
crowbot: you -> your

docs/running/holding_pen.md
```diff
@@ -0,0 +1,92 @@
...
+  an incoming email is addressed to *more than one* valid email addresses &mdash;
+  Alaveteli will direct it into the holding pen because it cannot decide
+  which, if any, is the correct one.
+
```
crowbot: Actually it [sends the response](https://github.com/mysociety/alaveteli/blob/rails-3-develop/app/mailers/request_mailer.rb#L258) to all of them

docs/running/holding_pen.md
```diff
@@ -0,0 +1,92 @@
...
+* **there's something unusual about the way it was sent**<br>
+  For example, if it was delivered here because the address is in the `Bcc:`
+  field, and is not the `To:` address.
+
+* **a partial email address may have been guessed**<br>
+  This may be because someone has guessed an email address either because they
+  have misunderstood how the addresses are formed, or due to a deliberate 
+  attempt to send spam.
+
```
crowbot: Also, requests can have their `handle_rejected_responses` field set to `holding_pen`, in which case incoming mail that [is not allowed](https://github.com/mysociety/alaveteli/blob/rails-3-develop/app/models/info_request.rb#L429) on the request goes into the holding_pen.

docs/running/holding_pen.md
```diff
@@ -0,0 +1,92 @@
...
+    Alaveteli tries to guess the right request to help you, so sometimes
+    you can just accept its suggestion. 
+    
+  * **delete the message**<br>
+    If the message is not a response, you can delete it.
+
+For instructions, see
+[removing a message from the holding pen]({{ site.baseurl }}docs/running/admin_manual/#removing-a-message-from-the-holding-pen).
+
+If the message is clearly spam, you can also add its `To:` email address to Alaveteli's
```
crowbot: only if it's not `to` a genuine request email address

PR #2154, opened 2015-02-10
Display translations on failed validation

garethrees:
ac907d7 built the translations in the controller rather than the view.

In a request that contains no params for a translation and fails
validation the translation is no longer available to the view. This
commit builds the translations in the case of a failed validation.

Fixes issue brought up in https://github.com/mysociety/alaveteli/pull/2140#issuecomment-73094018


garethrees:
moved to `2140-translations-failed-validation`

PR #2159, opened 2015-02-12
Set binary mode to false in call to pdftohtml

crowbot:


garethrees:
:+1:

garethrees:
Oh, might want to use a `hotfix` branch (though it doesn't really matter since it should get tagged).

Issue #2113, opened 2015-01-23
Clean up admin forms for PublicBodyCategories and Headings


crowbot:
These were based on the pattern used for public bodies, which we now know has [a bug](https://github.com/mysociety/sobanukirwa-theme/issues/60). This should be done before #2049.

garethrees:
Fixed by https://github.com/mysociety/alaveteli/pull/2238

PR #2166, opened 2015-02-16
Display translations on failed validation

### LINKED ISSUE #2112: Require a description for a PublicBodyCategory translation

garethrees:
~~**HOTFIX** (Moved from #2154)~~

**Merge in to `rails-3-develop`**

ac907d7 built the translations in the controller rather than the view.

In a request that contains no params for a translation and fails
validation the translation is no longer available to the view. This
commit builds the translations in the case of a failed validation.

Fixes issue brought up in https://github.com/mysociety/alaveteli/pull/2140#issuecomment-73094018

Also fixes #2112, #2113

### DIFF
--- a/app/controllers/admin_public_body_categories_controller.rb
+++ b/app/controllers/admin_public_body_categories_controller.rb
@@ -7,17 +7,39 @@ class AdminPublicBodyCategoriesController < AdminController
 
     def new
         @category = PublicBodyCategory.new
-        render :formats => [:html]
+        @category.build_all_translations
+    end
+
+    def create
+        I18n.with_locale(I18n.default_locale) do
+            @category = PublicBodyCategory.new(params[:public_body_category])
+            if @category.save
+                # FIXME: This can't handle failure (e.g. if a PublicBodyHeading
+                # doesn't exist)
+                if params[:headings]
+                    params[:headings].values.each do |heading_id|
+                        PublicBodyHeading.find(heading_id).add_category(@category)
+                    end
+                end
+                flash[:notice] = 'Category was successfully created.'
+                redirect_to admin_categories_path
+            else
+                @category.build_all_translations
+                render :action => 'new'
+            end
+        end
     end
 
     def edit
         @category = PublicBodyCategory.find(params[:id])
+        @category.build_all_translations
         @tagged_public_bodies = PublicBody.find_by_tag(@category.category_tag)
     end
 
     def update
         @category = PublicBodyCategory.find(params[:id])
         @tagged_public_bodies = PublicBody.find_by_tag(@category.category_tag)
+
         heading_ids = []
 
         I18n.with_locale(I18n.default_locale) do
@@ -43,6 +65,8 @@ class AdminPublicBodyCategoriesController < AdminController
                     end
 
                     added_headings.each do |heading_id|
+                        # FIXME: This can't handle failure (e.g. if a
+                        # PublicBodyHeading doesn't exist)
                         PublicBodyHeading.find(heading_id).add_category(@category)
                     end
                 end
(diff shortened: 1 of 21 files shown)


garethrees:
I think this works in most cases now, but if the `PublicBody` fails validation the `PublicBody::Translation` is persisted without a `public_body_id`. https://github.com/mysociety/alaveteli/pull/2142 should prevent this, so we'll have to include both in the hotfix.

app/controllers/admin_public_body_categories_controller.rb
```diff
@@ -60,7 +83,17 @@ def update
     def create
         I18n.with_locale(I18n.default_locale) do
             @category = PublicBodyCategory.new(params[:public_body_category])
+            # Build a translation for the default locale as it isn't available
```
garethrees: I think we can just call `@category.translation` here which seems to add the default locale in to the translations array:

```ruby
> category = FactoryGirl.build(:public_body_category)
=> #<PublicBodyCategory id: nil, title: "Example Public Body Category 6", category_tag: "example_tag_6", description: "Example Public body Description 6">
> category.translations
=> []
> category.translation
=> #<PublicBodyCategory::Translation id: nil, public_body_category_id: nil, locale: "en", description: nil, title: nil, created_at: nil, updated_at: nil>
> category.translations
=> [#<PublicBodyCategory::Translation id: nil, public_body_category_id: nil, locale: "en", description: nil, title: nil, created_at: nil, updated_at: nil>]
```

garethrees:
Strange behaviour of validations on the primary translation:

edit-category-validations

app/views/admin_public_body_categories/_form.html.erb
```diff
@@ -4,43 +4,21 @@
 
 <div id="div-locales">
   <ul class="locales nav nav-tabs">
-  <% I18n.available_locales.each_with_index do |locale, i| %>
-    <li><a href="#div-locale-<%=locale.to_s%>" data-toggle="tab" ><%=locale_name(locale.to_s) || "Default locale"%></a></li>
-  <% end %>
+    <% @category.ordered_translations.each do |translation| %>
+      <li>
+        <a href="#div-locale-<%= translation.locale.to_s %>" data-toggle="tab" >
+          <%= locale_name(translation.locale.to_s) || _("Default locale") %>
```
crowbot: Probably translation.locale.to_s would be a better fallback here than _("Default locale") as this is not necessarily the default locale.

app/views/admin_public_body_headings/_form.html.erb
```diff
@@ -4,37 +4,20 @@
 
 <div id="div-locales">
   <ul class="locales nav nav-tabs">
-  <% I18n.available_locales.each_with_index do |locale, i| %>
-    <li><a href="#div-locale-<%=locale.to_s%>" data-toggle="tab" ><%=locale_name(locale.to_s) || "Default locale"%></a></li>
-  <% end %>
+    <% @heading.ordered_translations.each do |translation| %>
+      <li>
+        <a href="#div-locale-<%= translation.locale.to_s %>" data-toggle="tab" >
+          <%= locale_name(translation.locale.to_s) || _("Default locale") %>
```
crowbot: Same thought as https://github.com/mysociety/alaveteli/pull/2166/files#r25874045

garethrees:
> going back to a pattern like 6b0a79a#diff-e9be0b7887688275035d3346ae523a01R18 is a cleaner fix.

Yeah, done.

Also highlighted the tab 3b0bfb5

garethrees:
> I don't think these if statements are needed - should be handled by built-in validation

Ah yep, removed.

I'll drop 3b0bfb5, merge the fixups and make a PR against rails-3-develop

garethrees:
See https://github.com/mysociety/alaveteli/pull/2166

Issue #2175, opened 2015-02-19
Create a diagram on how the process works on Alaveteli


JenMysoc:
showing the send email, publish, respond, publish

garethrees:
+1. Kinda related #2080

garethrees:
I think #2193 completed this ticket. The question now is whether to reuse the diagram on alaveteli.org. We'll have that discussion in #2193 and consider this ticket finished.

Issue #2192, opened 2015-02-24
add an "overview" diagram to the docs


davewhiteland:
...describing the general process, rather than a technical diagram.

(Design label on this for aesthetic input from design squad).

crowbot:
Duplicate of #2175

PR #2193, opened 2015-02-24
add alaveteli overview PNG (see #2192)

davewhiteland:
Fixes https://github.com/mysociety/alaveteli/issues/2175

Discuss what we need to sort out in this diagram here :-)


garethrees:
I'd reduce the contrast of everything outside of the middle channel by 50% or something.

garethrees:
What page do we want to actually put this on?

garethrees:
> How about on http://alaveteli.org/docs/getting_started/ ?

Works for me.

garethrees:
Looks good :+1:

PR #2203, opened 2015-03-02
Move translation population to model, include unsaved translations.

crowbot:


garethrees:
Cheers. cherry-picked as db9c40e

PR #2204, opened 2015-03-02
Remove obsolete instructions for not getting vendored rails.

### LINKED ISSUE #2200: Is "We don’t want to vendor Rails" step in the Mac OS X install documentation necessary?
Is [this step](http://alaveteli.org/docs/installing/macos/#clone-alaveteli) still necessary?

> #### Clone Alaveteli
>
> We don’t want to vendor Rails, as it causes problems locally.
> 
> ```git clone https://github.com/mysociety/alaveteli.git
> cd alaveteli
> git submodule init```
> 
> sed -i~ 's/\\&#91;submodule "vendor\/rails"\\&#93;//' .git/config
> 
> sed -i~ 's/url = git:\/\/github.com\/rails\/rails.git//' .git/config
> git submodule update

> Note: Due to Markdown bugs, the first sed command above does not display properly if it appears in blockquote.

If not, it could be an opportunity to cut a step.

crowbot:
Closes #2200.

### DIFF
--- a/docs/installing/macos.md
+++ b/docs/installing/macos.md
@@ -112,18 +112,11 @@ Then create the databases:
 
 ### Clone Alaveteli
 
-We don't want to vendor Rails, as it causes problems locally.
-
     git clone https://github.com/mysociety/alaveteli.git
     cd alaveteli
     git submodule init
-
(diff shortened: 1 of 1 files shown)


garethrees:
:+1:

PR #2207, opened 2015-03-02
Add a better note for new postgres users.

### LINKED ISSUE #2199: confusing PostgreSQL step in Mac installation docs
In the [Mac OS X docs, the step `createuser`](http://alaveteli.org/docs/installing/macos/#alaveteli) is confusing if you're new to PostgreSQL, like me.

I wasn't sure if the code provided would set the password or not. I had to ask someone for help because I didn't want to get it wrong because if left the password black I'd get some 'great confusion'.

> `createuser -s -P foi`
> Note: Leaving the password blank will cause great confusion if you’re new to PostgreSQL.

Maybe this could be something like:

> Note: After running this command you will be prompted to set a password for the user. Don't leave it blank if you are new to PostgreSQL, or it could be difficult to set later for you.

crowbot:
Closes #2199.

### DIFF
--- a/docs/installing/macos.md
+++ b/docs/installing/macos.md
@@ -96,8 +96,9 @@ Create a `foi` user from the command line, like this:
 
     createuser -s -P foi
 
-_Note:_ Leaving the password blank will cause great confusion if you're new to
-PostgreSQL.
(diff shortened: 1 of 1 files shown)


garethrees:
:+1:

PR #2214, opened 2015-03-03
new page about administering requests (metadata)

davewhiteland:
OK so you've seen this page before because it snuck into `gh-pages` -- but here it is as a pull request so the Alaveteli devs can fact-check it, etc. Thanks team!

### DIFF
--- a/_layouts/page.html
+++ b/_layouts/page.html
@@ -66,6 +66,7 @@ layout: default
           <li><a href="{{ site.baseurl }}docs/running/">Running</a>
             <ul>
               <li><a href="{{ site.baseurl }}docs/running/admin_manual/">Admin manual</a></li>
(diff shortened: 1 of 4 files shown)


davewhiteland:
@crowbot Yes, but I'm happy with that because I think it's an improvement to break such general things out from the one-big-admin-manual page; and I see this as really giving an overview of the kinds of things that can happen to requests, which are the core 'thing' (together with responses) in Alaveteli). We _can_ deep-link into the admin manual but ultimately I think it's better to deliberately break some things out of that big page. I'll have a look at what else from the admin manual about managing requests could be here, and linking between them -- the admin manual of course should say "you should do this" but I'm not averse to then linking to the instructions if they are on their own page.

PR #2233, opened 2015-03-12
Gh pages finish i18n

crowbot:


davewhiteland:
:+1:  I prefer your "orthodox"  alphabetic order to the one I was using

PR #2238, opened 2015-03-18
[#2140] translations failed validation r3d

garethrees:
Continued from https://github.com/mysociety/alaveteli/pull/2166


garethrees:
Squashed all the work from #2166 so just needs a quick once-over and then we should be able to merge.

PR #2239, opened 2015-03-18
Add rake task for uppdating locale files with slightly changed msgids

crowbot:
We don't really want to make all the translators retranslate strings
that are basically the same but slightly tweaked in the English version.
So this task just replaces the msgid in the app.po files for each locale.

Mappings at https://gist.github.com/crowbot/76f706150a3ee3754212

### DIFF
--- a/lib/tasks/gettext.rake
+++ b/lib/tasks/gettext.rake
@@ -35,12 +35,57 @@ namespace :gettext do
         end
    end
 
+   desc 'Update locale files with slightly changed English msgids using a csv file of old to new strings'
+   task :update_msgids_from_csv do
+       mapping_file = find_mapping_file(ENV['MAPPING_FILE'])
+       mappings = {}
+       CSV.parse(clean_csv_mapping_file(mapping_file)) do |csv_line|
+           from,to = csv_line
+           mappings[from] = to
+       end
+       Dir.glob("locale/**/app.po").each do |po_file|
+           lang_mappings = mappings.clone
+           lines = []
+           File.read(po_file).each_line do |line|
+               /^msgid "(.*)"/ =~ line
+               if $1 && mappings[$1]
+                    lines << "msgid \"#{lang_mappings.delete($1)}\""
+                 else
+                    lines << line
+                end
+           end
+           puts "Mappings unused in #{po_file}: #{lang_mappings.keys}" unless lang_mappings.empty?
+           File.open(po_file, "w") { |f| f.puts(lines) }
+       end
+   end
+
+   # Use a quote for quote-escaping as CSV errors on the \" with "Missing or stray quote"
+   def clean_csv_mapping_file(file)
+       data = ''
+       File.foreach(file) do |line|
+         data += line.gsub('\"', '""')
+       end
+       data
+   end
+
    def find_theme(theme)
        unless theme
            puts "Usage: Specify an Alaveteli-theme with THEME=[theme directory name]"
-           exit(0)
+           exit(1)
        end
        theme
+   end
+
+   def find_mapping_file(file)
+       unless file
+           puts "Usage: Specify a csv file mapping old to new strings with MAPPING_FILE=[file name]"
+           exit(1)
+       end
+       unless File.exists?(file)
+           puts "Error: MAPPING_FILE #{file} not found"
+           exit(1)
+       end
+       file
    end
 
    def theme_files_to_translate(theme)


lib/tasks/gettext.rake
```diff
@@ -43,6 +68,14 @@ namespace :gettext do
        theme
    end
 
+   def find_mapping_file(file)
+       unless file
```
garethrees: `unless File.exist?(file)` might be a better check

lib/tasks/gettext.rake
```diff
@@ -35,6 +35,31 @@ namespace :gettext do
         end
    end
 
+   desc 'Update locale files with slightly changed English msgids using a csv file of old to new strings'
+   task :update_msgids_from_csv do
+       mapping_file = find_mapping_file(ENV['MAPPING_FILE'])
+       mappings = {}
+       File.foreach(mapping_file) do |csv_line|
+           from,to = CSV.parse(csv_line.gsub('\"', '""')).first
```
garethrees: This looks like it would be easy to forget why the `gsub` is there.

I'd be inclined to extract the data cleaning in to `clean_csv_mapping_file`, and then parse the cleaned data:

```ruby
mappings = {}
CSV.parse(clean_csv_mapping_file(mapping_file)).each do |line|
  from, to = line
  mappings[from] = to
end

# Some descriptive comment about what the problem is…
def clean_csv_mapping_file(file)
  data = ''
  File.foreach(file) do |line|
    data += line.gsub('\"', '""')
  end
  data
end
```

Here's something you should be able to paste in to the console to verify the result is the same:

```ruby
require 'csv'
require 'tempfile'

MAPPING_CSV = <<-CSV
"{{reason}}, please sign in or make a new account.","{{reason}}, create an account or sign in"
"If you are thinking of using a pseudonym,\\n                please <a href=\\"{{url}}\\">read this first</a>.","<a href=\\"{{url}}\\">Thinking of using a pseudonym?</a>."
CSV

mapping_file = Tempfile.new('mappings.csv')
mapping_file.write(MAPPING_CSV)
mapping_file.close

## original

mappings = {}
File.foreach(mapping_file.path) do |csv_line|
  from,to = CSV.parse(csv_line.gsub('\"', '""')).first
  mappings[from] = to
end

mappings_original = mappings

## review

def clean_csv_mapping_file(file)
  data = ''
  File.foreach(file) do |line|
    data += line.gsub('\"', '""')
  end
  data
end

mappings = {}
CSV.parse(clean_csv_mapping_file(mapping_file.path)).each do |line|
  from, to = line
  mappings[from] = to
end

mappings == mappings_original

mapping_file.unlink
```

lib/tasks/gettext.rake
```diff
@@ -35,6 +35,31 @@ namespace :gettext do
...
+           mappings[from] = to
+       end
+       Dir.glob("locale/**/app.po").each do |po_file|
+           lang_mappings = mappings.clone
+           lines = []
```
garethrees: Could do with a more descriptive variable name here. `lines_to_replace`?

lib/tasks/gettext.rake
```diff
@@ -35,6 +35,31 @@ namespace :gettext do
...
+           File.read(po_file).each_line do |line|
+
```
garethrees: Don't need the newline here

lib/tasks/gettext.rake
garethrees: Actually I think I've misunderstood – we're writing the entire file, except swapping out the updated line on L53?

lib/tasks/gettext.rake
```diff
@@ -35,6 +35,31 @@ namespace :gettext do
...
+               /^msgid "(.*)"/ =~ line
+               if $1 && mappings[$1]
+                    lines << "msgid \"#{lang_mappings.delete($1)}\""
+                 else
+                    lines << line
+                end
+           end
+           puts "Mappings unused: #{lang_mappings.keys}" unless lang_mappings.empty?
+           File.open(po_file, "w"){ |f| f.puts(lines) }
```
garethrees:
```diff
- File.open(po_file, "w"){ |f| f.puts(lines) }
+ File.open(po_file, "w") { |f| f.puts(lines) }
```

lib/tasks/gettext.rake
```diff
@@ -35,6 +35,31 @@ namespace :gettext do
...
+           File.read(po_file).each_line do |line|
+
+               /^msgid "(.*)"/ =~ line
+               if $1 && mappings[$1]
+                    lines << "msgid \"#{lang_mappings.delete($1)}\""
+                 else
+                    lines << line
+                end
+           end
+           puts "Mappings unused: #{lang_mappings.keys}" unless lang_mappings.empty?
```
garethrees: These are mappings that were in `mappings.csv` that didn't get replaced in this file?

Should we print the filename that they didn't get used in?

```
Mappings unused in locale/cy/app.po:
  "Some mapping"
  "Another key"
```

garethrees:
A few minor suggestions, but looks great :cake:

lib/tasks/gettext.rake
crowbot: Yep

PR #2240, opened 2015-03-19
Add an autodetect feed for the frontpage list of requests

### LINKED ISSUE #434: Link the RSS/Atom feed of successful requests from homepage
Add the link as a link tag in the header (for auto-discovery and SEO); and also under the list of successful requests for humans. 

Previously (pre-Alaveteli) I believe the annotations feed was also linked from the homepage.

crowbot:
Closes #434.


garethrees:
:+1:

PR #2242, opened 2015-03-20
Set STAGING_SITE to 1 in the example config.

### LINKED ISSUE #2225: config/rails_env.rb created after vagrant up
Forces the app in to production mode even though it's creating a development environment.

Via @jacksonj04

crowbot:
The install script generates a config from this, which we want to
default to development. Closes #2225

### DIFF
--- a/config/general.yml-example
+++ b/config/general.yml-example
@@ -416,7 +416,7 @@ READ_ONLY: ''
 #   STAGING_SITE: 0
 #
 # ---
-STAGING_SITE: 0
+STAGING_SITE: 1
 
 # Recaptcha, for detecting humans. Get keys here:
 # http://recaptcha.net/whyrecaptcha.html


garethrees:
:+1: on this for now, as it fixes expected behaviour when running `vagrant up`. 9e16a9600aa5297b5ea42abdd3c574f27143fe8c also looks like it's broken the AMI expected behaviour; launching an instance and running `sudo service alaveteli start` does not bring the site up on the latest two AMI releases.

In general, I'm :-1: on this. The install script should do no more than effectively creating a ["build"](http://12factor.net/build-release-run); Putting the code in place, installing dependencies. The config should then get applied to the "build" when the app is run.

I don't think it should be doing any system-level stuff either (setting locale, adding apt sources etc) – that's the job of a CM tool.

PR #2244, opened 2015-03-23
Updated commonlib reference in .gitmodules

tuttinator:
Hi,

As a new cloner of this repo I found that git://git.mysociety.org/commonlib is unavailable when running `git submodule init && git submodule update`.

Switching it out with https://github.com/mysociety/commonlib solves this.

Tests will be run if mysociety/alaveteli#2250 is merged.

Thanks,

Caleb

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,3 @@
 [submodule "commonlib"]
     path = commonlib
-    url = git://git.mysociety.org/commonlib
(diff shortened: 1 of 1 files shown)


tuttinator:
Hi - sorry I realise now that this was just a problem with my office firewall, where the git:// protocol is blocked. Often we have to use `git config --global url."https://".insteadOf git://` as a workaround.

PR #2245, opened 2015-03-23
Update eventmachine due to install issue

tuttinator:
Hi again!

In this (also tiny) PR I updated `eventmachine` (change just in the Gemfile.lock) due to running into this error https://github.com/eventmachine/eventmachine/issues/509. I'm working on creating a [Docker container](https://github.com/nzherald/alaveteli-docker) (with a base image of Debian Jessie and using Ruby 2.2).

This also includes updating the sqlite gem, as the native extension fails to build on Jessie unless updates as well.

I also removed one additional occurrence of whitespace at the end of a line in the Gemfile.

Tests will be run if mysociety/alaveteli#2250 is merged.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -32,7 +32,7 @@ gem 'secure_headers', '~> 1.3.4'
 gem 'statistics2', '~> 0.54'
 gem 'syslog_protocol', '~> 0.9.2'
 gem 'thin', '~> 1.5.1'
-gem 'vpim', '~> 13.11.11' 
+gem 'vpim', '~> 13.11.11'
 gem 'will_paginate', '~> 3.0.5'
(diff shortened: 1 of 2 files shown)


tuttinator:
No problem. I've also noticed that Travis has a security feature in which Pull Requests from forks are unable to use secure environment variables. (Described here http://docs.travis-ci.com/user/pull-requests/  and http://docs.travis-ci.com/user/environment-variables/) - which means that tests won't run on my PRs.

(`script/wad` fails due to `ENV['S3_CREDENTIALS']` [being nil here](https://github.com/mysociety/alaveteli/blob/rails-3-develop/script/wad#L313) )

PR #2248, opened 2015-03-23
Issues/2218 filtering links

wrightmartin:
This is now ready for review


wrightmartin:
@crowbot I think alavateli-theme will need some fixes for this specifically

wrightmartin:
I've made a couple of changes and pushed them, I can't get alaveteli-theme working locally, could you check it for me @crowbot?

PR #2250, opened 2015-03-23
Add a guard in script/wad for the S3 credentials ENV vars being missing

tuttinator:
Hi,

From what I can tell `script/wad` was a tool designed for diagnosing bundler issues on Travis by tarring up an archive and sending to to S3. I wonder if this is still required (or an issue with Ruby 1.9.3 and below being deprecated). 

Also in the way of a suggestion would installing the `ca-certificates` apt package and using the `fog` gem possibly make the amount of required code smaller and remove the need to include certificates at the end of the file?

Unfortunately Travis includes a security feature which prevents secure environment variables being read for forks of repositories (http://docs.travis-ci.com/user/pull-requests/). The solution this PR proposes is to skip the `put` method call if `ENV['S3_CREDENTIALS']` are blank.

This should allow forks to be tested by Travis.

Thanks,

Caleb

### DIFF
--- a/script/wad
+++ b/script/wad
@@ -330,6 +330,8 @@ class Wad
   end
 
   def s3_configure
+    return log "ENV['S3_CREDENTIALS'] not present." unless ENV['S3_CREDENTIALS']
+
     Presss.config = {
       :region => s3_region,
       :bucket_name => s3_bucket_name,
@@ -424,7 +426,7 @@ class Wad
   end
 
   def self.setup
-    new.setup
+    ENV['S3_CREDENTIALS'] ? new.setup : new.install_bundle
   end
 end
 


tuttinator:
Hi @crowbot,

Just looking at this again, and it looks like Travis has implemented their own caching system[1] and the wad source seems to have noticed this too[2].

To implement it seems to only require adding `cache: bundler` to the `.travis.yml`.
This means that `script/wad` could be deleted. 

I'm testing this out on a branch now, as I'm trying to use Travis to run specs for different Ruby versions on my fork with Travis while I start working on refactoring.

One advantage of using Travis' cache over `script/wad` is that pull requests should be able to use the same bundler cache (unfortunately prevented due to the limitation of how Travis sets up secure environment variables). I've noticed a few of my PRs timing out during the bundle install phase.

Hope this is useful.

[1] http://docs.travis-ci.com/user/caching/
[2] https://github.com/Fingertips/WAD/commit/63a774a6612b81f326497593988b085a97302e9e

tuttinator:
Ah, I see. My bad, missed the part about private repositories only.

PR #2262, opened 2015-03-27
added new icons for different attachment types

### LINKED ISSUE #1918: Distribute more modern icons with Alaveteli
These are pretty early-2000's looking:

- https://github.com/mysociety/alaveteli/blob/rails-3-develop/app/assets/images/icon_application_pdf_large.png
- https://github.com/mysociety/alaveteli/blob/rails-3-develop/app/assets/images/icon_application_rtf_large.png
- https://github.com/mysociety/alaveteli/blob/rails-3-develop/app/assets/images/icon_application_vnd.ms-excel_large.png
- https://github.com/mysociety/alaveteli/blob/rails-3-develop/app/assets/images/icon_application_octet-stream_large.png
- etc…

wrightmartin:
Adds redesigned icons for attachments, closes #1918


garethrees:
Should have noticed this sooner sorry, but I'm having second thoughts about the `.ppt` icon. It makes me think its a video, like `.mp4` or whatever. I don't feel that strongly about it so I'm happy to merge as is, but just putting it out there if you have a really quick solution.

wrightmartin:
> I'm a bit concerned that (particularly in the context of the themes) matt grey says 'greyed out, can't click

I was worried about this, is it too much trouble to have different colours in each theme?

PR #2267, opened 2015-03-30
Use https protocol for git dependencies

garethrees:
Generally more reliable. Had one report of git:// being blocked by an
office firewall.


henare:
:+1: and also merge #2244

Issue #2210, opened 2015-03-03
Header search form has accessibility issues


equivalentideas:
The [search form in the header](https://github.com/mysociety/alaveteli/blob/rails-3-develop/app/views/general/_responsive_topnav.html.erb#L24) has no button to submit a search. This means there is no way to submit a search with the mouse, which is unpredictable behaviour. There's a bit of discussion about this here http://ux.stackexchange.com/questions/23440/why-is-it-inaccessible-to-create-a-form-with-no-submit-button with a link to the [w3c's thoughts](http://www.w3.org/TR/WCAG-TECHS/H32).

The magnifying glass icon could be the button, rather than a label. The label could be included and hidden so it’s only available for screen readers.

There's quite a nice pattern for this on [simplyaccessible.com](http://simplyaccessible.com/)

crowbot:
Sounds reasonable to me - would you consider making a pull request to make the changes you suggest?

equivalentideas:
I'd be happy to have a pass. I'll wait till after I've finished this push to make Right To Know responsive.

equivalentideas:
You can assign this to me if you like.

crowbot:
Thanks! I have done.

equivalentideas:
Here's the html for the Simply Accessible example:

```
<form class="search" action="/" method="get" role="search">
  <p>
    <label for="header-search" class="visuallyhidden">
      <span>Search</span>
    </label>
    <input type="text" class="text-input" id="header-search" name="s" placeholder="search" value="">
    <button type="submit">
      <i class="fa fa-search"></i> <span class="visuallyhidden">Submit Search</span>
    </button>
  </p>
</form>
```
here's what it looks like:
![screen shot 2015-04-01 at 11 54 53 am](https://cloud.githubusercontent.com/assets/1239550/6932598/3a520cfa-d867-11e4-8c50-ea3e0a65c9a0.png)


I think we can simplify the markup a bit more.

In addition to adding the accessible form markup, I think we should move the button over to the righthand side as this is a much more familiar pattern which is already used on the site in other places (such as the [search results page](http://demo.alaveteli.org/search/test/all)).

* [ ] add submit button
* [ ] add proper hidden label with text
* [ ] move the button to the righthand side

PR #2273, opened 2015-04-01
Accessibility improvements to header search form

equivalentideas:
There are a number of usability problems with the header search form that particularly impact citizens using assistive technology.

See issue https://github.com/mysociety/alaveteli/issues/2210

These commits improve the accessibility of the form by:

* adding a text label that can be associated with the input by assistive tech
* adds a submit button to the form to provide more ways to trigger the form with difference inputs (tap/mouse)
* adds the 'search' aria role to the form
* adds type and placeholder attributes to the text input

The submit button (using the magnifying-glass icon) has been moved from the left to righthand side. This works better for keyboard access and is a more familiar design pattern, on this site and the web generally.

Note that this changes both the form markup and it’s visual appearance.

## Before:

screen shot 2015-04-01 at 1 35 25 pm

screen shot 2015-04-01 at 3 18 33 pm

## After:

screen shot 2015-04-01 at 1 36 23 pm

screen shot 2015-04-01 at 3 17 54 pm

I thought of changing the style to incorporate the button visually into the input (like the form in the header of  [Simply Accessible](http://simplyaccessible.com/)), but I thought that might be something for theme developers to take care of. Maybe the button should look more like a button now?

Feedback very welcome :)

Closes https://github.com/mysociety/alaveteli/issues/2210

### DIFF
--- a/app/assets/stylesheets/responsive/_header_layout.scss
+++ b/app/assets/stylesheets/responsive/_header_layout.scss
@@ -131,25 +131,22 @@
   }
   form{
         @include grid-row;
-        padding-right: 1em;
+        padding: 1em 1em 0;
         @include lte-ie7 {
            display: inline;
         }
     }
     input{
         @include grid-column($columns:9);
-        margin:0;
+        margin-right:0;
         @include lte-ie7 {
           width: 10.063em;
         }
     }
-    label{
+    button[type="submit"]{
         @include prefix-postfix-base;
-        @include grid-column($columns:3,$float:left);
+        @include grid-column($columns:3,$float:right);
         border:none;
-        img{
-            max-width: 100%;
-        }
         @include lte-ie7 {
            width: 2.125em;
         }
--- a/app/assets/stylesheets/responsive/_header_layout.scss
+++ b/app/assets/stylesheets/responsive/_header_layout.scss
@@ -135,6 +135,9 @@
         @include lte-ie7 {
            display: inline;
         }
+        @include respond-min( $main_menu-mobile_menu_cutoff ){
+          padding-top: 0;
+        }
     }
     input{
         @include grid-column($columns:9);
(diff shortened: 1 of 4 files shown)


garethrees:
If we're looking at this, is it worth using a [search input type](http://api.rubyonrails.org/v3.2.21/classes/ActionView/Helpers/FormTagHelper.html#method-i-search_field_tag)?

equivalentideas:
I thought so ;-) https://github.com/equivalentideas/alaveteli/commit/644a665aabf7ca7e3a398a8e47a56653b2c1b9e6

I think anything that helps the machines understand what things are so they can help people input/access is "accessibility" :)

equivalentideas:
I see what you're saying though, nicer to do it with the ruby code than on top.

garethrees:
Oops sorry! just saw `text_field_tag 'query'` and didn't go past that. [`search_field_tag`](https://github.com/rails/rails/blob/4179470f0116a08a1838c474ba5541bb33658c9d/actionpack/lib/action_view/helpers/form_tag_helper.rb#L548-L550) is just `text_field_tag` with a default `:value` set anyway.

equivalentideas:
That’s great, thanks @crowbot I'm very happy to contribute this kind of stuff :smile: 

Thinking about the button thing a bit more, I think it probably should be by default a bit more buttony, there's no obvious affordance to kick on it currently. I think all that it would need is removing the `transparent` background colour (https://github.com/mysociety/alaveteli/pull/2273/files#diff-9e2e1825f3e68584bb451acfd5d77c51R8) and changing the colour of the icon to white. I think the background colour will default to blue for submit buttons. Or to keep things more white label you could make it a light grey and leave the icon black.

wrightmartin:
> Thinking about the button thing a bit more, I think it probably should be by default a bit more buttony, there's no obvious affordance to kick on it currently. I think all that it would need is removing the transparent background colour (https://github.com/mysociety/alaveteli/pull/2273/files#diff-9e2e1825f3e68584bb451acfd5d77c51R8) and changing the colour of the icon to white. I think the background colour will default to blue for submit buttons. Or to keep things more white label you could make it a light grey and leave the icon black.

Yep, I agree, I'd be happy with either solution, but if making it blue uses the Sass variable then it makes sense to do that, as it makes it better for customisation. Let me know if you need me to make the new icon.

PR #2283, opened 2015-04-04
Update app.po

jbaezni:
Se cambio "al contacto en"


garethrees:
I've made this change on Transifex

Issue #2268, opened 2015-03-30
Look at vertical spacing of definition lists and paragraphs


wrightmartin:
https://github.com/mysociety/alaveteli/commit/4a58b416616f15adb522e54dce87990de0455352

garethrees:
Fixed by https://github.com/mysociety/alaveteli/issues/2293

PR #2293, opened 2015-04-07
remove para declaration

wrightmartin:
Not needed, causing spacing problems on themes, so they just revert it anyway


garethrees:
:+1:

garethrees:
Fixes #2268

PR #2305, opened 2015-04-09
Allow configuration for SMTP settings fixes #2130

### LINKED ISSUE #2130: Incorporate cuttle.io / mailgun as outgoing SMTP
- Since now https://cuttlefish.io/ is an official Poplus component; would be great to incorporate it in so that it is not so painful setting up the email portion of Alaveteli 
- Be able to use alternatives like Mailgun as well (or other alternatives) to be plugged in

tuttinator:
Hi,

The production environment configuration had [hardcoded](https://github.com/mysociety/alaveteli/blob/3830325be71e5f1a823ef8930134f1480e390050/config/environments/production.rb#L20) `:sendmail` as the delivery method.

This Pull Request changes `:sendmail` to be the default, but also allows configuration options in `general.yml` to override this.

This should also fix #2130

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -17,7 +17,20 @@ Alaveteli::Application.configure do
 
   # Disable delivery errors, bad email addresses will be ignored
   # config.action_mailer.raise_delivery_errors = false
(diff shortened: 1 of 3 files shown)


tuttinator:
#2325 looks good to me!

PR #2311, opened 2015-04-11
Fix admin public body edit test

henare:
On my machine the integration test "Editing a Public Body can add a
translation for multiple locales" (http://git.io/vvv9t) was failing
because when it submitted the form to update one locale it created
empty records for all locales resulting in the test
(http://git.io/vvv93) for a nil locale to fail.

It was doing this because when it submitted the form it included a
notes field with a single newline character and so this method
evaluated as false and therefore a largely blank translation was
created.

I've changed it to check for blank values now which I think is more what we want anyway... and my tests pass :)

### DIFF
--- a/app/models/public_body.rb
+++ b/app/models/public_body.rb
@@ -775,7 +775,7 @@ class PublicBody < ActiveRecord::Base
 
     def empty_translation_in_params?(attributes)
         attrs_with_values = attributes.select do |key, value|
-            value != '' and key.to_s != 'locale'
+            !value.blank? and key.to_s != 'locale'
         end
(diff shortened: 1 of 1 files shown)


henare:
It's worth also noting that this test was failing on my Linux machine and also someone else's OS X machine. I've got no idea why the test was/is working on mySociety people's machines and Travis!

henare:
Great to have such a quick merge, thanks @crowbot :)

henare:
Yes that sounds right. It was one of the many horrible paths I'd gone down in trying to debug the problem! :-S

PR #2328, opened 2015-04-14
Add some explanation of the new WIDGET_PAGE config var.

crowbot:


docs/customising/config.md
```diff
@@ -994,6 +995,22 @@ SHARED_DIRECTORIES:
   </dd>
 
   <dt>
+    <a name="widget_page"><code>WIDGET_PAGE</code></a>
```
garethrees: To me this config key looks like it would accept a URL path. `ENABLE_WIDGETS` might be a bit clearer.

docs/customising/config.md
```diff
@@ -994,6 +995,22 @@ SHARED_DIRECTORIES:
...
+  </dt>
+  <dd>
+     If you would like to give users the opportunity to insert HTML 'widgets' into their other websites, advertising the
+     requests they've made in Alaveteli, set this to true.
```
garethrees: Should add a link to some documentation on adding a widget. Just reading this doesn't give me a sense of what a widget looks like and whether its something I'd want enabled.

crowbot: I've made this a bit more detailed so people can go and preview it on their own dev site. I'm not sure it actually merits more documentation than that at the moment. Perhaps we can ticket it for a later revisit.

PR #2332, opened 2015-04-15
Stop categorisation game causing mixed content warning

henare:
I started by fixing the mixed content warning then I noticed the HTML wasn't valid and that the alt text could be improved.


garethrees:
:+1: cheers!

henare:
:bouquet:

PR #2335, opened 2015-04-16
Stop AlaveteliTextMasker logger call throwing an exception

henare:
Stops exception `undefined local variable or method 'logger' for AlaveteliTextMasker:Module` being thrown by calling logger method explicitly because it's not in scope here.

Downstream issue https://github.com/openaustralia/righttoknow/issues/508.


henare:
Nice one, thanks @crowbot!

PR #2352, opened 2015-04-20
Fix Xapian macron highlighting

tuttinator:
It seems that `term.is_a?(String) ? term.force_encoding('UTF-8') : term` doesn't work with Regexp instances, so this PR fixes string encoding at the point of interpolation.

Also included is a test.


tuttinator:
Looks like it might need both for another code path. Will fix.

PR #2377, opened 2015-04-27
Email subject HTML entities

henare:
Previously requests with titles including symbols such as quote marks would generate email subject lines that had these HTML encoded. For example a request titled `Here's a request` would generate an email subject line such as, `New response to your FOI request - Here&#x27;s a request`. This pull request fixes that bug.

Factories made testing this soo much easier so thanks for that :) The request mailer spec could do with a tidy up but I decided against doing it as part of this because I didn't want to have a big crazy diff that was hard to understand.


garethrees:
:+1: cheers!

PR #2384, opened 2015-04-27
Reduce queries in InfoRequest#initial_request_text

garethrees:
Delegating to OutgoingMessage#get_text_for_indexing results in an
additional query to find the parent InfoRequest.

Also eliminates a COUNT on OutgoingMessage as we can just return nil if
nothing is returned from `outgoing_messages.first`.

### DIFF
--- a/app/models/info_request.rb
+++ b/app/models/info_request.rb
@@ -803,11 +803,20 @@ public
 
     # Text from the the initial request, for use in summary display
     def initial_request_text
-        if outgoing_messages.empty? # mainly for use with incomplete fixtures
-            return ""
-        end
-        excerpt = self.outgoing_messages[0].get_text_for_indexing
-        return excerpt
+        message = outgoing_messages.first
+        return nil if message.nil?
+
+        text = message.raw_body.strip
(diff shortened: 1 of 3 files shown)


garethrees:
> Did it appear to produce a significant saving in time?

A fair amount yeah:

```
|              | Original (0a29541) | New (9f752d6) |
|--------------|--------------------|---------------|
| Full Request | 6040ms             | 4210ms        |
| Templates    | 5000ms             | 2945ms        |
| AR Objects   | 28                 | 19-22         |
| Queries      | 26                 | 16-19         |
```

The original way is a bit all over the place. You find the `InfoRequest` in the controller, and then call `initial_request_text` on it. This then finds the `OutgoingMessage` and calls `get_text_for_indexing`. `get_text_for_indexing` then ends up calling the original `InfoRequest` and then many of its associations trying to find `CensorRule`s. The worst offender for this is `OutgoingMessage#body`. This commits reduces a few queries and lots of object initialization.

garethrees:
Closing in favour of #2394

Issue #2085, opened 2015-01-08
Move EIR UK specific logic into whatdotheyknow-theme


crowbot:
Similar to https://github.com/mysociety/alaveteli/issues/625

crowbot:
EIR is actually Europe wide, so we need to think about what's best to do here.

HelenWDTK:
This issue has been automatically closed due to a lack of discussion or resolution for over 12 months.
Should we decide to revisit this issue in the future, it can be reopened.

PR #2386, opened 2015-04-27
Don't load PublicBody when initalizing InfoRequest

garethrees:
Doesn't need to try to set the default law_used once it's persisted,
so we save a query each time we find an InfoRequest


garethrees:
I'm leaning towards making #2085 a priority so that we can do this nicely.

PR #2390, opened 2015-04-28
Use .example files for daemons.

crowbot:
Consistent with other example config files, and helps us solve a mySociety config
issue.


garethrees:
:+1:

PR #2394, opened 2015-04-29
Alternative approach for speeding up InfoRequest#initial_request_text

garethrees:
Pass the censor rules to outgoing message rather than duplicating the logic like in https://github.com/mysociety/alaveteli/pull/2384

Has a few failing specs that should be relatively easy to fix if we're happy with the general idea.

Could also improve further with https://github.com/mysociety/alaveteli/commit/e7bf72a5489e18cb2140f74ee6a2c6cc3bc61c27

### DIFF
--- a/app/models/info_request.rb
+++ b/app/models/info_request.rb
@@ -804,7 +804,9 @@ public
 
     # Text from the the initial request, for use in summary display
     def initial_request_text
-        outgoing_messages.first.try(:get_text_for_indexing) or ''
+        return '' if outgoing_messages.empty?
+        body_opts = { :censor_rules => applicable_censor_rules }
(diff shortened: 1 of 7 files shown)


app/models/info_request.rb
```diff
@@ -804,8 +804,8 @@ def get_last_public_outgoing_event
 
     # Text from the the initial request, for use in summary display
     def initial_request_text
-        return '' if outgoing_messages.empty? # mainly for use with incomplete fixtures
-        outgoing_messages.first.get_text_for_indexing
+        body_opts = { :censor_rules => applicable_censor_rules }
```
garethrees: Better to revert to using `return '' if outgoing_messages.empty?` so that the censor rules are only loaded if its going to use them? Not sure when `applicable_censor_rules` actually gets called.

PR #2403, opened 2015-04-30
Add the site user to the 'adm' group.

### LINKED ISSUE #2402: Allow load-mail-server-logs to be run as app user, not root
This is currently a blocker for https://github.com/mysociety/alaveteli/issues/1325 as we're using the app user to control the ruby version. @mlandauer pointed out a long time ago that it would be preferable to have all cron jobs run as the app user anyway.

crowbot:
This is so that they can access mail log files. `adm` group members
have access to read log files in /var/log.

Closes #2402.

### DIFF
--- a/script/site-specific-install.sh
+++ b/script/site-specific-install.sh
@@ -180,6 +180,9 @@ postfix reload
 
 install_website_packages
 
+# Give the unix user membership of the adm group so that they can read the mail log files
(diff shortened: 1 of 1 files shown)


garethrees:
:+1:

garethrees:
Probably need to add this to the manual install docs

PR #2404, opened 2015-04-30
Document adding alaveteli user to adm group.

crowbot:


docs/installing/manual_install.md
```diff
@@ -225,9 +225,12 @@ Refresh the sources after adding the extra repositories:
 
 ### Create Alaveteli User
 
-Create a new linux user to run the Alaveteli application.
+Create a new linux user to run the Alaveteli application. Add them to the 'adm' group
+so that they can read the mail log files.
 
     adduser --quiet --disabled-password --gecos "Alaveteli" alaveteli
+    adduser alaveteli adm
```
garethrees: In #2403 `usermod -a -G adm "$UNIX_USER"` is used. Would be better to be consistent as much as possible?

garethrees:
:+1:

PR #2408, opened 2015-05-05
Don't initialize PublicBody when initializing InfoRequest

garethrees:
Handles issue of moving a request to a public body with a different law.

### DIFF
--- a/spec/models/info_request_spec.rb
+++ b/spec/models/info_request_spec.rb
@@ -28,6 +28,26 @@ require File.expand_path(File.dirname(__FILE__) + '/../spec_helper')
 
 describe InfoRequest do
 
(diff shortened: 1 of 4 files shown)


spec/models/info_request_spec.rb
```diff
@@ -42,7 +188,7 @@
             info_request.errors[:title].should be_empty
         end
 
-        it 'should not accept a summary with no ascii or unicode characters' do
+         it 'should not accept a summary with no ascii or unicode characters' do
```
garethrees: typo

garethrees:
Looks like this has had some impact:

screen shot 2015-07-07 at 13 26 23

PR #2410, opened 2015-05-05
Make the namespace for memcached dependent on ruby version.

crowbot:
Items cached in 1.8 can't safely be retrieved in 1.9


garethrees:
:+1:

PR #2412, opened 2015-05-05
Remove monkeypatch required when running Rails 2.3 under Ruby 1.9

crowbot:
This was added because TMail calls String#is_binary_data? which
does not exist in Ruby 1.9.2 onwards, so is no longer needed.


garethrees:
:+1: but amend commit message

PR #2413, opened 2015-05-07
Add rack-insight

garethrees:


garethrees:
Closing as we decided it's not ready enough to add a new dependency

PR #2423, opened 2015-05-11
Send mail import errors to exception notification address.

crowbot:
Usually these are going to require some technical action. There's
not much regular admins can do about them.


hsenag:
+500 :-)

garethrees:
:+1:

PR #2425, opened 2015-05-12
[2422] incoming message spam

### LINKED ISSUE #2422: Handle incoming messages with no 'to' address or no 'from' address
Currently spam messages with either no `from` or no `to` address are causing error messages from the incoming mail script.

garethrees:
Fixes #2422 

**To be hotfixed**


garethrees:
Merged in https://github.com/mysociety/alaveteli/commit/aa67dc94f8c5631ae5da889fa2d976196ee63f91

PR #2426, opened 2015-05-12
Use mb_chars to prevent slicing multibyte chars under ruby 1.8

crowbot:
mb_chars provides a multibyte-aware wrapper for strings. It should
have no effect on ruby 1.9.3 and above. Although ruby 1.8.7 wouldn't
raise errors on a badly sliced multibyte string, on upgrading to ruby
1.9.3 and above, string operations such as gsub, match and join may
produce ArgumentErrors with the message "invalid byte sequence in UTF-8".
Additionally, a database with 'UTF-8' encoding may produce the error
"PG::CharacterNotInRepertoire: ERROR:  invalid byte sequence for encoding "UTF8""

### DIFF
--- a/app/models/incoming_message.rb
+++ b/app/models/incoming_message.rb
@@ -654,7 +654,7 @@ class IncomingMessage < ActiveRecord::Base
 
         # Save clipped version for snippets
         if self.cached_attachment_text_clipped.nil?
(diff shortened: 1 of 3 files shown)


garethrees:
Looks good :+1:

PR #2427, opened 2015-05-12
Add deprecation notice for ruby 1.8.7

crowbot:


config/initializers/alaveteli.rb
```diff
@@ -63,3 +63,7 @@
   require 'no_constraint_disabling'
 end
 
+unless RUBY_VERSION.to_f >= 1.9
+	ActiveSupport::Deprecation.warn "[DEPRECATION] You are using Ruby 1.8. This will not be supported " \
+									"as of Alaveteli release 0.23. Please upgrade your ruby version."
```
garethrees: I think we should have a blog / mailing list post to link to for a change of this size.

PR #2434, opened 2015-05-15
Add source file encoding for all ruby files.

crowbot:
This is important under ruby 1.9 in order to determine the
encoding that will be used for new strings created in the code in
the file.


garethrees:
:+1:

PR #2439, opened 2015-05-20
Exclude "lib/strip_attributes/test/*" from CodeClimate

jpmckinney:


garethrees:
Cheers! I had a look at moving the tests to the `spec` directory but it was a bit of a faff and couldn't get it working after 5 minutes. It would be best to just move to a gem release of the plugin, but we'll look at that if/when we work around this area.

PR #2440, opened 2015-05-20
Move common public body translation behavior into an ActiveSupport:Concern

jpmckinney:


jpmckinney:
The most recent Travis failure is bogus (timeout).

Gemfile
```diff
@@ -55,7 +55,7 @@ group :assets do
   gem 'compass-rails', '2.0.0'
   gem 'coffee-rails', '~> 3.2.1'
   gem 'uglifier', '>= 1.0.3'
-  gem 'therubyracer', '~> 0.12.0'
```
crowbot: Don't think you meant to include this

PR #2442, opened 2015-05-20
Use built in validation on AboutMeValidator model

henare:


henare:
This doesn't support Ruby 1.8.7 so closing in favour of #2451.

PR #2443, opened 2015-05-20
Install better_errors in development for awesome debugging

henare:
better_errors is amazing, especially with the live REPL :tada:

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -82,6 +82,8 @@ group :development do
   gem 'pry', '~> 0.9.6'
   gem 'quiet_assets', '~> 1.0.2'
   gem 'rdoc', '~> 3.12.2'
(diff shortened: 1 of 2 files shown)


garethrees:
```
Gem::InstallError: better_errors requires Ruby version >= 2.0.0.
```

Is there a version available that supports 1.9.3?

If not we could only attempt to install on 2.0+:

```ruby
if RUBY_VERSION >= '2.0.0'
  gem 'better_errors'
  gem 'binding_of_caller'
end
```

henare:
>Is there a version available that supports 1.9.3?

Ruby 1.9.3 is EOL so I guess that's why it's not supported :(

Rather than just hack around it I'll just close this since it's a nice to have.

PR #2450, opened 2015-05-21
Add guard-rspec and update RSpec

tuttinator:
Hi,

This PR adds guard-rspec, which also required a newer version of RSpec (2.99.0).

Guard is a personal preference for automating running tests by watching changes to the filesystem.

RSpec 2.99 also includes deprecation notices which may be helpful for refactoring tests, however if this causes problems I can investigate the best fit compatible versions.


garethrees:
Cheers! As mentioned in https://github.com/mysociety/alaveteli/pull/2450#commitcomment-11387441, I get a load of deprecation warnings after the spec run that I think we need to fix before merging the updated rspec gem. @crowbot not sure what the outcome of #2469 was?

garethrees:
@crowbot ah gotcha

PR #2451, opened 2015-05-21
About me built in validation

crowbot:


henare:
>Support old hash syntax

Thanks @crowbot, @jpmckinney just told me about this at breakfast! :)

PR #2453, opened 2015-05-21
[WIP] User Controller refactor

tuttinator:
Hi all,

This is a very immature pull request for the purpose if starting discussion. I've stripped out quite a number of stray commits and rebased.

The goals I had for the end state of this pull request:

- ~~Use ActiveRecord instead of Xapian for listing a user's InfoRequests and comments (as per the `TODO` comment)~~
- First extract pieces to smaller private methods, and then into a service object (`UserService` may not be the best name for what it actually ends up doing)
- Aim for controller methods of 10 lines or less
- Use a Presenter to clean up view logic for the user views
- Minimise the number of required instance variables
- Remove database-specific pieces like `ilike` and generally try to use more conventional ActiveRecord interfaces

If this is not too painful, I'd like to try to apply this kinds of refactoring to other controllers, and then work on implementing `delayed_job` for some in-request operations.

Any comments are very welcome.

### DIFF
--- a/app/controllers/user_controller.rb
+++ b/app/controllers/user_controller.rb
@@ -13,22 +13,20 @@ class UserController < ApplicationController
     # Show page about a user
     def show
         long_cache
+        set_view_instance_variables
 
         unless MySociety::Format.simplify_url_part(params[:url_name], 'user') == params[:url_name]
             redirect_to :url_name =>  MySociety::Format.simplify_url_part(params[:url_name], 'user'), :status => :moved_permanently
             return
         end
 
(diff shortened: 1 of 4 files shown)


tuttinator:
Ah, I see - it's full text search. You're right, keeping Xapian does makes sense. 

Perhaps extracting another class as a `XapianQueryService` could be one approach, and possibly reusable in other places

tuttinator:
re: hash syntax - my bad, force of habit. Will change to 1.8 syntax.

PR #2459, opened 2015-05-21
Extract a method to add conditions from extra_params in InfoRequest

jpmckinney:
Seem to have forgot this one.


jpmckinney:
Bogus failures.

PR #2460, opened 2015-05-21
Use *_was methods from Rails 3

jpmckinney:


jpmckinney:
Real failures - need to revisit.

jpmckinney:
Superceded/corrected by #2716

PR #2461, opened 2015-05-21
Avoid unnecessary ternary operations

jpmckinney:
Some of these ternary operations are entirely redundant (the value is already a boolean), others are unnecessary (truthy/falsy values are fine), and others are more idiomatic as `!!condition`.

### DIFF
--- a/app/controllers/admin_comment_controller.rb
+++ b/app/controllers/admin_comment_controller.rb
@@ -16,7 +16,7 @@ class AdminCommentController < AdminController
 
         old_body = @comment.body
         old_visible = @comment.visible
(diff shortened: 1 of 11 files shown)


jpmckinney:
Bogus failures.

garethrees:
Looks great apart from the uses of `!!` and the [one](https://github.com/mysociety/alaveteli/pull/2461#commitcomment-11388078) instance of wanting to specifically assign a `Boolean`.

jpmckinney:
I've eliminated the `!!`, and I think the one you refer to doesn't need to be boolean.

garethrees:
:beers: Thanks!

PR #2462, opened 2015-05-21
Simplify HealthCheckable API

jpmckinney:
I broke this out of #2461 since it involves a small module API change.


jpmckinney:
Bogus failures.

PR #2464, opened 2015-05-21
FoiAttachment: Use simpler Array#include? instead of Hash#[]

jpmckinney:
`Hash#[]` is faster, but harder to read. If this method can be proven to be a performance bottleneck (it would have to be called thousands/millions of times per request to make a difference), `Set#member?` can be used (both fast and easy to read).


jpmckinney:
Bogus failures.

henare:
Much nicer :+1:

PR #2494, opened 2015-05-27
2493 utf8 signup

gorm:
We are having trouble registering people with utf-8 in their names on test site. I guess all params might be passed as utf-8 but force_encoding seems to fix it. Please see if this could be done in any better way and test if you can reproduce the error, by testing with name "mimes brønn" on a test site.

### DIFF
--- a/app/views/user_mailer/already_registered.text.erb
+++ b/app/views/user_mailer/already_registered.text.erb
@@ -1,4 +1,4 @@
-<%= raw @name %>,
+<%= raw @name.force_encoding('utf-8') %>,
 
 <%= _('You just tried to sign up to {{site_name}}, when you
 already have an account. Your name and password have been
--- a/app/views/user_mailer/changeemail_confirm.text.erb
+++ b/app/views/user_mailer/changeemail_confirm.text.erb
@@ -1,4 +1,4 @@
-<%= raw @name %>,
+<%= raw @name.force_encoding('utf-8') %>,
 
 <%= _('Please click on the link below to confirm that you want to 
 change the email address that you use for {{site_name}}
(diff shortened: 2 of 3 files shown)


petterreinholdtsen:
The error was first reported in https://lister.ping.uio.no/pipermail/i18n-no/2015-May/008296.html , where you can see the complete error message.

gorm:
I can reproduce in the default vagrant build. The error message is: 

Encoding::CompatibilityError (incompatible character encodings: ASCII-8BIT and UTF-8):
  app/views/user_mailer/confirm_login.text.erb:3:in `_app_views_user_mailer_confirm_login_text_erb__2630094158524030762_50905920'
  app/mailers/user_mailer.rb:15:in `confirm_login'
  app/controllers/user_controller.rb:659:in `send_confirmation_mail'
  app/controllers/user_controller.rb:229:in `signup'
  app/controllers/application_controller.rb:111:in `record_memory'
  lib/whatdotheyknow/strip_empty_sessions.rb:15:in `call'

To reproduce:

My settings:
AVAILABLE_LOCALES: 'nb nn en'
DEFAULT_LOCALE: 'nb'

1. Goto Signup
2. In Your name paste: "Mimes brønn"
3. email and pw can be set to your liking
4. Press Sign up

I get the error message above.

gorm:
Btw, it would be interesting to find out the cause of this? is it because of file format of the templates (doesn't seem to be utf-8) or does it require some # coding: utf-8 magic? The utf-8 support in rails just some weird to me.

gorm:
@crowbot thats great! I tested your fix now and it seems to work. Thanks! Closing this one then.

PR #2502, opened 2015-05-29
Include Hong Kong's accessinfo.hk in deployments page

slygent:
No image yet


deployments.md
```diff
@@ -202,6 +202,14 @@ in any language, and within any legislation</p>
                         <p class="deployment__link"><a href="https://chiedi.dirittodisapere.it/">https://chiedi.dirittodisapere.it/</a></p>
                     </div>
                 </div>
+                --><div class="deployments__unit--minor">
```
garethrees: Looks like this `-->` may be a stray closing comment.

There should either be an opening comment (`<!--`) at the end of the previous line, or none at all.

slygent:
If you can make a screenshot that would be very helpful!

Issue #2288, opened 2015-04-07
Improve performance on public body pages on WDTK


crowbot:
These now seem to be performing poorly, routinely well over 1s to return.

garethrees:
Current status: Deploying #2505 to see what effect that has and then reassessing.

garethrees:
The biggest problem here was rendering the search results. We should make more use of the xapian document data to store the data needed for displaying the result to the user.

In this example the data has been serialized in to the xapian document for quick retrieval. We'd then load the full record when the user clicks the result they are interested in.

```sh
$ delve -r 1 -d db       # output has been reformatted
Data for record #1:
{
 "MEASUREMENTS": "",
 "DESCRIPTION": "Ansonia Sunwatch (pocket compas dial)",
 "PLACE_MADE": "New York county, New York state, United States",
 "id_NUMBER": "1974-100",
 "WHOLE_PART": "WHOLE",
 "TITLE": "Ansonia Sunwatch (pocket compas dial)",
 "DATE_MADE": "1922-1939",
 "COLLECTION": "SCM - Time Measurement",
 "ITEM_NAME": "Pocket horizontal sundial",
 "MATERIALS": "",
 "MAKER": "Ansonia Clock Co."
}
Term List for record #1: Q1974-100 Sansonia Scompas Sdial Spocket
Ssunwatch XDansonia XDcompass XDdial XDpocket XDsunwatch ZSansonia
ZScompas ZSdial ZSpocket ZSsunwatch ZXDansonia ZXDcompas ZXDdial
ZXDpocket ZXDsunwatch Zansonia Zcompass Zdial Zpocket Zsunwatch
ansonia compass dial pocket sunwatch
```

See http://getting-started-with-xapian.readthedocs.org/en/latest/practical_example/indexing/verifying_the_index.html for more.

HelenWDTK:
This issue has been automatically closed due to a lack of discussion or resolution for over 12 months.
Should we decide to revisit this issue in the future, it can be reopened.

PR #2505, opened 2015-05-29
[#2288] public body page performance

garethrees:
#2288 

Could only get minor improvements so far:

```
Old: Completed 200 OK in 2966.7ms (Views: 2637.3ms | ActiveRecord: 88.2ms)
New: Completed 200 OK in 2675.2ms (Views: 2514.1ms | ActiveRecord: 94.8ms)
```

The bulk of the response time happens when rendering each search result:

```diff
-                <%= render :partial => 'request/request_listing_via_event', :locals => { :event => result[:model] } %>
+                <%# render :partial => 'request/request_listing_via_event', :locals => { :event => result[:model] } %>
```

```
Completed 200 OK in 711.1ms (Views: 511.3ms | ActiveRecord: 8.3ms)
```

I'd like to release at this point and try to get some more detailed data from New Relic to see where to go next. I think the main issue is that we're [loading the whole result set](https://github.com/mysociety/alaveteli/blob/develop/app/controllers/application_controller.rb#L394-L395) after pulling them out from Xapian. We're then calling out to many association objects during the view rendering. Looks like we can pass an [`:eager_load` option](https://github.com/mysociety/alaveteli/blob/develop/lib/acts_as_xapian/acts_as_xapian.rb#L424) to the Xapian search, but we don't currently support passing options to the [search wrapper method](https://github.com/mysociety/alaveteli/blob/develop/app/controllers/application_controller.rb#L376).

### DIFF
--- a/app/controllers/public_body_controller.rb
+++ b/app/controllers/public_body_controller.rb
@@ -66,7 +66,7 @@ class PublicBodyController < ApplicationController
             begin
                 @xapian_requests = perform_search([InfoRequestEvent], query, sortby, 'request_collapse', requests_per_page)
                 if (@page > 1)
-                    @page_desc = " (page " + @page.to_s + ")"
+                    @page_desc = " (page #{ @page })"
                 else
                     @page_desc = ""
                 end
--- a/app/controllers/public_body_controller.rb
+++ b/app/controllers/public_body_controller.rb
@@ -82,6 +82,8 @@ class PublicBodyController < ApplicationController
                 @existing_track = TrackThing.find_existing(@user, @track_thing)
             end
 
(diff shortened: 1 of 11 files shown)


garethrees:
Merged in https://github.com/mysociety/alaveteli/commit/26dbf38319f22de55d6c8bfbf3376df0dbb297c2

garethrees:
Zoom! :racehorse: 

screen shot 2015-07-07 at 13 02 44

garethrees:
Also looks to have helped here:

screen shot 2015-07-07 at 13 30 52

Looks like much less time in `request/_request_listing_via_event.html.erb`.

PR #2522, opened 2015-06-04
Tidy public body

garethrees:


garethrees:
One spec failure but not sure what we can do about it

garethrees:
> Making these methods private seems like a nice to have at this point rather than a requirement, so my suggestion would be to drop that part of this PR.

Yeah. Its more of a concern that this particular spec will stop us deprecating other methods in future too.

I'll drop these for now but I think we should figure this out at some point soon.

PR #2533, opened 2015-06-05
Hotfix/0.21.0.38

crowbot:


gorm:
Nice patch :) It seems to work fine. I have tested it on master and did not notice any issues with an example import and with our full import, after a little testing. Thanks!

garethrees:
Looks good but does a bit too much for a hotfix.

I'd suggest dropping https://github.com/mysociety/alaveteli/commit/71d9980bfd3bc06fe1243131697d201819651a37 (covered elsewhere) and cherry-picking https://github.com/mysociety/alaveteli/commit/5801ae2440f749b2763b30ba88ba19196342a8d4 for a PR in to develop for further discussion.

PR #2534, opened 2015-06-08
Use rack-utf8_sanitizer to handle badly-formed UTF-8 in request URI a…

crowbot:
…nd headers.


spec/integration/parameter_stripping_spec.rb
```diff
@@ -0,0 +1,20 @@
+# -*- encoding : utf-8 -*-
+require File.expand_path(File.dirname(__FILE__) + '/../spec_helper')
+
+describe "When handling bad requests" do
+
+	it 'should return a 404 for GET requests to a malformed request URL' do
```
garethrees: Looks like this is using tabs rather than spaces

garethrees:
:+1:

PR #2548, opened 2015-06-11
Updating derived atts for translations 

crowbot:
Fixes the issue reported in https://github.com/mysociety/alaveteli/pull/2521

### DIFF
--- a/app/models/concerns/public_body_derived_fields.rb
+++ b/app/models/concerns/public_body_derived_fields.rb
@@ -1,0 +1,47 @@
+module PublicBodyDerivedFields
+
+    extend ActiveSupport::Concern
+
+    included do
+        before_save :set_first_letter
+
+        # When name or short name is changed, also change the url name
+        def short_name=(short_name)
+            super
+            update_url_name
+        end
+
+        def name=(name)
+            super
+            update_url_name
+        end
+
+    end
+
+    # Return the short name if present, or else long name
+    def short_or_long_name
+        if self.short_name.nil? || self.short_name.empty?
+            self.name.nil? ? "" : self.name
+        else
+            self.short_name
+        end
+    end
+
+    # Set the first letter, which is used for faster queries
+    def set_first_letter
+        unless name.blank?
+            # we use a regex to ensure it works with utf-8/multi-byte
+            first_letter = Unicode.upcase name.scan(/^./mu)[0]
+            if first_letter != self.first_letter
+                self.first_letter = first_letter
+            end
+        end
+    end
+
+    def update_url_name
+        if changed.include?('name') || changed.include?('short_name')
+            self.url_name = MySociety::Format.simplify_url_part(self.short_or_long_name, 'body')
+        end
+    end
+
+end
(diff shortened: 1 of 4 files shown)


app/models/public_body.rb
```diff
@@ -68,6 +68,11 @@ class PublicBody < ActiveRecord::Base
     }
 
     translates :name, :short_name, :request_email, :url_name, :notes, :first_letter, :publication_scheme
+    accepts_nested_attributes_for :translations, :reject_if => :empty_translation_in_params?
```
garethrees: This looks like it should get handled by [`Translatable`](https://github.com/mysociety/alaveteli/blob/develop/app/models/concerns/translatable.rb#L4-L6).

garethrees: If not, probably deserves its own commit so that the intention is clear

app/models/concerns/public_body_derived_fields.rb
```diff
@@ -0,0 +1,46 @@
+module PublicBodyDerivedFields
+    extend ActiveSupport::Concern
+
+    included do
+        before_save :set_first_letter
+
+        # When name or short name is changed, also change the url name
+        def short_name=(short_name)
```
garethrees: Looking at https://signalvnoise.com/posts/3372-put-chubby-models-on-a-diet-with-concerns and https://github.com/mysociety/alaveteli/blob/develop/app/models/concerns/translatable.rb, it looks like the `included` block is only needed for macros, so the method definitions should be outside of the block?

garethrees:
Looks good other than the minor comments. Should probably squash all the commits (retaining commit messages) in to 3ef4980659bdae8bcd723bbd4b7f649e60089603, so that we don't have commits that have a failing spec suite.

garethrees:
> The methods that are overriding accessors need to be in the `included` block - http://stackoverflow.com/questions/10436700/override-models-attribute-accessor-in-a-mixin-module

Weird! Looks good :+1: :ship:

gorm:
Tried to cherry-pick this into master (which is where I need this for our deploy), but it doesn't play too well because of missing autoload. Should it be in the publicbody folder instead of concerns? Not sure about autoload in ruby, but error seems real?
```
 NameError in GeneralController#frontpage

uninitialized constant PublicBody::PublicBodyDerivedFields

app/models/public_body.rb:69:in `<class:PublicBody>'
app/models/public_body.rb:34:in `<top (required)>'
app/views/general/_frontpage_search_box.html.erb:6:in `_app_views_general__frontpage_search_box_html_erb__2406349430524979121_15552660'
app/views/general/frontpage.html.erb:8:in `block in _app_views_general_frontpage_html_erb__2578638797913161574_15413000'
app/helpers/application_helper.rb:108:in `block in cache_if_caching_fragments'
app/helpers/application_helper.rb:108:in `cache_if_caching_fragments'
app/views/general/frontpage.html.erb:1:in `_app_views_general_frontpage_html_erb__2578638797913161574_15413000'
app/controllers/application_controller.rb:111:in `record_memory'
lib/whatdotheyknow/strip_empty_sessions.rb:14:in `call'
```

gorm:
Sorry for late reply here. I forgot about this one.

I tested the two commits slightly and it seems to generate first_letter for all public bodies I imported. So it looks good and hope to see this in master soon :)

PR #2568, opened 2015-06-17
Update jquery-rails to get the fix for CVE-2015-1840.

crowbot:


garethrees:
Looks okay as far as I can see.

PR #2575, opened 2015-06-18
Add json initialiser as workaround for CVE-2015-3226.

crowbot:


garethrees:
Looks like the recommended fix, so :+1:

lizconlan:
Yes, unfortunately the 4 series patch has no effect on 3 series code as `ActiveSupport::JSON::Encoding:: EscapedString` was introduced in Rails 4 :(

For now I've just ticketed it as "could be an issue for future development" in foi-reg

PR #2576, opened 2015-06-18
Switch xml backend to nokogiri to avoid CVE-2015-3227.

crowbot:


config/initializers/xml.rb
```diff
@@ -0,0 +1 @@
+ActiveSupport::XmlMini.backend = 'Nokogiri'
```
garethrees: _pedant:_ No newline at the end of the file.

garethrees:
Looks ok to me :+1:

PR #2579, opened 2015-06-18
Use :platform to add rack-utf8_sanitizer

crowbot:
Unfortunately, the conditional used in 090531bf2d2b763e5bb281658e91b58905912130
results in `Gemfile.lock` being inconsistent with `Gemfile` under ruby 1.8.7 so
we can't use it until after Alaveteli release 0.22, which is the last release
that will support ruby 1.8.7. If we merge this, I think we should immediately create a ticket for extending it to ruby 2.0.0 after release 0.22


garethrees:
Bah, :+1: 

> If we merge this, I think we should immediately create a ticket for extending it to ruby 2.0.0 after release 0.22

Yup sounds good

PR #2580, opened 2015-06-18
Use BASH_SOURCE to return the script path.

crowbot:
It is reliable in the case where the script is sourced from another
script.


garethrees:
:+1:

PR #2584, opened 2015-06-19
Add nokogiri to main group

crowbot:
As we're using nokogiri as the XmlMini backend as of bd6e94fc7c9c1790622ba29655f6e7e27044dd64,
it should be in the default gem group.


garethrees:
:+1:

PR #2598, opened 2015-06-25
Hotfix/0.21.0.40

crowbot:


spec/script/handle-mail-replies_spec.rb
```diff
@@ -12,11 +12,22 @@ def mail_reply_test(email_filename)
...
+        it "should not fail handling a bounce mail" do
+            xc = ExternalCommand.new("script/handle-mail-replies",
+                                     { :stdin_string => load_file_fixture("track-response-exim-bounce.email") })
+            xc.run
+            xc.err.should == ""
+        end
+
+        it 'should not fail handling a non-UTF8 encoded mail', :focus => true do
+            xc = ExternalCommand.new("script/handle-mail-replies",
+                                     { :stdin_string => load_file_fixture("russian_email.email") })
```
garethrees: The email fixture file actually specifies that its UTF-8 (don't know if the lower-case `utf-8` is significant)

```
Content-Type: text/plain; charset=utf-8; format=flowed
Content-Transfer-Encoding: 8bit
```

Is it the case that this example mail is badly encoded?

crowbot: Hmmm..actually it's valid - I'll rewrite the test description accordingly.

PR #2606, opened 2015-06-25
Add deprecation notice for ruby 1.8.7

crowbot:
Not in an initializer to avoid any risk of popping up in mail
handling contexts.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -35,6 +35,17 @@ class ApplicationController < ActionController::Base
     before_filter :set_vary_header
     before_filter :validate_session_timestamp
     after_filter  :persist_session_timestamp
+    before_filter :deprecation_notice
+
+    def deprecation_notice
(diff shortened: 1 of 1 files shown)


app/controllers/application_controller.rb
```diff
@@ -35,6 +35,15 @@ class RouteNotFound < StandardError
     before_filter :set_vary_header
     before_filter :validate_session_timestamp
     after_filter  :persist_session_timestamp
+    before_filter :deprecation_notice
+
+    def deprecation_notice
+        if RUBY_VERSION.to_f <= 1.9
+            ActiveSupport::Deprecation.warn "[DEPRECATION] You are using Ruby 1.8. This will not be supported " \
+                                    "as of Alaveteli release 0.23. Please upgrade your ruby version."
```
garethrees: Should we add a link to http://git.io/vLNg0?

crowbot: Good idea.

garethrees:
:+1:

PR #2611, opened 2015-06-29
Fix Widget

garethrees:


garethrees:
Added https://github.com/mysociety/alaveteli/pull/2611#commits-pushed-739e37b for review

wrightmartin:
@crowbot here's what I've done locally - is this enough, or am I missing a key part here 
screen shot 2015-07-03 at 16 03 44

wrightmartin:
If you're already tracking it

screen shot 2015-07-03 at 16 28 54

wrightmartin:
@crowbot ah, yep, easy - I didn't think it made sense as a button with that wording, I'll do that

wrightmartin:
With the wording as is, it will overflow and needs a tweak. If you change it to 'unsubscribe' it fits fine

screen shot 2015-07-03 at 16 44 11

Are there any other states to this button?

garethrees:
> Are there any other states to this button?

Yup https://github.com/mysociety/alaveteli/blob/release-0-22-fix-widget/app/views/widgets/show.html.erb#L25-L53

garethrees:
For reference, here are all the button states:

```html
<!-- track: subscribing to emails -->
<!-- vote: simply increments the vote count. No subscription -->
 
 
<!-- User owns request; they cannot make an action -->
<div class="alaveteli-widget__bottom">
  <span class="alaveteli-widget__bottom--user-owns-request">This is your request</span>
</div>
 
<!-- The signed in user has an existing track; clicking this deletes the track -->
<div class="alaveteli-widget__bottom">
  <a class="alaveteli-widget__bottom--unsubscribe">Unsubscribe</a>
</div>
 
<!-- The user has clicked "i want to know" so has a cookie set -->
<!-- but can click again to convert it to a track -->
<div class="alaveteli-widget__bottom">
  <a class="alaveteli-widget__bottom--sign-in-to-track">Sign in to track this request</a>
</div>
 
<!-- A signed in user who isn't currently tracking the request can create a track directly -->
<div class="alaveteli-widget__bottom">
  <a class="alaveteli-widget__bottom--create-track">I also want to know!</a>
</div>
 
<!-- A user who is not signed in can "vote" on a request without further action; -->
<!-- this just creates a cookie that they can convert to a track later. -->
<div class="alaveteli-widget__bottom">
  <input class="alaveteli-widget__bottom--create-vote" value="I also want to know!" />
</div>
```

garethrees:
Still need to fix https://github.com/mysociety/alaveteli/pull/2611#commitcomment-11997477 so just removing the 'awaiting deploy' marker until then

Issue #2263, opened 2015-03-27
Release 0.22


crowbot:
- [x] Look for deprecation notices - remove code
- [x] Run brakeman
- [ ] Bring AskTheEU back on to the master branch - need to check their theme (https://github.com/mysociety/asktheeu-theme/issues/4) and check that translations are up to date (https://github.com/mysociety/asktheeu-theme/issues/8), also merge the changes in the develop branch of their theme.

garethrees:
`Translatable#translated_versions=` is deprecated

garethrees:
Release announced https://groups.google.com/forum/#!topic/alaveteli-dev/3aHg7XLajTY

garethrees:
New brakeman warnings:

Don't think they're especially concerning.

One is in the admin interface and the other user input is the translation of the public body name; an admin input.

```json
"new": [
    {
        "code": "PublicBody.where([((get_public_body_list_translated_condition(\"public_body_translations\", (false or true), true) + \" AND \") + ((\"public_bodies.id <> #{PublicBody.internal_admin_body.id}\" + (\" AND (SELECT count(*) FROM has_tag_string_tags WHERE has_tag_string_tags.model_id = public_bodies.id AND has_tag_string_tags.model = 'PublicBody'\" + \" AND has_tag_string_tags.name in (#{PublicBodyCategory.get.tags.map do\n \"'#{c}'\"\n end.join(\",\")})) = 0\")) + \" AND (SELECT count(*) FROM has_tag_string_tags WHERE has_tag_string_tags.model_id = public_bodies.id AND has_tag_string_tags.model = 'PublicBody'\")), \"%#{(params[:public_body_query] or \"\")}%\", \"%#{(params[:public_body_query] or \"\")}%\", \"%#{(params[:public_body_query] or \"\")}%\", locale_from_params.gsub(\"-\", \"_\")])",
        "confidence": "Medium",
        "file": "app/controllers/public_body_controller.rb",
        "fingerprint": "333b2abcdd6293c947ace5d61fa877d4268749d35d177ebc509363f2f4deb1be",
        "line": 205,
        "link": "http://brakemanscanner.org/docs/warning_types/sql_injection/",
        "location": {
            "class": "PublicBodyController",
            "method": "list",
            "type": "method"
        },
        "message": "Possible SQL injection",
        "render_path": null,
        "user_input": "get_public_body_list_translated_condition(\"public_body_translations\", (false or true), true)",
        "warning_code": 0,
[...]

PR #2616, opened 2015-06-30
[#2387] Discard spam messages

### LINKED ISSUE #2387: Discard incoming mail with a SpamAssassin header over a threshold
- [x] Add configuration option
- [x] Add spec/code
- [x] Add changelog entry
- [x] Add docs to `gh-pages`
- [x] Find sensible threshold for WDTK

garethrees:
Fixes #2387 

If you filter incoming emails through a spam detector like SpamAssassin, you can configure Alaveteli to ignore messages with a high spam score. This feature requires the messages to contain a header with a numeric spam score, and both `INCOMING_EMAIL_SPAM_HEADER` and `INCOMING_EMAIL_SPAM_THRESHOLD` to be configured before the filtering will take effect.

I've made this PR to `master` because I think its important we get this in operation on WDTK quickly.
Its a good isolated candidate for trailing a [minor release](http://alaveteli.org/docs/running/upgrading/#minor-version-increases).

This should be merged after #2263 and released as `0.22.1.0`.

Will need to bump `ALAVETELI_VERSION` and add a changelog entry.


garethrees:
We talked In standup we thought it might be a good idea to just delete spam rather than sending it to the holding pen.

garethrees:
Added some commits to ignore mail above the spam threshold.

PR #2631, opened 2015-07-06
Redesign footer

wrightmartin:
This adds the back-ported footer from WDTK, with some minor changes

screen shot 2015-07-06 at 11 44 22


garethrees:
Looks great. I've made a PR to add a Facebook config option (#2648), so we might as well add that to the footer when its merged.

garethrees:
Facebook config is now available (33bd415400d55a3f69d9a9942857262304799e09) in develop.

PR #2634, opened 2015-07-06
Add new style locale switcher HTML code

wrightmartin:
Changes the locale switcher to include markup that will be used for JS dropdown functionality in themes


garethrees:
@crowbot could you review my commits to this once the build passes? Cheers.

garethrees:
> The locales now jump around as you select them - can we keep the order static, or does the current locale have to be first?

Current has to be first. I don't think its a major problem as users are unlikely to repeatedly switch around languages like they might a navigation bar.

PR #2636, opened 2015-07-07
Create and use template_utf8 during vagrant provisioning

### LINKED ISSUE #2501: Vagrant should create database with template_utf8
The vagrant build seems to create database with template0 and not template_utf8, so it's causing an issue with running tests on the vagrant machine (it will drop the alaveteli_test db, then try to recreate a new one with a non-existent template, template_utf8). 

So maybe best to change script/install-as-user to also use template_utf8 (like the docs and database.yml file does)?

stevenday:
This adds a new step to the install-as-user script to create a UTF-8 db
template, then use it when we create the dev/test/prod databases.

Closes #2501

### DIFF
--- a/script/install-as-user
+++ b/script/install-as-user
@@ -127,6 +127,9 @@ sed -r \
     -e "s,# constraint_disabling: false,  constraint_disabling: false," \
     config/database.yml-example > config/database.yml
 
+# create the template_utf8 template we'll use for our databases
+createdb -T template0 -E UTF-8 template_utf8
+
 for SUFFIX in production test development
 do
     REAL_DB_NAME="${DB_NAME}_$SUFFIX"
@@ -134,7 +137,7 @@ do
     # Create each database if it doesn't exist:
     if ! psql -l | egrep "^ *$REAL_DB_NAME *\|" > /dev/null
     then
-        createdb -T template0 --owner "$UNIX_USER" "$REAL_DB_NAME"
+        createdb -T template_utf8 --owner "$UNIX_USER" "$REAL_DB_NAME"
     fi
 done
 
--- a/script/install-as-user
+++ b/script/install-as-user
@@ -129,6 +129,9 @@ sed -r \
 
 # create the template_utf8 template we'll use for our databases
 createdb -T template0 -E UTF-8 template_utf8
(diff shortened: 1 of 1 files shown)


script/install-as-user
```diff
@@ -127,6 +127,9 @@ sed -r \
     -e "s,# constraint_disabling: false,  constraint_disabling: false," \
     config/database.yml-example > config/database.yml
 
+# create the template_utf8 template we'll use for our databases
+createdb -T template0 -E UTF-8 template_utf8
```
garethrees: I had to [explicitly set](http://alaveteli.org/docs/installing/manual_install/#configure-database) `datistemplate`. Not sure how much it matters

stevenday: Ah, I wondered about that, but it seemed to work fine for me without doing it, so I left it as-is just to make it simpler. Do you remember where it was an issue if you left it out?

garethrees: Off the top of my head, I don't think you're creating a "template" with this command. I think `set datistemplate=true` prevents some modifications like changing the locale and collate settings.

garethrees: I wonder if we should also be setting `datallowconn=false` to prevent connections to it.

garethrees:
```sql
select datname,datistemplate,datallowconn from pg_database;
                    datname                    | datistemplate | datallowconn
-----------------------------------------------+---------------+--------------
 template1                                     | t             | t
 template0                                     | t             | f
```

stevenday:
@garethrees - I've updated this to copy your `datistemplate` setting and set `datallowconn` too, it doesn't seem like it'll do any *harm* to set it, so why not..

garethrees:
Thanks! Merged in cdb4630

script/install-as-user
```diff
@@ -127,6 +127,12 @@ sed -r \
...
+echo "update pg_database set datistemplate=true datallowconn=false where datname='template_utf8';" > /tmp/update-template.sql
+sudo -u postgres psql -f /tmp/update-template.sql
```
dracos: This line will only run successfully if the user installing alaveteli has sudo rights/access. If sudo is needed, it needs to be in /site-specific-install.sh, not install-as-user. Worth a function in commonlib's install-site, if this might be useful elsewhere (similarly to the postgis function)?

(Also, you can run a command directly, no need for a temporary file here.)

PR #2640, opened 2015-07-07
Replace .attr("value", val) with .val(val)

crowbot:
Looks like this has changed behaviour in the later version of jquery
we now use https://github.com/jquery/jquery-migrate/blob/master/warnings.md#jqmigrate-jqueryfnattrvalue-val-no-longer-sets-properties


kingqueen3065:
I don't understand this whole system, other than how to submit bugs and feature reqeusts, so apologies that I don't know if and how I should respond to notifications sent to me by it! but thank you for working on it.

garethrees:
@kingqueen3065 no need for a response unless there's something you specifically want to add :)

PR #2645, opened 2015-07-10
Replace existing hero section with new one, based on WDTK

wrightmartin:
This adds the new hero section from WDTK. 

I've broken out the copy sections that are likely to be overridden, but I haven't thought up sensible defaults for them yet

screen shot 2015-07-10 at 10 20 05


garethrees:
Merged in 7c4f6f15774161f7c703df465a4bd12b796ad676

PR #2652, opened 2015-07-10
Added OpenGraph metadata

crowbot:


garethrees:
:+1:

owenblacker:
Please ping me when this actually makes it to the live site; it might actually be the first time any code of mine has made it to a production mySociety site since mySociety became a thing :grinning:

owenblacker:
_\*excited*_

PR #2654, opened 2015-07-12
Recognize Norwegian email marker.

petterreinholdtsen:
Recognize Norwegian marker between top posted messages and original text in email.

### DIFF
--- a/app/models/incoming_message.rb
+++ b/app/models/incoming_message.rb
@@ -294,9 +294,11 @@ class IncomingMessage < ActiveRecord::Base
         # http://www.whatdotheyknow.com/request/123/response/192
         # http://www.whatdotheyknow.com/request/235/response/513
         # http://www.whatdotheyknow.com/request/445/response/743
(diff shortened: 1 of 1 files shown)


petterreinholdtsen:
Is this the correct way to handle translated emails?

garethrees:
We don't really have a good way of handling this at the moment.

The current ["supported"](http://alaveteli.org/docs/customising/themes/#adding-or-overriding-models-and-controllers) way would be to override the method in your theme, adding the strings that are relevant to your language.

I'd like to see a way to extend what the quoting picks up, along the lines of:

```ruby
IncomingMessage.end_of_message_strings << '''----*\s*Opprinnelig melding\s*----*'''
```

garethrees:
Hey @petterreinholdtsen,

I've had a think about this and rather than accept this PR, I've made https://github.com/mysociety/alaveteli/issues/2662. Before we get to this you can use the [currently supported method](http://alaveteli.org/docs/customising/themes/#adding-or-overriding-models-and-controllers) of modifying methods.

Thanks for the contribution and prompting the discussion!

PR #2655, opened 2015-07-13
Add FACEBOOK_USERNAME configuration documentation

garethrees:


garethrees:
This might be an easy one to try translating before we merge it?

PR #2669, opened 2015-07-20
add backported print styles from WDTK

wrightmartin:
There may be one or selectors that don't exist here yet, but they will when other PRs are merged in


garethrees:
Will leave this one 'til last. Marking as needs clarification so that we know its blocked on other PRs.

garethrees:
Needs updating with develop

wrightmartin:
Have rebased from develop, and is now ready for review

garethrees:
:+1:

PR #2673, opened 2015-07-21
Retrofit header

wrightmartin:
This does two things

1) Makes the header and positioning as simple as on WDTK
2) Parameterises the logo using settings.scss


app/assets/stylesheets/responsive/_settings.scss
```diff
@@ -3,3 +3,9 @@ $row-width: 64em;
 $body-font-family: "Helvetica Neue", Arial, Helvetica, Helmet, Freesans, sans-serif;
 $form-label-font-color: #333333;
 $base-font-size: 16px;
+
+// Specify the filename and dimensions of your logo file here
+$logo-filename: 'logo.png';
+$logo-highdpi-filename: 'logo@2.png'; //Should be at least 2x the size, in the same proportions
```
garethrees: We should include a default logo for this file:

```sh
$ grep -r logo@2.png app/assets/images/ || echo 'Not Found'
Not Found
```

garethrees:
Looks great other than including a default retina logo

PR #2678, opened 2015-07-22
Set the return-path for the contact form mail.

### LINKED ISSUE #2635: Set Return-path and Reply-to on messages sent to admins via contact form
One of the site admins on WDTK is experiencing issues when a user whose email domain uses SPF records sends mail to the site admins via the contact form. The admin's mail server checks the SPF record for the sender and rejects the mail as it doesn't match the `envelope from`. We can set the `envelope from` by explicitly setting the `return-path` header, as we do in some other messages. We should also then set the `Reply-to` header to the user's email address.

crowbot:
If the MTAs for the people on the support alias for the site check
SPF, this is the field they check, so it should be something we
control. Closes #2635.

### DIFF
--- a/app/mailers/contact_mailer.rb
+++ b/app/mailers/contact_mailer.rb
@@ -10,6 +10,9 @@ class ContactMailer < ApplicationMailer
   def to_admin_message(name, email, subject, message, logged_in_user, last_request, last_body)
     @message, @logged_in_user, @last_request, @last_body = message, logged_in_user, last_request, last_body
 
+    # Return path is an address we control so that SPF checks are done on it.
+    headers('Return-Path' => blackhole_email, 'Reply-To' => MailHandler.address_from_name_and_email(name, email))
+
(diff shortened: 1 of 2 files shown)


app/mailers/contact_mailer.rb
```diff
@@ -10,6 +10,8 @@ class ContactMailer < ApplicationMailer
     def to_admin_message(name, email, subject, message, logged_in_user, last_request, last_body)
         @message, @logged_in_user, @last_request, @last_body = message, logged_in_user, last_request, last_body
 
+        # Return path is an address we control so that SPF checks are done on it.
+        headers('Return-Path' => blackhole_email, 'Reply-To' => MailHandler.address_from_name_and_email(name, email))
```
garethrees: Would be worth adding a quick spec for this

garethrees:
Looks good but seems to have the commit from #2687

PR #2687, opened 2015-07-27
Refactor repeated logic to method on InfoRequestEvent.

crowbot:


app/models/info_request_event.rb
```diff
@@ -427,4 +427,13 @@ def json_for_api(deep, snippet_highlight_proc = nil)
 
     return ret
   end
+
+  def set_calculated_state!(state)
```
garethrees: Would be great to add a spec for this

app/models/info_request_event.rb
```diff
@@ -427,4 +427,13 @@ def json_for_api(deep, snippet_highlight_proc = nil)
...
+    unless calculated_state == state
+      self.calculated_state = state
+      self.last_described_at = Time.now
+      save!
```
garethrees: I have a mild preference for this method not calling `save!` (or `save`).

It would be more flexible for it to just set the attributes and the programmer can save the record later (which happens [here](https://github.com/mysociety/alaveteli/blob/small-method-extraction/app/models/info_request.rb#L659-L664) though the `TODO` is quite right.).

The use of `save!` isn't ideal as we don't really handle the exception anywhere. Calling `save` and handling validation errors is usually preferable.

As an alternative, we could override `calculated_state=`?

I'm not that sure on whether this would be a pain elsewhere though.

```ruby
def calculated_state=(new_calculated_state)
  unless calculated_state == new_calculated_state
      @calculated_state = new_calculated_state
      self.last_described_at = Time.now
  end
end
```

Might also be preferable to use [`touch(:last_described_at)`](http://api.rubyonrails.org/v3.2.22/classes/ActiveRecord/Persistence.html#method-i-touch)?

This is still better than what's there though, so `¯\_(ツ)_/¯`

crowbot: Yeah, really my intention here was just to get a small, simple refactor that wouldn't change functionality done quickly.

PR #2690, opened 2015-07-28
Refactor users controller rebased

crowbot:


garethrees:
Few minor bits but :+1: overall

garethrees:
:+1:

PR #2693, opened 2015-07-30
Add some more info about how to identify a request from its address.

crowbot:


docs/running/admin_manual.md
```diff
@@ -385,7 +385,14 @@ the request's *title_url* will already be in the input box: click the
 
 If there is not a guess, or Alaveteli's guess is wrong, look at the  `To:`
 address of the raw email and the contents of the message itself. You need
-to figure out which request it belongs to. You can browse and search
+to figure out which request it belongs to. The special addresses generated by
+Alaveteli are of the form:
+
+<pre><code>&lt;foi+request-3-691c8388@example.com&gt;
+</code></pre>
```
garethrees: I think it would be worth adding a comment above that line, noting that the prefix and domain are `general.yml` config options and the `id` and `idhash` are database attributes of request?

```
# [INCOMING_EMAIL_PREFIX]request-[id]-[idhash]@[DOMAIN]
```

garethrees:
Looks good. Comment addition would help add even more clarity, but good as it is.

PR #2619, opened 2015-07-01
[#2482] Fix translated public body sort

garethrees:
Fixes #2482 

~~WIP for initial review~~

~~Specs will fail unless nn_NO language is available:~~

```sh
$ sudo apt-get install language-pack-nn
```

##### Todo

- [x] Changelog entry
- [x] Documentation on adding a collation

PR #2694, opened 2015-07-30
Add Internationalised Sorting documentation

garethrees:
Introduced in https://github.com/mysociety/alaveteli/pull/2619


garethrees:
Needs translation

PR #2705, opened 2015-08-03
[#2587] Transaction stats json

### LINKED ISSUE #2587: Include all transactions in version.json
Currently we don't have:

- annotations added
- 'track this request' signups
- widget votes
- User to User messages (not easily trackable https://github.com/mysociety/research/issues/61 would help)
- `public_body_change_requests` – where a user suggests a modification to a public body record
- `request_classifications` – where users classify requests (that aren't their own?)
- followup messages

@paullenz let us know if there's anything else that would be useful.

garethrees:
Fixes #2587 

Target version: `0.22.2.0`

### DIFF
--- a/app/controllers/general_controller.rb
+++ b/app/controllers/general_controller.rb
@@ -190,12 +190,22 @@ class GeneralController < ApplicationController
 
   def version
     respond_to do |format|
-      format.json { render :json => {
-                      :alaveteli_git_commit => alaveteli_git_commit,
-                      :alaveteli_version => ALAVETELI_VERSION,
-                      :ruby_version => RUBY_VERSION,
-                      :visible_request_count => InfoRequest.visible.count,
-                      :confirmed_user_count => User.where(:email_confirmed => true).count
+      format.json {
+        render :json => {
+          :alaveteli_git_commit => alaveteli_git_commit,
+          :alaveteli_version => ALAVETELI_VERSION,
+          :ruby_version => RUBY_VERSION,
+          :visible_public_body_count => PublicBody.visible.count,
+          :visible_request_count => InfoRequest.visible.count,
+          :confirmed_user_count => User.where(:email_confirmed => true).count,
+          :visible_comment_count => Comment.visible.count,
+          :track_thing_count => TrackThing.count,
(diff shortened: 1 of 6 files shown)


app/models/info_request_event.rb
```diff
@@ -32,6 +32,7 @@ class InfoRequestEvent < ActiveRecord::Base
   belongs_to :incoming_message
   belongs_to :comment
 
+  has_many :request_classifications
```
crowbot: I think this is probably better represented as a `has_one` relationship - there's only one `RequestClassification` per `InfoRequestEvent`

spec/controllers/general_controller_spec.rb
```diff
@@ -2,6 +2,44 @@
...
+          stub(:alaveteli_git_commit).
+            and_return(mock_git_commit)
+
+      expected = { :alaveteli_git_commit => mock_git_commit,
+                   :alaveteli_version => ALAVETELI_VERSION,
+                   :ruby_version => RUBY_VERSION,
+                   :visible_request_count => 13,
+                   :confirmed_user_count => 9,
+                   :visible_comment_count => 2,
+                   :track_thing_count => 2,
```
crowbot: Are these first few values implicitly based on the fixtures? Might be better to divorce them if so in order to make the test independent of changes to the fixtures.

app/controllers/general_controller.rb
```diff
@@ -190,12 +190,19 @@ def not_found
...
+          :alaveteli_version => ALAVETELI_VERSION,
+          :ruby_version => RUBY_VERSION,
+          :visible_request_count => InfoRequest.visible.count,
+          :confirmed_user_count => User.where(:email_confirmed => true).count,
+          :visible_comment_count => Comment.visible.count,
+          :track_thing_count => TrackThing.count,
+          :widget_vote_count => WidgetVote.count,
+          :public_body_change_request_count => PublicBodyChangeRequest.count,
+          :request_classification_count => RequestClassification.count,
+          :followup_message_count => OutgoingMessage.where(:message_type => 'followup').count
```
crowbot: Probably should also add a visible condition on `OutgoingMessage`.

PR #2706, opened 2015-08-04
[#2629] transaction stats rake

### LINKED ISSUE #2629: Add user signups to stats.rake
Related to #2587.

https://github.com/mysociety/alaveteli/blob/master/lib/tasks/stats.rake#L12 should include user signups

garethrees:
Fixes #2629 

Need to squash tidy up commits

### DIFF
--- a/doc/CHANGES.md
+++ b/doc/CHANGES.md
@@ -1,5 +1,9 @@
 # develop
 
+## Highlighted Features
+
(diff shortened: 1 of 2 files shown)


garethrees:
I think this is due a tidy up while we're here.

garethrees:
Merged in 7259b92140e9ef773d9550bb203fa3da0088140a

PR #2715, opened 2015-08-06
Use standard patterns in all admin controllers.

crowbot:
Standardising some patterns in the admin controllers.


jpmckinney:
I had named the before_filters `:set_` (instead of `:find_`), because that is what Rails 4 scaffolding generates.

jpmckinney:
Oops, it was `set_` not `get_`.

garethrees:
Sounds like this isn't ready for review yet so removed the label. Pop it back on after updates :)

garethrees:
Looks great. A couple of minor points but good to deploy when resolved.

PR #2717, opened 2015-08-07
Remove code duplication in admin_censor_rule_controller.rb with before_filter

jpmckinney:
Missed one in #2715


jpmckinney:
Once #2715 is merged, this PR is just https://github.com/mysociety/alaveteli/commit/0f8be902f1374771434847a0e67270f006871516

jpmckinney:
Good catch! Pushed.

jpmckinney:
I've rebased and fixed.

PR #2719, opened 2015-08-07
definition lists are easier to read and follow

wrightmartin:


garethrees:
Looks good. You could add a Changelog entry ([example here)](https://github.com/mysociety/alaveteli/commit/fa46557ca3e919ac44adb311cd45c3bd40f49c3b) as I  think this would appeal to re-users when they come to upgrade :) :+1:

Issue #619, opened 2012-09-23
request-creation-graph throws warnings on a new install


henare:
[request-creation-graph](https://github.com/mysociety/alaveteli/blob/develop/script/request-creation-graph) spits stuff like this out on a new install:

```
gnuplot> plot "/tmp/foi-creation-rate-graph-data-1331528155" using 1:2 with impulses lt 8 lw 4 title "waiting_response",          "/tmp/foi-creation-rate-graph-data-749023717" using 1:2 with impulses lt 3 lw 4 title "waiting_clarification",          "/tmp/foi-creation-rate-graph-data-406429876" using 1:2 with impulses lt 9 lw 4 title "not_held",          "/tmp/foi-creation-rate-graph-data-2632630602" using 1:2 with impulses lt 6 lw 4 title "rejected",          "/tmp/foi-creation-rate-graph-data-32609883" using 1:2 with impulses lt 2 lw 4 title "successful",          "/tmp/foi-creation-rate-graph-data-147129119" using 1:2 with impulses lt 10 lw 4 title "partially_successful",          "/tmp/foi-creation-rate-graph-data-1453118459" using 1:2 with impulses lt 5 lw 4 title "requires_admin",          "/tmp/foi-creation-rate-graph-data-30514681" using 1:2 with impulses lt 7 lw 4 title "gone_postal",          "/tmp/foi-creation-rate-graph-data-331711846" using 1:2 with impulses lt 4
  lw 4 title "internal_review",          "/tmp/foi-creation-rate-graph-data-12608695" using 1:2 with impulses lt 12 lw 4 title "error_message",          "/tmp/foi-creation-rate-graph-data-2291711105" using 1:2 with impulses lt 13 lw 4 title "user_withdrawn",     "< awk 'BEGIN { n = 0 } { n += $2; print $1, $2, n; }' /tmp/foi-creation-rate-graph-data-1331528155" using 1:3 axes x1y2 with lines lt 2 title "cumulative total number of requests"
                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                            ^
         line 0: warning: Skipping data file with no valid points
```

<!---
@huboard:{"order":478.2101903874427,"milestone_order":619,"custom_state":""}
-->

garethrees:
Had a look at this because its really annoying me, having set up a few new sites lately.

Tried adding some data to the files if the file was zero-length…

```diff
[...]

PR #2725, opened 2015-08-13
Quiet down graph generating cron job on new installations.

petterreinholdtsen:
The cron job complained about empty data files because there were
no requests of a given category used in the graph.  This change make
sure every data set have at least one zero value data point.

https://github.com/mysociety/alaveteli/issues/619


garethrees:
:+1: thanks!

PR #2731, opened 2015-08-14
WIP Use commonlib function to create template_utf8

### LINKED ISSUE #2501: Vagrant should create database with template_utf8

stevenday:
Another attempt at #2501, following [@dracos' comments](https://github.com/mysociety/alaveteli/pull/2636#discussion_r34871729) on the last one.

Adds a new function to commonlib (see related PR here
https://github.com/mysociety/commonlib/pull/42) to create the template and
then uses it in the `install-as-user` script as before.

Closes #2501

### DIFF
--- a/script/site-specific-install.sh
+++ b/script/site-specific-install.sh
@@ -190,10 +190,7 @@ usermod -a -G adm "$UNIX_USER"
 add_postgresql_user --superuser
 
 # create the template_utf8 template we'll use for our databases
-sudo -u postgres createdb -T template0 -E UTF-8 template_utf8
-sudo -u postgres psql <<EOF
-update pg_database set datistemplate=true, datallowconn=false where datname='template_utf8';
-EOF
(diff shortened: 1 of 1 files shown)


stevenday:
No worries, the only reason I took another look is because @garethrees re-assigned the original issue to me. It looks like he might have fixed it in the meantime and forgotten though..

stevenday:
Hmm, I'm going to scrap this because it's not working as I expected and I don't really have time to figure out why and it's not solving an actual problem in alaveteli anymore anyway (site-specific-install.sh can't find the function from commonlib for some reason).

Issue #2646, opened 2015-07-10
Release 0.23


garethrees:
Release notes: 
- [x] Add `RESTRICT_NEW_RESPONSES_ON_OLD_REQUESTS_AFTER_MONTHS`
- [x] Add `ENABLE_TWO_FACTOR_AUTH`
- [x] Add `GEOIP_DATABASE`
- [x] Reduce `RESTRICT_NEW_RESPONSES_ON_OLD_REQUESTS_AFTER_MONTHS` to `3`
- [ ] remove the extra instance of gaze on our server when African sites are upgraded.
- [x] Delete WDTK branches
- [x] Delete alavetelitheme branches

crowbot:
- [ ] remove the extra instance of gaze on our server.

garethrees:
- [ ] Reduce `RESTRICT_NEW_RESPONSES_ON_OLD_REQUESTS_AFTER_MONTHS` to `3`

crowbot:
Have made [a new ticket](https://github.com/mysociety/alaveteli/issues/2905) for removal of Gaze server - it has other dependencies.

PR #2737, opened 2015-08-18
2642 invalid byte seq

crowbot:


garethrees:
Haven't tested it on a Wheezy system, but the code looks good to me.

garethrees:
This probably should have had a changelog entry / upgrade notes – reminder for #2646

Issue #2582, opened 2015-06-19
Remove squeeze from supported OS list 


crowbot:
It's end of life is [Feb 2016](https://wiki.debian.org/LTS)

PR #2738, opened 2015-08-18
Remove support for Debian squeeze.

crowbot:
Fixes https://github.com/mysociety/alaveteli/issues/2582

It's end of life in Feb 2016 and doesn't package any ruby above
1.8.7.


garethrees:
:+1:

PR #2740, opened 2015-08-18
Drop ruby 1.8

crowbot:


garethrees:
Looks good other than minor comment change

PR #2741, opened 2015-08-18
Add missing spec helper method

garethrees:
Lost from 30ecc1111900765ee589d10743ce5a889e841775


garethrees:
Ah just noticed this is also in https://github.com/mysociety/alaveteli/commit/8c1c9fb4ed8d6ae6977c8141a1cbe861d55b39dc

PR #2752, opened 2015-08-24
add explanation to FORCE_REGISTRATION_ON_NEW_REQUEST docs

davewhiteland:
This arose because a user asked for clarification: in fact,
the config setting is a little misleadingly named since
(as hopefully described here) it's really about forcing
registration at the *start* of the new request process,
and doesn't mean no registration is required at all.

Please check that what I've put here (in the config setting's explanation) is true :-)

### DIFF
--- a/docs/customising/config.md
+++ b/docs/customising/config.md
@@ -206,11 +206,32 @@ indentation correct. If in doubt, look at the examples already in the file, and 
     <a name="force_registration_on_new_request"><code>FORCE_REGISTRATION_ON_NEW_REQUEST</code></a>
   </dt>
   <dd>
-    Does a user needs to sign in to start the New Request process?
+    Does a user need to sign in to start the New Request process?
+    <p>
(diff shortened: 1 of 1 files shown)


docs/customising/config.md
```diff
@@ -206,11 +206,32 @@ indentation correct. If in doubt, look at the examples already in the file, and
     <a name="force_registration_on_new_request"><code>FORCE_REGISTRATION_ON_NEW_REQUEST</code></a>
   </dt>
   <dd>
-    Does a user needs to sign in to start the New Request process?
+    Does a user need to sign in to start the New Request process?
```
garethrees: Maybe we could highlight "start" here:

``
Does a user need to sign in to _start_ the New Request process?
```

garethrees:
Looks good. A counterpart PR to keep the [example config file](https://github.com/mysociety/alaveteli/blob/develop/config/general.yml-example#L257-L262) docs in sync would be most awesome. :+1:

PR #2753, opened 2015-08-24
Sanitise mail server log lines before processing them

### LINKED ISSUE #2653: Sanitise lines of email log files whilst loading them
On the Panama site we've encountered a particular log file which causes errors when we try to parse it, as there's a line which is invalid UTF-8. (It looks to contain a bunch of escape characters perhaps).

After a bit of discussion, it seems like the best thing to do is to try to sanitise each line as we parse the files.

stevenday:
Fixes #2653

I'm not sure why I had to do ceed036 - if there's something I need to change
on my dev vm instead, let me know.

It would be nice to test the postfix version as well as the exim, but I only
had an example from the latter, any pointers on how to convert this would be
appreciated...


stevenday:
@crowbot - I can't use `load_file_fixture` as that reads the file into a string, rather than just returning the file handle, but I've switched it to use `file_fixture_name` from the same support module. I've also rebased out the require line. (A complete bundle reinstall didn't do it for me, but I'll trust Travis if it's wrong as I get similar issues with other branches/tags)

@garethrees - I've updated the test and the fixture so that it checks the content of the offending stored line. It's not quite as simple as your suggestion because the method doesn't return things directly, it creates records in the DB, but the effect is the same. Hopefully that's an improvement?

garethrees:
Looks good to me other than a few minor points

stevenday:
@garethrees - thanks for the comments, I think I've fixed them all in the latest fixup commit: 53c662d

I've also pushed a new commit to close the log file in the actual code (rather than the test) since it seemed a good idea to fix that at the same time...

garethrees:
@crowbot just needs a quick once over on the extra commits

garethrees:
Many thanks @stevenday!

PR #2762, opened 2015-09-02
Update some more deprecated syntax instances for rspec 3.

crowbot:
Replace `should` with `expect` as in 753c784fa9e568493819d830348f302886fc7bcd
and `any_instance.stub` with `allow_any_instance_of`


garethrees:
:+1:

PR #2764, opened 2015-09-02
Use regex that includes hostname for incoming mail pipe.

crowbot:
Otherwise mail for other domains matching the local part will
be directed into Alaveteli.


garethrees:
:+1:

PR #2765, opened 2015-09-02
Use more specific regular expressions in Postfix transports.

crowbot:
These transports are applied to all domains, not just the domains
handled by the server, so there's a risk of matching a local part
of a remote domain here. Matches #2764.


garethrees:
:+1:

PR #2770, opened 2015-09-04
[#2387] Refactor InfoRequest#receive

garethrees:
#2387

WIP refactoring work.

I'm reasonably happy with this as is, so making a PR for a first-pass review.

Flog score has dropped considerably:

```
# develop
130.0: InfoRequest#receive              app/models/info_request.rb:422
# 2387-refactor-info-request-receive
44.9: InfoRequest#receive              app/models/info_request.rb:422
```

I basically followed the ideas in Sandi Metz's [All the Little Things talk](http://confreaks.tv/videos/railsconf2014-all-the-little-things).

#### Could be better

- I couldn't find a good way of dynamically finding the `ResponseGatekeeper` because they need different initialisation arguments. I think we could get around this with more refactoring elsewhere. The main problem is needing to pass the `Mail` instance for the calculation. It would be better if we only needed the `InfoRequest` instance, because then we could always pass `self` and just ignore it if we don't need it.

```diff
- gatekeeper =
-   case allow_new_responses_from
-   when 'nobody'
-     ResponseGatekeeper::Nobody.new
-   when 'anybody'
-     ResponseGatekeeper::Anybody.new
-   when 'authority_only'
-     ResponseGatekeeper::AuthorityOnly.new(self, email)
-   else
-     raise ResponseGatekeeper::UnknownResponseGatekeeperError,
-           "Unknown allow_new_responses_from '#{ allow_new_responses_from }'"
-   end
+ gatekeeper = ResponseGatekeeper.for(allow_new_responses_from, self)
```

Doing this drops the flog score to `37.6`

- Still need to initialise a `ResponseRejection` with lots of things
- Lots of order-dependent work in `InfoRequest#create_response`, passing ids around. Should be able to use `build_association` etc. Just seems a bit yuck, but at least its isolated now.
- Still need hack around the `SpamChecker` to figure out if the current `InfoRequest` is the holding pen.

#### Todo

- [x] Move extracted classes to their own files
- [x] Add specs for extracted classes
- [x] Minor syntax tidying (Line length, maybe use symbol for spam_action)

### DIFF
--- a/app/models/info_request/response_gatekeeper/base.rb
+++ b/app/models/info_request/response_gatekeeper/base.rb
@@ -1,0 +1,22 @@
+# -*- encoding : utf-8 -*-
+class InfoRequest
+  module ResponseGatekeeper
+    class Base
+      attr_reader :info_request, :allow, :reason
+
+      def initialize(info_request)
+        @info_request = info_request
+        @allow = true
+        @reason = nil
+      end
+
+      def allow?(email)
+        allow
+      end
+
+      def rejection_action
+        info_request.handle_rejected_responses
+      end
+    end
+  end
+end
--- a/app/mailers/request_mailer.rb
+++ b/app/mailers/request_mailer.rb
@@ -194,10 +194,12 @@ class RequestMailer < ApplicationMailer
   # actual original mail sent by the authority in the admin interface (so
   # can check that attachment decoding failures are problems in the message,
   # not in our code). ]
-  def self.receive(raw_email)
-    logger.info "Received mail:\n #{raw_email}" unless logger.nil?
-    mail = MailHandler.mail_from_raw_email(raw_email)
-    new.receive(mail, raw_email)
+  def self.receive(raw_mail)
+    ActiveSupport::Notifications.instrument("receive.action_mailer") do |payload|
+      mail = MailHandler.mail_from_raw_email(raw_mail)
+      set_payload_for_mail(payload, mail)
+      new.receive(mail, raw_mail)
+    end
   end
 
   # Find which info requests the email is for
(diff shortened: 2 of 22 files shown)


app/models/info_request/response_gatekeeper/base.rb
```diff
@@ -0,0 +1,22 @@
...
+      def allow?(email)
+        allow
+      end
+
+      def rejection_action
+        info_request.handle_rejected_responses
+      end
+    end
+  end
+end
```
garethrees: Fixed in https://github.com/mysociety/alaveteli/commit/6b7fb04f9afb2546594b08cd34bbc672579367fb

garethrees:
Ready for review after build pass.

I think the thing still bugging me is figuring out whether the email is rejected (basically `#accept_incoming?`). Passing around the `Mail` instance and `raw_email_data` string is pretty clumsy. If we had one point at which we converted to a `Mail` instance I think we could clean things up. Here's a sketch of what I'd ideally like to see:

```ruby

def receive(raw_email, force = false)
  rejection = filter_through_gatekeepers(raw_email) unless force
  
  if rejection
    # Return false so that we can do:
    #
    #   if @info_request.receive(raw_email)
    #     # something
    #   else
    #     # handle failure
    #   end
    rejection.reject
    false
  else
    create_response(raw_email)
  end
end

def filter_through_gatekeepers(raw_email)
  gatekeepers.find do |gatekeeper|
    # Return a ResponseRejection if rejected, false if OK
    gatekeeper.reject(raw_email)
  end
end

def gatekeepers
  @response_gatekeeper = ResponseGatekeeper.for(allow_new_responses_from, self)
  @spam_checker = ResponseGatekeeper::SpamChecker.new
  [@response_gatekeeper, @spam_checker]
end
```

garethrees:
Just had a look at the new flog score:

```
# develop
130.0: InfoRequest#receive              app/models/info_request.rb:421
# 2387-refactor-info-request-receive
12.8: InfoRequest#receive              app/models/info_request.rb:421
```

garethrees:
We should be able to call [`raw_source`](https://github.com/mikel/mail/blob/master/lib/mail/message.rb#L377-L387) on the mail object rather than messing around with passing them around individually. That would make life much easier.

PR #2772, opened 2015-09-04
Explicitly force a mail reparse when we set the mail or raw_email.

crowbot:
Without forcing a mail reparse, these tests would fail when the
order of tests meant that there was already data stored on the
message from the usual raw_email.


garethrees:
:+1:

PR #2778, opened 2015-09-08
Fix theme_display_status call.

crowbot:
theme_display_status is a class level method.


garethrees:
Hmm hound… bit over eager…

PR #2781, opened 2015-09-08
Make sure fixtures are available in xapian for spec.

crowbot:
Otherwise create_request will fail due to the absence of
Geraldine Quango e.g. https://travis-ci.org/mysociety/alaveteli/jobs/79274938


garethrees:
:+1:

Issue #2697, opened 2015-07-31
Increase security of superuser accounts


RichardTaylor:
Superusers can access material hidden from public view which can include bulk accidental releases of sensitive personal information; they can also access users' email addresses. 

Consider eg. 

* Keys / tokens instead of or in addition to passwords.
* Preventing password reset email links being sent for superusers.

This was a suggestion made at an in-person discussion among WhatDoTheyKnow volunteers.

crowbot:
http://stackoverflow.com/questions/2848134/what-are-best-practices-for-securing-the-admin-section-of-a-website

garethrees:
Some authentication libraries worth looking in to:

- Warden
- Omniauth
- Devise

Should also look in to better authorisation. Looks like a [brief attempt](https://github.com/mysociety/alaveteli/blob/develop/lib/ability.rb) was made.

- CanCanCan (forked CanCan)
- [Six](https://github.com/randx/six) was popular for a time. Not sure on maintenance.

Maybe check out some gov.uk apps?

Haven't watched yet but might be relevant http://confreaks.tv/videos/railsconf2015-passwords-are-not-enough

garethrees:
A general overview of current state of rails security http://confreaks.tv/videos/railsconf2015-the-world-of-rails-security

garethrees:
https://github.com/thoughtbot/clearance

garethrees:
GitLab 2-factor https://about.gitlab.com/2015/05/22/gitlab-7-11-released/ (uses Google Authenticator and [devise-otp](https://github.com/wmlele/devise-otp))

garethrees:
> Federate Rights Elevation - e.g. email other admins when any admin's rights are upgraded. [JoeGeeky]

I really like the federated rights idea, but sites may have only one admin user.

garethrees:
https://github.com/heapsource/active_model_otp

garethrees:
https://rubygems.org/gems/rotp

crowbot:
http://www.pandurang-waghulde.com/2014/06/client-side-ssl-certificate.html

crowbot:
http://blog.meldium.com/home/2013/8/23/screw-up-two-factor-authentication

crowbot:
[...]

PR #2782, opened 2015-09-08
[#1902] change request spam

### LINKED ISSUE #1902: Improve spam handling on /change_request/new
> Any chance of adding a capcha here?
> https://www.whatdotheyknow.com/change_request/new
> 
> We get quite a bit of spam.
> 
> regards,
> 
> John

---

> Hi John,
> 
> We added a Honeypot fairly recently [1].
> 
> Are you still getting a fair amount to this form?
> 
> Cheers,
> 
> [1] https://github.com/mysociety/alaveteli/pull/1799

---

> We get spam via that form most days. As an example there were 4 messages yesterday and has already been one message sent through today.  There is a similar issue with the update email address form (see below). We get spam via that form most days as well (there have already been two sent today).
> 
> Regards,
> 
> Helen Cross

---

> Just to add that IMO the single quickest and easiest improvement to this
> would be adding a way to just delete the spam requests
> (https://github.com/mysociety/alaveteli/issues/1457).
> 
> At the moment legitimate ones are tangled up in them making it hard to
> check the backlog.
> 
> Obviously actually stopping the spam would be even better, but that may
> not be quite so simple and in any case we would want a way to clear out
> the existing spam.
> 
> Cheers,
> 
> Ganesh

garethrees:
Fixes #1902 

- Uses the latest reCAPTCHA; during dev (after the gem update) it just updated me. It worked for a while, then changed to the new reCAPTCHA and complained my keys were not valid for the domain. I generated new keys and continued with the new reCAPTCHA as normal.

screen shot 2015-09-08 at 18 23 53

- I was going to tidy up reCAPTCHA use in `UserController`, but:
  - the spec coverage is low, so refactoring was hard
  - we might end up changing a lot of that in #2697


garethrees:
Needs a changelog

garethrees:
Sorry, I think I rushed this one a bit. I've created a new PR (https://github.com/mysociety/alaveteli/pull/2789) as it was easier to re-write some of the commits now that we're just rendering the captcha for any non-logged in users.

PR #2785, opened 2015-09-09
Add shortcut to Holding Pen in admin nav

garethrees:
We make a big deal about the Holding Pen in the documentation but its
not really used as a first-class concept in the admin interface. This
makes it a bit more prominent and easier to find.


garethrees:
I think the holding pen deserves some sort of "page" of its own as its a key concept in the app. In that case I think a better solution might be to reword the summary as mentioned and create a `HoldingPen` controller; which basically renders an auto-expanded list of the messages that need moving,  and maybe some help info for the admin. Obviously not quite as quick as this PR.

Issue #2662, opened 2015-07-15
Extract quoted section removal to library and make quoted section markers translatable


garethrees:
@petterreinholdtsen submitted https://github.com/mysociety/alaveteli/pull/2654 to add a quoted section marker in their language.

I don't see the benefit in continually adding additional strings to this method, as many of them will be irrelevant for a given install.

The particular markers are theme-specific, so we should supply an interface for them to be customised from a theme. This would have the benefit of reducing the performance cost of processing messages.

A nice interface to add markers would be something along the lines of:

``` ruby
Alaveteli::QuotedSectionRemoval.end_of_message_strings << '''----*\s*Opprinnelig melding\s*----*'''
```

Usage of the library in alaveteli core should look something like:

``` ruby
class IncomingMessage
  def self.remove_quoted_sections(text, replacement = "FOLDED_QUOTED_SECTION")
   MessageFilter::RemoveQuotedSection.new(text, replacement).apply
  end
end
```

There also looks like some special handling for [other UK-specific stuff](https://github.com/petterreinholdtsen/alaveteli/blob/master/app/models/incoming_message.rb#L275-L289).

It would be awesome if you could build up a filter chain for sanitising your messages:

``` ruby
# config/initializers/my_theme_message_filtering.rb

IncomingMessage.filter_chain = [ RemoveLotusQuotes,
                                 RemoveQuotedSection,
                                 RemoveBBCFurtherCommsConsent,
                                 HideEmailAddresses ]
```

garethrees:
This is a similar problem to #2551.

garethrees:
https://github.com/Ibsciss/ruby-middleware

petterreinholdtsen:
I revisited #2654 today, to see what I could to do make the message view nicer on Mimes brønn, but reading through this issue again leave me mystified about how to solve it.  Is there some documentation somewhere explaining how to customize the original_message setting?

garethrees:
Hi Petter,

You can override the full method in your theme.

```ruby
# THEME_ROOT/lib/model_patches.rb

# -*- encoding : utf-8 -*-
Rails.configuration.to_prepare do
  InfoRequest.class_eval do
    def self.remove_quoted_sections(text, replacement = "FOLDED_QUOTED_SECTION")
      # method body omitted for demonstration
    end
  end
end
```

[...]

Issue #2551, opened 2015-06-11
Make default text masks localisable/configurable


schlos:
Alaveteli v0.20 (ImamoPravoZnati.org)

I've noticed that phone numbers from signature are not detected properly for redaction, but email address is redacted.

Could you share rules how you detect phone numbers? 
Do I need to include Croatian country code (+385 ...) on rule list?

Example where phone number from signature is not redacted:
http://imamopravoznati.org/request/naknada_za_okolis#incoming-495

crowbot:
@schlos - we only automatically redact [mobile phone numbers, email addresses, and alaveteli log in links]( https://github.com/mysociety/alaveteli/blob/develop/lib/alaveteli_text_masker.rb#L122-L129) in text type files and [email addresses in binary type files](https://github.com/mysociety/alaveteli/blob/develop/lib/alaveteli_text_masker.rb#L80). At the moment these patterns are hard-coded, but there are also [censor rules](http://alaveteli.org/docs/running/admin_manual/#hiding-certain-text-from-a-request-using-censor-rules) which you can create for a request or user in order to remove certain bits of information.

schlos:
Could I suggest to add this to configuration to allow localization - translation of words "mobile", "telephone" etc? Or even site wide redaction rules which would allow us to define localized phone redaction rules.
Thanks!

crowbot:
Yep, seems like a reasonable request. I'm changing the title of this ticket to reflect that.

garethrees:
```
 TEXT +-----+   
            |   
            |   
+-----------v--+
| InfoRequest  |
| Masks        |
+-----------+--+
            |   
+-----------v--+
| Global       |
| Masks        |
+-----------+--+
            |   
+-----------v--+
| InfoRequest  |
| CensorRule   |
+-----------+--+
            |   
+-----------v--+
| Global       |
| CensorRule   |
+-----------+--+
            |   
+-----------v--+
| PublicBody   |
| CensorRule   |
+-----------+--+
            |   
+-----------v--+
| User         |
| CensorRule   |
+--------------+
```

garethrees:
Action for now is to make all censor rule types editable through admin interface

Issue #2761, opened 2015-08-25
Facility to change order of censor rules


kingqueen3065:
It would be great to be able to change the order of censor rules. 
To give an example: when somebody called Fred Bloggs wants their account anonymising, I generally start by putting in a censor rule for "Fred Bloggs" (without the quotes), replacing with [name removed]. If I then find somewhere further in, "Dear Fred", I have to add a censor rule for "Fred". But as this is processed before the one for "Fred Bloggs", we then get "[name removed] Bloggs". It would be great if we could rearrange the rules to sort this sort of problem.

RichardTaylor:
Just adding a +1 to this. @kingqueen3065 has explained the issue very well. The case described of trying to censor variants of a name is the common one on WhatDoTheyKnow so any way to do this more intelligently would be great.

HelenWDTK:
This issue has been automatically closed due to a lack of discussion or resolution for over 12 months.
Should we decide to revisit this issue in the future, it can be reopened.

PR #2787, opened 2015-09-11
[#2551] Text Mask Customisation

garethrees:
#2551

I think I'm happy enough with where this is going for an initial review.

The implementation idea is very similar to rack middleware. We want to build a list of processors and pass some text through each, applying each processor on the way.

I've used the ruby-middleware library [1] as a foundation for building these middleware stacks. It's a forked version of Mitchell Hashimoto's middleware library [2, 3], which is in use in Vagrant.

The current filter chain has been retained. We'll want to inject the middleware stacks where shown. The code to do this is still a TODO (see discussion below).

```
                                TEXT +-----+                  
                                           |                  
                                           |                  
                               +--------------------+ CODE    
                                           |                  
                                           |                  
                               +-----------v--+               
INJECT STACK +-----------------> InfoRequest  |               
                               | Masks        |               
                               +-----------+--+               
                                           |                  
                               +-----------v--+               
INJECT STACK +-----------------> Global       |               
                               | Masks        |               
                               +-----------+--+               
                                           |                  
                                           |                  
                               +--------------------+ DATABASE
                                           |                  
                                           |                  
                               +-----------v--+               
                               | InfoRequest  |               
                               | CensorRule   |               
                               +-----------+--+               
                                           |                  
                               +-----------v--+               
                               | Global       |               
                               | CensorRule   |               
                               +-----------+--+               
                                           |                  
                               +-----------v--+               
                               | PublicBody   |               
                               | CensorRule   |               
                               +-----------+--+               
                                           |                  
                               +-----------v--+               
                               | User         |               
                               | CensorRule   |               
                               +--------------+               

```

There are some more related links from the thinking on #2662 [4]

[1] https://github.com/Ibsciss/ruby-middleware
[2] https://github.com/mitchellh/middleware
[3] http://goo.gl/sWCB35
[4] https://github.com/mysociety/alaveteli/issues/2662#issuecomment-132635228

The gist of this is that middleware stacks can be built up by supplying classes that respond to `#call` in to a `Middleware::Builder`.

``` ruby
stack = Middleware::Builder.new do |b|
  b.use SomeTextMask
  b.use AnotherTextMask
  b.use MoreMaskingTextMask
end

stack.call('Some text')
# => '[ITS ALL REDACTED]'
```
## What's been implemented so far?
1. A proof of concept `TextMasks` module, including a foundation `RegexpMasker` and a more real-world `EmailAddressMasker`. The `EmailAddressMasker` inherits from `RegexpMasker`, supplying it with a hard-coded email `Regexp`. I see us mostly creating domain-specific masks based on the `RegexpMasker`. For example:
   - `MobileNumberMasker`
   - `LandlineMasker`
   - `FaxNumberMasker`
   - `AlaveteliLoginLinkMasker`
2. A non-destructive `AlaveteliTextMasker.apply_text_masks` that compiles a middleware stack from the default text masks (now also basic middleware) and the `InfoRequest` masks (also basic middleware). The censor rules are applied after the middleware stack.
3. Deprecated lots of destructive `apply_masks!` methods and implemented non-destructive versions.
## What's still to do?
1. Implement the domain specific middlewares
2. Add the ability for a re-user to modify the middleware stacks (discussion below)
3. Use middleware stacks for binary masks
4. Allow censor rules to be injected in to the middleware stack
## Implementing configurable stacks

Here's a sketch of how I think the interface should look for a re-user:

``` ruby
# GLOBAL MASKS
# ============

# I want to add a global text mask to the end of the stack
AlaveteliTextMasker.global_text_masks.use SomeCustomTextMasker

# I want to add a global text mask in the middle of the stack
AlaveteliTextMasker.
  global_text_masks.
    insert_after EmailAddressMasker, SomeCustomTextMasker

# Option A: I want to replace the global text mask stack
custom_stack = Middleware::Builder.new do |b|
  b.use SomeCustomTextMasker
end

AlaveteliTextMasker.global_text_masks = custom_stack

# Option B (Preferred):  I want to replace the global text mask stack
# Not quite sure how to implement this, but I think it should be easy enough
# once I've read up on blocks and yields.
AlaveteliTextMasker.global_text_masks do |b|
  b.use SomeCustomTextMasker
end

# INFO REQUEST MASKS
# ==================

# Much the same as the global masks…

# I want to add an InfoRequest text mask to the end of the stack
AlaveteliTextMasker.text_masks.use SomeCustomTextMasker

# I want to add a InfoRequest text mask in the middle of the stack
AlaveteliTextMasker.text_masks.insert_after EmailAddressMasker, SomeCustomTextMasker

# Option B:  I want to replace the global text mask stack
InfoRequest.text_masks do |b|
  b.use SomeCustomTextMasker
end
```

I think this should actually be fairly simple. Something like:

``` ruby
class AlaveteliTextMasker
  def self.default_global_text_masks
    ::Middleware::Builder.new do |m|
      # the defaults…
    end
  end

  def self.global_text_masks=(middleware_stack)
    @global_text_masks = middleware_stack
  end

  def global_text_masks
    @global_text_masks ||= self.class.default_global_text_masks
  end
end
```
## Adding Censor Rules to the stack

Again, this should be fairly easy now the framework is in place:

``` ruby
# CENSOR RULES
# ============

# Censor rules can basically be converted to an
# AlaveteliTextMasker::TextMasks::RegexpMasker, as they just use a :regexp and
# a :replacement. We can inject these in to the middleware stack.
class CensorRule < ActiveRecord::Base
  def to_regexp_masker_options
    { :regexp => to_replace, :replacement => replacement }
  end
end

# Then we can build a middleware stack for them
class InfoRequest < ActiveRecord::Base
  def applicable_censor_rules
    ::Middleware::Builder.new do |b|
      # Do this for each type of rule
      censor_rules.each do |rule|
        b.use AlaveteliTextMasker::TextMasks::RegexpMasker,
              rule.to_regexp_masker_options
      end
    end
  end
end

# These can be added to the stack as noted in the TODO
class AlaveteliTextMasker
  def apply_text_masks(text, options = {})
    stack = Middleware::Builder.new do |s|
      s.use options[:masks] if options[:masks]
      s.use default_text_masks(true)
      # TODO: Add censor rules in to stack:
      # s.use options[:censor_rules] if options[:censor_rules]
    end

    stack.call(text)
  end
end
```
## Is this the right approach?

I think so. It makes it easy for us to add new masks (just add a class and `use` it in the middleware) and likewise for re-users.

I think it should make it a lot easier to figure out what will be applied to some text. Here`s a sketch of how I think we can do that:

``` ruby
# With a bit of re-jigging to make we should be able to call stack.inspect [1]
# to visualise the full middleware stack. After that it looks pretty trivial
# [2] to print the stack to screen.
#
# [1] https://github.com/Ibsciss/ruby-middleware#debug
# [2] http://git.io/vZgaF
class AlaveteliTextMasker
  def stack(opts = {})
    Middleware::Builder.new do |s|
      s.use options[:masks] if options[:masks]
      s.use default_text_masks(true)
      s.use options[:censor_rules] if options[:censor_rules]
    end
  end

  def apply_masks(text)
    stack.call(text)
  end
end

class IncomingMessage < ActiveRecord::Base
  def apply_masks(text, options = {})
    text_masker.apply_masks(text)
  end

  def text_masker(options = {})
    @text_masker ||= AlaveteliTextMasker.new(options)
  end
end
```

``` ruby
# Some erb view
@incoming_message.text_masker.inspect
# => Prints the middleware stack (including Censor Rules) for the admin to
#    figure out what will happen
```

There are still a few other places where `apply_masks!` is called in an `IncomingMessage` instance – ideally we should call it once and then always use that result.

There are other places where we run the text through a filter. The main thing that comes to mind are the quoted section methods (#2662). We could address these in the same way – add a `TextMask` class, add it to or build a middleware and then process the text through it. The same patterns for adding custom filters could be used.


garethrees:
I'll mull this over – some good points.

> I don't think we really have any evidence of demand for reordering text masks from reusers, or for removing them

Not masks, but censor rules (though I'm beginning to think of them as the same thing) https://github.com/mysociety/alaveteli/issues/2761

garethrees:
I'm also thinking that we should have one processing pipeline from "raw email" to "text on the request page". We might want to implement linked requests (e.g. [Stack Overflow](http://blog.stackoverflow.com/2010/04/new-linked-posts/) which transforms an Alaveteli URL in to a link with the request title, and dumping a list of linked requests in the sidebar. That should happen in the pipeline, along with any other modifications we make to the raw body.

garethrees:
> I think once we decide what we think a desirable end state for this code is, we should probably think a bit about whether it makes sense to break up into some smaller steps for implementation.

Yeah definitely. I also think I/we should spend a bit of time looking in to what other changes are applied to the raw email text before it makes it to the end result, and whether they might influence what we do here.

garethrees:
We can take two viewpoints to how we process emails through a pipeline:
  - (A) Do as much as possible before save
  - (B) Do the minimum to sensibly format before save

I'm leaning towards **B**.

The current situation is that we don't do _any_ processing (that I could see) before saving an incoming email. All the processing is performed lazily and cached.

I think we should do some pre-processing to store the message in a sensible format – things like encoding, stripping and compacting whitespace.

At this point we have a friendly version of the message to work with.

#### Masks vs Censor Rules

> I think I would say at the moment 'censor rules are a kind of text mask that is specific to the install'.

I'm thinking of them less like that and more as (a) how changeable they are and (b) who decides to add them. 

A text mask can be specific to the install (e.g. telephone number format) set by the programmer, but that doesn't change much. A Censor Rule on the other hand might change often, as the admin can just go and edit them in the DB.

I think of quote folding as a "mask", in that its decided by the application, rather than the admin.

There's totally a grey area between a mask and a _global_ censor rule though.

---

So working on the basis that we have an un-censored (but cleaned) version of the message in the database, we'll need to apply redaction on the fly.

I was originally a bit against this, but dealing with changes becomes much harder if our master record already redacts email addresses etc.

I think the best way to do this is a combination of role-based and format-based presenters.

For example, for an admin viewing a message we might apply the `AdminMessagePresenter` and pass that in to a `HtmlMessagePresenter`. The role-based presenter would apply appropriate redactions and masks, and the format-based presenter would take care of …the formatting (making links clickable, etc).

In another example, we might pass the message through a `SearchIndexMessagePresenter` and a `PlaintextMessagePresenter`.


```
                                ^                                                                                                        
                                |                                                                                                        
                                |                                                                                                        
                 +--------------+---+                                                                                                    
                 |                  |                                                                                                    
EMAIL IN  +------>  REJECT & HANDLE |                                                                                                    
                 |  OR ACCEPT       |                                                                                                    
                 |                  |                                                                                                    
                 +--------+---------+                                                                                                    
                          |                                                                                                              
                          |                                                                                                              
                 +--------v---------+                                                                                                    
                 |                  |                                                                                                    
                 |  SAVE RAW EMAIL  |                                                                                                    
                 |                  |                                                                                                    
                 +--+---------------+                                                                                                    
                    |                                                  +                          +                          +           
                    |                                                  |                          |                          |           
                    |    +------------+    +------------+   +------+   |   +------------------+   |   +------------------+   |           
                    |    |            |    | [1]        |   |      |   |   |                  |   |   |                  |   |           
                    +---->EXTRACT     +----> PRE        +-----SAVE +-------> ROLE PRESENTER   +-------> FORMAT PERSENTER +-------> OUTPUT
                    |    |BODY        |    | PROCESSING |   |      |   |   |                  |   |   |                  |   |           
                    |    +------------+    +------------+   +------+   |   +-+--------------^-+   |   +-+--------------^-+   |           
                    |                                                  |     |              |     |     |              |     |           
                    |                                                  |     | +---------+  |     |     | +---------+  |     |           
                    |                                                  |     | | [2]     |  |     |     | |         |  |     |           
                    |    +------------+    +------------+              |     +-> GLOBAL  |  |     |     +-> SANITIZE|  |     |           
                    |    |            |    |            |              |       | MASKS   |  |     |       | HTML    |  |     |           
                    +---->EXTRACT     +----> ETC…       |              |       +--+------+  |     |       +--+------+  |     |           
                         |ATTACHMENTS |    |            |              |          |         |     |          |         |     |           
                         +------------+    +------------+              |       +--v------+  |     |       +--v------+  |     |           
                                                                       |       | [3]     |  |     |       |         |  |     |           
                                                                       |       | INSTANCE|  |     |       | ETC     +--+     |           
                                                                       |       | MASKS   |  |     |       |         |        |           
                                                                       |       +--+-------  |     |       +---------+        |           
                                                                       |          |         |     |                          |           
                                                                       |       +--v------+  |     |                          |           
                                                                       |       | [4]     |  |     |                          |           
                                                                       |       | CENSOR  +--+     |                          |           
                                                                       |       | RULES   |        |                          |           
                                                                       |       +---------+        |                          |           
                                                                       |                          |                          |           
                                                                       +                          +                          +           

[1] Things like stripping and compacting whitespace, excessive newlines etc
[2] Replacing email addresses, phone numbers, folded sections
[3] Anything related to request attributes/status
[4] Apply all censor rules
```

Where and how we cache these, I'm not 100% sure. I expect initially we'll use the existing table columns as much as possible. If I was doing this from scratch I'd be thinking Memcached for anything after the `SAVE` point.

I think I'd still retain the idea of middleware stacks – they'd just be owned by the presenters.

Below is a list of the processing that we do on messages. I've focused more on the messages rather than attachments.

Now that we have some sort of list, I think the next step is to organise them (design-wise, not move code) in to the correct place in the above process and figure out where there may need to be customisation points.

```
# INCOMING MAIL

script/mailin
RequestMailer.receive # Turns the raw email String in to a Mail::Message
RequestMailer#receive
InfoRequest#receive
InfoRequest#create_response # Saves a RawEamil – should we change it to CRLF?
                            # Saves an IncomingMessage

# Doesn't process anything up to this point

---

# MAIL ALREADY SAVED IN DB

# These apply to a string argument – used elsewhere
InfoRequest#apply_censor_rules_to_text!
InfoRequest#apply_censor_rules_to_binary!


## INCOMING MESSAGES

# Strips uudecode parts
# Applies text masks
# Saves folded and unfolded versions
IncomingMessage#_cache_main_body_text 
  IncomingMessage#extract_attachments!   # TODO: not sure what happens here
  IncomingMessage#apply_masks            # 
  IncomingMessage#remove_lotus_quoting   # 3 matchers
  IncomingMessage#remove_quoted_sections # 9 matchers

# TODO Does some html sanitization?
# Adds a note if text is scrubbed
# Converts CRLF to LF
# Compresses spaces
IncomingMessage#_convert_part_body_to_text

# simplify_angle_bracketed_urls
# escapeHTML
# make_clickable
# Converts phone/email replacements added by masks in to links
# TODO Collapses folded quoted sections – not sure what happens here. Why is it
# different to e.g. remove_quoted_sections
# A bit more processing collapsing and linking to folded quoted sections
# strip!
# Replace newlines with <br> tags
# Remove excessive consecutive <br> tags
# returns a html_safe string
IncomingMessage#get_body_for_html_display  

# Replace the string "FOLDED_QUOTED_SECTION" with a space
# strip!
IncomingMessage#get_body_for_quoting

IncomingMessage#get_attachment_text_full
  IncomingMessage#_get_attachment_text_internal # extracts attachment text
                                                # make attachment text utf8
  apply_masks!
  Creates a snippet version

## OUTGOING MESSAGES

OutgoingMessage#body
  clean_text          # Remove excessive consecutive <br> tags
  apply censor rules
```

garethrees:
Waiting until we talk in person on 2015-10-16 before we go any further on this.

garethrees:
https://github.com/jch/html-pipeline

garethrees:
This PR is 2 years old, so closing as stale.

PR #2788, opened 2015-09-14
[#2569] use classes in markup

### LINKED ISSUE #2569: Rewrite markup to use classes as well as IDs
- Update markup - add classes as well as IDs
- Update core styles to work from those classes
- Update alavetelitheme to use the classes
- Update whatdotheyknow-theme to use the classes

wrightmartin:
Fixes #2569

This adds nothing new, but adds classes to the markup so in the future we can make theming easier

### DIFF
--- a/app/views/general/_frontpage_search_box.html.erb
+++ b/app/views/general/_frontpage_search_box.html.erb
@@ -5,9 +5,9 @@
   :number_of_requests => number_with_delimiter(InfoRequest.visible.count),
   :number_of_authorities => number_with_delimiter(PublicBody.visible.count)) %>
 </h2>
-<form id="search_form" method="get" action="<%= search_redirect_path %>">
+<form id="search_form" class="search_form" method="get" action="<%= search_redirect_path %>">
   <div>
-    <input id="query" type="text" size="30" name="query" title="type your search term here" >
+    <input id="query" class="query" type="text" size="30" name="query" title="type your search term here" >
     <input type="submit" value="<%= _('Search') %>">
   </div>
 </form>
--- a/app/views/general/_responsive_header.html.erb
+++ b/app/views/general/_responsive_header.html.erb
@@ -16,10 +16,11 @@
       <%= render :partial => 'general/locale_switcher' %>
 
       <% if ! (controller.action_name == 'signin' or controller.action_name == 'signup') %>
-        <div id="logged_in_bar">
-          <div id="logged_in_links">
+        <div id="logged_in_bar" class="logged_in_bar">
+          <div id="logged_in_links" class="logged_in_links">
             <% if @user %>
-              <span class="greeting"><%= _('Hello, {{username}}!', :username => @user.name) %></span>
+              <span class="greeting">
+              <%= _('Hello, {{username}}!', :username => @user.name) %></span>
               <%= link_to _("My requests"), show_user_requests_path(:url_name => @user.url_name) %>
               <%= link_to _("My profile"), show_user_profile_path(:url_name => @user.url_name) %>
               <%= link_to _("My wall"), show_user_wall_path(:url_name => @user.url_name) %>
--- a/app/views/general/_responsive_header.html.erb
+++ b/app/views/general/_responsive_header.html.erb
@@ -1,9 +1,9 @@
 <div class="only-show-for-print">
   <p class="print-information">Printed from <%= request.original_url %> on <%= Time.zone.now.to_formatted_s(:long)  %></p>
 </div>
-<div id="banner">
-  <div id="banner_inner">
-    <div id="banner_content">
+<div id="banner" class="banner">
+  <div id="banner_inner" class="banner_inner">
+    <div id="banner_content" class="banner_content">
       <div class="banner_site-title">
         <h1><%= link_to site_name, frontpage_path, :id => 'logo', :class => 'site-title__logo' %></h1>
       </div>
--- a/app/views/help/_sidebar.html.erb
+++ b/app/views/help/_sidebar.html.erb
@@ -1,4 +1,4 @@
-<div id="right_column_flip">
+<div id="right_column_flip" class="right_column_flip">
 <h2>Help pages</h2>
 <ul class="no_bullets">
   <li><%= link_to_unless_current "Introduction", help_about_path %></li>
@@ -11,7 +11,7 @@
   <li><%= link_to_unless_current "Advanced search", advanced_search_path %></li>
 </ul>
 
-<h2 id="contact">Contact us</h2>
+<h2 id="contact" class="contact">Contact us</h2>
 <p>If your question isn't answered here, or you just wanted to let us know
   something about the site, <a href="<%= help_contact_path %>">contact&nbsp;us</a>.
 </p>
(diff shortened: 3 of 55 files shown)


garethrees:
The code in this PR is good and we could deploy it as is (the minor comments I've made so far could be ignored, really).

I kinda think that we're not really buying ourselves anything though. I interpreted #2569 to be about making it easy to style the core components on the site. The classes added here probably make specificity issues a little easier, but I still fell that it looks a bit of a mish-mash.

I was thinking that we'd look for common elements around the site and do some work to make them in to reusable components. An example of this is the request summary that you see when you filter requests. Its used in a few different places, but the classes applied aren't really easy to re-use.

screen shot 2015-09-16 at 15 40 16


```erb
<!-- Old -->
<div class="request_listing">
  <div class="request_left">
    <span class="head">
      <!-- request title -->
    </span>

    <div class="requester">
      <!-- event description -->
    </div>

    <span class="bottomline icon_<%= event.info_request.calculate_status %>">
      <strong>
        <!-- request status -->
      </strong><br>
    </span>
  </div>

  <div class="request_right">
    <span class="desc">
      <!-- event excerpt -->
    </span>
  </div>
</div>

<!-- BEM -->
<div class="request-listing">
  <h1 class="request-listing__request-title">
    <!-- request title -->
  </h1>

  <div class="request-listing__request-event-description">
    <!-- event description -->
  </div>

  <div class="request-listing__request-status request-status--success">
    <!-- request status -->
  </div>

  <div class="request-listing__request-event-excerpt">
    <!-- event excerpt -->
  </div>
</div>
```

I think the BEM syntax here would give re-users a clear, documentable interface for styling the component.

I don't know if we're thinking that this kind of thing is a different piece of work, so happy to do this later, but I think it makes sense to do this before re-doing the themes after this PR only to need to do it again later after making modular components.

Totally open for discussion!

wrightmartin:
Actually the whole way through this is was trying to decide between just adding classes to fix the specificity problem, or breaking the whole thing down to components as you discussed above.

I think just switching IDs for classes is a low-hanging-fruit approach, it fixes the specificity problems, but it doesn't do much more. 

If we wanted to go further here I'd be up for that, but it will take a fair bit longer, we could call in a whole new piece of work for the future?

garethrees:
> I think just switching IDs for classes is a low-hanging-fruit approach, it fixes the specificity problems, but it doesn't do much more.

Yeah that makes sense.

> If we wanted to go further here I'd be up for that, but it will take a fair bit longer, we could call it a whole new piece of work for the future?

I definitely don't think we should add commits to this PR to achieve it. I think lets get this merged and then make extractions later. If we're doing that, then I think the class names should all be exactly the same as the IDs. At the moment a few differ, and I think that's really confusing if their only purpose is to fix specificity issues.

I think my main concern was re-implementing the themes twice, so we should have a discussion next sprint planning about where we want to go after this gets merged.

wrightmartin:
> At the moment a few differ, and I think that's really confusing if their only purpose is to fix specificity issues.

I'll go back and fix those ones, and we can resubmit

garethrees:
> I'll go back and fix those ones, and we can resubmit

Awesome, cheers. Could you use the auto-fixup format for the commit messages? e.g. if you're amending some code from "add classes to search bar", title the fixup commit "fixup! add classes to search bar" – that way its easier to figure out where the commits need to get squashed.

Also needs rebasing with develop. I'd do this before making further commits.

Cheers!

wrightmartin:
I'd like to pretend I was thinking about this outside of work hours, but do you think the refactored CSS should be part of this PR? eg. Swapping ```#footer``` for ```.footer``` and nothing else?

garethrees:
> I'd like to pretend I was thinking about this outside of work hours, but do you think the refactored CSS should be part of this PR? eg. Swapping ```#footer``` for ```.footer``` and nothing else?

Hmm, I think while its a conceptually similar change, this PR is pretty massive as it is. Maybe get this merged and then just do a new PR for changing the classes. There might be a few more :dragon: :dragon_face: :dragon: there.

app/views/general/_frontpage_search_box.html.erb
```diff
@@ -5,9 +5,9 @@
   :number_of_requests => number_with_delimiter(InfoRequest.visible.count),
   :number_of_authorities => number_with_delimiter(PublicBody.visible.count)) %>
 </h2>
-<form id="search_form" method="get" action="<%= search_redirect_path %>">
+<form id="search_form" class="search-form" method="get" action="<%= search_redirect_path %>">
```
garethrees: `search_form`, not `search-form`

app/views/general/_responsive_header.html.erb
```diff
@@ -1,10 +1,10 @@
...
   <p class="print-information">Printed from <%= request.original_url %> on <%= Time.zone.now.to_formatted_s(:long)  %></p>
 </div>
-<div id="banner">
-  <div id="banner_inner">
-    <div id="banner_content">
-      <div class="banner_site-title">
+<div id="banner" class="banner">
+  <div id="banner_inner" class="banner_inner">
+    <div id="banner_content" class="banner_content">
+      <div class="banner_site-title site-header__site-title">
```
garethrees: Extra class? Don't mind keeping the odd one if it really helps.

app/views/help/_sidebar.html.erb
```diff
@@ -1,6 +1,6 @@
-<div id="right_column_flip">
+<div id="right_column_flip" class="right_column_flip">
 <h2>Help pages</h2>
-<ul class="no_bullets">
+<ul class="no_bullets help__navigation">
```
garethrees: Typo?

app/views/request/list.html.erb
```diff
@@ -1,10 +1,10 @@
-<div id="header_left">
+<div id="header_left" class="header_left">
   <h1><%= @title %></h1>
   <%= render :partial => 'request/request_search_form',
              :locals => { :after_form_fields => render(:partial => 'request/request_filter_form') } %>
 </div>
 
-<div id="header_right" class="sidebar">
+<div id="header_right" class="header_right">
```
garethrees: Should retain the `sidebar` class here I think

app/views/request/new.html.erb
```diff
@@ -101,7 +101,7 @@
 
       <% unless @batch %>
         <% if @info_request.public_body.has_notes? %>
-          <div id="request_header_text">
+          <div id="request_header_text" class="request-header__text">
```
garethrees:
```diff
- request-header__text
+ request_header__text
```

app/views/user/show.html.erb
```diff
@@ -151,7 +151,7 @@
     <% if @xapian_requests %>
       <% if @xapian_requests.results.empty? %>
         <% if @page == 1 %>
-          <h2 class="foi_results" id="foi_requests">
+          <h2 class="foi_results foi_requests" id="foi_requests">
```
garethrees: don't need it twice

app/views/user/signchangepassword.html.erb
```diff
@@ -2,7 +2,7 @@
 
 <% raise "internal error" if not @user %>
 
-<div id="change_password">
+<div id="change_password" class="change-password">
```
garethrees:
```diff
- change-password
+ change_password
```

garethrees:
Couple of typos but looks good otherwise

wrightmartin:
Thanks @garethrees, I've corrected those mistakes, should be good to go now

Issue #2365, opened 2015-04-23
User reports adblock preventing button being shown


RichardTaylor:
A WhatDoTheyKnow.com user has [reported via Twitter](https://twitter.com/RichGreenhill/status/591139725806858241):
>The WDTK green “Follow” button is being hidden by AdBlock because of its HTML ID attribute: `<div id="follow_box">`

At WhatDoTheyKnow we regularly, but not in large numbers, get users saying they can't see the "submit" button too (and while in many cases this may be because they're being required to "preview" / correct an error) I'm wondering if adblocking is an issue this could be a possible cause too. 

This is really a bug with the adblock software not Alaveteli. 

Further clarification from the user on exactly what software they were using when they experienced this has been requested.

RichardTaylor:
Further details provided:

>AdBlock 2.29 (recently renamed BetaFish Adblocker) on Chrome for Linux. “#follow_box” is blacklisted in CSS…filter lists selected: AdBlock custom, EasyList, Antisocial, EasyPrivacy, Malware protection.

crowbot:
Fixed in 6884c3bea4c0573e05654e615fa656eb94498e72

PR #2798, opened 2015-09-16
Issues/remove visual button references

wrightmartin:
This removes all the ```link_button_green``` classes and replaces them with nicer, more contextual classes. 

This was smaller than I thought it would be.

### DIFF
--- a/app/views/public_body/show.html.erb
+++ b/app/views/public_body/show.html.erb
@@ -57,17 +57,17 @@
   <div class="authority__header__action-bar">
     <div class="action-bar__make-request">
       <% if @public_body.is_requestable? || @public_body.not_requestable_reason == 'bad_contact' %>
-        <%= link_to _("Make a request to this authority"), new_request_to_body_path(:url_name => @public_body.url_name), :class => "link_button_green" %>
+        <%= link_to _("Make a request to this authority"), new_request_to_body_path(:url_name => @public_body.url_name), :class => "link_button_green make-request__action" %>
       <% end %>
     </div>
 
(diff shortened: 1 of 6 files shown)


app/views/public_body/show.html.erb
```diff
@@ -57,17 +57,17 @@
...
 
     <div class="action-bar__follow">
       <div class="action-bar__follow-button">
         <% if @existing_track %>
-          <%= link_to _("Unsubscribe"), {:controller => 'track', :action => 'update', :track_id => @existing_track.id, :track_medium => "delete", :r => request.fullpath}, :class => "link_button_green" %>
+          <%= link_to _("Unsubscribe"), {:controller => 'track', :action => 'update', :track_id => @existing_track.id, :track_medium => "delete", :r => request.fullpath}, :class => "unsubscribe__action" %>
         <% else %>
           <div class="feed_link">
-            <%= link_to _("Follow"), do_track_path(@track_thing), :class => "link_button_green" %>
+            <%= link_to _("Follow"), do_track_path(@track_thing), :class => "follow__action" %>
```
garethrees: Will this get adblocked #2365 ?

garethrees:
I think we should retain `link_button_green` as well as the new class names for at least one major release. This PR should include some upgrade notes that the class is being removed, and to update any custom overrides to the new class names.

garethrees:
https://github.com/mysociety/alaveteli/commit/5a894a9a3bd0ac95e87f5acc6a273f06cd136afc is a good example of adding some upgrade notes

PR #2803, opened 2015-09-18
Set blackhole return-path for admin report mails.

crowbot:
The path should be checkable by SPF.


garethrees:
Looks good. Should this come with a changelog entry?

PR #2811, opened 2015-09-25
Only count visible requests when rendering the public body list

### LINKED ISSUE #2431: Discrepancy in number of requests when some are hidden
After setting "Prominence" of some requests to "requester_only", the statistics on numbers of requests diverge (see screenshot)
screenshot_2015-05-14_23-56-19

lizconlan:
Fixes #2431

### DIFF
--- a/app/models/info_request.rb
+++ b/app/models/info_request.rb
@@ -1170,14 +1170,15 @@ class InfoRequest < ActiveRecord::Base
   # an InfoRequest associated with the body:
   def update_counter_cache
     PublicBody.skip_callback(:save, :after, :purge_in_cache)
+    success_states = ['successful', 'partially_successful']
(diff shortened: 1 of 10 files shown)


lizconlan:
So, if I'm reading this right (on second attempt), the front page request count comes from this line...

https://github.com/mysociety/alaveteli/blob/develop/app/views/general/_frontpage_bodies_list.html.erb#L13

...which draws the value from `PublicBody#info_requests_count`, a precalculated db value which includes **all** the info_request, regardless of prominence. Hmm

lizconlan:
* "Additional fix" added
* controller spec repaired (although it could be removed, not sure if it's adding anything useful)
* view specs added for checking info_request counts on the front page and the public body list page

PR #2820, opened 2015-09-29
Issues/1467 improve authority list

wrightmartin:
This adds some CSS to allow a swapped column order, and swaps the column order of `public_body/list`


garethrees:
Looks great. Could you add a changelog entry with 509acb7 and add the template to the list of changed templates? :+1:

Issue #2056, opened 2015-01-05
add more information on censor rules


davewhiteland:
from admin via @crowbot 

Broken out from #1075

garethrees:
Current docs: http://alaveteli.org/docs/running/redaction/

schlos:
+1
Example for redacting simple one-liners is explained good enough.
Could you also add some examples how to redact multi-line information (like address)?

garethrees:
Hi @schlos, [this section](http://alaveteli.org/docs/running/redaction/#general-law) covers how we typically do multi-line redaction. "Multi-line redaction" would totally be a better header there!

crowbot:
You can also do request-specific multi-line redaction using censor rules though - it would be good to talk to the WDTK volunteers about how they use the rules, and more advanced examples

garethrees:
Try a post on https://groups.google.com/forum/#!forum/alaveteli-users / https://groups.google.com/forum/#!forum/alaveteli-dev :)

RichardTaylor:
Gareth's link in the [above comment](https://github.com/mysociety/alaveteli/issues/2056#issuecomment-93936843) doesn't appear to relate to admin created censor rules?

The documentation should include, with an example,

* if it's possible to censor from point "a" to point "b" over many lines?
* if it is possible to enter regexes with modifiers for m (multi-line) i (case insensitive) etc.  

The fact the example given in the current documentation proposes 
`[Aa]lice`
perhaps suggests there's no way to do something like:
`/alice/i` 
?

I'm wondering if there's a powerful feature here that isn't documented enough for site administrators to be able to use it. 

Documentation is at:
http://alaveteli.org/docs/running/redaction/

Some hints, examples and links to the full documentation could be added within the "New censor rule" form.

garethrees:
> Some hints, examples and links to the full documentation could be added within the "New censor rule" form.

Added what we have in https://github.com/mysociety/alaveteli/pull/5439

RichardTaylor:
The fact censor rules don't work on annotations should be made clear in in-line help text explaining annotations. 

On the subject of in-line help for admins, at 
https://github.com/mysociety/alaveteli/pull/5439

I wrote:

One key point for new admins that's missing here is that multi-line censor rules often won't work.

Line by line redaction is also more likely to work when a request is quoted in a response so should be encouraged. Actually the shorter the text to be redacted the more likely it is to catch occurrences of material in responses.

[...]

PR #2826, opened 2015-10-01
docs on hiding

### LINKED ISSUE #2062: document options for hiding messages and providing reasons for taking material down
running sites transparently: make this aspect of running a site more evident: cover all the (new) options to hide individual messages and provide public reasons for taking material down

Broken out from #1075

davewhiteland:
fixes #2062 -- unless Alavetelians think this is not strong enough on the "provide public reasons for taking material down" aspect?

Includes new overview of why/how things can be removed from Alaveteli, linking to existing instructions (which are in both requests or the admin_guide, depending on their nature) which themselves have been updated;  also expanded the Redaction page. Gareth had done sterling work on the Nicaraguan example (thanks for all the screenshots) but I felt it needed more context/qualification because it jumped straight into the Nicaraguan use-case. (Also did so style/rewording on that text).


davewhiteland:
includes some of #2056 but haven't explicitly dived into breaking down regular expressions which I think is needed to satisfy that, but maybe a simple (multiline?) example would fit nicely before the considerably more heavyweight Nicaraguan example now.

PR #2829, opened 2015-10-07
[#2697] Add 2factor auth

### LINKED ISSUE #2697: Increase security of superuser accounts

garethrees:
Fixes #2697

### DIFF
--- a/db/migrate/20151006101417_add_otp_enabled_to_users.rb
+++ b/db/migrate/20151006101417_add_otp_enabled_to_users.rb
@@ -1,0 +1,6 @@
+# -*- encoding : utf-8 -*-
+class AddOtpEnabledToUsers < ActiveRecord::Migration
+  def change
+    add_column :users, :otp_enabled, :boolean, :default => false
+  end
+end
--- a/Gemfile
+++ b/Gemfile
@@ -6,6 +6,7 @@ gem 'pg', '~> 0.17.1'
 
 # New gem releases aren't being done. master is newer and supports Rails > 3.0
 gem 'acts_as_versioned', :git => 'https://github.com/technoweenie/acts_as_versioned.git', :ref => '63b1fc8529d028'
+gem 'active_model_otp', :git => 'https://github.com/garethrees/active_model_otp.git', :ref => '0eb977b4c6dafd'
 gem 'charlock_holmes', '~> 0.6.9.4'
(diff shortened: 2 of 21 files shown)


db/migrate/20151006101417_add_otp_enabled_to_users.rb
```diff
@@ -0,0 +1,5 @@
+class AddOtpEnabledToUsers < ActiveRecord::Migration
```
garethrees: Fixed

garethrees:
> Only one one time password is available to the user at any one time here, and it doesn't expire on use. 

Ah, I had read [this](https://github.com/mysociety/alaveteli/issues/2697#issuecomment-140344057) as that we only wanted manual refreshes.

> I think it might be more secure for us to allow the user to print off a block of (say 5) passwords and for us to expire them one by one on use by incrementing the `otp_counter`. Regeneration would display a new set of passwords and automatically expire the old ones. 

I think this would add fairly significant complexity. We'd have to keep a record of which counter values have been used for a given secret key. `active_model_otp` doesn't provide an interface for interacting with the `ROTP::HOTP` instance – it just expects the counter to be incremented after each use.

Even if we were able to interface with the `HOTP`, we'd have a situation like this:

```ruby
hotp = ROTP::HOTP.new(ROTP::Base32.random_base32)
counter = 1

(counter..counter+4).each do |i|
  puts hotp.at(i)
end

# 111111
# 222222
# 333333
# 444444
# 555555

# User authenticates with `333333`
counter += 1

(counter..counter+4).each do |i|
  puts hotp.at(i)
end

# 222222
# 333333 # Still available for use
# 444444
# 555555
# 666666
```

garethrees:
> Why don't we just tell people to use the codes in order?

I don't think that would be reliable. We might as well just increment the counter after each use and require them take a note of the new OTP. I think this is the expected behaviour now that Google and banks are starting to use 2FA,

garethrees:
Breaking this up in to separate pull requests. Initially #2854 and #2856.

Issue #2675, opened 2015-07-22
Review and clean msgids


garethrees:
Lots of strings for translation have extra spaces and newlines due to formatting in erb templates:

```yaml
msgid "  This will appear on your {{site_name}} profile, to make it\\n            easier for others to get involved with what you're doing."
msgstr ""
```

The template that generates this msgid is:

```erb
        <p>
          <%= _('  This will appear on your {{site_name}} profile, to make it
            easier for others to get involved with what you\'re doing.', :site_name=>site_name)%>
        </p>
```

Here's what we should do to create a sensible msgstr:

```diff
        <p>
-          <%= _('  This will appear on your {{site_name}} profile, to make it
-            easier for others to get involved with what you\'re doing.', :site_name=>site_name)%>
+          <%= _("This will appear on your {{site_name}} profile, " \
+                "to make it easier for others to get involved with " \
+                "what you're doing.",
+                :site_name => site_name) %>
        </p>
```

This ticket does not consider making untranslatable strings translatable (see https://github.com/mysociety/alaveteli/issues/2676).

garethrees:
Was this fixed by https://github.com/mysociety/alaveteli/pull/2839?

crowbot:
Should be, yes. Can reopen if we find other occurences.

PR #2839, opened 2015-10-12
Whitespace in translations

crowbot:
To deploy as a minor release after #2646, with the corresponding changes to locale files such that translators won't have to retranslate the strings. #2675.


lizconlan:
:+1:

Think I've safely recreated this post the develop rebase, but I have a backup of the original if needs be

PR #2844, opened 2015-10-13
Remove fixture dependencies from integration/admin_spec

### LINKED ISSUE #2842: integration/admin_spec fails if controllers/general_controller_spec runs first
I imagine that this follows the introduction of a destroy cascade which deletes raw_email files from the disk... (sorry)

The `general_controller_spec` file starts by removing fixtures which are later required / assumed present by `integration/admin_spec`. When having a look around with pry, it seems to be failing at the point where it looks for a raw_email file with a specific number (in this case 3) to render the logged in user's post login page.

To recreate, either run all the specs with `bundle exec rake spec SPEC_OPTS="--seed 30731"` or just those 2 files with `bundle exec rspec ./spec/integration/admin_spec.rb ./spec/controllers/general_controller_spec.rb --seed 5011`

lizconlan:
Fixes #2842 (and hopefully stops the random spec fail)

### DIFF
--- a/spec/integration/admin_spec.rb
+++ b/spec/integration/admin_spec.rb
@@ -3,6 +3,9 @@ require File.expand_path(File.dirname(__FILE__) + '/../spec_helper')
 require File.expand_path(File.dirname(__FILE__) + '/alaveteli_dsl')
 
 describe "When administering the site" do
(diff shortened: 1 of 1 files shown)


garethrees:
Nice! :+1:

PR #2846, opened 2015-10-13
Use count rather than empty to determine if requests exist.

crowbot:
empty? was returning true for an unloaded collection.
This seems like a bug in the association. Similar, but not identical
behaviour is described in
http://stackoverflow.com/questions/14737148/empty-collection-check-returning-false-with-no-entries-ruby-on-rails#comment20620674_14737148

Update the method used in the destroy action for consistency.


garethrees:
:+1:

PR #2854, opened 2015-10-19
[#2697] refactor change password

garethrees:
Required for #2697 

If we're happy with this then two minor things to do:

- [x] Deprecate `UserController#signchangepassword`
- [x] Upgrade notes – I think just add the old route to the theme's `custom-routes.rb` file?


app/controllers/user_controller.rb
```diff
@@ -174,12 +174,30 @@ def confirm
...
+    when 'login_as'
+      @user = confirm_user!(post_redirect.user)
+      session[:user_id] = @user.id
+    when 'change_password'
+      unless session[:user_id] == post_redirect.user_id
+        clear_session_credentials
+      end
+
+      session[:change_password_post_redirect_id] = post_redirect.id
+    when *%w(normal change_email)
```
garethrees: Fixed in b22d66c

PR #2856, opened 2015-10-19
[#2697] Add User OTP

garethrees:
Required for #2697

This PR just adds the model-level stuff for a `User` instance being able to enable, set and validate an OTP.


garethrees:
> Looks like some of the OTP specs are failing here.

Ah looks like 09c2705 broke things. 97659da was passing – so just disregard 09c2705 and I'll take a look at that shortly.

garethrees:
> > Looks like some of the OTP specs are failing here.
>
> Ah looks like 09c2705 broke things. 97659da was passing – so just disregard 09c2705 and I'll take a look at that shortly.

Fixed in da00491

PR #2858, opened 2015-10-20
Pass HTML safe request title in track_thing's request_update params

### LINKED ISSUE #2736: Request title displayed with HTML entities in 'follow' page
See e.g. https://www.whatdotheyknow.com/track/request/the_national_youth_theatres_canc when not logged in.

lizconlan:
<strike>Passes a non-HTML escaped version of the title to the params used when redirecting for authentication. Also updates the strings used in the email to mirror the `title_in_email` param</strike>

Asks the view template to decode the title string before it is rendered to prevent double encoding of HTML tags

Fixes #2736

### DIFF
--- a/app/views/user/sign.html.erb
+++ b/app/views/user/sign.html.erb
@@ -28,7 +28,7 @@
         <% if @post_redirect.reason_params[:web].empty? %>
           <%= _('Please create an account or sign in') %>
         <% else %>
-          <%= _('{{reason}}, create an account or sign in', :reason => @post_redirect.reason_params[:web]) %>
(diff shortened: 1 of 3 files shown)


lizconlan:
Ah, looks like my fix only works if it's creating a new `PostRedirect` object from the `track_thing` params, not if it's fetching one that was already there. Intervening at the wrong point.

garethrees:
I think both of these are a valid fix – just looks like the first one will take a while to come in to play as old post redirects are purged. I think the first one is _probably_ a better fix, but might have more knock on effects. If you chose to go with 819e83a, rebase this branch so that you drop the two previous commits.

garethrees:
Could also do with a changelog entry btw! :)

PR #2859, opened 2015-10-20
Prevent user seeing field length SQL error when creating a search track

### LINKED ISSUE #937: Long track search query fails with database error PG::Error: ERROR: value too long for type character varying(255)

lizconlan:
New migration to extend the length of the `track_things.track_query` field from 255 chars to 500

Validates that the query fits within the 500 char limit to avoid the user seeing a database error if they exceed the field length

Fixes #937

### DIFF
--- a/spec/controllers/track_controller_spec.rb
+++ b/spec/controllers/track_controller_spec.rb
@@ -2,9 +2,9 @@
 require File.expand_path(File.dirname(__FILE__) + '/../spec_helper')
 
 describe TrackController do
+  let(:mock_cookie) { '0300fd3e1177127cebff' }
 
   describe 'GET track_request' do
-
     it 'clears widget votes for the request' do
       allow(AlaveteliConfiguration).to receive(:enable_widgets).and_return(true)
       @info_request = FactoryGirl.create(:info_request)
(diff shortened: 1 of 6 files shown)


spec/models/track_thing_spec.rb
```diff
@@ -70,6 +70,13 @@
     track_thing = TrackThing.create_track_for_search_query('fancy dog', 'bodies')
     expect(track_thing.track_query).to match(/variety:authority/)
   end
+
+  it "will check that the query isn't too long to store" do
+    long_query = "Lorem ipsum dolor sit amet, consectetur adipiscing elit. Duis blandit consequat enim, sit amet facilisis lorem bibendum in. Pellentesque cursus pretium tristique. Mauris nec venenatis justo. Nam facilisis, lectus et blandit eleifend, felis arcu massa nunc. Lorem ipsum dolor sit amet, consectetur adipiscing elit. Duis blandit consequat enim, sit amet facilisis lorem bibendum in. Pellentesque cursus pretium tristique. Mauris nec venenatis justo. Nam facilisis, lectus et blandit eleifend, felis arcu massa nunc."
```
garethrees: Maybe just `"Lorem " * 100`?

lizconlan: oh yes, that's much less horrible - thanks!

lizconlan:
bah, I was hoping to get a `humanize`d version of the message e.g. "Track query is too long" rather than just the message ("is too long") or the unexpected actual outcome "TrackThing|Track query is too long" which is pretty nasty :(

spec/controllers/track_controller_spec.rb
```diff
@@ -16,295 +16,367 @@
...
   end
 
-  it "should set no-cache headers on the login redirect" do
-    get :track_request, :url_title => @ir.url_title, :feed => 'track'
-    response.headers["Cache-Control"].should == 'no-cache, no-store, max-age=0, must-revalidate'
-    response.headers['Pragma'].should == 'no-cache'
-    response.headers['Expires'].should == '0'
-  end
+  describe "when making a search track" do
+    let(:track_thing) { FactoryGirl.create(:search_track,
```
lizconlan: I think there's an exemption for this https://github.com/bbatsov/rubocop/issues/376

garethrees: I don't disagree with the rule as it happens – you can still do:

```ruby
let(:blah) do
  # ...
end
```

If you rebase on develop you should get the updated config that disables hound checks by default, other than the ones we've explicitly enabled though :)

garethrees:
Looks good to me other than the minor comment (https://github.com/mysociety/alaveteli/pull/2859#commitcomment-14018483) – :shipit: :ship:

garethrees:
Could also do with a changelog entry btw! :)

Issue #2699, opened 2015-07-31
include something pictorial in the default logo


davewhiteland:
The default theme has a "logo" which is the text word Alaveteli.
Inserts something -- anything even a wee squiggle -- into that file which is plainly a non-text thing, so it is abundantly clear to new owners that that element is a graphic, not a text, which can therefore be replaced.

We (mySociety staff) got caught out in a redeploy by not realising this was a graphic file change requirement and not the site name being dropped in (yes we would have "inspected element" if we had reason to suspect it was not just text :-) )

davewhiteland:
oops I think perhaps this really belongs in https://github.com/mysociety/alavetelitheme and is kind-of related to https://github.com/mysociety/alavetelitheme/issues/46 ...

Leaving this here unless Alavetelians want to move it across.

garethrees:
The default logo for Alaveteli is the word "LOGO" as an image https://github.com/mysociety/alaveteli/blob/develop/app/assets/images/logo.png

Probably needs a similar change in both places? I thought we used an "A" logo mark somewhere, but that might have been on the old-style theme.

zarino:
Two PRs: one for core https://github.com/mysociety/alaveteli/pull/2861, and one for alavetelitheme https://github.com/mysociety/alavetelitheme/pull/64

![screen shot 2015-10-21 at 13 31 27](https://cloud.githubusercontent.com/assets/739624/10636583/9025ecce-77f8-11e5-8eca-733bbe2c4cd3.png)

![screen shot 2015-10-21 at 13 20 10](https://cloud.githubusercontent.com/assets/739624/10636377/6b7a028a-77f7-11e5-8b93-6d9ab6289561.png)

PR #2861, opened 2015-10-21
Include pictorial element in logo.png

zarino:
Part of #2699.

Adding an icon to the logo makes it clear that it’s not just text, and that it needs to be replaced when you are making your own custom theme.

screen shot 2015-10-21 at 13 31 27

(Sorry if I wasn't meant to branch off master – feel free to rebase etc, it's only a tiny change).


zarino:
@crowbot – So the image is currently included just via an `` tag. I can put a double-size image in there, but since it's not in a media query or anything, it'll get downloaded even by non-hi-dpi devices. I've just exported the 2x image here, and it’s exactly twice the file size of the 1x. Are you ok with that?

Or do you want me to edit the theme so it uses CSS image replacement (with media queries) rather than just an inline `` tag?

zarino:
Ah! The CSS background image treatment has been added in `develop` already. I'll close this PR, and open a new one off the `develop` branch.

PR #2865, opened 2015-10-22
Use Rails' built in method to clear session

garethrees:
I don't know if there's a reason we're not doing this?

There is a subtle behaviour change from:

```ruby
session[:remember_me] = false
```

After this we'd get `nil`. I can't see anything obvious that would break because of this?


garethrees:
Hmm. Doesn't actually work…

PR #2866, opened 2015-10-23
Include pictorial element in logo.png

zarino:
Part of #2699.

Adding an icon to the logo makes it clear that it should be replaced when you customise the theme.

screen shot 2015-10-23 at 10 32 29


zarino:
Also makes the logo bigger, as it turns out, because, as everyone knows, bigger logos are better.

garethrees:
Looks great, thanks!

PR #2869, opened 2015-10-27
Tidy up InfoRequest

### LINKED ISSUE #2822: InfoRequest first pass cleanup/refactoring
- [x] Remove unecessary calls to `self`, '!..nil?' combinations. 
- [x] Consistent spacing between methods. 
- [x] Remove duplication in `law_used_*` methods (deprecate `law_used_with_a` as it doesn't seem to be used)
- [x] `fully_destroy` - can some of this logic be done with `:dependent => :destroy' clauses on the associations? 
- [ ] `title_formatting` - convert to built-in validations? 
- [x] `recent_requests`
  - [x] remove unused `sortby` variable, use `max_count` in limit of first `ActsAsXapian::Search` call
  - [x] Factor out `ActsAsXapian::Search` calls into a common method.

lizconlan:
Fixes #2822

Note - opted not to move all the functionality of `title_formatting` to a built-in validation as the RegEx required for "does not match this pattern" is not obvious and would probably be harder to read than the current arrangement.

### DIFF
--- a/app/models/info_request.rb
+++ b/app/models/info_request.rb
@@ -310,7 +310,7 @@ class InfoRequest < ActiveRecord::Base
   def email_subject_request(opts = {})
     html = opts.fetch(:html, true)
     _('{{law_used_full}} request - {{title}}',
-      :law_used_full => law_used_full,
+      :law_used_full => law_used_human(:full),
       :title => (html ? title : title.html_safe))
   end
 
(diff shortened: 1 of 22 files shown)


app/models/info_request.rb
```diff
@@ -39,7 +39,8 @@ class InfoRequest < ActiveRecord::Base
   validates_presence_of :title, :message => N_("Please enter a summary of your request")
   validates_format_of :title, :with => /[[:alpha:]]/,
     :message => N_("Please write a summary with some text in it"),
-    :if => Proc.new { |info_request| !info_request.title.nil? && !info_request.title.empty? }
+    :unless => Proc.new { |info_request| info_request.title.blank? }
+  validate :title, :max_length => 200, :message => _('Please keep the summary short, like in the subject of an email. You can use a phrase, rather than a full sentence.')
```
lizconlan: not sure whether this will break any older requests which weren't originally subject to the length restriction, if so I should add an `:on => :create` so that it only applies to new requests

garethrees:
Couple of minor tweaks left but looking really good. Will want a changelog entry, and some upgrade notes about the deprecations and what to use instead

garethrees:
Looks like this is ready to go other than a typo in the Changelog and a rebase with develop, so marking as awaiting deploy.

lizconlan:
Have done the rebase and squashed the fixups. Will have another read through in the morning to double-check rather than merge it in now.

PR #2874, opened 2015-10-28
Add the pry-nav gem to enable debugging steps with pry

### LINKED ISSUE #2873: Enable debugging steps for pry

lizconlan:
Enables simple next and step commands to allow pry to function as
a basic debugger (more complex versions are available but they're
ruby version specific).

Also adds a config file to make single letter command aliases for
easier usage

Close #2873


.pryrc
```diff
@@ -0,0 +1,5 @@
+if defined?(PryNav)
+  Pry.commands.alias_command 'c', 'continue'
+  Pry.commands.alias_command 's', 'step'
+  Pry.commands.alias_command 'n', 'next'
+end
```
garethrees: Missing newline here. Just fix that up on merge :+1:

PR #2876, opened 2015-10-30
add "simple" example of censor rule (word match and regexp)

davewhiteland:
Note this doesn't cover the case-insensitive example well: as @schlos asked https://github.com/mysociety/alaveteli/issues/2056#issuecomment-93829106
about how to add the "options" that regexp users might expect.
I don't think the Nicargauan example really covers this because it adds the regexp within the source code -- is there a case for adding "options" as a (JS-revealed?) selection in the admin if someone chooses "regexp"?

If we can show it (that is: explicit case insensitivity and multiline regexps) in the admin interface, that would be great but I'm not sure how to do that: as it happens, I've simulated

* case-insenstitive (by using character class [Aa] on one letter)
* multiline (by using whitespace \s)

both of which would be better if I was bolder about saying "this is how to do it."  But I'm not convinved it's the best way... is it?

### DIFF
--- a/docs/running/redaction.md
+++ b/docs/running/redaction.md
@@ -163,7 +163,7 @@ will have the rules applied.
 This example removes the name "Alice Alaveteli" from messages relating to a
 specific request.
 
-* click on **Requests** in the admin and searc or find the one you want 
+* click on **Requests** in the admin and search or find the one you want 
 * click on the request's title to go to its page in the admin
 * scroll to the bottom: *Censor rules*
 * click **New censor rule** and fill in the form:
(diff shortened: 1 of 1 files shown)


docs/running/redaction.md
```diff
@@ -150,6 +151,113 @@ site, ask us for help.
...
+hidden from messages associated with a particular request. You can also
+add censor rules to a user (rather than a request), so any requests they make
+will have the rules applied.
+
+### Add a censor rule to hide a name using an exact match
+
+This example removes the name "Alice Alaveteli" from messages relating to a
+specific request.
+
+* click on **Requests** in the admin and searc or find the one you want
```
crowbot: typo "searc"

docs/running/redaction.md
```diff
@@ -150,6 +151,113 @@ site, ask us for help.
...
+regardless of the case of the first letter, and even if it has double-l in the
+spelling.
+
+<div class="attention-box warning">
+   We recommend you <strong>do not</strong> use regular expressions unless you
+   really need them, because they can radically slow down the display of
+   requests on your site.
+</div>
+
+* click on **Requests** in the admin and searc or find the one you want
```
crowbot: typo "searc"

PR #2882, opened 2015-11-03
Handle UTF-7 encoding mail parts.

crowbot:
Solution taken from
http://programmers.stackexchange.com/questions/163930/how-relevant-is-utf-7-when-it-comes-to-parsing-emails


garethrees:
:+1: merged in https://github.com/mysociety/alaveteli/commit/e5d6ac3e34771d2e4ff43f689979912d4ff53cc4

PR #2890, opened 2015-11-10
Brakeman fixes

crowbot:
Fixes suggested by Brakeman for the new release.


lizconlan:
Was just checking that the change to the locale handling didn't affect the locale switching functionality (it looks fine). :+1:

PR #2901, opened 2015-11-13
Implementation of @lizconlan's original fix for #2736.

crowbot:
Looking at this again, I think I was wrong in pushing back on this
original solution. The problem is that strings containing quotes etc,
interpolated into translated strings, get html escaped and marked as
`html_safe`. This is fine when they are immediately displayed in a
template as the `html_safe` marker means they don't automatically get
escaped again. However, when the resulting strings are stored in a `PostRedirect`
`reason_params` field as YAML and then pulled out again for display
on the signup page, the html_safe attribute is lost, so the string
is double escaped.

The `web` param of a track is only ever displayed via a `PostRedirect`
on the `sign` template, I think it's better to mark interpolated values
as safe originally, so that the whole string gets escaped in the template
like other non-translated strings. This won't fix any existing `PostRedirects`,
but these will be deleted in 3 months. It doesn't require us to make
special use of `html_entities`, so is more consistent with what's going
on in general.

Sorry for the bad review @lizconlan!

### DIFF
--- a/app/models/track_thing.rb
+++ b/app/models/track_thing.rb
@@ -174,7 +174,7 @@ class TrackThing < ActiveRecord::Base
                          :request_title => info_request.title),
       # Authentication
       :web => _("To follow the request '{{request_title}}'",
-                :request_title => info_request.title),
+                :request_title => info_request.title.html_safe),
       :email => _("Then you will be updated whenever the request '{{request_title}}' is updated.",
                   :request_title => info_request.title),
       :email_subject => _("Confirm you want to follow the request '{{request_title}}'",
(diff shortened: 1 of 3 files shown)


lizconlan:
:+1: looks good to me (but I would say that!)

I note that I appear have taken it upon myself to [spoil the indention on this line though](https://github.com/mysociety/alaveteli/commit/13cf302bbd2c362afda0d7d7d0a5d7f9f42efe05#diff-b353e0f520c744f61c05c1669d9fb2c3R177)

app/models/track_thing.rb
```diff
@@ -174,7 +174,7 @@ def request_updates_params
                          :request_title => info_request.title),
       # Authentication
       :web => _("To follow the request '{{request_title}}'",
-                :request_title => info_request.title),
+                  :request_title => info_request.title.html_safe),
```
lizconlan: indentation increased in error by the looks of things

PR #2935, opened 2015-12-01
Hotfix/0.23.0.3

crowbot:
Translations in Transifex were those for the previous release - this PR re-syncs Transifex and git.


garethrees:
Looks okay to me :+1:

Issue #2843, opened 2015-10-13
Admin summary page timing out due to length of time taken by 'old unclassified' query


crowbot:
e.g. 
```
LOG:  duration: 94670.055 ms  execute <unnamed>: SELECT  *, (SELECT info_request_events.created_at
                  FROM info_request_events, incoming_messages
                  WHERE info_request_events.info_request_id = info_requests.id
                   AND info_request_events.event_type = 'response'
                  AND incoming_messages.id = info_request_events.incoming_message_id
                               AND incoming_messages.prominence = 'normal'
                  ORDER BY created_at desc
                  LIMIT 1) AS last_response_time FROM "info_requests"  WHERE (awaiting_description = 't'
                                            AND (SELECT info_request_events.created_at
                  FROM info_request_events, incoming_messages
                  WHERE info_request_events.info_request_id = info_requests.id
                   AND info_request_events.event_type = 'response'
                  AND incoming_messages.id = info_request_events.incoming_message_id
                               AND incoming_messages.prominence = 'normal'
                  ORDER BY created_at desc
                  LIMIT 1) < '2015-09-21 11:45:48.319245'
                                            AND url_title != 'holding_pen'
                                            AND user_id IS NOT NULL AND prominence = 'normal') ORDER BY last_response_time LIMIT 20
```

Query plan for the above:

```
 Limit  (cost=11270998.38..11270998.43 rows=20 width=204) (actual time=100057.108..100057.114 rows=20 loops=1)
   ->  Sort  (cost=11270998.38..11271025.31 rows=10772 width=204) (actual time=100057.107..100057.110 rows=20 loops=1)
         Sort Key: ((SubPlan 1))
         Sort Method: top-N heapsort  Memory: 30kB
         ->  Index Scan using index_info_requests_on_user_id on info_requests  (cost=0.00..11270711.74 rows=10772 width=204) (actual time=0.102..99993.337 rows=27295 loops=1)
               Index Cond: (user_id IS NOT NULL)
               Filter: (awaiting_description AND (url_title <> 'holding_pen'::text) AND ((prominence)::text = 'normal'::text) AND ((SubPlan 2) < '2015-09-21 11:45:48.319245'::timestamp without time zone))
               SubPlan 1
                 ->  Limit  (cost=36.57..36.58 rows=1 width=8) (actual time=0.034..0.035 rows=1 loops=27295)
[...]

PR #2937, opened 2015-12-02
Stop wrapping html version of outgoing emails at 67 chars

### LINKED ISSUE #2913: Wrapping at 67 chars makes messages hard to read
Currently, all outgoing emails are hard wrapped at 67 characters.

When a user with a phone (or similarly narrow viewport) tries to read a message, they end up with a fairly ugly and hard to read experience.

The culprit seems to be https://github.com/mysociety/commonlib/blob/master/rblib/format.rb#L22

I notice that *incoming* emails are not subject to the same restrictions and therefore are much more pleasant to read.

Firstly, I'm confused as to why it's 67 characters.  [RFC 2822](http://www.faqs.org/rfcs/rfc2822.html) says:

> Each line of characters MUST be no more than 998 characters, and SHOULD be no more than 78 characters, excluding the CRLF.

Secondly, if the wrap *does* have to be made when sending, would it still make more sense to display it **un**wrapped?

(Possibly related to #707)

lizconlan:
Now behaves more like the code for incoming messages, skips the 67 char line break code and squishes extra line breaks. (Would it better for the preview screen not to do that second bit?)

Closes #2913

### DIFF
--- a/spec/models/outgoing_message_spec.rb
+++ b/spec/models/outgoing_message_spec.rb
@@ -143,12 +143,45 @@ describe OutgoingMessage, " when making an outgoing message" do
     expect(@outgoing_message.get_text_for_indexing).not_to include("foo@bar.com")
   end
 
-  it "should not display email addresses on page" do
-    expect(@outgoing_message.get_body_for_html_display).not_to include("foo@bar.com")
-  end
+  context "when formatting the message for html display" do
 
-  it "should link to help page where email address was" do
-    expect(@outgoing_message.get_body_for_html_display).to include('<a href="/help/officers#mobiles">')
+    it "does not display email addresses on page" do
+      expect(@outgoing_message.get_body_for_html_display).not_to include("foo@bar.com")
+    end
+
+    it "links to help page where email address was" do
+      expect(@outgoing_message.get_body_for_html_display).to include('<a href="/help/officers#mobiles">')
+    end
+
+    it "does not force long lines to wrap" do
+      long_line = "long string of 125 characters, set so the old line break " \
+                  "falls here, and making sure even longer lines are not " \
+                  "affected either"
+      @outgoing_message.body = long_line
+      expect(@outgoing_message.get_body_for_html_display).to eq("<p>#{long_line}</p>")
+    end
+
+    it "interprets single line breaks as <br> tags" do
+      split_line = "Hello,\nI am a test message\nWith multiple lines"
+      expected = "<p>Hello,\n<br />I am a test message\n<br />With multiple lines</p>"
+      @outgoing_message.body = split_line
+      expect(@outgoing_message.get_body_for_html_display).to include(expected)
+    end
+
+    it "interprets double line breaks as <p> tags" do
+      split_line = "Hello,\n\nI am a test message\n\nWith multiple lines"
+      expected = "<p>Hello,</p>\n\n<p>I am a test message</p>\n\n<p>With multiple lines</p>"
+      @outgoing_message.body = split_line
+      expect(@outgoing_message.get_body_for_html_display).to include(expected)
+    end
+
+    it "removes excess linebreaks" do
+      split_line = "Line 1\n\n\n\n\n\n\n\n\n\nLine 2"
+      expected = "<p>Line 1</p>\n\n<p>Line 2</p>"
+      @outgoing_message.body = split_line
+      expect(@outgoing_message.get_body_for_html_display).to include(expected)
+    end
+
   end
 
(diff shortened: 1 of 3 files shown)


garethrees:
Implementation looks good. Would be great to add a spec and a Changelog entry.

lizconlan:
@garethrees I keep forgetting the Changelog entries, I think I missed one for #2843 as well :cry:

garethrees:
> I think I missed one for #2843 as well

Just make a new commit for it.

spec/models/outgoing_message_spec.rb
```diff
@@ -143,12 +143,36 @@
...
+    end
+
+    it "links to help page where email address was" do
+      expect(@outgoing_message.get_body_for_html_display).to include('<a href="/help/officers#mobiles">')
+    end
+
+    it "does not force long lines to wrap" do
+      long_line = "123456789 123456789 123456789 123456789 123456789 123456789 1234"
+      @outgoing_message.body = long_line
+      expect(@outgoing_message.get_body_for_html_display).to include(long_line)
```
garethrees: This spec passes irrespective of whether the change to `app/models/outgoing_message.rb` is applied.

I can make the spec fail before the code change is applied by changing the spec to:

```diff
- @outgoing_message.body = long_line
+ @outgoing_message.body = long_line * 2
```

Since we're making a change here, could we make the example string a bit clearer? Something like `"String with more than 78 characters"` or something?

I find that writing specs first – or at least without the code change in place – usually catches this sort of thing.

lizconlan: Ah, the string content is an old typesetter's trick of having `1234567890` as an easy count of 10 chars. Except I got the count wrong. And the space placement /o\

lizconlan: (but apart from that, and the fact that that only makes sense to recovering mainframe programmers who had to set fixed width layouts for dot matrix printers... \*puts character count ruler back in the drawer\*)

spec/models/outgoing_message_spec.rb
```diff
@@ -143,12 +143,45 @@
...
+    it "interprets single line breaks as <br> tags" do
+      split_line = "Hello,\nI am a test message\nWith multiple lines"
+      expected = "Hello,<br>I am a test message<br>With multiple lines"
+      @outgoing_message.body = split_line
+      expect(@outgoing_message.get_body_for_html_display).to include(expected)
+    end
+
+    it "interprets double line breaks as <p> tags" do
+      split_line = "Hello,\n\nI am a test message\n\nWith multiple lines"
+      expected = "Hello,<p>I am a test message<p>With multiple lines"
```
garethrees: I'd expect, at a minimum:

```
"Hello,<p>I am a test message</p><p>With multiple lines"
```

Shouldn't it really be:

```
"<p>Hello,</p><p>I am a test message</p><p>With multiple lines</p>"
```

I would not be adverse to directly calling [`simple_format`](to do this).

spec/models/outgoing_message_spec.rb
```diff
@@ -143,12 +143,45 @@
...
+    it "links to help page where email address was" do
+      expect(@outgoing_message.get_body_for_html_display).to include('<a href="/help/officers#mobiles">')
+    end
+
+    it "does not force long lines to wrap" do
+      long_line = "long string of 125 characters, set so the old line break " \
+                  "falls here, and making sure even longer lines aren't " \
+                  "affected either"
+      @outgoing_message.body = long_line
+      expect(@outgoing_message.get_body_for_html_display).not_to include('<br>')
```
garethrees: What about `\n`s? I'd maybe make this:

```ruby
expect(@outgoing_message.get_body_for_html_display).to eq(long_line)
```

I don't know – on one hand its a bit more explicit that we're saying we don't want the two specific breaking characters. On the other, I like that we're stating that the string should be unchanged.

Any thoughts @crowbot?

spec/models/outgoing_message_spec.rb
lizconlan: That would be more correct, should the [equivalent code for incoming messages](https://github.com/mysociety/alaveteli/blob/develop/app/models/incoming_message.rb#L581-L582) be updated in parallel?

garethrees: Probably. I'll ping you tomorrow to take a quick look at it?

lizconlan: @garethrees sounds good, thanks :)

lizconlan:
unmerged!

garethrees:
Open a new PR :)

Issue #2665, opened 2015-07-17
Remove UK-specific code to mysociety/whatdotheyknow-theme


garethrees:
Ticket to track UK-specific code

- [ ] ~~https://github.com/mysociety/alaveteli/issues/2085 (EIR)~~
- [x] https://github.com/mysociety/alaveteli/issues/625 (Schools)
- [ ] ~~[Health Checks](https://github.com/mysociety/alaveteli/blob/develop/config/initializers/health_checks.rb) should be configured in a theme~~
- [x] https://github.com/mysociety/alaveteli/blob/develop/config/initializers/session_store.rb#L3 (Cookie name)
- [x] https://github.com/mysociety/alaveteli/blob/develop/config/application.rb#L83-L85, https://github.com/mysociety/alaveteli/blob/e0fc3d2d6348e16b34471cdb7d8fad11e1b8b3cc/lib/configuration.rb#L2 (WDTK Empty Sessions)
- [ ] https://github.com/mysociety/alaveteli/issues/2662 (Extract quoted section removal)
- [x] https://github.com/mysociety/alaveteli/issues/2396 (Heather Brooke)

garethrees:
Colloquial language, e.g. `(<strong>no ranty</strong> politics, read our <a href="{{url}}">moderation policy</a>)`. Related to https://github.com/mysociety/alaveteli/issues/2676

garethrees:
Checking the comment uses mixed capitals probably doesn't work so well for non-latin?

- https://github.com/mysociety/alaveteli/blob/e0fc3d2d6348e16b34471cdb7d8fad11e1b8b3cc/app/models/comment.rb#L88-L94
- https://github.com/mysociety/alaveteli/blob/e0fc3d2d6348e16b34471cdb7d8fad11e1b8b3cc/app/models/comment.rb#L34

garethrees:
The [laws](https://github.com/mysociety/alaveteli/blob/e0fc3d2d6348e16b34471cdb7d8fad11e1b8b3cc/app/models/info_request.rb#L35-L45) are pretty UK-specific, but quite a difficult extraction. We talked about this in #2819.

garethrees:
The `very_overdue` state is pretty UK-specific but might be a tough one to deprecate.

garethrees:
[...]

PR #2939, opened 2015-12-02
[#2665] More generic cookie name

garethrees:
Part of #2665


garethrees:
Need to try a bit of testing on this to see what happens when we change this.

garethrees:
Closing as stale.

Issue #2891, opened 2015-11-11
Submitting new request can produce error


crowbot:
NameError in UserController#signin

uninitialized constant ActionView::CompiledTemplates::HTMLEntities

PR #2940, opened 2015-12-03
Get htmlentities from Gemfile, not commonlib.

crowbot:
Fixes https://github.com/mysociety/alaveteli/issues/2891


garethrees:
:+1:

PR #2889, opened 2015-11-10
Improve handling of duplicate url_titles

lizconlan:
* Light refactor of the code for `InfoRequest#update_url_title` so that it's easier to follow
* Sneak in a few tests for the above
* Add a new validation so that if there is a `url_title` clash the user will be prompted to try again rather than receiving an app error page
* Prevent the new request form submit button from being pressed twice which should hopefully prevent the error condition from occurring

Fixes #390

PR #2956, opened 2015-12-15
Add a config option to turn off comments/annotations site-wide

### LINKED ISSUE #2955: Allow comments/annotations to be disabled site-wide
Panama have requested this, I have a PR forthcoming, just creating this ticket to identify it.

stevenday:
This adds a new config option `ENABLE_ANNOTATIONS` which defaults to `true`
but which allows you to turn off the commenting on InfoRequests if desired.

I've implemented this in two places - firstly in the "after actions" view, to
hide or show the comment link, and then on the `before_filter` for the actual
comment form action. Is that everywhere we offer the ability to comment?
(There are various open tickets about providing bulk annotations, or
redirecting straight to the annotation form after submission, but I couldn't
find anything like that actually in the code at the moment).

Closes #2955

### DIFF
--- a/app/views/general/_advanced_search_tips.html.erb
+++ b/app/views/general/_advanced_search_tips.html.erb
@@ -9,7 +9,9 @@
   <li><%= _('<strong><code>variety:</code></strong> to select type of thing to search for, see the <a href="{{varieties_url}}">table of varieties</a> below.', :varieties_url => "#varieties") %></li>
   <li><%= _('<strong><code>requested_from:home_office</code></strong> to search requests from the Home Office, typing the name as in the URL.')%></li>
   <li><%= _('<strong><code>requested_by:julian_todd</code></strong> to search requests made by Julian Todd, typing the name as in the URL.') %></li>
-  <li><%= _('<strong><code>commented_by:tony_bowden</code></strong> to search annotations made by Tony Bowden, typing the name as in the URL.')%></li>
+  <% if AlaveteliConfiguration.enable_annotations %>
+    <li><%= _('<strong><code>commented_by:tony_bowden</code></strong> to search annotations made by Tony Bowden, typing the name as in the URL.')%></li>
+  <% end %>
   <li><%= _('<strong><code>request:</code></strong> to restrict to a specific request, typing the title as in the URL.')%>
   <li><%= _('<strong><code>filetype:pdf</code></strong> to find all responses with PDF attachments. Or try these: <code>{{list_of_file_extensions}}</code>', :list_of_file_extensions => IncomingMessage.get_all_file_extensions)%></li>
   <li><%= _('Type <strong><code>01/01/2008..14/01/2008</code></strong> to only show things that happened in the first two weeks of January.')%></li>
@@ -49,7 +51,9 @@
   # the initial request
   _('Follow up message sent by requester') %></td></tr>
   <tr><td><strong><%=search_link('variety:response')%></strong></td><td><%= _('Response from a public authority') %></td></tr>
-  <tr><td><strong><%=search_link('variety:comment')%></strong></td><td><%= _('Annotation added to request') %></td></tr>
+  <% if AlaveteliConfiguration.enable_annotations %>
+    <tr><td><strong><%=search_link('variety:comment')%></strong></td><td><%= _('Annotation added to request') %></td></tr>
+  <% end %>
   <tr><td><strong><%=search_link('variety:authority')%></strong></td><td><%= _('A public authority') %></td></tr>
   <tr><td><strong><%=search_link('variety:user')%></strong></td><td><%= _('A {{site_name}} user', :site_name=>site_name) %></td></tr>
 </table>
--- a/app/views/request/_after_actions.html.erb
+++ b/app/views/request/_after_actions.html.erb
@@ -5,7 +5,7 @@
     <strong><%= _('Anyone:') %></strong>
 
     <ul>
-      <% if @info_request.comments_allowed? %>
+      <% if AlaveteliConfiguration.enable_annotations && @info_request.comments_allowed? %>
         <li>
           <%= raw(_('<a href="{{url}}">Add an annotation</a> (to help the ' \
                       'requester or others)',
--- a/app/views/request/_request_sent.html.erb
+++ b/app/views/request/_request_sent.html.erb
@@ -40,14 +40,22 @@
 
         <h2><%= _("Keep your request up to date") %></h2>
 
-        <p>
-          <%= _('If you write about this request ' \
-                '(for example in a forum or a blog) ' \
-                'please link to this page, and <a href="{{url}}">add an ' \
-                'annotation</a> below telling people ' \
-                'about your writing.',
-                :url => new_comment_url(:url_title => @info_request.url_title).html_safe) %>
-        </p>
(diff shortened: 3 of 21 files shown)


garethrees:
Build failure looks like its because of https://github.com/mysociety/alaveteli/pull/2889#issuecomment-164815995 – I'll check it locally but should be safe to ignore.

garethrees:
Cheers! Looks good, but there are quite a few other places that annotations are mentioned (e.g. https://github.com/mysociety/alaveteli/blob/develop/app/controllers/request_controller.rb#L496) – probably need to cover these off.

stevenday:
For my future self: @garethrees was linking to https://github.com/mysociety/alaveteli/blob/208c1f766d623109c02c984d04e2be40b3999e3a/app/controllers/request_controller.rb#L496 as an example. It looks like (annoyingly) we'll have to create a whole bunch of new translation strings for when this is turned off :(

stevenday:
@garethrees - wow, that got complicated fast. 

I've been through and made every mention of annotations conditional on this new config, but it's a lot of changes. Before I start writing tests for all of these, I just want to check whether you and @crowbot will be happy to merge this?

stevenday:
Note: I haven't removed mention of annotations from the admin because I assumed that was unecessary.

Neither have I made any changes to the mailers that send emails about events (that might include annotations). I think that means there's is a possible "race condition" where you allow comments, then turn them off, and people get alerts about them inbetween, but I've assumed that that doesn't really matter.

I'm happy to revisit either of these assumptions if needs be.

app/views/request/_after_actions.html.erb
```diff
@@ -5,7 +5,7 @@
     <strong><%= _('Anyone:') %></strong>
 
     <ul>
-      <% if @info_request.comments_allowed? %>
+      <% if AlaveteliConfiguration.enable_annotations && @info_request.comments_allowed? %>
```
garethrees: You could bake `AlaveteliConfiguration.enable_annotations` in to `InfoRequest#comments_allowed?`.

app/views/request/_request_sent.html.erb
```diff
@@ -40,14 +40,22 @@
...
 
-        <p>
-          <%= _('If you write about this request ' \
-                '(for example in a forum or a blog) ' \
-                'please link to this page, and <a href="{{url}}">add an ' \
-                'annotation</a> below telling people ' \
-                'about your writing.',
-                :url => new_comment_url(:url_title => @info_request.url_title).html_safe) %>
-        </p>
+        <% AlaveteliConfiguration.enable_annotations %>
```
garethrees: missing if

app/views/request/_after_actions.html.erb
stevenday: I don't have a strong preference, but I would find that more confusing I think - it's a boolean field on the model, so the database and the ruby object would be telling different stories about what's happening. It also seems like giving the model too much extra knowledge about the application state?

app/views/request/_request_sent.html.erb
stevenday: Thanks! I've fixed that up in 7d91c43

stevenday:
I've added specs (a83b588), fixed up https://github.com/mysociety/alaveteli/pull/2956#discussion_r57137166 in 7d91c43 and also fixed up a couple of other templates which needed some conditionals in 1e5f4cc

I note from this that alavetelitheme also contains copies of some of these templates which should (I guess?) have the conditionals too. Would you like me to make a matching PR there?

BenJam:
If there's any way this could be reviewed tomorrow it would be greatly appreciated. There's an inside chance that we will need to launch Infórmate Panamá April 1st.

app/views/request/_after_actions.html.erb
garethrees: Yeah, you're right there.

app/views/general/_advanced_search_tips.html.erb
```diff
@@ -9,7 +9,9 @@
   <li><%= _('<strong><code>variety:</code></strong> to select type of thing to search for, see the <a href="{{varieties_url}}">table of varieties</a> below.', :varieties_url => "#varieties") %></li>
   <li><%= _('<strong><code>requested_from:home_office</code></strong> to search requests from the Home Office, typing the name as in the URL.')%></li>
   <li><%= _('<strong><code>requested_by:julian_todd</code></strong> to search requests made by Julian Todd, typing the name as in the URL.') %></li>
-  <li><%= _('<strong><code>commented_by:tony_bowden</code></strong> to search annotations made by Tony Bowden, typing the name as in the URL.')%></li>
+  <% AlaveteliConfiguration.enable_annotations %>
```
garethrees: Missing `if`

stevenday: Gah, I thought I'd caught that, fixed now.

config/general.yml-example
```diff
@@ -897,3 +897,14 @@ ENABLE_TWO_FACTOR_AUTH: false
 #
 # ---
 EXTERNAL_REVIEWERS: ''
+
+# Enable annotations on requests.
+# If ENABLE_ANNOTATIONS is set to true, Alaveteli will include a link in the
+# "Things to do with this request" section to allow users to add extra
+# comments or information to it. This is a standard feature which is usually
+# turned on, but if you want to disable it, you can set this to false.
```
garethrees: I think we should add a line like "We don't recommend disabling annotations after users have started making comments"

garethrees:
Its worth noting that existing comments are still shown, even if comments are disabled:

screen shot 2016-03-31 at 09 10 58

I _think_ this is what we should do in that situation. If a site admin disables comments after the site has been live, it seems like it would make sense to preserve the history – they can always delete specific comments through the admin interface. 

If we did want to stop hiding them, we could apply the following diff:

```diff
diff --git a/app/views/request/_correspondence.html.erb b/app/views/request/_correspondence.html.erb
index a6263c8..8a80bdf 100644
--- a/app/views/request/_correspondence.html.erb
+++ b/app/views/request/_correspondence.html.erb
@@ -7,5 +7,7 @@
 <% when 'resent', 'followup_resent' %>
   <%= render :partial => 'request/resent_outgoing_correspondence', :locals => { :outgoing_message => info_request_event.outgoing_message, :info_request_event => info_request_event }%>
 <% when 'comment' %>
-  <%= render :partial => 'comment/single_comment', :locals => { :comment => info_request_event.comment } %>
+  <% if AlaveteliConfiguration.enable_annotations %>
+    <%= render :partial => 'comment/single_comment', :locals => { :comment => info_request_event.comment } %>
+  <% end %>
 <% end %>
```

garethrees:
@stevenday looks great. A couple of things left to do:

- [ ] [Small fixup](https://github.com/mysociety/alaveteli/pull/2956#discussion_r58020849) to the config documentation
- [ ] Add a `doc/CHANGES.md` entry
- [ ] Squash & Merge
- [ ] Make a PR to the [configuration documentation](http://alaveteli.org/docs/customising/config/)

Thanks!

stevenday:
@garethrees - thanks! I've done everything except squash and merge - happy for you to merge as much as you'd like. The documentation PR is here: https://github.com/mysociety/alaveteli/pull/3135,

I also made an alavetelitheme PR, which is here: https://github.com/mysociety/alavetelitheme/pull/76

PR #2961, opened 2015-12-16
Add missing translation wrappers

garethrees:


lizconlan:
:+1: (I'm ignoring the collation "error")

PR #2963, opened 2015-12-17
[#625] Extract school logic

### LINKED ISSUE #625: Move school specific logic out of the core of Alaveteli
In `InfoRequest#date_very_overdue_after` there is UK specific logic which calculates when a request is very overdue which makes specific different allowances for schools.

There is also an associated method `PublicBody#is_school?` which uses the tag to figure out whether this is a school.

It would be good to move this stuff out of the core of Alaveteli because it's hardcoded UK specific stuff.

garethrees:
Fixes #625

### DIFF
--- a/app/models/info_request.rb
+++ b/app/models/info_request.rb
@@ -724,7 +724,9 @@ class InfoRequest < ActiveRecord::Base
   # last_event_forming_initial_request. There may be more obscure
   # things, e.g. fees, not properly covered.
   def date_response_required_by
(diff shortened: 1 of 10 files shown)


garethrees:
> Should this PR also remove the config variable `SPECIAL_REPLY_VERY_LATE_AFTER_DAYS`, and be accompanied by a documentation PR that does the same? 

+1

> It probably also merits a line in the release notes.

+1

PR #2964, opened 2015-12-17
Update gems

garethrees:


garethrees:
Updating mail needs further investigation (#2984).

PR #2969, opened 2015-12-23
Refactor RequestController#show_response into new controller

### LINKED ISSUE #2868: Refactor RequestController#show_response into it's own RESTful controller
We identified this as a refactoring that will help us implement the ombudsman feature being discussed in #2819.

lizconlan:
Reworks the `RequestController#show_response` action into the `new_followup` (GET), `preview_followup`, and `create_followup` (POST) actions of `ResponseController`. These actions are still known as `show_response` and `preview_response` in the routes file in an effort to minimise the disruption to themes (although as view files have moved and/or been renamed, maybe this isn't as useful as I first thought).

There is potential both to refactor more controller actions into `RequestController` to make it more generally useful, or to rename it to `FollowupController` and simplify the existing actions to merely `new`, `preview` and `create`.

As a preliminary step to refactoring, I have also rewritten the message/alert sending sections of the `RequestController` specs as `spec/integration/request_mailer_spec.rb` which hopefully removes the need for `ApplicationController.test_code_redirect_by_email_token` although I have left it in place for now.

Closes #2868


garethrees:
> or to rename it to `FollowupController` and simplify the existing actions to merely `new`, `preview` and `create`

Yeah, I think this is what I was expecting to end up with, since these actions are about managing followups.

garethrees:
As for some general notes on reviewing a big refactoring PR like this:

- Would have been easier if big (> 10 LOC) extractions were first extracted as one commit, then changed. The combination of extract and change in one commit is difficult to review, because it's hard to see where behaviour has changed, and whether that behaviour change matches the original.
- Better commit messages are super helpful. For example, I don't really know the context of what was broken in (d894a4f7282), so I don't know why its being "fixed". Was it broken before the branch was started, or was the breakage introduced in the branch? What was happening because of the breakage that shouldn't have been happening?
- Some specs have been removed without comment. I don't know if thats a mistake or intentional. If the latter, why do we no longer need them?
- More of a preference point, but for anything more than 3 filter actions, I find it easier to follow if the symbols are listed one per line. Its also nice if the actual method definitions is in the same order as the filters, so that you can read down through the methods and understand the filter chain in one reading.
- Again, more style/preference but I try to avoid calling instance variables in helper methods as much as possible. I find it better to pass them in as method arguments so that you know what dependencies the methods have. This is obviously not possible with filter methods, which is why keeping them simple seems better.
- Specs should pass on each commit, because it makes it easier to verify that the commit's changes are working as intended. Would be really nice for Travis to check this.

app/models/info_request.rb
```diff
@@ -1011,6 +1011,23 @@ def foi_fragment_cache_directories
     directories
   end
 
+  def is_followupable?(incoming_message)
+    if is_external?
+      @followup_bad_reason = "external"
+      return false
+    end
+    unless OutgoingMailer.is_followupable?(self, incoming_message)
+      if public_body.is_requestable?
```
lizconlan: @houndci one of us is not reading that properly

app/controllers/followup_controller.rb
```diff
@@ -0,0 +1,189 @@
...
+    if params[:reedit]
+      render :action => 'new'
+      return
+    end
+  end
+
+  def check_request_matches_incoming_message
+    if @incoming_message and @info_request != @incoming_message.info_request
+      raise ActiveRecord::RecordNotFound.
+              new("Incoming message #{@incoming_message.id} does not belong " \
```
lizconlan: I... I'm not sure. I take your point but I like that new lines up under the calling object. Will wait for a human adjudicator

app/models/info_request.rb
garethrees: It means do this:

```ruby
@followup_bad_reason = if public_body.is_requestable?
  "unexpected followupable inconsistency"
else
  public_body.not_requestable_reason
end
```

app/controllers/followup_controller.rb
garethrees: ¯\_(ツ)_/¯

app/models/info_request.rb
lizconlan: aha!

garethrees:
Superseded by https://github.com/mysociety/alaveteli/pull/3053

PR #2977, opened 2016-01-05
Add Thin cluster setup steps to the manual install page

### LINKED ISSUE #2952: Instructions for running thin with more than one process
Currently we just give instructions for setting up one thin server, not a cluster. Presumably this will struggle to handle larger amounts of traffic.

lizconlan:
Closes #2952

### DIFF
--- a/docs/installing/manual_install.md
+++ b/docs/installing/manual_install.md
@@ -401,6 +401,7 @@ Run the following to get the server running:
       --environment=production \
       --user=alaveteli \
       --group=alaveteli \
+      --servers=1 \
       start
 
 By default the server listens on all interfaces. You can restrict it to the
@@ -498,6 +499,8 @@ Start the application:
   e.g. `alaveteli`
 * `site`: a string to identify your alaveteli instance
 * `user`: the user that the software runs as
+* `cpus`: the number of CPU cores your server has - run `nproc` to find what
+  this number should be. This controls how many thin servers the daemon starts.
 
 There is a rake task that will help to rewrite this file into one that is
 useful to you. Change the variables to suit your installation.
@@ -507,6 +510,7 @@ useful to you. Change the variables to suit your installation.
       DEPLOY_USER=alaveteli \
       VHOST_DIR=/var/www \
       VCSPATH=alaveteli \
+      CPUS=1 \
       SITE=alaveteli \
       SCRIPT_FILE=/var/www/alaveteli/config/sysvinit-thin.example > /etc/init.d/alaveteli
     popd
@@ -522,7 +526,7 @@ Start the application:
 
 One of the cron jobs refers to a script at `/etc/init.d/alaveteli-alert-tracks`. This
 is an init script, which can be generated from the
-`config/alert-tracks-debian.example` template. This script sends out emails to users subscribed to updates from the site – known as [`tracks`]({{ page.baseurl }}/docs/installing/email/#tracks-mail) – when there is something new matching their interests.
+`config/alert-tracks-debian.example` template. This script sends out emails to users subscribed to updates from the site – known as [`tracks`]({{ page.baseurl }}/docs/installing/email/#tracks-mail) – when there is something new matching their interests
 
 **Template Variables:**
 
@@ -636,7 +640,7 @@ occurrences of `www.example.com` to your URL
       /etc/apache2/sites-available/alaveteli
 
 Disable the default site and enable the `alaveteli` VirtualHost
-  
+
     a2dissite default
     a2ensite alaveteli
 
@@ -728,6 +732,21 @@ production server**. Replace `www.example.com` with your domain name.
       -subj /CN=www.example.com
     chmod 640 /etc/ssl/certs/www.example.com.cert
 
+If you have configured thin to use more than one server, you will need to edit the alaveteli upstream
+directive in your <code>/etc/nginx/sites-enabled/alaveteli_https</code> file so
+that it has a `server` line for each instance you are running (otherwise nginx
+will only know how to send requests to the first server process). Each address
+line needs to have a unique port number - start at `3000` and add `1` for each
+new process. For example if you started your cluster with 4 servers, the
+upstream directive should look like this:
+
+    upstream alaveteli {
+        server 127.0.0.1:3000;
+        server 127.0.0.1:3001;
+        server 127.0.0.1:3002;
+        server 127.0.0.1:3003;
+    }
+
 Check the configuration and fix any issues
 
     service nginx configtest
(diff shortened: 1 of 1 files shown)


lizconlan:
I'm regarding this as a first attempt - I'm not convinced that the changes described in a070c58 (to edit the sysVinit file) are easy to follow. It might be simpler to modify the `.example` file to include a 
`--servers=1` option for the start and stop commands, which looks odd but is valid syntax.

This could make for a less intimidating (and easier to follow) change if it's just changing a number in a couple of places in a file rather than asking people to paste things in and alter them to suit (with potentially confusing results).

garethrees:
> It might be simpler to modify the `.example` file to include a `--servers=1`

Yeah, then its obvious what to change. I'll give this some thought while reviewing.

docs/installing/manual_install.md
```diff
@@ -411,6 +411,13 @@ site in action.
 
 You can daemonize the process by starting it with the `--daemonize` option.
 
+
+If you anticipate getting a lot of traffic, you can configure thin to run as a cluster
```
garethrees: I think think we should generate the daemon through the rake task as suggested https://github.com/mysociety/alaveteli/issues/2952#issuecomment-169300575, but without dynamically setting the number of servers.

I think setting a default of 1 is the best idea, so that the init file has the `--servers` flag in all the correct places.

garethrees: This section can then be moved to Generate application daemon/THIN/Template Variables

docs/installing/manual_install.md
```diff
@@ -514,6 +521,33 @@ useful to you. Change the variables to suit your installation.
     chown root:alaveteli /etc/init.d/alaveteli
     chmod 754 /etc/init.d/alaveteli
 
+**Set up the cluster (optional):**
```
garethrees: This section can be removed, as we'll be making the `--server` flag available to the rake task.

docs/installing/manual_install.md
```diff
@@ -728,6 +762,18 @@ production server**. Replace `www.example.com` with your domain name.
       -subj /CN=www.example.com
     chmod 640 /etc/ssl/certs/www.example.com.cert
 
+If you are running a thin cluster, you will need to edit the top section of
```
garethrees:
```diff
- need to edit the top section of
+ need to edit the alaveteli upstream directive
```

docs/installing/manual_install.md
```diff
@@ -728,6 +762,18 @@ production server**. Replace `www.example.com` with your domain name.
...
+your <code>nginx-ssl.conf.example</code> file so that it has a `server` line
```
garethrees: Probably want to edit the generated nginx.conf, rather than the example which is checked in to source control

docs/installing/manual_install.md
```diff
@@ -728,6 +762,18 @@ production server**. Replace `www.example.com` with your domain name.
...
+for each instance you are running (otherwise nginx will only know how to send
+requests to the first server process). If you started your cluster with 3 servers,
```
garethrees: use 4 as an example, because 3 is an unlikely number given you're likely to have a 1,2 or 4 core machine

docs/installing/manual_install.md
```diff
@@ -728,6 +762,18 @@ production server**. Replace `www.example.com` with your domain name.
...
+the top section of your file should look like this.
```
garethrees:
```diff
- top section
+ alaveteli upstream directive
```

garethrees: Would also note that you should start at `3000` and add `1` for each additional thin process

garethrees:
Looks good. Made some inline suggestions, but up for discussion :+1:

docs/installing/manual_install.md
lizconlan: I used 3 as it's used throughout the thin docs, I imagine - although I've not seen it explained anywhere - that it's to avoid tying up all the CPU resources running server processes but I'm not attached to it

garethrees: I think keeping it simple "1 server per CPU core" is the better message here http://serverfault.com/a/222858.

docs/installing/manual_install.md
```diff
@@ -507,6 +509,7 @@ useful to you. Change the variables to suit your installation.
       DEPLOY_USER=alaveteli \
       VHOST_DIR=/var/www \
       VCSPATH=alaveteli \
+			CPUS=1 \
```
garethrees: Is this a _tab_ :scream::scream::scream:

docs/installing/manual_install.md
```diff
@@ -498,6 +499,7 @@ Start the application:
   e.g. `alaveteli`
 * `site`: a string to identify your alaveteli instance
 * `user`: the user that the software runs as
+* `cpus`: the number of CPU cores your server has - run `nproc` to find what this number should be
```
garethrees: I'd add the following to the end of this:

```diff
- * `cpus`: the number of CPU cores your server has - run `nproc` to find what this number should be
+ * `cpus`: the number of CPU cores your server has - run `nproc` to find what this number should be. This controls how many thin servers the daemon starts.
```

docs/installing/manual_install.md
```diff
@@ -728,6 +731,21 @@ production server**. Replace `www.example.com` with your domain name.
       -subj /CN=www.example.com
     chmod 640 /etc/ssl/certs/www.example.com.cert
 
+If you are running a thin cluster, you will need to edit the alaveteli upstream
```
garethrees:
```diff
- If you are running a thin cluster
+ If you have configured thin to use more than one server
```

docs/installing/manual_install.md
```diff
@@ -728,6 +731,21 @@ production server**. Replace `www.example.com` with your domain name.
...
+will only know how to send requests to the first server process). Each address
+line needs to have a unique port number - start at `3000` and add `1` for each
+new process. For example if you started your cluster with 4 servers, the
+upstream directive should look like this:
+
+    upstream alaveteli {
+        server 127.0.0.1:3000;
+        server 127.0.0.1:3001;
+        server 127.0.0.1:3002;
+				server 127.0.0.1:3003;
```
garethrees: tab

docs/installing/manual_install.md
```diff
@@ -755,6 +773,21 @@ Disable the default site and enable the `alaveteli` server
     ln -s /etc/nginx/sites-available/alaveteli \
       /etc/nginx/sites-enabled/alaveteli
 
+If you are running a thin cluster, you will need to edit the alaveteli upstream
```
garethrees:
```diff
- If you are running a thin cluster
+ If you have configured thin to use more than one server
```

docs/installing/manual_install.md
```diff
@@ -755,6 +773,21 @@ Disable the default site and enable the `alaveteli` server
...
+will only know how to send requests to the first server process). Each address
+line needs to have a unique port number - start at `3000` and add `1` for each
+new process. For example if you started your cluster with 4 servers, the
+upstream directive should look like this:
+
+    upstream alaveteli {
+        server 127.0.0.1:3000;
+        server 127.0.0.1:3001;
+        server 127.0.0.1:3002;
+				server 127.0.0.1:3003;
```
garethrees: tab

garethrees:
Couple of minor comments, but looks great!

docs/installing/manual_install.md
lizconlan: \*fires Markdown editor\*

PR #2979, opened 2016-01-05
Fix spanish translation for "This person's annotations"

alfonsocora:
"This person's annotations" is currently translated to "Tus comentarios" ("Your annotations") in spanish, which leads to confusion in the user profile page. This commit changes the translation to "Comentarios de esta persona" which is the literal translation.


garethrees:
Thanks! We translate Alaveteli through [Transifex](http://alaveteli.org/docs/customising/translation/), so I've added these changes there. Let us know if you have a Transifex account and would like to be added to the Spanish translators team. If not, we still appreciate pull requests like this.

alfonsocora:
@garethrees gotcha! I just joined the team in Transifex. Thanks!

PR #2991, opened 2016-01-11
Allow sysvinit to be generated with support for thin clustering

### LINKED ISSUE #2952: Instructions for running thin with more than one process

lizconlan:
Supplying a value for CPUS to the generation script will start
(and stop) the specified number of thin processes. If no value
is supplied, it will default to launching a single process.

Part of fix #2952

### DIFF
--- a/lib/tasks/config_files.rake
+++ b/lib/tasks/config_files.rake
@@ -49,6 +49,7 @@ namespace :config_files do
       :vhost_dir => ENV['VHOST_DIR'],
       :vcspath => ENV.fetch('VCSPATH') { 'alaveteli' },
       :site => ENV.fetch('SITE') { 'foi' },
+      :cpus => ENV.fetch('CPUS') { '1' },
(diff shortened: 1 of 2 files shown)


lib/tasks/config_files.rake
```diff
@@ -49,6 +49,7 @@ namespace :config_files do
       :vhost_dir => ENV['VHOST_DIR'],
       :vcspath => ENV.fetch('VCSPATH') { 'alaveteli' },
       :site => ENV.fetch('SITE') { 'foi' },
+      :cpus => ENV.fetch('CPUS') {'1'},
```
garethrees: Hashes should be padded:

```diff
- :cpus => ENV.fetch('CPUS') {'1'},
+ :cpus => ENV.fetch('CPUS') { '1' },
```

garethrees:
Looks great :+1:

PR #2992, opened 2016-01-12
Symlink down file instead of copying

garethrees:
Fixes https://github.com/mysociety/Non-Site-Specific/issues/672


garethrees:
Also need to update `/d/s/vhosts/alaveteli/sysvinit-passenger.ugly` on merge

PR #2993, opened 2016-01-13
Release/0.23.2.0

lizconlan:


lizconlan:
@crowbot good point, I'll revert 04e0e08

lizconlan:
The original class methods have been renamed to `old_unclassified_params_old` and `find_old_unclassified_old` and can be removed safely once benchmarking against the replacement methods has been completed (assuming there is a genuine performance improvement)

garethrees:
Don't know if its worth looking in to https://github.com/github/scientist to have real metrics on this?

lizconlan:
Forgot to add - possible down time when deploying this as it's a non trivial db migration on a large dataset https://github.com/mysociety/alaveteli/blob/50dd31997238e10bd0a55caae68037ee3cdb27fb/db/migrate/20151104131702_add_last_public_response_at_to_info_request.rb

PR #2996, opened 2016-01-13
Add APPDIR to daemon files.

crowbot:
This ensures that the daemons are running in the app directory. When
used with an .rbenv-version file in the application to control ruby
version, this will ensure that the correct ruby version is used.

### DIFF
--- a/config/alert-tracks-debian.example
+++ b/config/alert-tracks-debian.example
@@ -20,6 +20,7 @@ PIDFILE=$PIDDIR/!!(*= $daemon_name *)!!.pid
 LOGDIR=!!(*= $vhost_dir *)!!/!!(*= $vcspath *)!!/log
 LOGFILE=$LOGDIR/!!(*= $daemon_name *)!!.log
 DUSER=!!(*= $user *)!!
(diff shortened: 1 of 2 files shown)


config/alert-tracks-debian.example
```diff
@@ -20,6 +20,7 @@ PIDFILE=$PIDDIR/!!(*= $daemon_name *)!!.pid
 LOGDIR=!!(*= $vhost_dir *)!!/!!(*= $vcspath *)!!/log
 LOGFILE=$LOGDIR/!!(*= $daemon_name *)!!.log
 DUSER=!!(*= $user *)!!
+APPDIR=!!(*= $vhost_dir *)!!/!!(*= $vcspath *)!!
```
garethrees: Might be worth tidying up the above paths to be based on APPDIR?

crowbot: Yep, but I would rather do that outside a hotfix.

garethrees:
Merged in to hotfix PR https://github.com/mysociety/alaveteli/pull/3012

PR #2997, opened 2016-01-13
Fix collation support connection

garethrees:


garethrees:
Merged in to hotfix PR https://github.com/mysociety/alaveteli/pull/3012

PR #3000, opened 2016-01-13
Move graph generation to ruby

### LINKED ISSUE #675: /admin/stats request-creation-graph and user-use-graph require additional config
These scripts [don't read the password](https://github.com/mysociety/alaveteli/blob/develop/script/request-creation-graph#L20) out of the `database.yml` file so you need to make sure your deployment user can authenticate to the database another way, e.g. a [pgpass file](http://www.postgresql.org/docs/8.1/static/libpq-pgpass.html).

We need to either make these read the password or document the fact you need to set up a `.pgpass` file.

lizconlan:
Ports the graph generation shell scripts to rake tasks (currently without description so they don't clutter up the list of rake tasks - happy to change this if it doesn't feel right), the shell scripts now call the appropriate rake task.

Assumes that we don't want to continue to send site admins messages about empty y ranges but are happy for the graph to be generated if not all the possible datasets are populated (e.g. there is request data but nothing has gone postal). Retains a warning about there being no data at all as that could indicate something going wrong.

Fixes #675


garethrees:
PR 3000! :tada: :birthday:

lizconlan:
note to self - will require a line in the changelog

henare:
So good :ok_hand:

garethrees:
:birthday:

PR #3004, opened 2016-01-14
Remove package requirement for xlhtml

### LINKED ISSUE #2566: xlhtml is no longer supported and has open security issues
xlhtml is in Alaveteli's packages list, but 'It's dead upstream, is orphaned, has open security issues and alternatives exist.' (from https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=743429)

There's no xlhtml package in the Debian Jessie repo, so we need to use an alternative.

crowbot:
The executable provided isn't used anywhere.

Closes #2566.

### DIFF
--- a/config/packages
+++ b/config/packages
@@ -41,4 +41,3 @@ uuid-dev
 wkhtmltopdf-static
 wv
 xapian-tools
-xlhtml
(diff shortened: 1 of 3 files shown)


garethrees:
:+1:

PR #3005, opened 2016-01-14
Support and test against later versions of ruby

crowbot:


garethrees:
The build passed, so, guess its good!

PR #3007, opened 2016-01-19
Send public holiday reminder email to site admins

### LINKED ISSUE #2896: Send automated "update holidays" email direct to site admins
And change wording of the automated "holidays need updating" email so that it can be sent to site admins rather than (just) mySociety staffers for the 2017 notifications on.

May depend on #2052 (i.e. getting everyone onto at least v0.21 so the Holiday upkeep feature is available in the admin interface).

lizconlan:
Sends a longer email which is suitable for sending to the site admins rather than the Alaveteli team.

Moves the email sending to a rake task and uses a mailer template so that it can be overridden by a theme to account for variations in practice or local language-specific instructions.

Currently doesn't attempt to translate the subject line which may not be correct.

Closes #2896

### DIFF
--- a/app/mailers/reminder_mailer.rb
+++ b/app/mailers/reminder_mailer.rb
@@ -1,0 +1,16 @@
+# -*- encoding : utf-8 -*-
+
+class ReminderMailer < ApplicationMailer
+  # Send reminder message to administrator
+  def public_holidays(name, email, subject)
+    # Return path is an address we control so that SPF checks are done on it.
+    headers(
+      'Return-Path' => blackhole_email,
+      'Reply-To' => MailHandler.address_from_name_and_email(name, email)
+    )
+
+    mail(:from => MailHandler.address_from_name_and_email(name, email),
+         :to => MailHandler.address_from_name_and_email(name, email),
+         :subject => subject)
+  end
+end
--- a/config/crontab-example
+++ b/config/crontab-example
@@ -37,8 +37,7 @@ MAILTO=!!(*= $mailto *)!!
 48 2 * * * !!(*= $user *)!! !!(*= $vhost_dir *)!!/!!(*= $vcspath *)!!/script/user-use-graph
 
 # Once a year :)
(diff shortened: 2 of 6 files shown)


config/crontab-example
```diff
@@ -37,7 +37,8 @@ MAILTO=!!(*= $mailto *)!!
 48 2 * * * !!(*= $user *)!! !!(*= $vhost_dir *)!!/!!(*= $vcspath *)!!/script/user-use-graph
 
 # Once a year :)
-0 0 1 11 * !!(*= $user *)!! /bin/echo "A year has passed, please update the public holidays for the Freedom of Information site, thank you."
+
+0 0 1 11 * !!(*= $user *)!! !!(*= $vhost_dir *)!!/!!(*= $vcspath *)!!/script/runner reminder:public_holidays
```
garethrees: `rake reminder:public_holidays`?

app/mailers/reminder_mailer.rb
```diff
@@ -0,0 +1,15 @@
+# -*- encoding : utf-8 -*-
+
+class ReminderMailer < ApplicationMailer
+  # Send reminder message to administrator
+  def public_holidays(name, email, subject)
+    @message = message
```
garethrees: Where's `message` coming from?

app/mailers/reminder_mailer.rb
```diff
@@ -0,0 +1,15 @@
...
+
+    # Return path is an address we control so that SPF checks are done on it.
+    headers('Return-Path' => blackhole_email, 'Reply-To' => MailHandler.address_from_name_and_email(name, email))
```
garethrees: Line length <= 80 :)

config/crontab-example
garethrees: Not sure but don'I expect you'll need to be in the app directory for this task as we don't have binstubs installed to be abled to call the bundled rake. Definitely worth testing this in your VM's crontab

garethrees:
> Currently doesn't attempt to translate the subject line which may not be correct.

We don't currently translate the admin interface, so no need here either

garethrees:
:+1:

PR #3008, opened 2016-01-19
Add a temporary rake task to output requests made to top 20 bodies

jacksonj04:
This is to support research work on semantic tagging.

### DIFF
--- a/lib/tasks/temp.rake
+++ b/lib/tasks/temp.rake
@@ -71,6 +71,20 @@ EOF
 
   end
 
+  desc 'Output all the requests made to the top 20 public bodies'
+  task :get_top20_body_requests => :environment do
+
+    require 'csv'
+    puts CSV.generate_line(["public_body_id", "public_body_name", "request_created_timestamp", "request_title", "request_body"])
+
+    PublicBody.limit(20).order('info_requests_visible_count DESC').each do |body|
+      body.info_requests.where(:prominence => 'normal').find_each do |request|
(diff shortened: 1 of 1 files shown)


lib/tasks/temp.rake
```diff
@@ -71,6 +71,20 @@ EOF
...
 
+  desc 'Output all the requests made to the top 20 public bodies'
+  task :get_top20_body_requests => :environment do
+
+    require 'csv'
+    puts CSV.generate_line(["public_body_id", "public_body_name", "request_created_timestamp", "request_title", "request_body"])
+
+    PublicBody.limit(20).order('info_requests_visible_count DESC').each do |body|
+      body.info_requests.where(:prominence => 'normal').find_each do |request|
+        puts CSV.generate_line([request.public_body.id, request.public_body.name, request.created_at, request.url_title, request.initial_request_text.gsub("\r\n", " ").gsub("\n", " "), '"'])
```
garethrees: Seems to be adding an extra field at the end of your desired line:
> `Thanks!  Bob",""""`


```csv
public_body_id,public_body_name,request_created_timestamp,request_title,request_body
2,Geraldine Quango,2007-10-11 19:15:57 +0100,why_do_you_have_such_a_fancy_dog,"Hi,  Why do you have such a fancy dog?  Thanks!  Bob",""""
```

garethrees: FYI, if the time format isn't ideal you can use [`strftime`](http://ruby-doc.org/core-1.9.3/Time.html#method-i-strftime)

garethrees:
Looks good other than the extra column thats getting added.

PR #3011, opened 2016-01-20
Improvements to the Romanian translation

andreicristianpetcu:
I just changed the app.po file, I don't know why the "new commonlib" commit appears  b39a083.


andreicristianpetcu:
I wanted to use transifex. We were at a sort of hackathon somewhere outside Bucharest and snow started, Internet and electricity fell and we translated offline. I know transifex is the preferred way.

andreicristianpetcu:
I'll try that.

Issue #2989, opened 2016-01-08
Tweak Alaveteli base design to better handle long strings in translations


crowbot:
Some languages are more verbose than others - ideally, the alaveteli design would be robust to longer (and shorter) strings

crowbot:
The main nav and the language picker and logged in nav are particularly susceptible to overflow

zarino:
Example of alavetelitheme header in Turkish, at the widest screen size:

![screen shot 2016-01-21 at 14 24 21](https://cloud.githubusercontent.com/assets/739624/12483042/c2e65172-c04a-11e5-92f9-9b3009637935.png)

And at the narrowest size before the mobile menu kicks in:

![screen shot 2016-01-21 at 14 25 13](https://cloud.githubusercontent.com/assets/739624/12483062/d66ab800-c04a-11e5-9f28-84266c59ebc2.png)

The language switcher doesn't support long language names particularly well either: (the dropdown menu is limited to the width of the currently active language plus the width of the triangle)

![screen shot 2016-01-21 at 14 29 39](https://cloud.githubusercontent.com/assets/739624/12483188/6fdb7e52-c04b-11e5-931b-e37962e78543.png)

We also had a problem recently, where French translations of the man nav bar items weren't fitting in asktheeu-theme (which is a very light customisation of alavetelitheme). In the end we picked shorter words for the links, but I'm told the shorter versions aren’t ideal.

PR #3014, opened 2016-01-21
Stop user nav overlapping logo when text strings are long

zarino:
The "logged_in_bar" now has a limited width, and if translations result in links that are too long to fit 60% of the header width, the links will break onto a second line. Whether they are on one line or two, the links are still vertically centred.

Part of #2989.

Note that there is also a corresponding change in https://github.com/mysociety/alavetelitheme/issues/72, where the Alavetelitheme logged_in_bar is updated to benefit from these changes.

screen shot 2016-01-21 at 16 48 05

screen shot 2016-01-21 at 16 45 09

### DIFF
--- a/app/assets/stylesheets/responsive/_header_layout.scss
+++ b/app/assets/stylesheets/responsive/_header_layout.scss
@@ -201,33 +201,33 @@
 }
 
 #logged_in_bar{
-  @include respond-min( $main_menu-mobile_menu_cutoff ){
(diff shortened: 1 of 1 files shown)


garethrees:
Merged in https://github.com/mysociety/alaveteli/commit/614f93bb937bd6fd4ce4ee520e431974fe40d459 – thanks!

PR #3015, opened 2016-01-22
Stop long nav links from disappearing behind search box

zarino:
Part of #2989.

This still means we have to massage nav link wording if we want the links to fit on a single line. But at least now, if the links *are* too long, they’ll wrap onto a new line so people can still click them.

Before:

before

After:

after


garethrees:
Thanks!

Issue #2011, opened 2014-12-01
Reduce complexity around finding a CensorRule subject


garethrees:
https://github.com/mysociety/alaveteli/pull/1987 will introduce complexity for finding the record a `CensorRule` should be associated with. Once the routes have been converted to something more standard we should be able to reduce the complexity as described in https://github.com/mysociety/alaveteli/commit/bb233e381c2607788a91534a46c89f5fae6dce06#commitcomment-8658339.

Here's a code snapshot for immediate reference, but see the commit comments for further context:

``` ruby
module AdminHelper
  def public_and_admin_links(subject)
      "#{ public_link(subject) } #{ admin_link(subject) }"
  end

  def public_link(subject)
      title = "View #{ subject.class.model_name.human } on public website"
      link_to(eye, polymorphic_path(subject), :title => title)
  end

  def admin_link(subject)
      title = "View #{ subject.class.model_name.human } on admin website"
      link_to(h(subject.name), polymorphic_path([:admin, subject]), :title => title)
  end
end
```

``` ruby
    def new
        if params[:info_request_id]
            @subject = InfoRequest.find(params[:info_request_id])
            @censor_rule = @subject.censor_rules.build
            @form_url = admin_info_request_censor_rules_path(@subject)
        end

        if params[:user_id]
            @subject = User.find(params[:user_id])
            @censor_rule = @subject.censor_rules.build
            @form_url = admin_user_censor_rules_path(@subject)
        end

        @subject ||= nil
        @censor_rule ||= CensorRule.new
        @form_url ||= admin_rule_create_path
    end
```

``` erb
<%= form_for @censor_rule, :url => @form_url, :class => "form form-horizontal" do %>
  <%= render :partial => 'form', :locals => { :subject => @subject } %>
```

``` erb
<div class="well">
  <%= _("Applies to") %>
    <%= public_and_admin_links(subject) if subject %>
  <% end %>
</div>
```

We could even strip the controller down further if the routes were in a format `polymorphic_path` could work with:

``` ruby
if params[:info_request_id]
  @subject = InfoRequest.find(params[:info_request_id])
end

if params[:user_id]
  @subject = User.find(params[:user_id])
end

@censor_rule = @subject.censor_rules.build || CensorRule.new
@form_url =  polymorphic_path([:admin, subject, @censor_rule.class]) || admin_rule_create_path
```

garethrees:
[...]

PR #3021, opened 2016-01-25
[#2551] Manage global and public body censor rules through admin UI

### LINKED ISSUE #2551: Make default text masks localisable/configurable

garethrees:
Fixes #2551

screen shot 2016-01-25 at 12 38 46

### DIFF
--- a/app/controllers/admin_censor_rule_controller.rb
+++ b/app/controllers/admin_censor_rule_controller.rb
@@ -63,6 +63,12 @@ class AdminCensorRuleController < AdminController
       @censor_rule = @censor_user.censor_rules.build(censor_rule_params)
       @form_url = admin_user_censor_rules_path(@censor_user)
     end
+
+    if params[:body_id]
+      @public_body = PublicBody.find(params[:body_id])
+      @censor_rule = @public_body.censor_rules.build(censor_rule_params)
+      @form_url = admin_body_censor_rules_path(@public_body)
+    end
   end
 
   def set_editor
@@ -92,6 +98,8 @@ class AdminCensorRuleController < AdminController
     elsif @censor_rule.user
       @censor_rule.user.expire_requests
       redirect_to admin_user_url(@censor_rule.user)
+    elsif @censor_rule.public_body
+      redirect_to admin_body_url(@censor_rule.public_body)
     end
   end
 end
--- a/app/controllers/admin_censor_rule_controller.rb
+++ b/app/controllers/admin_censor_rule_controller.rb
@@ -8,7 +8,7 @@
 class AdminCensorRuleController < AdminController
 
   before_filter :set_editor, :only => [:create, :update]
-  before_filter :find_and_check_rule, :only => [:edit, :update, :destroy]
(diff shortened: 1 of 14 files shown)


app/controllers/admin_censor_rule_controller.rb
```diff
@@ -63,18 +63,25 @@ def set_info_request_and_censor_rule_and_form_url
       @censor_rule = @censor_user.censor_rules.build(censor_rule_params)
       @form_url = admin_user_censor_rules_path(@censor_user)
     end
+
```
crowbot: `set_info_request_and_censor_rule_and_form_url` could really now do with a rename.

app/views/admin_censor_rule/index.html.erb
```diff
@@ -0,0 +1,39 @@
...
+    <div class="page-header">
+      <h1><%= @title %></h1>
+    </div>
+  </div>
+</div>
+
+<%= render :partial => 'show',
+           :locals => { :censor_rules => @censor_rules } %>
+
+<hr />
```
crowbot: I think there needs to be some clear notice on this page that all that's being displayed here is global censor rules and that user, request and public body specific censor rules may also exist and can be seen, edited and created on the admin pages for those users, requests and public bodies.

app/controllers/admin_censor_rule_controller.rb
```diff
@@ -85,6 +92,10 @@ def censor_rule_params
     end
   end
 
+  def global_censor_rule_params
+    censor_rule_params.merge(:allow_global => true)
```
crowbot: the `:allow_global` flag was introduced to limit the creation of global rules from the admin interface. Given that we're now allowing that, I don't think it's really serving much purpose.

app/views/admin_censor_rule/index.html.erb
garethrees: Fixed in 93e0406

app/controllers/admin_censor_rule_controller.rb
garethrees: Its verbose, but it's doing what it says on the tin. A way to avoid this would be to make `CensorRule` a [polymorphic association](https://github.com/mysociety/alaveteli/issues/2011).

crowbot: Ah, my point is that it's not doing what it says on the tin anymore - it's setting a user or a public body instead of an info request, depending on circumstances. So the method name is now actually misleading - I'd prefer `set_instance_variables` or something similar.

garethrees: Ah yeah, you're right. Fixed in https://github.com/mysociety/alaveteli/pull/3061.

Issue #2898, opened 2015-11-12
Remove dynamically constructed translation strings


lizconlan:
Part of #2676

lizconlan:
In-progress branch updated, reluctantly parked due to hand tendon issues :disappointed:

The code itself is pretty much there (although the commits aren't as tidy as I'd like) - was part way reorganising the work into a more manageable series of commits and attempting to regenerate and manually fix up the `.po` files

lizconlan:
(some of the work in my branch may have been overtaken by changes made elsewhere, it was created a while ago)

PR #3023, opened 2016-01-25
[#3022] Fix untranslatable string

### LINKED ISSUE #3022: Untranslatable string 'Please tell us more:
In `http://www.site.com/request/slug-1/describe/error_message`:

```
Please tell us more:
```

Translation is available on Transifex (string id 588) but it is not included on web.

Screenshot (reproducible on all themes, screen is WTDK official theme):
https://qsnapnet.com/snaps/ibi7lm805dzvkj4

Link with (Croatian issue): https://github.com/codeforcroatia/alaveteli/issues/10

garethrees:
Fixes #3022

### DIFF
--- a/app/views/request/describe_state_message.html.erb
+++ b/app/views/request/describe_state_message.html.erb
@@ -10,7 +10,9 @@
 
 <%= form_for :incoming_message, :url => describe_state_url(:id => @info_request.id) do |f| %>
   <p>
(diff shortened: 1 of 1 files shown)


garethrees:
> If there are several of these untranslated strings, we should probably make a hotfix with them.

There are probably quite a few, but should get handled in https://github.com/mysociety/alaveteli/issues/2898

Issue #2998, opened 2016-01-13
http://alaveteli.org/docs/running/upgrading/ should talk about file permissions


crowbot:
Specifically that `git pulls` and `bundle` calls should be run as the alaveteli user.

schlos:
Adding some suggestions and ideas:

1) upgrade guide is simultaneously describing procedure for Capistrano deployments and non-Capistrano deployments
- I would suggest clearly split whole guide into titles "Guide for Capistrano deployments" and "Guide for normal deployments" (w/o automation deployment tools)
- this includes only changes for titles "How to upgrade the code" and "Run the post-deploy script"

I'll focus only on non-Capistrano procedure below:

2) "you can simply upgrade by running git pull" should also state that this command should be run from directory /var/www/your-deployment-name/alaveteli/ and can be run as *root* user (or any other current user) (?) (TBC, I'm missed to note that)

3) "you should always run the script scripts/rails-post-deploy after each deployment" should also state that this command should be run from directory /var/www/your-deployment-name/alaveteli/ with "sudo -u alaveteli ..."

4) guide might include some known (or sometime occurring problems and how to solve it). i.e. if you get permission error running "sudo -u alaveteli RAILS_ENV=production script/rails-post-deploy" like "Errno::EACCES: Permission denied - name-x.x.x.gem" then make sure that folders in "/home/alaveteli/" have right user/group permissions (it must be alaveteli:alaveteli). Check by running "ls -altr /home/alaveteli/bundle/ruby/x.x.x/gems" where x.x.x represents your Ruby version. Set correct permissions by running "sudo chown -R alaveteli:alaveteli /home/alaveteli/" then try running post-deploy-script again
- this can be added as (i) (information) text block, and it can be place to add more helpful tips there.

lizconlan:
>  guide might include some known (or sometime occurring problems and how to solve it

This is probably best addressed by creating a "**Troubleshooting**" section, in the first instance at the bottom of the current file (or page, from the reader's perspective). At this point it occurs to me that my misspent youth appears to have involved instruction manuals.

Putting this off for this evening as I'm not awake enough to attempt to extract useful examples from recent emails.

garethrees:
> This is probably best addressed by creating a "Troubleshooting" section

We do have a bit of one at http://alaveteli.org/docs/installing/manual_install/#troubleshooting.

[...]

PR #3036, opened 2016-02-01
Add a generic task for converting config files

garethrees:
Doesn't check for required variables, other than the script to convert.

Useful for less-used scripts like config/run-with-rbenv-path.example

### DIFF
--- a/lib/tasks/config_files.rake
+++ b/lib/tasks/config_files.rake
@@ -94,4 +94,36 @@ namespace :config_files do
     end
   end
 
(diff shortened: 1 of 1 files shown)


garethrees:
This will probably be useful when talking about Capistrano stuff in https://github.com/mysociety/alaveteli/issues/2998. Related to https://github.com/mysociety/alaveteli/pull/3026.

lizconlan:
:+1: 

Slightly nervous that a required variable will get missed but that seems pretty unavoidable, the supplied defaults should cover most eventualities

Issue #624, opened 2012-09-24
automated checking of mail delivery logs


confirmordeny:
Sometimes public bodies claim not to have received FOI requests.  When requesters write to us a skilled member of them (i.e. not me because I don't know how!) writes back and says something like:

I've checked the mail delivery logs for
http://www.whatdotheyknow.com/request/[...] and I can
confirm that both your messages were successfully delivered to [public authority]'s mail servers.

I just wonder whether it might  be possible to show on the request page a little green light or something that shows each outward message was successful delivered.  Perhaps it could also show the date and time of delivery. I have no idea how difficult it is.

It is far from high priority.  I feel it would add something to the trustworthiness of the record because people could think to themselves "yes they really did received this message"

RichardTaylor:
ICO decision notice FS50559082 concluded, after considering message delivery evidence from WhatDoTheyKnow

>The Commissioner considers that there is sufficient evidence to show that the Cabinet Office had received the request on the date it was sent by the complainant. As it did not issue its refusal notice within 20 working days, it breached section 17(5) of the Act.

https://www.whatdotheyknow.com/request/appointment_of_qualified_persons#comment-64694

zarino:
How big a deal do we want to make of this? A very subtle solution might be:

![unknown-status](https://cloud.githubusercontent.com/assets/739624/12112841/0b231864-b397-11e5-8ecd-785d752705c5.jpg)

(where the question mark is a link to an FAQ item about how we track delivery, and why we've said "Sent" instead of "Delivered" in this case)

And then when it’s accepted by the mail server:

![delivered](https://cloud.githubusercontent.com/assets/739624/12112844/0d3e4e3e-b397-11e5-9230-5c6903a3329b.jpg)

(where the tick reveals a popup or something showing what we mean by "delivered" perhaps even with some geeky mail server log for people who want to use it as some form of evidence)

Just my first thought. Any feedback?

crowbot:
I'm in favour of having the mail server log available, but hidden unless you actively seek it out - I think it would be good to have an indication in the 'delivered' mode that there is more to see - like the question mark in 'sent' mode. This could link to an explanation of the complexities of mail delivery, and an appropriately redacted version of the delivery log. Maybe we should show the requester an uncensored version that they could forward to the authority.

zarino:
@crowbot: What are the situations in which somebody would seek out this proof of an email being delivered?

crowbot:
The authority claiming that they never received the request.

[...]

PR #3039, opened 2016-02-03
[#624] Add OutgoingMessage#mail_server_logs

garethrees:
Part of #624 

Basic framework for collecting exim MTA logs for a specific `OutgoingMessage`.

Based on my current understanding, the last bits to add to get an `OutgoingMessage#delivered?` value are:

```ruby
  def delivered?
    delivery_message ? true : false
  end

  def delivery_message
    m = delivery_line.match(/C="(.*)"/i)
    m ? m[0] : nil
  end

  def delivery_line
    mail_server_logs.
      where("line ILIKE :mta AND line ILIKE :msgtype",
             mta: "%#{ mta_id.last }%", msgtype: "%=>%").
        last.
          line
  end
```

### DIFF
--- a/app/models/mail_server_log.rb
+++ b/app/models/mail_server_log.rb
@@ -67,10 +67,8 @@ class MailServerLog < ActiveRecord::Base
   def self.load_exim_log_data(f, done)
     order = 0
     f.each do |line|
-      sanitised_line = scrub(line)
       order = order + 1
-      emails = email_addresses_on_line(sanitised_line)
-      create_mail_server_logs(emails, sanitised_line, order, done)
+      create_exim_log_line(line, done, order)
     end
   end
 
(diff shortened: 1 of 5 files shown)


garethrees:
Marking as `5 - awaiting review` for a first pass look. The specs are (sometimes) failing because of some sort of ordering issue (5a30cd3) but I don't t think its worth diving in to unless we're happy with the general progress of this.

I'm interested in edge cases that make `mail_server_logs` not produce the output you'd expect.

garethrees:
> I wonder if there's going to be some value in storing the association between the outgoing message and the log lines

Made a note on the original ticket https://github.com/mysociety/alaveteli/issues/624#issuecomment-185202271 to do this at a later stage once we've used the on-the-fly way to check a random sample of messages and logs.

PR #3046, opened 2016-02-09
Don't generate local code coverage reports

garethrees:
Don't think anyone really uses these. Removing because manually
excluding the coverage directory in project-wide searches is tiresome.

Happy to keep these if they help someone.


lizconlan:
:+1: I have a slight reluctance for getting rid of this but can't remember the last time I referred to local code coverage to check something and it's definitely in the way during searches

garethrees:
Great, I'll merge. Next time someone needs local coverage, a PR for a rake task that generates it on-demand would be very welcome :)

Issue #2819, opened 2015-09-29
Workflow for appeals to a regulator


crowbot:
https://github.com/mysociety/whatdotheyknow-theme/issues/17
https://github.com/mysociety/alaveteli/issues/703
https://github.com/mysociety/ipvtheme/issues/7
https://github.com/mysociety/alaveteli/wiki/Complaining-to-ombudsman---feature-requirements

garethrees:
https://www.whatdotheyknow.com/help/unhappy#complaining
  - https://ico.org.uk/concerns/getting/
  - http://www.itspublicknowledge.info/YourRights/YourRights.aspx

garethrees:
> **Feature 1 - letting the regulator know about overdue requests**
>
> 1. Currently, where a response is overdue the site sends a reminder email to the requester. The proposed new feature is that at this point the requester should be able to press a button to "Let the [name of the appropriate regulator] know that [name the public authority] has not responded to this request on time."
> 2. …
>
> https://github.com/mysociety/alaveteli/wiki/Complaining-to-ombudsman---feature-requirements#feature-1---letting-the-regulator-know-about-overdue-requests

I think this feature would be best served by a manually sent email from the Alaveteli admin team. We could write a rake task that groups late requests by authority and prints them with their URL and number of days late. I think this batched-style would be more effective, as if the ICO spots a pattern emerging after several warnings they have more grounds to take substantial action.

```sh
$ rake stats:overdue_requests
Another Body
  https://alaveteli.example.org/request/abc, 60 days late
  https://alaveteli.example.org/request/cde, 20 days late

Some Body
  https://alaveteli.example.org/request/jhwe, 4 days late
  https://alaveteli.example.org/request/cwec, 1 day late
```

garethrees:
Simplest UI workflow

![doc - 2 oct 2015 16-34](https://cloud.githubusercontent.com/assets/282788/10250635/e2d4a430-6923-11e5-96b4-392d08f84ee4.jpg)

garethrees:
I think this is going to be pretty tricky without dealing with public/private. I don't really see the advantage to private complaints – people may as well just fill in the ICO's form themselves. We should deal with this in #8.

garethrees:
[...]

PR #3047, opened 2016-02-09
[#2819] outgoing message what doing specs

garethrees:
Work on #2819 

Adding some coverage so that I can add `external_review` to the values allowed in `OutgoingMessage#what_doing`.

### DIFF
--- a/spec/models/outgoing_message_spec.rb
+++ b/spec/models/outgoing_message_spec.rb
@@ -37,6 +37,35 @@ describe OutgoingMessage do
 
   end
 
+  describe '#what_doing' do
+
+    it 'allows a value of normal_sort' do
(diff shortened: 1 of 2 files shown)


spec/models/outgoing_message_spec.rb
```diff
@@ -37,6 +37,35 @@
...
+      expect(message).to be_valid
+    end
+
+    it 'allows a value of internal_review' do
+      message =
+        FactoryGirl.build(:initial_request, :what_doing => 'internal_review')
+      expect(message).to be_valid
+    end
+
+    it 'allows a value of internal_review' do
```
crowbot: Should read `new_information`, I think.

PR #3049, opened 2016-02-09
Add instructions for including logo@2.png files with themes

### LINKED ISSUE #3019: Add information about logo@2.png to http://alaveteli.org/docs/customising/themes/#changing-the-logo
You need to have a logo@2.png file in your theme assets directory as well as a logo.png - logo@2.png should be the same logo image but for use on high resolution screens, such as retina. It should be at least 2x the size as logo.png, in the same proportions.

lizconlan:
Closes #3019

### DIFF
--- a/docs/customising/themes.md
+++ b/docs/customising/themes.md
@@ -194,7 +194,7 @@ this means that if you put your own copy of the "about us" template
 in <code>lib/themes/<em>yourtheme</em>/lib/views/help/about.html.erb</code>,
 then that will appear instead of the core "about us" file.
 
-### Changing the logo
+### Changing the site logo
 
 Alaveteli uses Rails' [asset pipeline](http://guides.rubyonrails.org/asset_pipeline.html)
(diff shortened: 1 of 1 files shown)


lizconlan:
I think this is a bit better, now suggests starting with a double size image and making a smaller one from it.

garethrees:
+1, but could you change the dimensions to be the same as those of the current default logo files:

- https://github.com/mysociety/alaveteli/blob/develop/app/assets/images/logo%402.png
- https://github.com/mysociety/alaveteli/blob/develop/app/assets/images/logo.png

Sorry to move the goalposts, but it would be awesome if we could get a really quick (1-liner) mention of the [opengraph logo](https://github.com/mysociety/alaveteli/blob/develop/app/assets/images/logo-opengraph.png) too.

lizconlan:
one more read-through just in case?

garethrees:
:+1: don't forget to squash :)

PR #3052, opened 2016-02-10
2868 refactor request controller lc

lizconlan:


lizconlan:
experimental only, closed in favour of #3053

PR #3051, opened 2016-02-10
2868 refactor request controller gr

garethrees:

PR #3053, opened 2016-02-10
[#2868] Refactor RequestController#show_response

garethrees:
Supersedes https://github.com/mysociety/alaveteli/pull/2969 and https://github.com/mysociety/alaveteli/pull/3051
Fixes https://github.com/mysociety/alaveteli/issues/2868

This PR cherry-picks each commit from #3015 (#2969 with fixups applied) with some amends.


garethrees:
Still needs to move/remove the specs in [`request_controller_spec.rb`](https://github.com/mysociety/alaveteli/blob/2868-refactor-request-controller-gr-cp/spec/controllers/request_controller_spec.rb#L1771-L1857).

lizconlan:
Thinking about it, I'd still prefer `FollowupController` over `FollowupsController` as followup is an action, there's no Followup object

garethrees:
> there's no Followup object

The `OutgoingMessage` instance is the followup object.

I've had to do more work than I should have in [the past](https://github.com/mysociety/alaveteli/issues/2011) with forms because the controllers/routes don't follow convention.

In general sticking to the rails conventions helps with all sorts of things, which is why I've done that here.

PR #3061, opened 2016-02-23
Rename filter for setting subject

garethrees:
The filter no longer only sets an InfoRequest.


garethrees:
Hmm, maybe s/subject/association/ ?

Issue #2769, opened 2015-09-03
Check boxes for delete spam from holding pen


kingqueen3065:
The majority of posts that hit WhatDoTheyKnow's holding pen are spam. It's a pfaf to load each individual one, select Delete and confirm, when there are a lot of them to do at once, particularly as this seems to invoke some system lag. It would be great to have a system like on webmail inboxes, whereby admins could select the responses they want to delete all in one go using checkboxes, then perhaps a confirmation screen listing the first line of each spam to be doubly sure, then a button to delete the whole lot.

garethrees:
For search: bulk action

garethrees:
Hopefully not such an issue once we get #2387 done.

kingqueen3065:
this would still be a very lovely thing to have and save me a load of time :D

lizconlan:
Additional: add a cron job for the `rake:cleanup` task in case that improves the speed (separate PR)

lizconlan:
Closed via #3063 and #3064

PR #3063, opened 2016-02-24
Add checkboxes to allow deletion of incoming messages from a request

lizconlan:
Using some hacky javascript to avoid overlapping forms (due to the hidden edit panels in the message list), adds some checkboxes to the "Incoming messages" section of the admin's edit request screen to allow deletion of one or more messages.

screen shot 2016-02-24 at 12 08 16

Selecting a message enables the "Delete selected messages" button (the button does not appear if there are no incoming messages).

screen shot 2016-02-24 at 12 08 28

Clicking through submits to a summary screen where pressing "Yes" will destroy the message(s) and redirect back to the edit request screen ("No" redirects back with no action taken).

screen shot 2016-02-24 at 12 08 58

Connected to #2769

### DIFF
--- a/app/controllers/admin_incoming_message_controller.rb
+++ b/app/controllers/admin_incoming_message_controller.rb
@@ -36,6 +36,39 @@ class AdminIncomingMessageController < AdminController
     redirect_to admin_request_url(@incoming_message.info_request)
   end
 
+  def bulk_destroy
+    if params[:commit] == "No"
+      redirect_to(admin_request_url(params[:request_id]))
+    end
+
+    @incoming_messages = IncomingMessage.where(:id => params[:ids].split(","))
+    if params[:commit] == "Yes"
+      errors = []
+      info_request = InfoRequest.find(params[:request_id])
+      @incoming_messages.each do |message|
+        begin
+          message.destroy
+          info_request.log_event("destroy_incoming",
+                                 { :editor => admin_current_user,
+                                   :deleted_incoming_message_id => message.id })
+        rescue
+          errors << message.id
+        end
+      end
+      info_request.expire
+      if errors.empty?
+        flash[:notice] = "Incoming messages successfully destroyed."
+      else
+        flash[:error] = <<-EOF.strip_heredoc
+          Only some incoming messages were destroyed.
+          Incoming Messages #{ errors.join(', ') } could not be destroyed.
+          Try destroying them individually.
+        EOF
+      end
+      redirect_to(admin_request_url(params[:request_id]))
+    end
+  end
+
   def redeliver
 
     message_ids = params[:url_title].split(",").each {|x| x.strip}
--- a/app/controllers/admin_incoming_message_controller.rb
+++ b/app/controllers/admin_incoming_message_controller.rb
@@ -31,7 +31,7 @@ class AdminIncomingMessageController < AdminController
                                              { :editor => admin_current_user,
                                               :deleted_incoming_message_id => @incoming_message.id })
     # expire cached files
-    @incoming_message.info_request.expire
+    @incoming_message.info_request.expire(:preserve_database_cache => true)
     flash[:notice] = 'Incoming message successfully destroyed.'
     redirect_to admin_request_url(@incoming_message.info_request)
   end
@@ -55,7 +55,7 @@ class AdminIncomingMessageController < AdminController
           errors << message.id
         end
       end
-      info_request.expire
+      info_request.expire(:preserve_database_cache => true)
       if errors.empty?
         flash[:notice] = "Incoming messages successfully destroyed."
       else
--- a/app/views/admin_incoming_message/bulk_destroy.html.erb
+++ b/app/views/admin_incoming_message/bulk_destroy.html.erb
@@ -1,0 +1,23 @@
+<h1>Confirm incoming message deletion</h1>
+
+<% @incoming_messages.each do |incoming_message| %>
+  <b><%= h(incoming_message.mail_from) %></b>
+  at <%=admin_value(incoming_message.sent_at)%>
+  <blockquote class="incoming-message">
+    <% if !incoming_message.cached_main_body_text_folded.nil? %>
+      <%= truncate(
+            incoming_message.cached_main_body_text_folded.
+              gsub('FOLDED_QUOTED_SECTION', ''), :length => 400) %>
+    <% end %>
+  </blockquote>
+<% end %>
+
+Delete these messages?
+<%= form_tag bulk_destroy_admin_incoming_messages_path(@info_request) do %>
+  <%= hidden_field_tag(:ids, params[:ids]) %>
+  <%= hidden_field_tag(:request_id, params[:request_id]) %>
+  <%= submit_tag("Yes",
+                 { :class => "btn btn-danger",
+                   :confirm => 'Are you sure?' }) %>
+  <%= submit_tag("No", { :class => "btn btn-button" }) %>
+<% end %>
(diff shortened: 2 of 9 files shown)


garethrees:
Had a quick first-pass over this and it looks like a nice solution!

Some comments that I think should be addressed before a thorough review:

- This should post to `AdminIncomingMessageController`. I'm in two minds about whether it should be to `#destroy` or a new collection action (`#bulk_destroy`). The former is good because its standard REST. The latter might make it easier to implement the confirmation page (which I think is a good addition for this case).
- I think rather than having several `expect` calls in each spec, a better test would just be to find the messages and assert whether they exist or not. I say this because we're not testing the relationship between the `InfoRequest` and the `IncomingMessage`. In theory, I could make that spec pass by doing something like `@incoming_messages.each { |message| message.update_attribues(:info_request_id => nil) }`. Some examples:
  - `expect(IncomingMessage.find(ids)).to be_empty`. 
  - `expect(IncomingMessage.find(ids)).to eq([spam1])`
- Minor, and I know its an existing pattern, but I think `request` or `info_request` is a better variable name than `ir`. YMMV on that one; always takes me a bit of extra time to parse.

garethrees:
p.s. feel free to make fixups locally and force push new commits at this stage if it makes life easier.

lizconlan:
> p.s. feel free to make fixups locally and force push new commits at this stage if it makes life easier.

Cool, ta - will push some fixups around for a bit then squash when it's working properly

lizconlan:
> This should post to `AdminIncomingMessageController`.

Er... Yes, good point!

> I'm in two minds about whether it should be to `#destroy` or a new collection action (`#bulk_destroy`). 

I'm inclined to go with the latter

[snip spec fixes]

> Minor, and I know its an existing pattern, but I think request or info_request is a better variable name than ir. YMMV on that one; always takes me a bit of extra time to parse.

lazily went with `ir` to save typing (which I have seriously undermined by replying at length), not fussed so compromising with `request` to make for a simpler/better read and a marginally shorter line

lizconlan:
The bonus fixup commit adds logging, request expiry and a flash message confirmation

app/controllers/admin_incoming_message_controller.rb
```diff
@@ -36,6 +36,26 @@ def destroy
...
+    @incoming_messages = IncomingMessage.where(:id => params[:ids].split(","))
+    if params[:commit] == "Yes"
+      info_request = InfoRequest.find(params[:request_id])
+      @incoming_messages.each do |message|
+        message.destroy
+        info_request.log_event("destroy_incoming",
+                               { :editor => admin_current_user,
+                                 :deleted_incoming_message_id => message.id })
+      end
+      info_request.expire
```
garethrees: Calling `#expire` looks like the right thing to do here, but its clearing the incoming message cached body without regenerating it:

screen shot 2016-02-25 at 10 48 26


screen shot 2016-02-25 at 10 48 57

lizconlan: Hmm (consider this a note to self) - wonder if deleting an individual message through the existing delete button does the same thing

garethrees: Seems this happens on `develop` too

garethrees: Yeah I'm just having a bit of a dig in to this

garethrees: Yeah this happens at least as far back as `0.19.0.14`

app/views/admin_incoming_message/bulk_destroy.html.erb
```diff
@@ -0,0 +1,22 @@
+<h1>Confirm incoming message deletion</h1>
+
+<% @incoming_messages.each do |incoming_message| %>
+  <b><%= h(incoming_message.mail_from) %></b>
+  <%=_("at")%> <%=admin_value(incoming_message.sent_at)%>
+  <blockquote class="incoming-message">
+    <% if !incoming_message.cached_main_body_text_folded.nil? %>
+      <%= truncate(
+            incoming_message.cached_main_body_text_folded.
```
garethrees: I think this is causing https://github.com/mysociety/alaveteli/pull/3063/files#r54095767; I think it should be calling `#get_main_body_text_folded`, but it wouldn't surprise me if the the conditional was put there for performance reasons.

app/controllers/admin_incoming_message_controller.rb
```diff
@@ -36,6 +36,26 @@ def destroy
...
+  def bulk_destroy
+    if params[:commit] == "No"
+      redirect_to(admin_request_url(params[:request_id]))
+    end
+
+    @incoming_messages = IncomingMessage.where(:id => params[:ids].split(","))
+    if params[:commit] == "Yes"
+      info_request = InfoRequest.find(params[:request_id])
+      @incoming_messages.each do |message|
+        message.destroy
```
garethrees: What happens if a message can't be destroyed (e.g. some weird DB constraint issue)?

I think this should be more like:

```ruby
@incoming_messages.each do |message|
  if message.destroy
    # log event
  else
    errors < message.id
  end
end

# ...

if errors.empty?
  flash[:notice] = 'Incoming messages successfully destroyed.'
else
  flash[:error] = <<-EOF.strip_heredoc
  Only some incoming messages were destroyed.
  Incoming Messages #{ errors.join(', ') } could not be destroyed.
  Try destroying them individually.
  EOF
end

File: app/views/admin_incoming_message/bulk_destroy.html.erb [hunk 4]
```diff
@@ -0,0 +1,22 @@
+<h1>Confirm incoming message deletion</h1>
+
+<% @incoming_messages.each do |incoming_message| %>
+  <b><%= h(incoming_message.mail_from) %></b>
+  <%=_("at")%> <%=admin_value(incoming_message.sent_at)%>
```
garethrees: Don't need to translate admin strings

app/views/admin_request/show.html.erb
```diff
@@ -320,6 +321,16 @@
...
     </div>
   <% end %>
+  <% if @info_request.incoming_messages.count > 0 %>
+    <%= form_tag bulk_destroy_admin_incoming_messages_path do %>
+      <%= hidden_field_tag(:ids) %>
+      <%= hidden_field_tag(:request_id, @info_request.id) %>
+      <%= submit_tag("Delete selected messages",
+                     {
+                        :class => "btn btn-danger",
+                        :disabled => true }) %>
```
garethrees: formatting looks a bit weird here

config/routes.rb
```diff
@@ -291,6 +291,11 @@
     :only => [:edit, :update, :destroy] do
       post 'redeliver', :on => :member
     end
+    resource :incoming_messages,
+      :controller => 'admin_incoming_message',
+    :only => [:bulk_destroy] do
+      post 'bulk_destroy'
+    end
```
garethrees: The last 3 lines need another indent here

garethrees:
Some minor tweaks but on the whole, really nice :+1:

app/controllers/admin_incoming_message_controller.rb
lizconlan: Am trying something a bit odd - allowing `#expire` to optionally bypass the expiry of the incoming messages.

garethrees:
Good to go after tweaking the `#expire` argument and adding a changelog entry :+1:

garethrees:
Just a changelog entry and then :rocket: :ship:

PR #3073, opened 2016-03-02
Add tracking code to widgets

### LINKED ISSUE #3059: Add analytics to widget view

lizconlan:
Tracks widget clicks as an event of category 'widgets', sets the action as 'track', 'vote' or 'unsubscribe' and attempts to populate the label with the address of the page it's embedded in inside the iframe. Does not fire if the GA code for the site has not been set.

Closes #3059


garethrees:
Changelog entry plz! :)

PR #3078, opened 2016-03-03
Remove deprecated methods

garethrees:


lizconlan:
:+1: looks good to me, watching with interest to see whether taking that amount of code out noticeably improves the coverage stats :rocket:

PR #3083, opened 2016-03-07
Stop using text wrapping formatters for plain text emails

### LINKED ISSUE #3065: Unpleasant word wrap on outgoing emails
I think this is related to https://github.com/mysociety/alaveteli/issues/2913 

Lines are artificially truncated on outgoing emails to users. This results in an unpleasant reading experience especially on mobile and other devices with narrow viewports.

lizconlan:
Produces email without any artificial line breaks, leaving the email client to sort it out as this seems to be the [advice these days](https://www.campaignmonitor.com/dev-resources/guides/design/#two) (and using the Unicode equivalent soft line break character doesn't seem to be well understood by clients - Mutt ignores it and Gmail treats it as an extra space):

    Liz,
    
    Please click on the link below to confirm your email address. Then you can change your password on Alaveteli
    
    http://example.com/link
    
    We will not reveal your email address to anybody unless you
    or the law tell us to.
    
    -- the Alaveteli team

Note that this can still lead to imperfect formatting, as per the last paragraph of this example - the text is drawn from the translation file which, on my machine at time of testing, contained a line break. Although it is possible to strip out line breaks, it could lead to worse problems such as:

    Follow these instructions:
     - Step 1
     - Step 2

Being incorrectly formatted as:

    Follow these instructions: - Step 1 - Step 2

Which would be harder to catch and fix. 

Fixes #3065

### DIFF
--- a/app/mailers/outgoing_mailer.rb
+++ b/app/mailers/outgoing_mailer.rb
@@ -17,7 +17,6 @@ class OutgoingMailer < ApplicationMailer
   # Email to public body requesting info
   def initial_request(info_request, outgoing_message)
     @info_request, @outgoing_message, @contact_email = info_request, outgoing_message, AlaveteliConfiguration::contact_email
-    @wrap_lines_as_paragraphs = true
     headers["message-id"] = OutgoingMailer.id_for_message(outgoing_message)
 
     mail(:from => info_request.incoming_name_and_email,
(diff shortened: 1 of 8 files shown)


garethrees:
As discussed in IRC, I think we're safe to remove the [`@wrap_lines_as_paragraphs` ivars](https://github.com/mysociety/alaveteli/blob/develop/app/mailers/outgoing_mailer.rb#L20) as they were phased out in https://github.com/mysociety/alaveteli/commit/1890f41864adf2b6945a66636fbc2772b9f018cd#diff-bd05b1cfa8a0f86eb6d558efb8aa7269L1, and this PR removes the wrapping anyway.

Issue #2941, opened 2015-12-03
Release 0.24


garethrees:
- [x] Update crontab
- [x] Check translations (#2935)
- [x] Remove `SPECIAL_REPLY_VERY_LATE_AFTER_DAYS` config

garethrees:
Remember to carefully check translations (https://github.com/mysociety/alaveteli/pull/2935).

garethrees:
Got an exception sending a request with alavetelitheme:

```
lib/themes/alavetelitheme/lib/views/outgoing_mailer/initial_request.text.erb:10:in `_lib_themes_alavetelitheme_lib_views_outgoing_mailer_initial_request_text_erb___2811866733010843248_82362020'
app/mailers/outgoing_mailer.rb:24:in `initial_request'
app/controllers/request_controller.rb:364:in `new'
app/controllers/application_controller.rb:105:in `record_memory'
lib/strip_empty_sessions.rb:13:in `call'
```

garethrees:
> Got an exception sending a request with alavetelitheme:

Fixed in https://github.com/mysociety/alavetelitheme/commit/1450d48d17bfbc03510698c0c717d7be48e36316

garethrees:
Done. https://www.mysociety.org/2016/03/17/alaveteli-release-0-24/

- Haven't regenerated the AMI.
- Haven't upgraded any partner sites (other than colombia to test mysociety/alavetelitheme#74)

PR #3088, opened 2016-03-09
Add +x to script/cleanup-holding-pen

garethrees:
Add to #2941


lizconlan:
:+1:  (sorry for missing it in the original :disappointed:)

Issue #3087, opened 2016-03-08
Transient spec failures


garethrees:
A ticket to log any failed travis builds.

Please copy from "Failures:" to "Randomized with seed 23014" and paste the URL to the job.

garethrees:
```
Failures:
  1) MailServerLog Exim .load_exim_log_data sanitizes each line in the log file
     Failure/Error: expect(ir.mail_server_logs[1].line).to eq expected_log_line
       
       expected: "2015-07-09 15:41:40 [29933] foi+request-1234@example.com SMTP protocol synchronization error (next input sent too soon: pipelining was not advertised): rejected \"EHLO 0]C\u000E\u000E\u0003\u001C<\u0006\u0019~\u0006|='\u0016)\u0006\u0005\" H=remote.comagex.be [91.183.116.119]:53191 I=[46.43.39.78]:25 next input=\"\\f\\227\\212\\016\\314\\246\\r\\n\"\n"
            got: "2015-07-09 15:41:40 [13707] foi+request-1234@example.com SMTP connection from [91.183.116.119]:53191 I=[46.43.39.78]:25 (TCP/IP connection count = 3)\n"
       
       (compared using ==)
       
       Diff:
       @@ -1,2 +1,2 @@
       -2015-07-09 15:41:40 [29933] foi+request-1234@example.com SMTP protocol synchronization error (next input sent too soon: pipelining was not advertised): rejected "EHLO 0]C<~|=')" H=remote.comagex.be [91.183.116.119]:53191 I=[46.43.39.78]:25 next input="\f\227\212\016\314\246\r\n"
       +2015-07-09 15:41:40 [13707] foi+request-1234@example.com SMTP connection from [91.183.116.119]:53191 I=[46.43.39.78]:25 (TCP/IP connection count = 3)
       
     # ./spec/models/mail_server_log_spec.rb:143:in `block (4 levels) in <top (required)>'
Finished in 3 minutes 9.4 seconds (files took 11.57 seconds to load)
2581 examples, 1 failure, 1 pending
Failed examples:
rspec ./spec/models/mail_server_log_spec.rb:114 # MailServerLog Exim .load_exim_log_data sanitizes each line in the log file
Randomized with seed 23014
```

garethrees:
```
Failures:
  1) ActsAsXapian::Search#words_to_highlight should return a list of words used in the search
     Failure/Error: get_fixtures_xapian_index
     ActiveRecord::RecordNotFound:
[...]

PR #3092, opened 2016-03-11
Clear purge requests before test

crowbot:


garethrees:
:+1:. Is this work on #3087?

PR #3100, opened 2016-03-16
Code cleanup

garethrees:
Misc minor code cleanup; was on the train with no internet access.


app/models/outgoing_message.rb
```diff
@@ -383,7 +386,7 @@ def replying_to_incoming_message?
           incoming_message_followup.valid_to_reply_to?
   end
 
-  def format_of_body
+  def template_changed
```
crowbot: We should probably retain a deprecated version of `format_of_body` in case anyone's calling it from their theme.

Issue #3107, opened 2016-03-18
Migrate to modern Google Analytics code


lizconlan:
The Google Developer docs for Analytics suggest we are using a ["legacy" version of the tracking code](https://developers.google.com/analytics/devguides/collection/gajs/), and that the `_getTracker` method we are currently using to setup the tracker object on each page [has also been deprecated in favour of `_createTracker`](https://developers.google.com/analytics/devguides/collection/gajs/methods/gaJSApi_gat#_gat-object-methods).

It's unlikely to cause an immediate problem - there's no suggestion as yet that anything is about to stop working - but it's something we should consider before we get too far into adding more GA code to our sites. However there is a suggestion that there are some shiny new features we might not have access to until we upgrade to the new "Universal Analytics" code.

**Note**: there is a specific warning against using a mix of the old and new tracking code on the same site. [Full upgrade/migration guide can be found in the developer docs](https://developers.google.com/analytics/devguides/collection/upgrade/)

lizconlan:
[Check for use of config that has moved to the Admin page](https://developers.google.com/analytics/devguides/collection/upgrade/reference/gajs-analyticsjs#exclusions)
- [x] `_setCampaignCookieTimeout` or `_setSessionCookieTimeout`
- [x] `_addOrganic`
- [x] `_addIgnoredRef`
- [x] `_addIgnoredOrganic`

- [ ] [Replace code snippet](https://developers.google.com/analytics/devguides/collection/upgrade/reference/gajs-analyticsjs#snippet)
- [ ] [Fix event code](https://developers.google.com/analytics/devguides/collection/upgrade/reference/gajs-analyticsjs#events)
- [ ] [Check for custom variables](https://developers.google.com/analytics/devguides/collection/upgrade/reference/gajs-analyticsjs#custom-vars)
- [ ] [Check for virtual pageviews](https://developers.google.com/analytics/devguides/collection/upgrade/reference/gajs-analyticsjs#virtual-pageviews)
- [ ] [Check for social interactions](https://developers.google.com/analytics/devguides/collection/upgrade/reference/gajs-analyticsjs#social)
- [ ] [Check for user timings](https://developers.google.com/analytics/devguides/collection/upgrade/reference/gajs-analyticsjs#timings)
- [ ] [Check for cross domain tracking](https://developers.google.com/analytics/devguides/collection/upgrade/reference/gajs-analyticsjs#cross-domain)
[...]

PR #3110, opened 2016-03-18
Track social media clicks

lizconlan:
Creates a Google Analytics event whenever someone exits the site using a social media link (Experiment 71)

Events reporting example:

screen shot 2016-03-18 at 15 22 28

Also adds a new helper for constructing the event code string to go in the `onclick` attribute

### DIFF
--- a/app/helpers/analytics_helper.rb
+++ b/app/helpers/analytics_helper.rb
@@ -1,0 +1,101 @@
+# -*- encoding : utf-8 -*-
+module AnalyticsHelper
+
+  # helpers for embedding Google Analytics code
+  #
+  # Event categories and actions should be drawn from the list in the
+  # lib/analytics_events.rb file (add your own there when making new ones)
+
+
+  # Public: Constructs a String consisting of a Google Analytics (GA) tracking
+  # event function call with the (mandatory) event category and action params
+  # and optional label and value params.
+  #
+  # event_category - The String to be sent to GA as the tracking event category.
+  #                  Ideally this should be an AnalyticsEvent::Category::THING
+  #                  to avoid having magic strings everywhere
+  # event_action   - The String to be sent to GA as the tracking event action.
+  #                  Ideally this should be an AnalyticsEvent::Action::THING
+  #                  to avoid having magic strings everywhere
+  # options        - Hash of optional values, expects:
+  #                   label           - String, an optional label for the event
+  #                   label_is_script - Boolean, whether to treat the label
+  #                                     String as literal or browser-
+  #                                     interpreted (Javascript)
+  #                   value           - Integer, Google insists that a numerical
+  #                                     value param is supplied if label is used;
+  #                                     if you don't care about this value and
+  #                                     don't want to set it, leave it blank
+  #                                     and a default of 1 will be sent
+  #                  Any other supplied options will be ignored
+  #
+  # Examples
+  #
+  #   track_analytics_event("test", "button clicked")
+  #   # => "if (ga) { ga('send','event','test','button clicked') };"
+  #
+  #   track_analytics_event("test", "vote button", :label => "sidebar")
+  #   # => "if (ga) { ga('send','event','test','vote button','sidebar',1) };"
+  #
+  #   track_analytics_event("test", "Points Scored", :label => "Bonus", :value => 100)
+  #   # => "if (ga) { ga('send','event','test','Points Scored','Bonus',100) };"
+  #
+  #   track_analytics_event("test",
+  #                         "Embedded",
+  #                         :label => "window.location.href",
+  #                         :label_is_script => true)
+  #   # => "if (ga) { ga('send','event','test','Embedded',window.location.href,1) };"
+  #
+  # Returns a string of a GA JavaScript function to drop into an :onclick handler
+  def track_analytics_event(event_category, event_action, options={})
+    begin
+      value = if options[:value].nil?
+        1
+      else
+        Integer(options[:value])
+      end
+    rescue ArgumentError
+      raise ArgumentError, %Q(:value option must be an Integer: "#{ options[:value] }")
+    end
+
+    label_is_script = options[:label_is_script] == true
+    label = options[:label]
+    if label
+      label_string = ",#{format_event_label(label, label_is_script)},#{value}"
+    end
+    event_args = "'#{event_category}','#{event_action}'#{label_string}"
+    "if (ga) { ga('send','event',#{event_args}) };"
+  end
+
+  private
+
+  # Private: Format the event label by wrapping in single quotes if it is
+  # going to be used as a String literal (e.g. "'Button clicked'") or without if
+  # it is a JavaScript string (e.g. "window.location.href") that needs to be run
+  # and interpreted by the browser.
+  #
+  # label     - The label text (String) to by used
+  # is_script - Boolean indicating whether the supplied label text is intended
+  #             to be interpreted as JavaScript by the browser (defaults
+  #             to false)
+  #
+  # Examples
+  #
+  #   format_event_label("Vote Button Clicked", false)
+  #   # => "'Vote Button Clicked'"
+  #
+  #   format_event_label("window.top.location.href", true)
+  #   # => "window.top.location.href"
+  #
+  # Returns the label String with or without containing single quotes,
+  # depending on the value of the is_script param default behaviour:
+  #   is_script is evaluated as false, string returned wrapped in single quotes)
+  def format_event_label(label, is_script=false)
+    if is_script
+      label
+    else
+      "'#{label}'"
+    end
+  end
+
+end
--- a/app/views/widgets/show.html.erb
+++ b/app/views/widgets/show.html.erb
@@ -39,24 +39,43 @@
       <% elsif @existing_track %>
         <%= link_to update_url(:track_id => @existing_track.id, :track_medium => "delete", :r => show_user_profile_url(:url_name => @user.url_name)), :target => '_top',
           :class => 'alaveteli-widget__bottom__action alaveteli-widget__button--unsubscribe',
-          :onclick => "if (ga) { ga('send', 'event', 'Widget Clicked', 'Unsubscribe', window.top.location.href, 1)};" do %>
+          :onclick => track_analytics_event(
+                        AnalyticsEvent::Category::WIDGET_CLICK,
+                        AnalyticsEvent::Action::WIDGET_UNSUB,
+                        :label => "window.top.location.href",
+                        :label_is_script => true) do %>
             <%= _('Unsubscribe') %>
         <% end %>
       <% elsif @existing_vote %>
         <%= link_to track_request_url(:url_title => @info_request.url_title, :feed => 'track'), :target => '_top',
           :class => 'alaveteli-widget__bottom__action alaveteli-widget__button--sign-in-to-track',
-          :onclick => "if (ga) { ga('send', 'event', 'Widget Clicked', 'Sign in to track', window.top.location.href, 1)};" do %>
+          :onclick => track_analytics_event(
+                        AnalyticsEvent::Category::WIDGET_CLICK,
+                        AnalyticsEvent::Action::WIDGET_SIGNIN,
+                        :label => "window.top.location.href",
+                        :label_is_script => true) do %>
           <%= _('Sign in to track this request') %>
         <% end %>
       <% elsif @user %>
         <%= link_to do_track_path(@track_thing), :target => '_top',
-          :class => 'alaveteli-widget__bottom__action alaveteli-widget__button--create-track',
-          :onclick => "if (ga) { ga('send', 'Widget Clicked', 'Vote', window.top.location.href, 1)};" do %>
+              :class => 'alaveteli-widget__bottom__action alaveteli-widget__button--create-track',
+              :onclick => track_analytics_event(
+                            AnalyticsEvent::Category::WIDGET_CLICK,
+                            AnalyticsEvent::Action::WIDGET_VOTE,
+                            :label => "window.top.location.href",
+                            :label_is_script => true) do %>
           <%= _('I also want to know!') %>
         <% end %>
       <% else %>
-        <%= form_tag request_widget_votes_url(@info_request), :method => 'post', :target => '_top' do %>
-          <%= submit_tag _('I also want to know!'), :class => 'alaveteli-widget__bottom__action alaveteli-widget__button--create-vote', :onclick => 'if (ga) { ga("send", Widget Clicked", "Vote", window.top.location.href, 1)};' %>
+        <%= form_tag request_widget_votes_url(@info_request),
+              :method => 'post', :target => '_top' do %>
+          <%= submit_tag _('I also want to know!'),
+                :class => 'alaveteli-widget__bottom__action alaveteli-widget__button--create-vote',
+                :onclick => track_analytics_event(
+                              AnalyticsEvent::Category::WIDGET_CLICK,
+                              AnalyticsEvent::Action::WIDGET_VOTE,
+                              :label => "window.top.location.href",
+                              :label_is_script => true) %>
         <% end %>
       <% end %>
     </div>
--- a/lib/analytics_event.rb
+++ b/lib/analytics_event.rb
@@ -1,0 +1,20 @@
+# -*- encoding : utf-8 -*-
+module AnalyticsEvent
+
+  # modules for standardising Strings used for event categories and actions
+
+  module Category
+    WIDGET_CLICK = "Widget Clicked"
+    OUTBOUND = "Outbound Link"
+  end
+
+  module Action
+    FACEBOOK_EXIT = 'Facebook Exit'
+    TWITTER_EXIT = 'Twitter Exit'
+    WORDPRESS_EXIT = 'WordPress Exit'
+    WIDGET_VOTE = 'Vote'
+    WIDGET_SIGNIN = 'Sign in to track'
+    WIDGET_UNSUB = 'Unsubscribe'
+  end
+
+end
--- a/spec/helpers/analytics_helper_spec.rb
+++ b/spec/helpers/analytics_helper_spec.rb
@@ -1,0 +1,89 @@
+# -*- encoding : utf-8 -*-
+require File.expand_path(File.dirname(__FILE__) + '/../spec_helper')
+
+describe AnalyticsHelper do
+
+  include AnalyticsHelper
+
+  describe "#track_analytics_event" do
+    it "returns correctly formatted event javascript" do
+      expect(track_analytics_event(
+        AnalyticsEvent::Category::OUTBOUND,
+        AnalyticsEvent::Action::FACEBOOK_EXIT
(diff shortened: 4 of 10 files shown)


app/helpers/link_to_helper.rb
```diff
@@ -254,6 +254,20 @@ def unhappy_url(info_request = nil)
     end
   end
 
+  # Google Analytics
+
+  # if used, value must an integer
+  def analytics_event_code(category, action, label=nil, value=nil, label_is_script=false)
```
garethrees: An options hash would be much clearer here. This could also do with some [documentation](http://tomdoc.org/) about what the arguments do. I wouldn't know how to use this by only looking at the method signature.

app/helpers/link_to_helper.rb
```diff
@@ -254,6 +254,20 @@ def unhappy_url(info_request = nil)
...
+    value = 1 unless value
```
garethrees: > `# if used, value must an integer`

Lets handle this:

```ruby
irb(main):006:0> Integer(1)
=> 1
irb(main):004:0> Integer('1')
=> 1
irb(main):005:0> Integer('Word')
ArgumentError: invalid value for Integer(): "Word"
        from (irb):5:in `Integer'
        from (irb):5
        from /usr/bin/irb:12:in `<main>'
```

garethrees: I'd probably do something like:

begin
  value = Integer(value)
rescue ArgumentError
  raise ArgumentError, %Q(:value option must be an Integer: "#{ options[:value] }")
end

app/helpers/link_to_helper.rb
garethrees: I'd name the required arguments `event_category` and `event_action` so that they map to the titles used [by Google](https://github.com/mysociety/alaveteli/pull/3110#issue-141910287)

app/helpers/link_to_helper.rb
```diff
@@ -254,6 +254,20 @@ def unhappy_url(info_request = nil)
...
+    label_string = if label
+      ",#{format_event_label(label, label_is_script)},#{value}"
+    else
+      ""
```
garethrees: Don't need the else, since `nil` will interpolate as an empty string

app/helpers/link_to_helper.rb
```diff
@@ -273,4 +287,12 @@ def message_path(message)
     message_url(message, :only_path => true)
   end
 
+  def format_event_label(label, is_script)
```
garethrees: It would be worth adding a line of documentation about why you might need to use this.

spec/helpers/link_to_helper_spec.rb
```diff
@@ -146,4 +146,32 @@
 
   end
 
+  describe "when constructing a google analytics event string" do
```
garethrees: Much easier to find the spec you're after when the description is the method name:

```diff
- describe "when constructing a google analytics event string" do
+ describe '#analytics_event_code' do
```

spec/helpers/link_to_helper_spec.rb
```diff
@@ -146,4 +146,32 @@
...
+
+      it "uses the supplied value if there is one" do
+        expect(analytics_event_code("test category", "test", "test label", 42)).to eq(
+          "if (pageTracker) { pageTracker._trackEvent('test category','test','test label',42) };"
+        )
+      end
+
+      it "doesn't wrap the label in quotes if passed label_is_script=true" do
+        expect(analytics_event_code("test category", "test", "window.location.href", 42, true)).to eq(
+          "if (pageTracker) { pageTracker._trackEvent('test category','test',window.location.href,42) };"
```
garethrees: Why might I want to use `window.location.href` as the 3rd param vs `'test label'` as in the previous spec above? Presumably the former is going to dynamically grab the window URL; Is it valid to supply a String?

app/helpers/link_to_helper.rb
garethrees: I have a mild preference for calling this method `track_analytics_event` as its a bit more "Tell, don't ask" (okay, not exactly, but I think better describes what will _happen_ when used, rather than what it _is_). Shrug! :lollipop:

spec/helpers/link_to_helper_spec.rb
lizconlan: The use case for this is the widget code - which does something like `window.top.location.href` which tells us where the code's been embedded, not essential but a nice to have. If we wrap it as a String, it will literally log "window.top.location.href" which isn't what we want to capture (I have accidentally proved this a couple of times).

So yes, depending on the scenario, if we want to use the optional label/value feature we can supply a we can supply a String literal if that's useful or some javascript so that the browser assembles the value that the `onclick` event uses instead.

(or to put it another way... I agree, I need to write some helpful documentation!)

app/helpers/link_to_helper.rb
lizconlan: note to self: proposed new method signature:

`def analytics_event_code(event_category, event_action, options={})`

(original method name used for readability purposes)

garethrees:
The general gist of this is good, with some comments in line.

I'm concerned that allowing "any old strings" to the tracking method will result in a complete mess of duplicated values over time (e.g. someone may use `'Facebook exit'` when they mean `'Facebook Page Exit'`). @lizconlan commented that this has already happened during testing.

[Google's advice](https://support.google.com/analytics/answer/1033068?hl=en) is pretty clear on this, and cites a similar example:

>  For example, if you plan to call your video tracking category "Video" and later forget and use the plural "Videos," you will have two separate categories for video tracking.

TL;DR: Put some thought in to your naming conventions.

My thinking is that we should have some sort of "dictionary" of allowed trackable actions.

Working backwards, I think we want to be passing a value object in to the helper:

```ruby
@event = AnalyticsEvent.new(:category => 'C', :action => 'A', :label => 'L', :value => 1, :label_is_script => false)
```
```erb
<%= track_analytics_event(@event) %>
```

So about this dictionary…

I think the interface should be specifying a _Thing_, rather than a `String`.

My initial thinking was something like:

```ruby
module AnalyticsEvent
  module Facebook
    module PageExit
      SHARE_REQUEST = 'Share Request'
      SHARE_BATCH_REQUEST = 'Share Batch Request'
    end
  end
end

# Convert data from constant to an AnalyticsEvent instance under the hood?
track_analytics_event(AnalyticsEvent::Facebook::PageExit::SHARE_REQUEST)
```

I like the idea of modules and constants because its easy for a theme to define additional tracking values. The problem here is that `SHARE_REQUEST` is a String, so we lose the category/label/value attributes. We could make it a `Hash`, but then we're duplicating information that the module is giving us (`PageExit` in this case might be a category). We could have a giant `Hash` (`AnalyticsEvent::ALAVETELI_EVENTS`) but that _should_ be frozen, so adding to it is problematic.

I think the real problem here is that we don't have any schema for our event tracking (<q>Determine in advance all elements for which you want to track data – https://goo.gl/vq8I63</q>) so its pretty hard to model that in code.

I don't think its the _end of the world_ if we just roll with supplying strings, but we need to be happy that we're making a conscious decision to introduce technical debt, and if we're doing that we should have a plan for what we're going to do to repay it.

lizconlan:
> I'm concerned that allowing "any old strings" to the tracking method will result in a complete mess of duplicated values over time (e.g. someone may use 'Facebook exit' when they mean 'Facebook Page Exit'). @lizconlan commented that this has already happened during testing.

Worse, it's on the live site, see: https://github.com/mysociety/alaveteli/pull/3110/commits/ea9390d8f0a10904504265d297017409dae034b4#diff-6afc1c2b49c42f3d28d7a868fe7b43a5L52 / https://github.com/mysociety/alaveteli/pull/3110/commits/ea9390d8f0a10904504265d297017409dae034b4#diff-6afc1c2b49c42f3d28d7a868fe7b43a5L57

(I'm more than happy to pull these out into a separate PR/ticket so that we can treat fixing that as a separate issue)

lizconlan:
> @event = AnalyticsEvent.new(:category => 'C', :action => 'A', :label => 'L', :value => 1, :label_is_script => false)

Torn, the initial reason for wanting to have `category` and `action` as params was to bake into the method signature that those 2 params are mandatory whereas the rest are optional extras. But I like the object creation syntax.

garethrees:
> Torn, the initial reason for wanting to have `category` and `action` as params was to bake into the method signature that those 2 params are mandatory whereas the rest are optional extras.

They could totally be optional within the value object: `AnalyticsEvent.new(:category => 'C', :action => 'A')`.

garethrees:
Maybe a better idea for a "dictionary" is to actually map directly to the attributes:

```ruby
module AnalyticsEvent
  module Category
    # "Code" name = "Human" name
    OUTBOUND = 'Outbound'
  end

  module Action
    FACEBOOK_EXIT = 'Facebook Exit'
    TWITTER_EXIT = 'Twitter Exit'
  end
end
```

This improves things slightly:

```erb
<%= track_analytics_event(AnalyticsEvent::Category::OUTBOUND, AnalyticsEvent::Action::FACEBOOK_EXIT) %>
```

lizconlan:
>> Torn, the initial reason for wanting to have `category` and `action` as params was to bake into the method signature that those 2 params are mandatory whereas the rest are optional extras.
>
> They could totally be optional within the value object: `AnalyticsEvent.new(:category => 'C', :action => 'A')`.

Not immediately obvious that `category` and `action` are required though (raising an `ArgumentError` or similar would help, of course)

garethrees:
> Not immediately obvious that `category` and `action` are required though (raising an `ArgumentError` or similar would help, of course)

Ah, with you. Can make those arguments of a value object and use option keys for the other parts.

Note there's also http://api.rubyonrails.org/v3.2.22/classes/Hash.html#method-i-assert_valid_keys, but yeah, I think I prefer the method signature option.

lizconlan:
Right so we need...

- [ ] An opinionated "model" for tracking this data (and a place in the code for it to live)
- [ ] Some library/helper code to make constructing the string to put in the `onclick` attribute simpler / less error prone for the outbound link tracking
- [ ] To use this to track outbound links
- [ ] A fix for "widget" vs "widgets" as a category before we collect a ton of data which is awkward to use

garethrees:
> Some library/helper code to make constructing the string to put in the onclick attribute simpler / less error prone for the outbound link tracking

I'm actually thinking maybe not bother with the intermediary object, given that [this solution](https://github.com/mysociety/alaveteli/pull/3110#issuecomment-199881842) no longer uses magic strings.

My remaining concern with the above is that it would still be possible to provide an "invalid" combination, e.g.

```erb
<%= track_analytics_event(AnalyticsEvent::Category::INBOUND, 
                          AnalyticsEvent::Action::FACEBOOK_EXIT) %>
```

Maybe that's okay? I think its preferable to both the existing implementation (magic strings) and using an intermediary value object that complicates "just tracking the damn event".

lizconlan:
\*starts in the middle\*

How about:

A dedicated helper for outbound link tracking:
```
analytics_track_outbound_link_event(
  AnalyticsEvent::Action::FACEBOOK_EXIT,
  {:label => "Footer link"})
```
(clunky method name, counter-suggestions welcome)

As well as exposing the generic method (which this would call)

garethrees:
Yeah, something like that would work, but y'know what; I think that might be abstracting too soon. We only have a handful of events defined so I don't think its the end of the world to use the "generic" method for everything. Abstracting later is easier once we've got the constants sending the same values to google, so changing the methods called can be done later.

lizconlan:
At the risk of over-complicating things, I have half an idea about making Events part of Categories so that it's either immediately apparent that something odd is going on if we end up calling

```erb
<%= track_analytics_event(AnalyticsEvent::Category::INBOUND, 
                          AnalyticsEvent::Category::OUTBOUND::EXIT_ACTION) %>
```

and could even raise an error if the Action is not in an allowable choice for the category

then maybe we could have a simplified version of the call so that we just pass the Action and use the parent category by default

(could well be more trouble than it's worth tho)

garethrees:
> At the risk of over-complicating things, I have half an idea about making Actions part of Categories

I don't think we really have a good enough idea of how all these map to each other at the moment. My main concern here is that we have a dictionary of known names, rather than having to grep the entire codebase next time we want to add a tracking event. Honestly, I think the right solution for now is to just have a [module per component](https://github.com/mysociety/alaveteli/pull/3110#issuecomment-199881842).

Sorry for any confusion around this. I think the problem here was that we didn't have a ticket that set out what we were trying to do, so as _my_ understanding (as a reviewer) of that improved during discussions around the initial review, my ideas changed here, too.

app/helpers/analytics_helper.rb
```diff
@@ -20,7 +20,7 @@ def track_analytics_event(event_category, event_action, options={})
       label_string = ",#{format_event_label(label, label_is_script)},#{value}"
     end
     event_args = "'#{event_category}','#{event_action}'#{label_string}"
-    "if (pageTracker) { pageTracker._trackEvent(#{event_args}) };"
+    "if (ga) { ga('send','event',#{event_args}) };"
```
garethrees: Just noting to myself that https://github.com/mysociety/alaveteli/pull/3160 is the PR for #3107.

spec/helpers/analytics_helper_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+    context "when passed a label" do
+      it "includes the default value 1 if none is supplied" do
+        expect(track_analytics_event("test category", "test", :label => "test label")).to eq(
+          "if (ga) { ga('send','event','test category','test','test label',1) };"
+        )
+      end
+
+      it "uses the supplied value if there is one" do
+        expect(track_analytics_event(
+          "test category", "test", {:label => "test label", :value => 42}
```
garethrees: Don't need the braces around the options:

```ruby
track_analytics_event "test category", "test", :label => "test label", :value => 42
# => "if (ga) { ga('send','event','test category','test','test label',42) };"
```

spec/helpers/analytics_helper_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+        )).to eq(
+          "if (ga) { ga('send','event','test category','test','test label',42) };"
+        )
+      end
+
+      it "doesn't wrap the label in quotes if passed label_is_script=true" do
+        expect(track_analytics_event(
+          "test category",
+          "test",
+          {:label => "window.location.href", :value => 42, :label_is_script => true}
```
garethrees: Don't need the braces around the options

spec/helpers/analytics_helper_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+        )).to eq(
+          "if (ga) { ga('send','event','test category','test',window.location.href,42) };"
+        )
+      end
+
+      it "raises an ArgumentError if value is not nil or an Integer" do
```
garethrees: "is not nil"?

app/views/widgets/show.html.erb
```diff
@@ -37,24 +37,40 @@
       <% elsif @existing_track %>
         <%= link_to update_url(:track_id => @existing_track.id, :track_medium => "delete", :r => show_user_profile_url(:url_name => @user.url_name)), :target => '_top',
           :class => 'alaveteli-widget__bottom__action alaveteli-widget__button--unsubscribe',
-          :onclick => "if (pageTracker) { pageTracker._trackEvent('widgets', 'unsubscribe', window.top.location.href, 1)};" do %>
+          :onclick => "#{track_analytics_event(
```
garethrees: Not sure why the helper calls in this file are interpolated where others aren't.

Can also drop the braces on the options hash.

app/views/widgets/show.html.erb
```diff
@@ -37,24 +37,40 @@
...
+                          AnalyticsEvent::Category::WIDGET_CLICK,
+                          AnalyticsEvent::Action::WIDGET_VOTE,
+                          {:label => "window.top.location.href",
+                           :label_is_script => true})}" do %>
           <%= _('I also want to know!') %>
         <% end %>
       <% else %>
         <%= form_tag request_widget_votes_url(@info_request), :method => 'post', :target => '_top' do %>
-          <%= submit_tag _('I also want to know!'), :class => 'alaveteli-widget__bottom__action alaveteli-widget__button--create-vote', :onclick => 'if (pageTracker) { pageTracker._trackEvent("widget", "vote", window.top.location.href, 1)};' %>
+          <%= submit_tag _('I also want to know!'), :class => 'alaveteli-widget__bottom__action alaveteli-widget__button--create-vote', :onclick => '#{track_analytics_event(
```
garethrees: You can't interpolate in single quotes.

app/helpers/analytics_helper.rb
```diff
@@ -0,0 +1,36 @@
+# -*- encoding : utf-8 -*-
+module AnalyticsHelper
+
+  # helpers for embedding Google Analytics code
+
+  def track_analytics_event(event_category, event_action, options={})
```
garethrees: As mentioned in https://github.com/mysociety/alaveteli/pull/3110#discussion_r56967008, it would be good to get some  [documentation](http://tomdoc.org/) about what the arguments do, particularly mentioning where to find the established categories and actions (i.e. `AnalyticsEvent::Category` and `AnalyticsEvent::Action`)

spec/helpers/analytics_helper_spec.rb
garethrees: Just to clarify, I'm not sure what the intent of "is not nil" was here.

Given the expectation in this spec, I'd expect its description to read: `it "raises an ArgumentError if the :value option is not an Integer"`.

Should we be testing for some `nil` case in a new spec?

garethrees:
Needs a bit of extra work, and is waiting on https://github.com/mysociety/alaveteli/pull/3160 before it can be merged anyway. Also needs a changelog entry.

lib/analytics_event.rb
```diff
@@ -13,7 +13,7 @@ module Action
     TWITTER_EXIT = 'Twitter Exit'
     WORDPRESS_EXIT = 'WordPress Exit'
     WIDGET_VOTE = 'Vote'
-    WIDGET_SIGNIN = 'Sign in to vote'
+    WIDGET_SIGNIN = 'Sign in to track'
```
lizconlan: better/more accurate summary of click intention - from https://github.com/mysociety/alaveteli/pull/3160#discussion_r61054212

spec/helpers/analytics_helper_spec.rb
lizconlan: the valid values are Integer or nil/no value &lt;/not_helping&gt;

lizconlan: (rewriting tests)

app/views/widgets/show.html.erb
lizconlan: \*sigh\* overly literal thinking, of course I don't have to tell `:onclick =>` to wrap the results of the method call in quotes, the Rails helper will sort that out for itself - fixed in d19479d

lizconlan:
Stalling on the CHANGELOG for the moment, will add when the code feels less subject to change.

garethrees:
👍

Issue #3106, opened 2016-03-18
Problem with PDF attachment download on specific message


RichardTaylor:
Attempting to download the PDF files from:

https://www.whatdotheyknow.com/request/order_inviting_selected_accredit#incoming-784119

results in a file containing one " " character. 

The PDF files are present in the incoming message. "View as HTML" works, as does downloading the raw message via the admin interface and viewing it via an email application. 

As far as I know this isn't a more general problem, but it might be worth understanding in case it is.

RichardTaylor:
Potentially another case of the same bug at:

https://www.whatdotheyknow.com/request/west_sussex_youth_offending_serv#incoming-784527

FOIMonkey:
And another: https://www.whatdotheyknow.com/request/may_i_see_the_permission_which_a#incoming-783539

FOIMonkey:
I picked ten of the most recent responses with attached PDFs at random and 7 were affected.

More examples:

- https://www.whatdotheyknow.com/request/employment_tribunals_18#incoming-784973
- https://www.whatdotheyknow.com/request/humberside_police_fraud_unit#incoming-784972
- https://www.whatdotheyknow.com/request/unoccupied_commercial_properties_4#incoming-784966
- https://www.whatdotheyknow.com/request/vice_chancellor_appointment_of_t_6#incoming-784964
- https://www.whatdotheyknow.com/request/organisational_structure_charts_231#incoming-784998
- https://www.whatdotheyknow.com/request/patients_medical_records_and_isl#incoming-784991
- https://www.whatdotheyknow.com/request/organisational_structure_charts_231#incoming-784998
- https://www.whatdotheyknow.com/request/deputy_district_judge_christophe#incoming-784988

garethrees:
Here's some code to pull out the varnish URLs and filesystem paths that we should clear:

```ruby
# For clearing varnish
url_paths = []
FoiAttachment.joins(:incoming_message).
  where(["incoming_messages.created_at > ?", 6.days.ago]).
    where(["content_type = ?", "application/pdf"]).
      find_each do |attachment|
        url_paths << 
          app.get_attachment_path(:id => attachment.incoming_message.info_request_id,
                                  :incoming_message_id => attachment.incoming_message_id,
                                  :part => attachment.url_part_number,
                                  :file_name => attachment.filename)
end;0


# For clearing filesystem
controller = ApplicationController.new
controller.request = ActionDispatch::TestRequest.new

file_paths = []
[...]

PR #3113, opened 2016-03-21
Hotfix/0.24.0.3

garethrees:
Fixes https://github.com/mysociety/alaveteli/issues/3106

### DIFF
--- a/lib/alaveteli_text_masker.rb
+++ b/lib/alaveteli_text_masker.rb
@@ -144,6 +144,8 @@ module AlaveteliTextMasker
         else
           recompressed_text
         end
+      else
+        text
       end
     end
   end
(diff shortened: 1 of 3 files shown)


lib/alaveteli_text_masker.rb
```diff
@@ -139,11 +139,9 @@ def apply_pdf_masks(text, options = {})
           Rails.logger.warn "Unable to compress PDF; problem with your pdftk version?"
         end
 
-        if recompressed_text.blank?
-          text
-        else
-          recompressed_text
-        end
+        recompressed_text
```
crowbot: I think, to duplicate the logic of the destructive version of this method, we still want the `.blank?` check here and to return text if it returns true - it covers the case where we're using `censored_uncompressed_text` which is in itself blank.

garethrees: Fixed in https://github.com/mysociety/alaveteli/pull/3113/commits/791b676330316a78eca32287c1a11117b73794d5.

PR #3115, opened 2016-03-21
Allow timeout configuration for reading in the blog feed

lizconlan:
Set the blog feed timeout via the `BLOG_TIMEOUT` configuration option (expressed as a number of seconds, defaults to 60 which is the standard read timeout for `OpenURI.open`)

e.g:

```
BLOG_TIMEOUT: 100
```

Connects to mysociety/asktheeu-theme#14

### DIFF
--- a/spec/lib/quiet_open_spec.rb
+++ b/spec/lib/quiet_open_spec.rb
@@ -1,0 +1,26 @@
+# -*- encoding : utf-8 -*-
+require File.expand_path(File.dirname(__FILE__) + '/../spec_helper')
+
+describe "quietly_try_to_open" do
+
+  let(:controller) { double(ApplicationController) }
+  let(:empty_stream) { double(URI::HTTP) }
+  let(:uri) { "http://example.com/feed" }
+
+  before do
+    allow(empty_stream).to receive(:read).and_return("")
+  end
+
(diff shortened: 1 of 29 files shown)


spec/lib/quiet_open_spec.rb
```diff
@@ -0,0 +1,26 @@
+# -*- encoding : utf-8 -*-
+require File.expand_path(File.dirname(__FILE__) + '/../spec_helper')
+
+describe "quietly_try_to_open" do
+
+  let(:fakeController) { double(ApplicationController) }
```
crowbot: Any particular reason for using camel case here? Looks odd to me.

lizconlan: Hmm, looks almost like I carelessly renamed it (and `emptyStream`) as an afterthought, or I had an oddly formatted example open in another window. No idea why I did that. Distractingly odd. Will fix.

lizconlan: Although interestingly, if you call it `FakeController`, the spec doesn't pass.

lizconlan:
Closed in favour of [hotfix/0.24.0.6](https://github.com/mysociety/alaveteli/tree/hotfix/0.24.0.6)

PR #3138, opened 2016-03-31
Stop triggering gnuplot warning messages

### LINKED ISSUE #3126: Empty range warnings in graph generation
Warning: empty y range [1:1], adjusting to [0.99:1.01] e.g on colombia site

lizconlan:
Sets a default y-range start point of 0 and avoids trying to plot a cumulative value line from a single datapoint.

Closes #3126

### DIFF
--- a/lib/tasks/graphs.rake
+++ b/lib/tasks/graphs.rake
@@ -103,15 +103,19 @@ namespace :graphs do
 
         plot_datasets(graph_param_sets, plot.data)
 
-        # plot cumulative user totals
-        options.merge!({
-          :title => "cumulative total number of users",
-          :axes => "x1y2",
-          :with => "lines",
-          :linewidth => 1,
-          :linecolor => COLOURS[:lightgreen],
-          :using => "1:3"})
-        plot_data_from_columns(all_users, options, plot.data)
+        # skip this if there is just a single datapoint
+        # (counts the number of values in the first column)
+        if all_users[0].size > 1
+          # plot cumulative user totals
+          options.merge!({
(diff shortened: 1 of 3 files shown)


lib/tasks/graphs.rake
```diff
@@ -45,11 +45,13 @@ namespace :graphs do
         # primary y-axis
         plot.set("ytics nomirror")
         plot.ylabel("number of users on the calendar day")
+        plot.yrange("[0:]")
 
         # secondary y-axis
         plot.set("y2tics tc lt 2")
         plot.set('y2label "cumulative total number of users" tc lt 2')
         plot.set('format y2 "%.0f"')
+        plot.y2range("[0:]")
```
lizconlan: I'm not sure whether setting the `y2` range as well is strictly necessary but it doesn't seem to hurt

lizconlan:
Now quits `generate_request_creation_graph` with a warning message that the task was skipped if there is a single datapoint (I thought I was already doing that but I'd got the check in the wrong place).

script/request-creation-graph
```diff
@@ -6,5 +6,5 @@
 # Email: hello@mysociety.org. WWW: http://www.mysociety.org/
 
 cd `dirname $0`
-bundle exec rake --silent "$@" graphs:generate_request_creation_graph
+bundle exec rake "$@" graphs:generate_request_creation_graph
```
crowbot: Should the default for `request-creation-graph` and `user-use-graph` be the same here?

lizconlan: Good catch - thanks! (fixing...)

PR #3149, opened 2016-04-04
[#2819] OutgoingMessage #to, #from and #subject

garethrees:
Work on #2819

### DIFF
--- a/spec/models/outgoing_message_spec.rb
+++ b/spec/models/outgoing_message_spec.rb
@@ -167,11 +167,99 @@ describe OutgoingMessage do
 
   describe '#subject' do
 
-    it 'uses the request title with the law prefixed' do
-      request = FactoryGirl.create(:info_request, :title => 'Example Request')
-      message = FactoryGirl.build(:initial_request, :info_request => request)
-      expected = 'Freedom of Information request - Example Request'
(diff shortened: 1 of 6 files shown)


spec/models/outgoing_message_spec.rb
```diff
@@ -161,11 +161,99 @@
 
   describe '#subject' do
```
lizconlan: Are these effectively a replacement for the `"when working out follow up subjects"` specs in [/spec/mailers/outgoing_mailer_spec.rb](https://github.com/mysociety/alaveteli/blob/2819-outgoing-message-to-from-subject-2/spec/mailers/outgoing_mailer_spec.rb#L61)? If the functionality's moved we should probably get rid of the `OutgoingMailer` versions.

lizconlan:
:+1: Looks good, I'm currently a bit confused about similar functionality between `OutgoingMessage` and `OutgoingMailer` but that's probably nothing to worry about here

garethrees:
Yeah, the idea was to extract from `OutgoingMailer`, but some places still have dependencies on the methods in it. I've added [a note in #2891](https://github.com/mysociety/alaveteli/issues/2819#issuecomment-205720203) that we'll want to clean this up at some point later on.

PR #3150, opened 2016-04-04
[#2819] Allow an OutgoingMessage to be an external_review

garethrees:
Work on #2819

TODO: For a better spec, requires #3149 so that the factory we create is a followup, rather than initial request; that said, we use `:initial_request` for the external review spec.


lizconlan:
:+1:

garethrees:
Merging as-is. There's no actual constraint on what state an `OutgoingMessage` needs to be in for a given `what_doing` value, so decided not to try to imply one with the spec.

PR #3151, opened 2016-04-05
Fix the corrupt cookie bug

### LINKED ISSUE #2760: Corrupt cookie
Every so often, I get logged out of the website, and attempts to log back in fail - I am just returned to the log-in screen. To resolve this, I have to delete the WhatDoTheyKnow cookies in my browser; I can then immediately log in again. This seems to occur without any specific prompt, though it may happen sometimes on my return when I've been away from the computer for a bit. There have been two users who have reported similar over the last couple of weeks. 

I am using Firefox 39.0 on Windows 8.1 with all the updates installed.

lizconlan:
Fixes a bug where the `:ttl` session value was never unset meaning that users who had logged in having checked the **"remember me"** box would be logged out on their next page visit if the session expiry time indicated by the `:ttl` timestamp had already passed, or at some seemingly random interval if the expiry was still in the future at that point. (The `:ttl` value is not updated during a `:remember_me` session.)

The `:ttl` is now unset on signout and forcibly on the first page visit after signing in with "remember me" checked in case the session had expired. There is still a small risk that this could reoccur if `validate_session_timestamp` is called before `session_remember_me` as `validate_session_timestamp` relies on the `:ttl` value rather than checking whether `:remember_me` is present (I am happy to introduce an extra element to the existing if statement or a `return if` guard if this feels safer).

Improves the signin specs.

Closes #2760

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -129,7 +129,7 @@ class ApplicationController < ActionController::Base
   # is not replayable forever
   SESSION_TTL = 3.hours
   def validate_session_timestamp
-    if session[:user_id] && session.key?(:ttl) && session[:ttl] < SESSION_TTL.ago
+    if session[:user_id] && session[:ttl] && session[:ttl] < SESSION_TTL.ago
       clear_session_credentials
     end
   end
@@ -147,6 +147,7 @@ class ApplicationController < ActionController::Base
     session[:admin_name] = nil
     session[:change_password_post_redirect_id] = nil
     session[:post_redirect_token] = nil
+    session[:ttl] = nil
   end
 
   def render_exception(exception)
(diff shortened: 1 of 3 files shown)


garethrees:
Turkey have reported an issue that sounds like it could be this, so would be good to make this a hotfix to master instead of in to develop.

lizconlan:
>  @crowbot: seems like it could be simpler to set `session[:ttl]` to `nil` when setting remember_me for the session though?

Hmm, I can see why past me got suckered into thinking otherwise but I'm going to give that a try as it might resolve the "not set until next page loaded" oddity.

I'd prefer to keep d0c87aa as-is though as that feels like it's completing the signout functionality by resetting all the things

lizconlan:
Closed in favour of [the hotfix branch](https://github.com/mysociety/alaveteli/tree/hotfix/0.24.1.2)

PR #3153, opened 2016-04-06
Non uk help pages

crowbot:
This PR, in conjunction with https://github.com/mysociety/alavetelitheme/pull/78, moves the help page examples to alavetelitheme. It also standardises the way we create links from the app to those help pages that include anchor specifications, to make it easier to find them in the code in future.


lizconlan:
Nice! 👍

PR #3157, opened 2016-04-07
Documentation update for https://github.com/mysociety/alaveteli/pull/…

crowbot:
…3153 and https://github.com/mysociety/alavetelitheme/pull/78.


lizconlan:
👍

Not noted inline but a slight word of caution about using 2 spaces between sentences in Markdown that (aside from slightly confusing a particularly literal-minded reviewer) if this falls at the end of a line it gets interpreted as a `<br>`. Er, which it doesn't here so feel free to ignore this comment entirely!

PR #3160, opened 2016-04-07
[#3107] Migrate analytics code from `ga.js` to `analytics.js`

### LINKED ISSUE #3107: Migrate to modern Google Analytics code

lizconlan:
Checking whether we were using any custom code took longer than making any of the changes.

Will still have to double check that none of the analytics code has been overridden in themes before deploying to a site.

May have a knock-on effect for some tickets which are currently in review (see links in the ticket itself)

Fixes #3107

### DIFF
--- a/app/views/widgets/show.html.erb
+++ b/app/views/widgets/show.html.erb
@@ -3,12 +3,14 @@
 </head>
 <body>
 <% unless AlaveteliConfiguration::ga_code.empty?  %>
-  <script type="text/javascript">
-    var gaJsHost = (("https:" == document.location.protocol) ? "https://ssl." : "http://www.");
-    document.write(unescape("%3Cscript src='" + gaJsHost + "google-analytics.com/ga.js' type='text/javascript'%3E%3C/script%3E"));
-  </script>
-  <script type="text/javascript">
-    var pageTracker = _gat._getTracker("<%= AlaveteliConfiguration::ga_code %>");
+  <script>
+  (function(i,s,o,g,r,a,m){i['GoogleAnalyticsObject']=r;i[r]=i[r]||function(){
+  (i[r].q=i[r].q||[]).push(arguments)},i[r].l=1*new Date();a=s.createElement(o),
+  m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)
+  })(window,document,'script','https://www.google-analytics.com/analytics.js','ga');
+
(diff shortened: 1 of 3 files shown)


app/views/widgets/show.html.erb
```diff
@@ -37,24 +39,24 @@
...
           :class => 'alaveteli-widget__bottom__action alaveteli-widget__button--unsubscribe',
-          :onclick => "if (pageTracker) { pageTracker._trackEvent('widgets', 'unsubscribe', window.top.location.href, 1)};" do %>
+          :onclick => "if (ga) { ga('send', 'event', 'Widget Clicked', 'Unsubscribe', window.top.location.href, 1)};" do %>
             <%= _('Unsubscribe') %>
         <% end %>
       <% elsif @existing_vote %>
         <%= link_to track_request_url(:url_title => @info_request.url_title, :feed => 'track'), :target => '_top',
           :class => 'alaveteli-widget__bottom__action alaveteli-widget__button--sign-in-to-track',
-          :onclick => "if (pageTracker) { pageTracker._trackEvent('widgets', 'track', window.top.location.href, 1)};" do %>
+          :onclick => "if (ga) { ga('send', 'event', 'Widget Clicked', 'Sign in to vote', window.top.location.href, 1)};" do %>
```
crowbot: Should this be "sign in to track" ? They've already voted.

lizconlan: Yes, it probably should, thanks! I'll pick that up in #3110

lizconlan: (as well, not instead)

PR #3162, opened 2016-04-08
WIP - upgrade gems

### LINKED ISSUE #3161: Upgrade gems to latest Rails 3.2 compatible versions
Split out from #2968

- [x] Isolate the cause of the breakage after foundation / scss update
- [x] Fix problem or pin gem version of foundation / scss after above
- [x] Update secure headers (#3196)

lizconlan:
Slowly upgrading gems to the latest Rails 3.2 versions that support Ruby 1.9.3. This is more likely to be an integration test/cross-check than an actual reviewable PR

Closes #3161


Gemfile.lock
```diff
@@ -170,7 +170,7 @@ GEM
     holidays (2.2.0)
     htmlentities (4.3.4)
     i18n (0.7.0)
-    icalendar (1.4.3)
+    icalendar (2.3.0)
```
lizconlan: There's a compatibility warning install message but it looks like we're only using `#parse` which doesn't appear to be affected

lizconlan: I stand corrected: 
```
Failures:

  1) HolidayImport when populating a set of holidays to import from a feed should add an error if the calendar cannot be parsed
     Failure/Error: @holiday_import.populate
     NoMethodError:
       undefined method `events' for nil:NilClass
     # ./app/models/holiday_import.rb:68:in `populate_from_ical_feed'
     # ./app/models/holiday_import.rb:33:in `populate'
     # ./spec/models/holiday_import_spec.rb:146:in `block (3 levels) in <top (required)>'
```

lizconlan: (Fixed in `a6debad`)

lizconlan:
Capistrano needs to be upgraded to 3.1 to work with Rails 4 but that's a major task as all the setup and scripting has changed. Will not be addressed as part of this pre-upgrade ticket.

lizconlan:
superseded by individual PRs

PR #3181, opened 2016-04-14
[#3161] Update to will_paginate 3.0.7

lizconlan:
Version 3.1.0 is the latest but [removes the `:order` param](https://github.com/mislav/will_paginate/releases/tag/v3.1.0) in favour of respecting the order of the dataset so will involve quite a bit of work to implement. There are also [possible issues with translations](https://github.com/mislav/will_paginate/issues/466).

[Rails 4.1 support is introduced in 3.0.6](https://github.com/mislav/will_paginate/releases/tag/v3.0.6) with 3.0.7 providing a bug fix.

Connects to #3161


garethrees:
> Version 3.1.0 is the latest but removes the :order param in favour of respecting the order of the dataset so will involve quite a bit of work to implement. There are also possible issues with translations.
>
> Rails 4.1 support is introduced in 3.0.6 with 3.0.7 providing a bug fix.

Would be good to have comments like this in the commit message so that you can `git blame` and figure out why a gem is at a particular version.

PR #3194, opened 2016-04-14
[#3161] Upgrade sass/css gems

lizconlan:
Connects to #3161


lizconlan:
We have the opportunity to upgrade the bootstrap gem here but [bootstrap 3 is not backwards compatible with bootstrap 2 ](http://getbootstrap.com/migration/) so we without taking the time to go through a proper migration process, this could be a breaking change.

garethrees:
> We have the opportunity to upgrade the bootstrap gem here

I'd stick with the latest version of 2, and ticket upgrading to 3.

lizconlan:
\*nudges bootstrap to the latest minor point version\*

Gemfile
```diff
@@ -33,6 +33,7 @@ gem 'rails-i18n', '~> 0.7.3'
 gem 'recaptcha', '~> 0.4.0', :require => 'recaptcha/rails'
 gem 'rmagick', '~> 2.14.0'
 gem 'ruby-msg', '~> 1.5.0',  :git => 'https://github.com/mysociety/ruby-msg.git', :ref => 'f9f928ed76c024b4bc3a08bc1a59beb62df36663'
+gem 'sass', '3.4.21'
```
garethrees: Isn't this a dependency of `compass-rails`? Shouldn't we let that handle the version of the `sass` gem?

Gemfile
```diff
@@ -11,7 +11,7 @@ gem 'charlock_holmes', '~> 0.6.9.4'
 gem 'dynamic_form', '~> 1.1.4'
 gem 'exception_notification', '~> 3.0.1'
 gem 'fancybox-rails', '~> 0.2.1'
-gem 'foundation-rails', '~> 5.2.1.0'
+gem 'foundation-rails', '~> 5.5.3.2'
```
garethrees: Latest version of `foundation-rails` is [6.2.1.0](https://rubygems.org/gems/foundation-rails). Worth adding notes to the commit message about why can't we go further.

Gemfile.lock
```diff
@@ -76,7 +76,7 @@ GEM
       activerecord (>= 3.2, < 6.0)
       rake (~> 10.4)
     arel (3.0.3)
-    bootstrap-sass (2.3.1.2)
+    bootstrap-sass (2.3.1.3)
```
garethrees: Can't we move to [2.3.2.2](https://rubygems.org/gems/bootstrap-sass/versions/2.3.2.2)? Commit message comments as to why not would be worth doing.

Gemfile
lizconlan: Just checking that the deprecation warning from sass 3.4.22 doesn't break any tests (will result in either a fixup that removes sass from the Gemfile or a commit note to explain why it's pinned)

lizconlan: Ah

    Failure/Error: expect(r.out).to eq("")

       expected: ""
            got: "DEPRECATION WARNING:\nSass 3.5 will no longer support Ruby 1.9.3.\nPlease upgrade to Ruby 2.0.0 or greater as soon as possible.\n\n"

       (compared using ==)

       Diff:
       @@ -1 +1,4 @@
       +DEPRECATION WARNING:
       +Sass 3.5 will no longer support Ruby 1.9.3.
       +Please upgrade to Ruby 2.0.0 or greater as soon as possible.

     # ./spec/script/mailin_spec.rb:28:in `block (2 levels) in <top (required)>'

(spec run abandoned at that point)

Gemfile
lizconlan: For much the same reason as not moving to Bootstrap 3, Foundation 6 is a whole new framework (and as I had to look that up, I should add a note)

Gemfile.lock
lizconlan: Probably because I didn't update the version matcher in the Gemfile correctly \*sigh\*
Cheers!

Gemfile
lizconlan: Foundation upgrade ticket: #3230

Gemfile
```diff
@@ -11,7 +11,7 @@ gem 'charlock_holmes', '~> 0.6.9.4'
 gem 'dynamic_form', '~> 1.1.4'
 gem 'exception_notification', '~> 3.0.1'
 gem 'fancybox-rails', '~> 0.2.1'
-gem 'foundation-rails', '~> 5.5.3.2'
+gem 'foundation-rails', '~> 5.5.3.2' # 6.x.x.x versions cause problems (see blame)
```
garethrees: Don't really need the comment due to the nice commit message. I know I added one in the past, but I think now we've done a few of these its already convention to check the blame for why a gem might be pinned at a particular version.

PR #3196, opened 2016-04-14
[#3161] Upgrade secure_headers to 3.1.2

### LINKED ISSUE #2986: Update secure_headers gem
They've [released version 3](https://github.com/twitter/secureheaders/blob/master/upgrading-to-3-0.md) which has some deprecations that we need to address. I've updated to 2.4.x in https://github.com/mysociety/alaveteli/pull/2964 (2eb4cfa) but updating further will require a bit of work. Probably reasonably easy though.

lizconlan:
Upgrade guide (from 2.x to 3.x) here - https://github.com/twitter/secureheaders/blob/master/README.md

Not entirely convinced I've translated the old settings to the new format correctly

Connects to #3161
Fixes #2986

### DIFF
--- a/config/initializers/secure_headers.rb
+++ b/config/initializers/secure_headers.rb
@@ -1,12 +1,13 @@
 # -*- encoding : utf-8 -*-
-::SecureHeaders::Configuration.configure do |config|
+::SecureHeaders::Configuration.default do |config|
 
   # https://tools.ietf.org/html/rfc6797
   if AlaveteliConfiguration::force_ssl
-    config.hsts = { :max_age => 20.years.to_i, :include_subdomains => true }
+    config.hsts = "max-age=#{20.years.to_i}; includeSubdomains"
   else
-    config.hsts = false
+    config.hsts = SecureHeaders::OPT_OUT #don't send on non https sites
   end
+
   # https://tools.ietf.org/html/draft-ietf-websec-x-frame-options-02
   config.x_frame_options = "sameorigin"
 
@@ -14,11 +15,16 @@
   config.x_content_type_options = "nosniff"
 
   # http://msdn.microsoft.com/en-us/library/dd565647%28v=vs.85%29.aspx
-  config.x_xss_protection = { :value => 1 }
+  config.x_xss_protection = "1"
 
   # https://w3c.github.io/webappsec/specs/content-security-policy/
-  config.csp = false
+  config.csp = SecureHeaders::OPT_OUT
 
   # https://www.nwebsec.com/HttpHeaders/SecurityHeaders/XDownloadOptions
-  config.x_download_options = false
+  config.x_download_options = SecureHeaders::OPT_OUT
+end
+
+ # Allow individual actions to allow frames
+::SecureHeaders::Configuration.override(:allow_frames) do |config|
+  config.x_frame_options = SecureHeaders::OPT_OUT
 end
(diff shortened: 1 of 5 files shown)


config/initializers/secure_headers.rb
```diff
@@ -1,11 +1,9 @@
 # -*- encoding : utf-8 -*-
-::SecureHeaders::Configuration.configure do |config|
+::SecureHeaders::Configuration.default do |config|
 
   # https://tools.ietf.org/html/rfc6797
   if AlaveteliConfiguration::force_ssl
-    config.hsts = { :max_age => 20.years.to_i, :include_subdomains => true }
-  else
-    config.hsts = false
+    config.hsts = "max-age=#{20.years.to_i}; includeSubdomains; preload"
```
garethrees: As far as I can tell, we weren't using the `"preload"` directive, as it's only set [if given the `:preload` key](https://github.com/twitter/secureheaders/blob/2.x/lib/secure_headers/headers/strict_transport_security.rb#L35).

I think we also want to use `SecureHeaders::OPT_OUT` when we're not using SSL, as:

> All `nil` values will fallback to their default values. `SecureHeaders::OPT_OUT` will disable the header entirely.

Since we're not setting `config.hsts` when `force_ssl` is false, I'm assuming we're falling back to the default.

config/initializers/secure_headers.rb
```diff
@@ -14,11 +12,17 @@
   config.x_content_type_options = "nosniff"
 
   # http://msdn.microsoft.com/en-us/library/dd565647%28v=vs.85%29.aspx
-  config.x_xss_protection = { :value => 1 }
+  config.x_xss_protection = "1; mode=block"
```
garethrees: We're not matching the previous behaviour here.

The default value has changed between [3.0](https://github.com/twitter/secureheaders/blob/master/lib/secure_headers/headers/x_xss_protection.rb#L5) and [2.0](https://github.com/twitter/secureheaders/blob/2.x/lib/secure_headers/headers/x_xss_protection.rb#L6). Don't know if we should be using the new default here.

config/initializers/secure_headers.rb
```diff
@@ -14,11 +12,17 @@
...
 
   # https://w3c.github.io/webappsec/specs/content-security-policy/
-  config.csp = false
+  config.csp = {
```
garethrees: Should we be using `SecureHeaders::OPT_OUT` here to match the previous behaviour?

config/initializers/secure_headers.rb
```diff
@@ -14,11 +12,17 @@
...
+    # "meta" values. these will shaped the header, but the values are not included in the header.
+    report_only:  true,
+
+    # directive values: these values will directly translate into source directives
+    default_src: %w(https: 'self'),
+  }
 
   # https://www.nwebsec.com/HttpHeaders/SecurityHeaders/XDownloadOptions
-  config.x_download_options = false
+  config.x_download_options = nil
```
garethrees: Should we be using `SecureHeaders::OPT_OUT` here to match the previous behaviour?

lizconlan: Potentially, although this is an odd one...

(from https://www.nwebsec.com/HttpHeaders/SecurityHeaders/XDownloadOptions)

> The X-Download-Options is specific to IE 8, and is related to how IE 8 handles downloaded HTML files. Turns out if you download an HTML file from a web page and chooses to "Open" it in IE, it will execute in the context of the web site. That means that any scripts in that file will also execute with the origin of the web site.

So I'm tempted to set it to "noopen" which I'm taking to mean protect IE8 users from <strike>themselves</strike> accidentally running scripts locally. Although sending extra bytes to everyone else isn't massively useful.

lizconlan: (Set to `SecureHeaders::OPT_OUT` for now as those kind of discussions feel like they're out of scope for a gem upgrade task)

lizconlan:
Headers (as viewed in Chrome) over HTTPS...

screen shot 2016-05-18 at 10 09 04

and plain HTTP...

screen shot 2016-05-18 at 10 11 04

PR #3198, opened 2016-04-14
[#3161] Update the internationalisation gems

lizconlan:
Connects to #3161


Gemfile
```diff
@@ -45,8 +45,8 @@ gem 'xml-simple', '~> 1.1.2', :require => 'xmlsimple'
 gem 'zip', '~> 2.0.2'
 
 # Gems related to internationalisation
-gem 'gettext_i18n_rails', '~> 0.9.4' # Later versions cause error (see blame)
-gem 'gettext', '~> 2.3.9'
+gem 'gettext_i18n_rails', '~> 1.5.0'
+gem 'gettext', '~> 3.0.2'
```
garethrees: Latest version of gettext is [3.2.2](https://rubygems.org/gems/gettext). How come we can't go further? Worth adding to the commit message.

lizconlan: Er, I have no idea why I did that. Testing against 3.2.2 to see what happens.

garethrees:
👍

PR #3202, opened 2016-04-15
[#3161] Upgrade development and  test gems

lizconlan:
Connects to #3161


spec/factories/info_requests.rb
```diff
@@ -6,34 +6,34 @@
     public_body
     user
 
-    after_create do |info_request, evaluator|
-      FactoryGirl.create(:initial_request, :info_request => info_request)
+    after(:create) do |info_request, evaluator|
+      create(:initial_request, :info_request_id => info_request.id)
```
garethrees: Out of interest, why did `:info_request` need to get changed to `:info_request_id` here?

Gemfile
```diff
@@ -75,12 +75,12 @@ end
...
   gem 'bullet', '~> 5.0.0'
   gem 'factory_girl_rails', '~> 4.7.0'
-  gem 'pry', '~> 0.10.3'
   gem 'rspec-activemodel-mocks', '~> 1.0.1'
   gem 'rspec-rails', '~> 3.4.0'
   gem 'test-unit', '~> 3.1.0'
   gem 'spork-rails', '~> 4.0.0'
-  gem 'pry-nav', '~> 0.2.4'
+  gem 'pry-debugger', '~> 0.2.3', :platforms => :ruby_19
+  gem 'pry-byebug',  '~> 3.3.0', :platforms => [:ruby_20, :ruby_21]
```
garethrees: The version of bundler that we use (in production) is too old to recognise `:ruby_21` as a platform:

```sh
$ bundle
`ruby_21` is not a valid platform. The available options are: [:ruby, :ruby_18, :ruby_19, :ruby_20, :mri, :mri_18, :mri_19,
:mri_20, :rbx, :jruby, :mswin, :mingw, :mingw_18, :mingw_19, :mingw_20]
$ bundle --version
Bundler version 1.3.5
```

lizconlan: Bah, had similar issues with `:ruby_22` and `:ruby_23` on slightly more recent bundler versions. Take it out for now I guess.

spec/factories/info_requests.rb
lizconlan: It was causing an odd deprecation warning along the lines of

```
DEPRECATION WARNING: Accessing attributes from initialize_with when duplicate assignment is enabled is deprecated; use FactoryGirl.duplicate_attribute_assignment_from_initialize_with = false.
```

...but that might have been specific to one of the 3.x.x branches I needed to go through to get to 4.7.0. I can put it back if you'd prefer. In fact as I've just put it back to test whether it breaks, might as well add a fixup commit for it

garethrees: I don't "care" either way as such – whichever works and doesn't cause a breakage/warning :)

lizconlan: In fact this may even have been a "let's try something strange to see if it fixes it" test line which got committed by mistake. If I made notes at the time, I appear to have misplaced them.

Gemfile
```diff
@@ -80,7 +80,7 @@ group :test, :development do
   gem 'test-unit', '~> 3.1.0'
   gem 'spork-rails', '~> 4.0.0'
   gem 'pry-debugger', '~> 0.2.3', :platforms => :ruby_19
-  gem 'pry-byebug',  '~> 3.3.0', :platforms => [:ruby_20, :ruby_21]
+  gem 'pry-byebug',  '~> 3.3.0', :platforms => [:ruby_20]
```
garethrees: What's going to happen if you're using ruby 2.1? Guess you just won't get pry?

garethrees:
Looks good providing there are no conflicts/failures once rebased. 👍

Issue #3204, opened 2016-04-15
Release-Manager's-checklist error


crowbot:
The instructions linked to from "You may need to manually remove obsolete msgids" may delete translations from Transifex if they're run when the ```locale/en/app.po``` file has fuzzy translations. The command removes any fuzzy strings entirely, and if the translations are subsequently uploaded to Transifex, it uses the ```en``` translation file as a source, rather than a ```pot``` file. This results in the ```msgids``` that were marked as fuzzy being removed from all translations, along with any corresponding ```msgstrs```. An example of this can be seen in https://github.com/mysociety/alaveteli/commit/89b743219df674a9b90311c94027c5aa48a9e772

crowbot:
Fixed.

garethrees:
Thanks!

PR #3206, opened 2016-04-18
Add task for removing fuzzy translations.

crowbot:
RE: https://github.com/mysociety/alaveteli/issues/3204


garethrees:
Reviewed by adding some commits. I've moved the logic in to a tested class and called that from the rake task. I think we want to handle fuzzy plural translations, otherwise its going to ignore them.

garethrees:
👍

PR #3209, opened 2016-04-21
[#624] Add a MailServerLogsController

garethrees:
Work on #624 

This PR adds `MailServerLogsController#index` which shows the mail server logs for an `OutgoingMessage`. It responds to Text (to render in to the flashy JS dropdown) and HTML (no-JS fallback).

I wanted to display the message body in context, but was too difficult as the correspondence partials depend on an `InfoRequestEvent`. Here's a hacked in example:

screen shot 2016-04-21 at 10 40 55

Instead, I just went with the simple thing:

screen shot 2016-04-21 at 12 33 25

Link from correspondence box:

screen shot 2016-04-21 at 12 54 45

I also wanted to link back to `/request/:url_name/mail_server_logs` in the event that there are no logs for the `OutgoingMessage` – hence setting `@subject` rather than `@outgoing_message` – but I'm undecided whether this would make sense to a user. Anyway, the infrastructure is there, so we'd just have to add some extra ability to handle a `:request_id` param if we decide to do that.


garethrees:
Oops messed a spec up – fixing.

garethrees:
Wonder if we want to add the route to `robots.txt` to try to prevent crawlers seeing the emails in the mail logs?

zarino:
Pull request that adds ajax loading of the logs: https://github.com/mysociety/alaveteli/pull/3222

app/controllers/mail_server_logs_controller.rb
```diff
@@ -0,0 +1,34 @@
...
+class MailServerLogsController < ApplicationController
+  before_filter :set_subject, :check_prominence
+
+  def index
+    @title =
+      case @subject.class.to_s
+      when 'OutgoingMessage'
+        _('Mail Server Logs for Outgoing Message #{{id}}', :id => @subject.id)
+      end
+    @mail_server_logs = @subject.mail_server_logs
```
garethrees: This should be:

```ruby
@mail_server_logs = @subject.mail_server_logs.map(&:line)
```

app/views/mail_server_logs/index.html.erb
```diff
@@ -0,0 +1,16 @@
...
+  <%= _("These logs show what happened to this message as it passed through " \
+        "our {{mta_type}} mail server. You can send these logs to an " \
+        "authority to help them diagnose problems receiving email from " \
+        "{{site_name}}.",
+        :mta_type => AlaveteliConfiguration.mta_log_type.capitalize,
+        :site_name => AlaveteliConfiguration.site_name) %>
+</p>
+
+<% if @mail_server_logs.any? %>
+<pre><%= @mail_server_logs.join("\n") %></pre>
```
garethrees: Seems like lines are terminated with `"\n"` so can call `join` with no args.

spec/controllers/mail_server_logs_controller_spec.rb
```diff
@@ -0,0 +1,86 @@
+# -*- encoding : utf-8 -*-
+require File.expand_path(File.dirname(__FILE__) + '/../spec_helper')
+
+describe MailServerLogsController do
+
+  before do
+    @logs = ['2015-09-22 17:36:56 [2035] 1ZeQYq-0000Wm-1V => body@example.com F=<request@example.com> P=<request@example.com> R=dnslookup T=remote_smtp S=1685 H=mail.example.com [62.208.144.158]:25 C="250 2.0.0 Ok: queued as 95FC94583B8" QT=0s DT=0s',
```
garethrees: These should be `mock_model(MailServerLog, :line => '…')`

garethrees:
Probably also want to filter lines through the text masks and censor rules

garethrees:
Some screenshots:

## Not signed in

not-signed-in

## Signed in as request owner

signed-in-as-request-owner

## Signed in as admin

signed-in-as-admin

## Signed in as other user

signed-in-as-other-user

I've been doing something like this to mock the log output for testing:

```diff
diff --git a/app/models/outgoing_message.rb b/app/models/outgoing_message.rb
index 2337b27..542ed42 100644
--- a/app/models/outgoing_message.rb
+++ b/app/models/outgoing_message.rb
@@ -207,6 +207,7 @@ class OutgoingMessage < ActiveRecord::Base
   #
   # Returns an Array.
   def mail_server_logs
+    return MailServerLog.where(:info_request_id => 123)
     case AlaveteliConfiguration.mta_log_type.to_sym
     when :exim
       exim_mail_server_logs
```

app/controllers/mail_server_logs_controller.rb
```diff
@@ -0,0 +1,41 @@
...
+
+  def set_subject
+    @subject =
+      if params[:outgoing_message_id]
+        OutgoingMessage.find(params[:outgoing_message_id])
+      end
+  end
+
+  def check_prominence
+    unless @subject.user_can_view?(authenticated_user)
```
lizconlan: what happens if `@subject` has not been set? (i.e if there's not an `OutgoingMessage`) will this cause an error?

lizconlan:
The user access logic looks sound, not sure that it will work as intended if there's no associated `OutgoingMesage` though (however unlikely that may be with the current view logic as a link shouldn't appear if there's no message for it)

app/controllers/mail_server_logs_controller.rb
garethrees: The page will 404, which I think is the correct/expected behaviour in that scenario.

lizconlan:
👍

Issue #2061, opened 2015-01-05
document dealing with inbound spam


davewhiteland:
[from @henare]

Broken out from #1075

crowbot:
Done in ce59cb124ccd9cf63fbb72961f9d34b417506487

PR #3210, opened 2016-04-21
[#3146] Allow general use of reCAPTCHA for signup

### LINKED ISSUE #3146: Allow reCAPTCHA to be used for all new signups
Currently reCAPTCHA used for signups from another country. Allow an install of Alaveteli to use reCAPTCHA for all signups.

lizconlan:
If `USE_RECAPTCHA_FOR_REGISTRATION` is set to true (defaults to false) then reCAPTCHA will be used on the signup form regardless of whether the user is visiting the site from the same country or not.

Also moves the country check to a controller filter so that that signin and signup can both just check `@use_recaptcha`

Closes #3146

### DIFF
--- a/config/general.yml-example
+++ b/config/general.yml-example
@@ -273,7 +273,7 @@ THEME_URLS:
 # ---
 THEME_BRANCH: false
 
-# Does a user needs to sign in to start the New Request process?
+# Does a user need to sign in to start the New Request process?
 #
 # FORCE_REGISTRATION_ON_NEW_REQUEST - Boolean (default: false)
 #
--- a/config/general.yml-example
+++ b/config/general.yml-example
@@ -280,6 +280,17 @@ THEME_BRANCH: false
 # ---
 FORCE_REGISTRATION_ON_NEW_REQUEST: false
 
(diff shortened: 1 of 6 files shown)


lizconlan:
holding note to self that a new section for this to the new `docs/running/handling_spam` page in the `gh-pages` branch (see #2061)

config/general.yml-example
```diff
@@ -280,6 +280,17 @@ THEME_BRANCH: false
 # ---
 FORCE_REGISTRATION_ON_NEW_REQUEST: false
 
+# Does a user needs to pass a RECAPTCHA test before being allowed to
```
garethrees: s/needs/need/

garethrees:
👍

config/general.yml-example
lizconlan: d'oh! thanks!

lizconlan: Oh, hang on - I pinched that from the line above `FORCE_REGISTRATION_ON_NEW_REQUEST` fixing both while I'm here

PR #3214, opened 2016-04-22
[#624] delivery status parser

### LINKED ISSUE #624: automated checking of mail delivery logs

garethrees:
Fixes #624 

- [x] Add different "Sent" icon (blue tick)
- [ ] Improve wording of delivery status view partials?
- [ ] Check performance of request pages before/after deploy


garethrees:
Might have a performance hit on the request pages – remember to compare before/after deploy.

zarino:
Here's a PR based on this work, but with the additional styling: https://github.com/mysociety/alaveteli/pull/3313

app/models/outgoing_message.rb
```diff
@@ -261,6 +261,13 @@ def mail_server_logs
     end
   end
 
+  def delivery_status
+    mail_server_logs.
+      map { |log| log.line(:decorate => true).delivery_status }.
+        compact.
+          last
```
crowbot: I wonder how reliable line order is going to be - I guess we'll find out when we try this live.

doc/CHANGES.md
```diff
@@ -2,6 +2,7 @@
 
 ## Highlighted Features
 
+* Show the delivery status of outgoing messages (Gareth Rees, Zarino Zappia)
```
crowbot: If we did a release now, we'd need to note that this doesn't work for Postfix, but actually I think we should delay the release until Postfix works as well.

app/controllers/outgoing_messages/delivery_statuses_controller.rb
```diff
@@ -22,7 +26,7 @@ def set_outgoing_message
   end
 
   def check_prominence
-    unless @outgoing_message.is_owning_user?(@user)
+    unless @outgoing_message.user_can_view?(@user)
```
crowbot: This is actually just going to check the visibility of the message itself. I think we should also check the visibility to the user of the info_request itself:

```
@outgoing_message.user_can_view?(@user) && @outgoing_message.info_request.user_can_view(@user) 
```

spec/models/mail_server_log/exim_line_spec.rb
```diff
@@ -0,0 +1,118 @@
...
+    it 'parses an :additional_address_in_same_delivery line' do
+      line = '2016-06-03 14:22:11 [31522] 1b8p3C-0008CN-Vl -> authority@example.com <Authority@example.com> F=<request-123-abc987@example.net> P=<request-123-abc987@example.net> R=dnslookup T=remote_smtp S=5745 H=cluster5.eu.messagelabs.com [85.158.138.179]:25 X=TLS1.2:DHE_RSA_AES_128_CBC_SHA1:128 CV=no DN="C=US,ST=California,L=Mountain View,O=Symantec Corporation,OU=Symantec.cloud,CN=mail169.messagelabs.com" C="250 ok 1464960131 qp 11694 server-9.tower-169.messagelabs.com!1464960131!41697448!1" QT=1s DT=0s'
+      expected =
+        MailServerLog::EximDeliveryStatus.
+          new(:additional_address_in_same_delivery)
+      expect(described_class.new(line).delivery_status).to eq(expected)
+    end
+
+    it 'parses a :cutthrough_message_delivery line' do
+      pending("Can't find an example for this")
```
crowbot: Would be good to have a live spec here to demonstrate what the code is doing, even if it's reverse engineered from the code.

app/controllers/outgoing_messages/delivery_statuses_controller.rb
```diff
@@ -0,0 +1,30 @@
+# -*- encoding : utf-8 -*-
+class OutgoingMessages::DeliveryStatusesController < ApplicationController
+  before_filter :set_outgoing_message, :check_prominence
+
+  def show
+    @title = _('Delivery Status for Outgoing Message #{{id}}',
```
crowbot: I'm not sure that we want to expose the `id` here - could distract from the main point of the page. Don't feel super-strongly about it though.

app/models/outgoing_message.rb
garethrees: Seems to have been fine when looking through some logs for https://github.com/mysociety/alaveteli/pull/3039

app/controllers/outgoing_messages/delivery_statuses_controller.rb
garethrees: Fixed in f3a58cd

app/controllers/outgoing_messages/delivery_statuses_controller.rb
garethrees: Yeah it was more that I couldn't think of anything better to get the message across that the status is for the individual message rather than the requests itself. The ID is exposed in the URL so I don't think there's a problem in that respect. Going to just go with this for now.

doc/CHANGES.md
garethrees: Removed in 1e72a4d

spec/models/mail_server_log/exim_line_spec.rb
garethrees: Added in ecb5f7d

garethrees:
@zarino if you get chance could you commit a blue tick for "sent" as a fixup to c6ab8e61000ac51342f8fa23cc9e00a8652e4403?

zarino:
> @zarino if you get chance could you commit a blue tick for "sent" as a fixup to c6ab8e6?

@garethrees – done!

PR #3222, opened 2016-04-26
Load delivery log into request page with ajax

zarino:
The delivery log is pulled in via ajax. If javascript isn't available, the direct link to the log still works. And if the ajax request fails for some reason, a direct link is shown as a fallback.

Includes some basic styling for the delivery log, and some very rough error display messages.

delivery

**What this doesn’t include:**

* A nice clear heading like "This message has been delivered to the authority’s mailservers" inside the fold-out section.
* Pale background colour should reinforce the delivery state – eg: pale green for successful delivery, pale red for unknown status.
* Trigger should read "Delivered" or "Sent", depending on whether we know the message was delivered or not.
* Better wording when no logs are found. (Eg: we could fairly intelligently show "Come back later" if we know the message was sent too recently to be caught by the log scraper).


zarino:
See also https://github.com/mysociety/alavetelitheme/pull/79 which is the corresponding styling for alavetelitheme.

app/assets/javascripts/general.js
```diff
@@ -65,4 +65,54 @@ $(document).ready(function() {
...
+      dataType: "html"
+    }).done(function(html){
+      var $deliveryDiv = $(html).find('.controller_mail_server_logs');
+      $correspondence_delivery.html( $deliveryDiv.html() );
+      $correspondence_delivery.slideDown(200);
+
+    }).fail(function(){
+      // TODO: This string needs to be translated!!
+      var msgHtml = '<p>We couldn&rsquo;t load the logs for this message.</p>';
+      msgHtml += '<p>Try <a href="' + url + '" target="_blank">opening the logs in a new window</a>.</p>';
```
garethrees: We could put them in in the markup (hidden by default) and then unhide based on what happens here?

app/assets/javascripts/general.js
```diff
@@ -65,4 +65,54 @@ $(document).ready(function() {
     $this.unbind("mouseup");
     return false;
   });
+
+  $('.js-toggle-delivery-log').on('click', function(e){
```
garethrees: Probably worth changing "Show delivery log" to "Hide delivery log" when its open?

garethrees:
Looks good. Two minor tweaks to resolve but this is a good step towards showing Delivered vs Failed.

app/assets/javascripts/general.js
zarino: I resisted adding logic for this, partly because it's get another string that needs to be translated, but mostly because this is just placeholder wording, and the end goal is that the link text should be the message status like "delivered", rather than "show/hide".

app/assets/javascripts/general.js
zarino: I resisted adding logic for this, partly because it's get another string that needs to be translated, but mostly because this is just placeholder wording, and the end goal is that the link text should be the message status like "delivered", rather than "show/hide".

app/assets/javascripts/general.js
zarino: What do we do for javascript translations elsewhere? Or is the first time?

app/assets/javascripts/general.js
garethrees: Fair comment – not the end of the world 👍

app/assets/javascripts/general.js
garethrees: https://www.whatdotheyknow.com/select_authority is probably the only similar example – we just [render the view](https://github.com/mysociety/alaveteli/blob/develop/app/views/public_body/_search_ahead.html.erb) which contains the translation strings.

garethrees: Or, yuck, embed the script in the erb https://github.com/mysociety/alaveteli/blob/develop/app/views/user/set_draft_profile_photo.html.erb#L34-L36

zarino: I guess just including the entire blob of HTML markup in the document, and making sure it's completely ignored by screenreaders etc until we actually want it displayed is as good a solution as any.

zarino: Done in 6995541b68e559f486a35a8e200c96f02df7975b.

garethrees:
Cheers! Will merge in to `624-logs-controller` shortly.

garethrees:
Cheers, merged in cf8a368

schlos:
+1
cant wait to see it in action

PR #3224, opened 2016-04-26
Use reply-to for user email addresses to better handle DMARC.

### LINKED ISSUE #3205: In mails sent by alaveteli 'from' users e.g. the contact from, use 'reply-to' rather than 'from' for the user's email address
This would work around the issue described in https://groups.google.com/forum/#!topic/alaveteli-dev/Poo5drN7cso and https://github.com/openaustralia/righttoknow/issues/551.

crowbot:
If the from address is used, mails from domains with strict DMARC
policies may be refused. Fixes #3205


garethrees:
👍

PR #3234, opened 2016-04-29
[#3163] cleanup holding pen bugfix

### LINKED ISSUE #3128: script/cleanup-holding-pen has output, which gets printed when run via cron
Looks like the [`--silent`](https://github.com/mysociety/alaveteli/blob/develop/script/cleanup-holding-pen#L4) option is not surprising the rake task's [stdout](https://github.com/mysociety/alaveteli/blob/develop/lib/tasks/cleanup.rake#L13).

### LINKED ISSUE #3163: cleanup:holding_pen NoMethodError: undefined method `info_request_events' for nil:NilClass
If the holding pen doesn't exist yet:

```
Cron <FILTERED@FILTERED> /FILTERED/colombia.alaveteli.org/alaveteli/script/cleanup-holding-pen

This is a dryrun - nothing will be deleted
rake aborted!
NoMethodError: undefined method `info_request_events' for nil:NilClass
/data/vhost/colombia.alaveteli.org/alaveteli-2016-03-17T14-17-42/lib/tasks/cleanup.rake:10:in `block (2 levels) in <top (required)>'
Tasks: TOP => cleanup:holding_pen
(See full trace by running task with --trace)
```

lizconlan:
Avoids the possibility of getting a Rails error if the holding_pen request does not exist by fetching it using the `InfoRequest.holding_pen_request` method so that it will be created if it's missing.

Tidies up the rake task a bit and suppresses output - in theory using `--silent` should be enough as the generated output is being piped to `$stderr` but hiding it with checks on `verbose` and `dryrun` to be certain.

Closes #3163 
Closes #3128

### DIFF
--- a/lib/tasks/cleanup.rake
+++ b/lib/tasks/cleanup.rake
@@ -7,9 +7,9 @@ namespace :cleanup do
       $stderr.puts "This is a dryrun - nothing will be deleted"
     end
     holding_pen = InfoRequest.holding_pen_request
-    old_events = holding_pen.info_request_events.find_each(:conditions => ['event_type in (?)',
-                                                      ['redeliver_incoming',
-                                                      'destroy_incoming']]) do |event|
+    holding_pen.info_request_events.find_each(:conditions => ['event_type in (?)',
+                                                ['redeliver_incoming',
+                                                 'destroy_incoming']]) do |event|
       puts event.inspect
       if not dryrun
         event.destroy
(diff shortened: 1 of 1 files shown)


lizconlan:
The block

```ruby
else
    $stderr.puts "Holding pen not found - no action taken" if verbose
```
 
could also be used to create the holding pen - I can't remember what we decided it should do 😕

PR #3235, opened 2016-04-30
Discourage crawlers on change_request

schlos:
Robots file should discourage crawlers from storing and accessing change request pages, i.e. */change_request/new?body=turisticka_zajednica_knin


garethrees:
Looks good to me, unless @crowbot can think of a reason why we wouldn't merge this?

PR #3236, opened 2016-05-03
Redirect to frontpage when no 'comment' param supplied.

### LINKED ISSUE #2951: help#contact (NoMethodError) "undefined method `[]' for nil:NilClass"
```
undefined method `[]' for nil:NilClass
  app/controllers/help_controller.rb:81:in `catch_spam'
```

URL is /help/contact, looks like it's triggered by a `POST` action without the form params.

crowbot:
Given that the comment field is hidden visually, I think it's OK
to expect it to be supplied as blank - the real life cases where it
isn't supplied at all are spam. Also now handles the case where no
params are supplied with a POST request by rendering the form.

Closes #2951

### DIFF
--- a/app/controllers/help_controller.rb
+++ b/app/controllers/help_controller.rb
@@ -79,8 +79,10 @@ class HelpController < ApplicationController
   private
 
   def catch_spam
-    if request.post? && !params[:contact][:comment].empty?
-      redirect_to frontpage_url
+    if request.post? && params[:contact]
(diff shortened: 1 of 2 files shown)


garethrees:
👍

garethrees:
Not showing as merged for some reason. It is in https://github.com/mysociety/alaveteli/commit/d62498d26fe3250665ba0751e680b0c843adf65f.

Issue #3180, opened 2016-04-13
Clean out spam user profiles


crowbot:
Depends on using #3146 or #3147 or some other way to stop the inflow of new spam in order for this not to become a perpetual maintenance task.

lizconlan:
Hmm, not simple to programmatically sift spam from not with certainty. Would it be worth trying to assign some kind of score? I've found a few things which seem to pretty much guarantee a spammer at the moment but this covers quite a small percentage.

I would also suggest some obfuscation of our exact rules and methods to avoid accidentally making a list of quirks for future spammers to avoid.

garethrees:
> I've found a few things which seem to pretty much guarantee a spammer

Does it look like https://github.com/mysociety/alaveteli/issues/3146 slowed the inflow of these obvious spammers?

> I would also suggest some obfuscation of our exact rules and methods

Send an email to [the group](https://groups.google.com/a/mysociety.org/forum/#!forum/alaveteli) to start the discussion behind a log in.

> Would it be worth trying to assign some kind of score?

I'd be inclined to use something like [rakismet](https://github.com/joshfrench/rakismet). It looks easy enough to hack it to use independently (rather than via an `include`). In any case, we'd want some interactivity in the rake task before we actually ban/delete a user, so for now I think using your initial heuristics and then a confirmation step would be a good start.

garethrees:
So here's what I was going to do, but Akismet requires an IP address for the user. Sure, we could fake it, but I'd imagine that would skew the results.

```ruby
Rakismet.key = ENV['AKISMET_API_KEY']
Rakismet.url = 'http://10.10.10.30:3000/'
Rakismet.host = 'rest.akismet.com'
Rakismet.validate_key

data = {
  :comment_author => 'MelindaGates11',
  :comment_author_email => 'MelindaGates11@mail.ru',
  :comment_content => "http://XN--12C1CC9AL4E0G7A.COM\r\n\r\nHow to dislodge the lash The birds live and then need the eyelid area with a minimum thickness of 3 mm, when the birds are the gel fibre Lotus girls legs similar to a long cable to balance it on the fly. This is where the innovation is to import another type of characteristics of the domain renfin is made from stainless steel",
  :comment_date_gmt => '2016-05-18T13:10:28+01:00',
  :blog_lang => 'en,cy',
  :blog_charset => 'UTF-8'
}

Rakismet.akismet_call('comment-check', data)
# => "Missing required field: user_ip."

[...]

PR #3244, opened 2016-05-09
WIP - identify spam accounts

lizconlan:
Not for review!

Some example methods for fetching potential spam accounts to be run through an as-yet undefined process which will hopefully cut down the list to those most likely to be spammy.

`display_detail` was just for testing to make sure that random samples from the list look plausible but could be adapted for use by the eventual rake task if required.

Connects to #3180


garethrees:
Superseded by #3263. Thanks!

PR #3248, opened 2016-05-10
[#624] requester only mail logs

garethrees:
Part of #624 

Makes the mail server logs "public" to admins and request owners. Hidden from anyone else.


lizconlan:
👍

PR #3257, opened 2016-05-18
[#624] Redact idhash from mail server logs

garethrees:
https://github.com/mysociety/alaveteli/issues/624#issuecomment-219257191

Redacts the `idhash` for anyone other than admins


lizconlan:
👍 the `MailServerLog#line` override threw me for a bit in isolation but makes perfect sense in use - vastly lowers the risk of getting unredacted text by mistake!

PR #3264, opened 2016-05-23
(Re)upgrade compass, sass, and foundation-rails

lizconlan:
Found the bug - [this commit in foundation-rails 5.5.3.1](https://github.com/zurb/foundation-rails/commit/9d41db670df159a689ad0e5597151476503c079c#diff-d2e9f795423b3ea3f470109683fb1972R131) changes what `$last-column:true` does and breaks the new request layout with an unhelpful float spec.

Connects to #3161


garethrees:
👍

PR #3265, opened 2016-05-24
Reinstate the --silent flag for the graph generation scripts

### LINKED ISSUE #3250: Stats graphs have output
We moved the graph generation to ruby in https://github.com/mysociety/alaveteli/pull/3000 (fixing https://github.com/mysociety/alaveteli/issues/675).

I think that then caused https://github.com/mysociety/alaveteli/issues/3126 – or at least didn't fix it – which got addressed by https://github.com/mysociety/alaveteli/issues/3138.

Somewhere in this is https://github.com/mysociety/alaveteli/issues/619, which I _think_ should be fixed by #3138, but the scripts now seem to output even when they have enough data:

```
$ script/user-use-graph
(in /home/vagrant/alaveteli)
```

We're getting this from WDTK:

```
[cron-alaveteli] Cron <USER@SERVER> /APP_ROOT/www.whatdotheyknow.com/alaveteli/script/user-use-graph
(in /APP_ROOT/www.whatdotheyknow.com/alaveteli-2016-05-12T14-11-51)
```

As I understand it these scripts should be silent by default, and can be passed a `--verbose` option in some way?

lizconlan:
Fixes #3250

### DIFF
--- a/script/request-creation-graph
+++ b/script/request-creation-graph
@@ -6,5 +6,5 @@
 # Email: hello@mysociety.org. WWW: http://www.mysociety.org/
 
 cd `dirname $0`
-bundle exec rake "$@" graphs:generate_request_creation_graph
+bundle exec rake --silent "$@" graphs:generate_request_creation_graph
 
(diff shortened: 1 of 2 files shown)


garethrees:
👍

PR #3270, opened 2016-05-24
Check `default_letter` when checking if OutgoingMessage body has changed

### LINKED ISSUE #3258: Require users to add content to default request text
Currently users can make a request without adding any content to the default request text in the message body. When a user tries preview such a request they should be prompted to add some content. 

eg. AskTheyEu allows a request such as the following to be made.
>Dear Council of the EU,

>Under the right of access to documents in the EU treaties, as developed in Regulation 1049/2001, I am requesting documents which contain the following information:

>Yours faithfully,

>Richard Taylor

or on WhatDoTheyKnow

>Dear Doug's Test Authority,

>Yours faithfully,

>Richard Taylor

screen shot 2016-05-18 at 19 13 37

This issue has been raised following cases of a user of AskTheEU describing a request in the subject/title line and sending the unedited body template. 

When users make an internal review request they are forced by the system to replace place-holder text which prompts them to explain their concerns.  This occasionally confuses some users though so I wouldn't suggest including a [write your request here] place-holder and detecting that as the way to address this ticket.

lizconlan:
Fixes #3258

May need to be reworked if a hotfix is required.

### DIFF
--- a/app/models/outgoing_message.rb
+++ b/app/models/outgoing_message.rb
@@ -462,7 +462,7 @@ class OutgoingMessage < ActiveRecord::Base
   end
 
   def template_changed
-    if body.empty? || body =~ /\A#{Regexp.escape(letter_template.salutation(default_message_replacements))}\s+#{Regexp.escape(letter_template.signoff(default_message_replacements))}/ || body =~ /#{Regexp.escape(Template::InternalReview.details_placeholder)}/
+    if body.empty? || body =~ /\A#{regex_escape(letter_template.body(default_message_replacements))}/
       if message_type == 'followup'
         if what_doing == 'internal_review'
           errors.add(:body, _("Please give details explaining why you want a review"))
@@ -475,6 +475,15 @@ class OutgoingMessage < ActiveRecord::Base
         raise "Message id #{id} has type '#{message_type}' which validate can't handle"
       end
     end
+  end
+
+  def regex_escape(content)
+    text = content.gsub("\r", "\n") #shouldn't happen, but just in case
+    # feels like this should need a gsub(/\//, '\/') but doesn't seem to
+    Regexp.escape(text.squeeze("\n")).
+      gsub("\\n", '\s*').
+      gsub('\ \s*', '\s*').
+      gsub('\s*\ ', '\s*')
   end
 
   def body_has_signature
--- a/app/models/outgoing_message.rb
+++ b/app/models/outgoing_message.rb
@@ -462,7 +462,7 @@ class OutgoingMessage < ActiveRecord::Base
   end
 
   def template_changed
-    if body.empty? || body =~ /\A#{regex_escape(letter_template.body(default_message_replacements))}/
+    if body.empty? || HTMLEntities.new.decode(body) =~ /\A#{regex_escape(letter_template.body(default_message_replacements))}/
       if message_type == 'followup'
         if what_doing == 'internal_review'
           errors.add(:body, _("Please give details explaining why you want a review"))
--- a/app/models/outgoing_message.rb
+++ b/app/models/outgoing_message.rb
@@ -462,7 +462,7 @@ class OutgoingMessage < ActiveRecord::Base
   end
 
   def template_changed
-    if body.empty? || HTMLEntities.new.decode(body) =~ /\A#{regex_escape(letter_template.body(default_message_replacements))}/
+    if body.empty? || HTMLEntities.new.decode(raw_body) =~ /\A#{template_regex(letter_template.body(default_message_replacements))}/
       if message_type == 'followup'
         if what_doing == 'internal_review'
           errors.add(:body, _("Please give details explaining why you want a review"))
@@ -477,8 +477,8 @@ class OutgoingMessage < ActiveRecord::Base
     end
   end
 
-  def regex_escape(content)
-    text = content.gsub("\r", "\n") #shouldn't happen, but just in case
+  def template_regex(template_text)
+    text = template_text.gsub("\r", "\n") # in case we have '\r\n' or even '\r's all the way down
     # feels like this should need a gsub(/\//, '\/') but doesn't seem to
     Regexp.escape(text.squeeze("\n")).
       gsub("\\n", '\s*').
@@ -487,7 +487,7 @@ class OutgoingMessage < ActiveRecord::Base
   end
 
   def body_has_signature
-    if body =~ /#{letter_template.signoff(default_message_replacements)}\s*\Z/m
+    if raw_body =~ /#{template_regex(letter_template.signoff(default_message_replacements))}\s*\Z/m
       errors.add(:body, _("Please sign at the bottom with your name, or alter the \"{{signoff}}\" signature", :signoff => letter_template.signoff(default_message_replacements)))
     end
   end
--- a/doc/CHANGES.md
+++ b/doc/CHANGES.md
@@ -2,6 +2,13 @@
 
 ## Highlighted Features
 
+* Fixed bug in `OutgoingMessage.template_changed` which allowed a new request to
+  be submitted without changes to the default text if:
+   - the site (theme) overrode the core default text via `default_letter`
+   - the authority name contained any characters which were encoded as
+     HTMLEntities
+   - a global censor rule changed the template text
+  Only the first case is known to affect a live site (Liz Conlan)
 * Add an interface to calculate transaction stats per user (Gareth Rees)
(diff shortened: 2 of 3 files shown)


garethrees:
I think we have two issues to address here:

1. The validation doesn't work with specific authorities on WDTK, i.e. using the default code
1. The validation doesn't work on AskTheEU which uses the `default_letter` attribute

## 1. WDTK

Looking at the before and after (https://github.com/mysociety/alaveteli/commit/fa9895eb5ab8ec888e1fafda3dfddb21fcd779aa) the actual `Regexp` has not changed.

I've effectively done this:

```ruby
# Before OutgoingMessage::Template introduction
salutation      = Regexp.escape(get_salutation)
signoff         = Regexp.escape(get_signoff)
internal_review = Regexp.escape(get_internal_review_insert_here_note)

body =~ /\A#{salutation}\s+#{signoff}/ || body =~ /#{internal_review}/


# After OutgoingMessage::Template introduction
salutation      = Regexp.escape(letter_template.salutation(default_message_replacements))
signoff         = Regexp.escape(letter_template.signoff(default_message_replacements))
internal_review = Regexp.escape(Template::InternalReview.details_placeholder)

body =~ /\A#{salutation}\s+#{signoff}/ || body =~ /#{internal_review}/
```

Now, what the new template-based interpolations _return_ might be different, so that's definitely a point to check.

I think you might be along the right lines when you [suggest](https://github.com/mysociety/alaveteli/issues/3258#issuecomment-222673950) that this might be caused by something like the apostrophe in the authority name.

- [ ] Verify whether the validation fails given an authority name that includes an apostrophe on a version of Alaveteli prior to fa9895eb5ab8ec888e1fafda3dfddb21fcd779aa
- [ ] Verify that the new template methods return the same values as the old methods. I'd have thought I'd have tested these, but I might have missed something like an extra `\n` or something?

## 2. AskTheEU

Given that the original `Regexp` didn't include the "body" I'm not _sure_ we want to do that here. Again, stress on _not sure_.

It looks like `format_of_body` should have also included `OutgoingMessage#get_default_letter` as part of the check.

- [ ] Verify whether the validation fails when a `default_letter` is set on a version prior to fa9895eb5ab8ec888e1fafda3dfddb21fcd779aa

---

I think we need to complete the todos before I can think of where to go next with this.

Now we have templates with a `#body` method, I think we could simplify this a whole lot with something like:

```ruby
def template_changed
  if body =~ /\A#{Regexp.escape(letter_template.body(default_message_replacements))}/
    # …
end
```

lizconlan:
## WDTK

### Verify whether the validation fails given an authority name that includes an apostrophe on a version of Alaveteli prior to fa9895e

It appears that `OutgoingMessage.body` gets encoded as e.g. "Doug&amp;#x27;s Test Authority [...]" so it will not match the apostrophe in the RegExp. I was _half_ right - it's not matching the apostrophe but not for the reason I expected. This does not appear to be a consequence of fa9895e, just the first time we've seen it happen. (Tested under an older code version just in case, same result.)

### Verify that the new template methods return the same values as the old methods
| | |
| ----- | --- |
| before | `/\ADear\ Department\ for\ Humpadinking,\s+Yours\ faithfully,/`
| after | `/\ADear\ Department\ for\ Humpadinking,\s+Yours\ faithfully,/`

looks legit!

## AskTheEU

### Verify whether the validation fails when a `default_letter` is set on a version prior to fa9895e

Yep, running this code in the console locally from 0cf49e7:

```rb
public_body = PublicBody.first #"Department for Humpadinking"
info_request = InfoRequest.new(:public_body => public_body,
                               :url_title => 'a_test_title',
                               :title => 'a test title')
outgoing_message = OutgoingMessage.new(:status => 'ready',
                             :message_type => 'initial_request',
                             :what_doing => 'normal_sort',
                             :info_request => info_request)

outgoing_message.valid?
outgoing_message.errors.messages
```

...the only error message returned is that I didn't sign it.

Message body => `"Dear Department for Humpadinking,\n\nUnder the right of access to documents in the EU treaties, as developed in Regulation 1049/2001, I am requesting documents which contain the following information:\n\nYours faithfully,"`

lizconlan:
> now we have templates with a #body method, I think we could simplify this a whole lot with something like:
>
> ```rb
def template_changed
  if body =~ /\A#{Regexp.escape(letter_template.body(default_message_replacements))}/
    # …
end

Yes! I think that gets me the whole thing. So something like...

```rb
/\A#{Regexp.escape(letter_template.body(default_message_replacements)).
    squeeze("\n")).gsub("\\n", '\s*')}/`
```

...to get `"Dear\ Department\ for\ Humpadinking,\s*Under\ the\ right\ of\ access\ to\ documents\ in\ the\ EU\ treaties,\ as\ developed\ in\ Regulation\ 1049/2001,\ I\ am\ requesting\ documents\ which\ contain\ the\ following\ information:\s*Yours\ faithfully,\s*`

(so that just fiddling with the spacing won't count as an edit)

lizconlan:
note to self: might be best to the Regexp-ifying in a method call rather than having an unreasonably long `.gsub`/`squeeze` chain making a mess of the if statement

garethrees:
> (so that just fiddling with the spacing won't count as an edit)

Yeah, that's a pretty good idea. I think what I originally meant to say (and then promptly forgot!) was that rather than the regexp, we could just check equality with the template:

```ruby
if body == letter_template.body(default_message_replacements)
  # …
```

But I think your regexp approach seems a bit better.

You might want to check whether its possible to end up with other types of line ending, so might want to convert it all to `\n` before comparison, or indeed earlier in the process if it makes sense to do so.

lizconlan:
I should rename the first commit as part of the rebase, that's no longer a good description

app/models/outgoing_message.rb
```diff
@@ -455,7 +455,7 @@ def format_of_body
   end
 
   def template_changed
-    if body.empty? || body =~ /\A#{Regexp.escape(letter_template.salutation(default_message_replacements))}\s*#{Regexp.escape(letter_template.letter(default_message_replacements)).gsub(/\//, "\\/").gsub('\n', '')}\s*#{Regexp.escape(letter_template.signoff(default_message_replacements))}/ || body =~ /#{Regexp.escape(Template::InternalReview.details_placeholder)}/
+    if body.empty? || body =~ /\A#{regex_escape(letter_template.body(default_message_replacements))}/ || body =~ /#{Regexp.escape(Template::InternalReview.details_placeholder)}/
```
garethrees: I _think_ this should work without the last check (`|| body =~ /#{Regexp.escape(Template::InternalReview.details_placeholder)}/`) since that would be included in the `letter_template.body` of an internal review.

app/models/outgoing_message.rb
```diff
@@ -470,6 +470,11 @@ def template_changed
     end
   end
 
+  def regex_escape(content)
+    text = content.gsub("\r", "\n") #shouldn't happen, but just in case
```
garethrees: I _think_ you'd want `"\r\n"` as the first argument here. It doesn't technically matter as you're squashing them down later anyway, but it would make the intent clearer.

app/models/outgoing_message.rb
lizconlan: it looks as though it should work but it's failing my regex for reasons I don't entirely understand - working on a fix

lizconlan: ah spaces vs greedy regex matching

app/models/outgoing_message.rb
lizconlan: as it's technically possible that it's carriage returns all the way down, I thought this was safer

garethrees: 👍

lizconlan: http://stackoverflow.com/a/1761064

lizconlan: May add a line comment to clarify what's going on there

garethrees: Note in the commit message would be fine I think :)

garethrees:
Figured it out. I had a global censor rule in place.

screen shot 2016-06-10 at 09 47 02

Will need to take that in to account in the validation

lizconlan:
rebase for clarity/sanity

lizconlan:
actually considering squashing all of these commits into one at the end of this process

garethrees:
When signed out its possible to submit a request with no real body by adding your name under the signoff.

screen shot 2016-06-21 at 15 51 03

garethrees:
👍

PR #3271, opened 2016-05-24
Issues/3266 add aria landmark roles

wrightmartin:
This adds various [ARIA landmark roles](https://www.w3.org/WAI/GL/wiki/Using_ARIA_landmarks_to_identify_regions_of_a_page) to appropriate places in alaveteli core. 

These commits can probably be squashed into a single one, but I thought it might be clearer for reviewing purposes to keep them separate.


garethrees:
Dropped 78900aa57889d896f1f174c1d3a0bc475fb7fe8a because its now just a placeholder.

garethrees:
Thanks!

PR #3273, opened 2016-05-25
Add advice panel line asking users to check info not already published

### LINKED ISSUE #3259: Urge users to check informatoin isn't published before making a request
On the make a request page we should explicitly urge users to check the information they're seeking isn't already published. 

We already point to related requests; but could include links to the authority's home page and the relevant part of the help text: https://www.whatdotheyknow.com/help/requesting#responsible

I suspect this might a line we used to have but lost in a redesign?

lizconlan:
Closes #3259

Introduces a new translation - should I do anything extra in this PR for that?


garethrees:
> Introduces a new translation - should I do anything extra in this PR for that?

No. We currently build new translations on release.

Issue #1276, opened 2013-12-06
:ruby_20 platform requires bundler 1.3


mhl:
In 3b86cb6129140fc123 I added `byebug` (a debugger suitable for Ruby 2.0) to the `Gemfile` with `:platforms: -> :ruby_20`. Unfortunately:
* bundler's behaviour on seeing a platform that it doesn't understand [is to error, rather than to ignore it](https://github.com/bundler/bundler/issues/2428)
* mySociety's servers currently only have version 1.1.4 of bundler and
* `:ruby_20` platform was only [introduced in version 1.3.0 of bundler](https://github.com/bundler/bundler/blob/master/CHANGELOG.md#130pre3-dec-21-2012).

Possible solutions include:

* Just reverting the addition of byebug for the moment (e.g. by cherry-picking 93646e1eda5cbe8f35504ffd07) - note that [even wheezy doesn't have 1.3 though](http://packages.debian.org/search?keywords=bundler&searchon=names&suite=all&section=all), though.
* Installing bundler on our servers from [jessie](http://packages.debian.org/jessie/bundler).
* Installing bundler on our servers for the deploying user as a gem, by setting `GEM_HOME` and adding the gem's bin directory to `PATH`.

garethrees:
Reopening as we now officially support Ruby 2.0 (https://github.com/mysociety/alaveteli/commit/c65af092e7fc96225cc932e2c3503bc3b390d5c3) and recommend it as the default Ruby version.

PR #3282, opened 2016-05-31
Work around Bundler 1.1.4 not recognizing ruby_20 as a platform.

crowbot:


garethrees:
+1 assuming a rebuild on 1.9.3 is successful – looks transient. Related to https://github.com/mysociety/alaveteli/issues/1276.

PR #3285, opened 2016-05-31
Don't require a specific order on array match.

crowbot:
Fixes errors like https://travis-ci.org/mysociety/alaveteli/jobs/134157087


lizconlan:
👍 🎉

PR #3289, opened 2016-06-01
Revert #3282

### LINKED ISSUE #3288: Conditional in Gemfile causes git changes
Since #3282 I'm getting a modified `Gemfile.lock` after running `bundle`:

```diff
diff --git a/Gemfile.lock b/Gemfile.lock
index 85dee5b..1362131 100644
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -83,7 +83,6 @@ GEM
     bullet (5.0.0)
       activesupport (>= 3.0.0)
       uniform_notifier (~> 1.9.0)
-    byebug (9.0.3)
     capistrano (2.15.4)
       highline
       net-scp (>= 1.0.0)
@@ -225,9 +224,6 @@ GEM
       coderay (~> 1.1.0)
       method_source (~> 0.8.1)
       slop (~> 3.4)
-    pry-byebug (3.4.0)
-      byebug (~> 9.0)
-      pry (~> 0.10)
     pry-debugger (0.2.3)
       debugger (~> 1.3)
       pry (>= 0.9.10, < 0.11.0)
@@ -403,7 +399,6 @@ DEPENDENCIES
   nokogiri (~> 1.6)
   open4 (~> 1.3.4)
   pg (~> 0.18.4)
-  pry-byebug (~> 3.4.0)
   pry-debugger (~> 0.2.3)
   quiet_assets (~> 1.1.0)
   rack (~> 1.4.6)
```

garethrees:
Fixes #3288 

I _think_ if we deploy to production with the "Ruby 2.0" version of `Gemfile.lock` checked in (which it is as it includes pry-byebug) we'll get a deployment error in production, since that's running 1.9 and will be running bundle with the `--deployment` flag.


lizconlan:
Hmm, the only other way around this seems to be a multiple Gemfiles approach, which can either be environment/platform targeted Gemfiles (can't find a good example at the moment) or having [supplementary Gemfiles](http://abstractcoder.com/2013/08/23/using-multiple-gemfiles.html) so that devs can keep e.g. debug tools out of the main Gemfile.lock.

garethrees:
> multiple Gemfiles approach

Thoughtbot have a tool for this https://github.com/thoughtbot/appraisal – not sure if it fits our exact criteria.

garethrees:
> supplementary Gemfiles so that devs can keep e.g. debug tools out of the main Gemfile.lock.

Hmm, this sounds like it would get confusing if we needed to do this for an application gem.

lizconlan:
> I _think_ if we deploy to production with the "Ruby 2.0" version of Gemfile.lock checked in (which it is as it includes pry-byebug) we'll get a deployment error in production, since that's running 1.9

We _might_ be able get around it by using the `:require => false` flag, the Internets are split on this. Safest thing is to sigh and take it out.

lizconlan:
👍 let's reset this back to something that works for now and open a ticket for investigating whether there's an alternative approach that doesn't cause too much overhead (or just wait until we get to use the newer version of Bundler)

PR #3294, opened 2016-06-02
3261 handle rejection at smtp time

### LINKED ISSUE #3261: Allow rejection at mail server level to be shown on request in admin interface

crowbot:
Fixes #3261. 
To do:
- [x] Documentation for what to do with the reject list in Exim and Postfix #3314

### DIFF
--- a/app/models/info_request.rb
+++ b/app/models/info_request.rb
@@ -1231,7 +1231,7 @@ class InfoRequest < ActiveRecord::Base
     AND url_title <> 'holding_pen'
     EOF
 
-    # 'very_old' months since last change requests, don't allow any new
+    # 'very_old' months since last change to request, don't allow any new
     # incoming messages
     InfoRequest.update_all <<-EOF.strip_heredoc.delete("\n")
     allow_new_responses_from = 'nobody'
--- a/app/models/info_request.rb
+++ b/app/models/info_request.rb
@@ -1201,6 +1201,23 @@ class InfoRequest < ActiveRecord::Base
     true
   end
 
+  def self.reject_incoming_at_mta(options)
+    query = InfoRequest.where(["updated_at < (now() -
+                                interval '#{options[:age_in_months]} months')
+                                AND allow_new_responses_from = 'nobody'
+                                AND rejected_incoming_count >= ?
+                                AND reject_incoming_at_mta = ?
+                                AND url_title <> 'holding_pen'",
+                                options[:rejection_threshold], false])
+    yield query.pluck(:id) if block_given?
+
+    if options[:dryrun]
+      0
+    else
+      query.update_all(:reject_incoming_at_mta => true)
+    end
+  end
+
   # This is called from cron regularly.
   def self.stop_new_responses_on_old_requests
     old = AlaveteliConfiguration.restrict_new_responses_on_old_requests_after_months
--- a/app/models/info_request.rb
+++ b/app/models/info_request.rb
@@ -1447,6 +1447,7 @@ class InfoRequest < ActiveRecord::Base
     will_be_rejected = (response_rejector && response_rejection) ? true : false
 
     if will_be_rejected && response_rejection.reject(response_rejector.reason)
+      self.increment!(:rejected_incoming_count)
       logger.info "Rejected incoming mail: #{ response_rejector.reason } request: #{ id }"
       false
     else
--- a/app/models/info_request.rb
+++ b/app/models/info_request.rb
@@ -1,6 +1,5 @@
 # -*- encoding : utf-8 -*-
 # == Schema Information
-# Schema version: 20151104131702
 #
 # Table name: info_requests
 #
@@ -24,6 +23,8 @@
 #  comments_allowed          :boolean          default(TRUE), not null
 #  info_request_batch_id     :integer
 #  last_public_response_at   :datetime
+#  reject_incoming_at_mta    :boolean          default(FALSE), not null
+#  rejected_incoming_count   :integer          default(0)
 #
 
 require 'digest/sha1'
--- a/lib/tasks/config_files.rake
+++ b/lib/tasks/config_files.rake
@@ -126,9 +126,37 @@ namespace :config_files do
     end
   end
 
+  desc 'Set reject_incoming_at_mta on old requests that are rejecting incoming mail'
+  task :set_reject_incoming_at_mta => :environment do
+    example = 'rake config_files:set_reject_incoming_at_mta REJECTED_THRESHOLD=5 AGE_IN_MONTHS=12'
+    check_for_env_vars(['REJECTED_THRESHOLD', 'AGE_IN_MONTHS'], example)
+    dryrun = ENV['DRYRUN'] != '0'
+    if dryrun
+      STDERR.puts "Only a dry run; info_requests will not be updated"
+    end
+    options = {:rejection_threshold => ENV['REJECTED_THRESHOLD'],
+               :age_in_months => ENV['AGE_IN_MONTHS'],
+               :dryrun => dryrun}
+
+    updated_count = InfoRequest.reject_incoming_at_mta(options) do |ids|
+      puts "Info Request\tRejected incoming count\tLast updated"
+      ids.each do |id|
+        info_request = InfoRequest.find(id)
+        puts "#{info_request.id}\t#{info_request.rejected_incoming_count}\t#{info_request.updated_at}"
+      end
+    end
+    puts "Updated #{updated_count} info requests"
+  end
 
   desc 'Produce a list of email addresses for which the MTA should reject messages at RCPT time'
   task :generate_mta_rejection_list => :environment do
+    example = 'rake config_files:generate_mta_rejection_list MTA=(exim|postfix)'
+    check_for_env_vars(['MTA'], example)
+    mta = ENV['MTA'].downcase
+    unless ['postfix', 'exim'].include? mta
+      puts "Error: Unrecognised MTA"
+      exit 1
+    end
     InfoRequest.where(:reject_incoming_at_mta => true).each do |info_request|
       puts info_request.incoming_email
     end
--- a/lib/tasks/config_files.rake
+++ b/lib/tasks/config_files.rake
@@ -148,6 +148,22 @@ namespace :config_files do
     puts "Updated #{updated_count} info requests"
   end
 
+  desc 'Unset reject_incoming_at_mta on a request'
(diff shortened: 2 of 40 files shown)


db/migrate/20160602143125_add_reject_mail_at_mta_to_info_request.rb
```diff
@@ -0,0 +1,10 @@
+# -*- encoding : utf-8 -*-
+class AddRejectMailAtMtaToInfoRequest < ActiveRecord::Migration
+ def up
+    add_column :info_requests, :reject_mail_at_mta, :boolean, :default => false, :null => false
```
garethrees: Might be worth calling this `reject_incoming_at_mta` to match the language used elsewhere (e.g. `IncomingMessage`/`OutgoingMessage`, `:rejected_incoming_count`)

app/models/info_request.rb
```diff
@@ -1430,6 +1430,8 @@ def accept_incoming?(email, raw_email_data)
     will_be_rejected = (response_rejector && response_rejection) ? true : false
 
     if will_be_rejected && response_rejection.reject(response_rejector.reason)
+      self.rejected_incoming_count += 1
```
garethrees: Could use [`#increment!`](http://api.rubyonrails.org/v3.2.22/classes/ActiveRecord/Persistence.html#method-i-increment-21) here – I don't think running the validations is important after this?

lib/tasks/config_files.rake
```diff
@@ -127,4 +127,14 @@ namespace :config_files do
   end
 
 
+  desc 'Produce a list of email addresses for which the MTA should reject messages at RCPT time'
+  task :generate_mta_rejection_list => :environment do
+    example = 'rake config_files:generate_mta_rejection_list FILE=/tmp/rejection_list.txt'
+    check_for_env_vars(['FILE'], example)
+    f = File.open(ENV['FILE'], 'w')
```
garethrees: I don't think this needs to handle files – just print to stdout and then use shell redirection if you want it in a file?

```
bundle exec rake config_files:generate_mta_rejection_list > /tmp/rejection_list.txt
```

Makes it easier to preview your list.

app/models/info_request.rb
```diff
@@ -1201,6 +1201,20 @@ def indexed_by_search?
     true
   end
 
+  def self.set_reject_mail_at_mta(options)
```
garethrees: This doesn't return the array if `dryrun` is `false`. I'm not actually sure why, because `update_all` didn't seem to exhibit this behaviour while having a play around in the console.

In any case, I think yielding the IDs of records to be updated would be better:

```diff
diff --git a/app/models/info_request.rb b/app/models/info_request.rb
index a673496..2d1bc6e 100644
--- a/app/models/info_request.rb
+++ b/app/models/info_request.rb
@@ -1210,10 +1210,14 @@ class InfoRequest < ActiveRecord::Base
                                 AND reject_mail_at_mta = 'f'
                                 AND url_title <> 'holding_pen'",
                                 options[:rejection_threshold]])
+
+    yield query.pluck(:id) if block_given?
+
     if !options[:dryrun]
       query.update_all("reject_mail_at_mta = 't'")
+    else
+      0
     end
-    query.to_a
   end

   # This is called from cron regularly.
```

- The return value is always the number of records updated
- You avoid loading all the records in to memory

```ruby
updated_count = InfoRequest.set_reject_mail_at_mta(options) do |ids|
  puts "Info Request\tRejected incoming count\tLast updated"
  ids.each do |id|
    info_request = InfoRequest.find(id)
    puts "#{info_request.id}\t#{info_request.rejected_incoming_count}\t#{info_request.updated_at}"
  end
end
puts "Updated #{updated_count} info requests"
```

Some other miner comments:

We're a bit beyond supporting MySQL now, but a hash argument would read better here:

```ruby
query.update_all(:reject_mail_at_mta => true)
```

Could also use `unless options[:dryrun]` instead of `if !options[:dryrun]`.

Super minor, but I'd drop the `set_`.

spec/models/info_request_spec.rb
```diff
@@ -95,6 +95,42 @@
 
   end
 
+  describe '.set_reject_mail_at_mta', :focus => true do
```
garethrees: Unintentionally committed `:focus` tag.

lib/tasks/config_files.rake
```diff
@@ -144,6 +144,21 @@ namespace :config_files do
     end
   end
 
+  desc 'Unset reject_mail_at_mta on a request'
+  task :unset_reject_mail_at_mta => :environment do
```
garethrees: Lets say we use `:unset_reject_mail_at_mta` to set `:reject_mail_at_mta` to `false`on a request that's rejecting responses at MTA level. Running `config_files:set_reject_mail_at_mta` is going to reset this to `true` again the next time its run, ergo `config_files:unset_reject_mail_at_mta` will basically get ignored. 

I think `:unset_reject_mail_at_mta` also needs to change `allow_new_responses_from` so that it does not appear in the list for rejection.

garethrees:
> Should an error be thrown if a message is received by a request that is set to reject at the MTA level?

Only reason not to that I can think of is that an error would be raised in the case of an admin redelivering an incoming message to that request. Sounds pretty unlikely, but it wouldn't be obvious that doing so wouldn't work?

lib/tasks/config_files.rake
crowbot: Yeah, the idea was that changes in `allow_new_responses_from` and `handle_rejected_responses` should still be kept to the admin interface - to make it clearer in the case where there are different people who have access to the command line and the admin interface. But I think you're probably right - it's clearer just to set the request to `authority_only` here.

garethrees: Yeah, doing anything else would add a load of complexity.

I did start thinking about this. Here's a bit of a mind dump of my thought process when I looked in to it.

Feel free to completely ignore this, but I guess its worth posting if we're ever wondering why we did it this way.

---

Lets say we use `:unset_reject_mail_at_mta` to set `:reject_mail_at_mta` to `false`on a request that's rejecting responses at MTA level. Running `config_files:set_reject_mail_at_mta` is going to reset this to `true` again the next time its run, ergo `config_files:unset_reject_mail_at_mta` will basically get ignored. 

I don't think I have a _great_ idea of how to solve this, but here are some options.

I'll add specific comments that I noted when I was going through the commits, but they might be ignorable depending on how we deal with this situation.

I think we're trying to handle three situations:

- Not rejecting at MTA because its under the threshold
- Rejecting at MTA
- Not rejecting at MTA because we explicitly said not to

My initial thought was to invert the logic of `reject_mail_at_mta`; make everything `true` by default ("fair game for rejection"), generate the list based on requests that are `true` and over the threshold and exclude requests that are explicitly marked as `false`.

```ruby
# By default, anything is fair game to get rejected at MTA level
# Actually want this done in the migration, but YSWIM
InfoRequest.update_all(:reject_mail_at_mta => true)

# Have a query that finds all that are over thereshold and "fair game"
class InfoRequest
  def self.over_rejection_at_mta_threshold(options)
    where(["updated_at < (now() -
            interval '#{options[:age_in_months]} months')
            AND allow_new_responses_from = 'nobody'
            AND rejected_incoming_count >= ?
            AND reject_mail_at_mta = 't'
            AND url_title <> 'holding_pen'",
            options[:rejection_threshold]])
  end

#
# Rake Tasks
#

# Generate a list of any requests over the threshold and OK to be rejected
task :generate_mta_rejection_list => :environment do
  options = { :rejection_threshold => ENV['REJECTED_THRESHOLD'],
              :age_in_months => ENV['AGE_IN_MONTHS'],
              :dryrun => dryrun }
  puts InfoRequest.
         over_rejection_at_mta_threshold(options).
           pluck(:incoming_email)
end

# Manually stop mail getting rejected for a given request
# The `false` option is a manual marker to stop it getting
# included in the rejection list, unless explicitly changed back
# to `true`
task :unset_reject_mail_at_mta => :environment do
  info_request = InfoRequest.find(ENV['REQUEST_ID'])
  info_request.update_attributes!(:reject_mail_at_mta => false)
end
```

However, doing it this way around makes it trickier to handle the UI elements, and installs who don't configure MTA rejection. i.e. what happens here?

```erb
<% if @info_request.reject_mail_at_mta? %>
  <p><strong>Mail for this request is automatically being rejected at the mail server due to spam.</strong></p>
<% else %>
```

I guess the dumb way of doing this is having a config flag and some instance method like:

```erb
<% if AlaveteliConfiguration.reject_at_mta && @info_request.over_rejection_threshold? %>
  <p><strong>Mail for this request is automatically being rejected at the mail server due to spam.</strong></p>
<% else %>
  …
<% end %>
```

```ruby
def over_rejection_threshold?(options = {})
  self.class.over_rejection_at_mta_threshold(options).
    where(:id => id).
      any?
end
```

Downsides to this are:

- "is being rejected" logic is much more confusing
- Would probably end up having to add config flags for the thresholds, given that we don't know what good defaults are yet
- The current initiation of MTA rejection is simpler – just run the rake task

Another alternative is not setting a default value for `reject_at_mta`.

| `reject_at_mta` value | Implication |
| -------------------|----------|
| `nil` | "fair game" for rejection |
| `true` | Is being rejected |
| `false` | Do not reject at MTA |

This allows `InfoRequest#reject_at_mta?` to behave in a similar way to how it does now, and we'd just need to change the query to use `AND reject_mail_at_mta IS NULL`.

I'm not sure whether I like the idea of a boolean field allowing a `NULL` value, but it does sort of model what we're trying to say. I thought about using a `String` field instead, but what would the values be? `%q(can_be_rejected yes no)`? Meh.

---

So yeah, that's about the point where I just figured you can change `allow_new_responses_from` in `unset_reject_mail_at_mta` and make the whole thing a lot less of a faff.

lib/tasks/config_files.rake
```diff
@@ -166,8 +166,19 @@ namespace :config_files do
 
   desc 'Produce a list of email addresses for which the MTA should reject messages at RCPT time'
   task :generate_mta_rejection_list => :environment do
+    example = 'rake config_files:generate_mta_rejection_list MTA=(exim|postfix)'
+    check_for_env_vars(['MTA'], example)
+    mta = ENV['MTA'].downcase
+    unless ['postfix', 'exim'].include? mta
+      puts "Error: Unrecognised MTA"
+      exit 0
```
garethrees: `exit 1` for error?

garethrees:
👍

PR #3303, opened 2016-06-06
Remove mention of FOI Register

### LINKED ISSUE #3295: Update API documentation to reflect the fact that foi-register is defunct.

garethrees:
Fixes #3295 

We've discontinued FOI Register, so remove the mention of it as the link is now broken.

Also labelled the "Write API" as experimental now that we have no real users of it.


lizconlan:
:+1:

lizconlan:
I think you can avoid sending documentation updates to travis by calling the branch gh-pages-something

PR #3306, opened 2016-06-07
Add download option to incoming message admin page.

### LINKED ISSUE #3280: Make downloading raw emails possible without rendering first in browser
Following https://github.com/mysociety/alaveteli/issues/3279 I attempted to download the raw email in question so I could use my mail client to unencode it and extract the spreadsheet to upload to the server.

I struggled because one must click "View raw email" on https://www.whatdotheyknow.com/admin/incoming_messages/782209/edit before being presented with the download link. The resultant 13mb+ webpage crashed my browser.

I therefore found the download link for another message, and noted that it is of format https://www.whatdotheyknow.com/admin/raw_emails/(message number).txt, so I attempted to go to URL https://www.whatdotheyknow.com/admin/raw_emails/782209.txt, but that was not the correct message. 

The "view raw email" button on https://www.whatdotheyknow.com/admin/incoming_messages/782209/edit links to https://www.whatdotheyknow.com/admin/raw_emails/782057 - I don't know why this is a different number (782057 versus 782209). I therefore successfully downloaded the message using https://www.whatdotheyknow.com/admin/raw_emails/782057.txt and was able to extract the information from it.

To avoid this situation, it would be of great benefit to have the "download raw email" link on the message's admin page, so that one doesn't have to click "view raw email" first - this avoids large emails crashing the browser.

crowbot:
Also standardize on a button representation for it.
Fixes #3280

### DIFF
--- a/app/views/admin_incoming_message/_actions.html.erb
+++ b/app/views/admin_incoming_message/_actions.html.erb
@@ -54,6 +54,15 @@
       <%= link_to "View raw email", admin_raw_email_path(incoming_message.raw_email_id), :class => "btn" %>
     </div>
   </div>
+  <div class="control-group">
+    <label class="control-label">Download</label>
+    <div class="controls">
+      <%= link_to "Download",  admin_raw_email_path(incoming_message.raw_email, :format => 'txt'), :class => "btn" %>
+    </div>
+  </div>
+
+
 <% end %>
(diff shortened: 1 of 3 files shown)


garethrees:
👍  Don't forget a changelog entry :)

PR #3311, opened 2016-06-09
Remove :debug Gemfile group

garethrees:
annotate is a development dependency.

Nothing left in :debug after the move of annotate, so remove it.


lizconlan:
👍

PR #3313, opened 2016-06-09
Styling for message delivery status display

zarino:
Branched off [`624-delivery-status-parser`](https://github.com/mysociety/alaveteli/pull/3214).

Part of https://github.com/mysociety/alaveteli/issues/624.

Here are a few shots of how the different states look in the core "none" theme:

core-sent

core-delivered

core-failed

core-info

And in alavetelitheme:

alavetelitheme-delivered

alavetelitheme-info


zarino:
If you want an easy way to test this locally, add the quick monkeypatches @garethrees outlined in his comment https://github.com/mysociety/alaveteli/issues/624#issuecomment-224318063

app/assets/stylesheets/responsive/_request_style.scss
```diff
@@ -117,14 +121,88 @@ a img.attachment_image {
...
+    position: absolute;
+    top: 0;
+    left: 0;
+    background: transparent image-url('loader-000-on-fff-16px.gif') 0 0 no-repeat;
+    background-size: 16px 16px;
+  }
+}
+
+.toggle-delivery-log--sent {
+  background-image: image-url('delivery-status/sent.png');
```
garethrees: The question mark makes it look like we don't know what the app is doing. How about making this a blue tick?

lizconlan: Verified by mySoc!

garethrees:
Added some commits to make this all a bit clearer:

- Everyone can see the delivery status next to the date
- **Normal Users** can only see the delivery status explanation
- **Request owners** can see the redacted mail server logs in addition to the above
- **Admins** can see the both the delivery status and full mail server logs

screen shot 2016-06-10 at 19 01 26

screen shot 2016-06-10 at 19 01 16

@zarino if you could just swap the sent icon out with a blue tick icon that would be great!

PR #3314, opened 2016-06-10
Add instructions for MTA rejection of some recipients.

crowbot:


garethrees:
Looks good. Should probably mention `unset_reject_incoming_at_mta` at the end?

Issue #3283, opened 2016-05-31
Get User transactions per month


garethrees:
> Track transactions per user per month. Run rake task on AsktheEU first as we host them, and on IPV (after upgrade), and on WDTK

Some things that would be useful:

- Add a counter cache to `User` for each type of transaction association.
- Add some methods so that its easy to get the transactions for a given time period (.e.g. "comments in the last 6 months for User 1234")
- Build this in to alaveteli, but also paste a script here so that we can run it through `rails runner` for installs that haven't updated.

garethrees:
RE: profiling, after a bit of experimentation in the console here's an initial sketch run over WDTK users:

```ruby
def user_type(data)
  return :ghost          if data.values.sum.zero?
  return :single_request if data[:info_requests] == 1 && data.values.sum == 1
  return :commenter      if data[:comments] > data.dup.slice!(:comments).values.sum
  return :requester      if data[:info_requests] > data.dup.slice!(:info_requests).values.sum
  return :classifier     if data[:request_classifications] > data.dup.slice!(:request_classifications).values.sum
  return :follower       if data[:track_things] > data.dup.slice!(:track_things).values.sum
  #
  # Doesn't work yet because of missing association
  # return :body_updater if data[:public_body_change_requests] > data.dup.slice!(:public_body_change_requests).values.sum
  return :other
end

User.where(:ban_text => '').count
# => 97052

puts Time.zone.now
# => 2016-06-16 12:15:47 +0100

User.where(:ban_text => '').find_each do |user|
  data = {
    :comments => user.comments.count,
    :info_requests => user.info_requests.count,
    :request_classifications => user.request_classifications.count,
    :track_things => user.track_things.count
  }

  profiled_users[user_type(data)] ||= 0
  profiled_users[user_type(data)] += 1
end

puts Time.zone.now
# => 2016-06-16 12:51:37 +0100

profiled_users
# Manually sorted
# => { :single_request => 30306,
#      :ghost          => 25953,
#      :requester      => 20649,
#      :follower       => 11018,
#      :other          =>  7046,
#      :commenter      =>  1674,
#      :classifier     =>   400 }
```

- The distribution _feels_ about right. Lots who have done nothing and made a single request; outside of that, mainly people who make requests. A fair amount of followers and less of everyone else.
- It would be interesting to look in to `:other`, but it probably wouldn't affect the overall picture. An example of someone that we might want to class as a commenter, but is getting caught in `:other`:

```ruby
data = {
[...]

PR #3317, opened 2016-06-13
[#3283] Transaction counter caches

garethrees:
Work on #3283


doc/CHANGES.md
```diff
@@ -2,6 +2,7 @@
 
 ## Highlighted Features
 
+* Add counter caches for `User` transaction attributes (Gareth Rees)
```
crowbot: Probably not worth a changelog note in itself - really that should be for #3283 as a whole, which is the level at which people might use it.

PR #3320, opened 2016-06-14
Add new script for replicating Xapian full text search databases

hazelesque:
PR for review, per https://github.com/mysociety/sysadmin/issues/654

### DIFF
--- a/script/replicate-xapian-database
+++ b/script/replicate-xapian-database
@@ -33,7 +33,7 @@ trap "echo Caught SIGCHLD; jobs -l" SIGCHLD
 xapian-replicate-server -p "$PORT" --one-shot "$XAPIAN_DB_DIR" &
 
 # Start client on remote host
-ssh -n "$REMOTEUSER"@"$REMOTEHOST" -- xapian-replicate -h "$LOCALHOST" -p "$PORT" --one-shot -m "$DBNAME" "$REMOTEPATH/$DBNAME" &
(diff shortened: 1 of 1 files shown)


hazelesque:
> We mostly avoid reopen(), but have one call to it. Do we need to fix that, or does it not matter because it's the master database, so isn't being updated, is just being read from?

@crowbot: In our current "cold standby" scenario, there should be no need.

Getting rid of that reopen() call could be useful if we wished to have a "hot standby" instance of WDTK running against the replicated index, or load balance across multiple WDTK instances (e.g. with one machine acting as index master, and the others using replicas for read access), though.

hazelesque:
> Looking at http://xapian.org/docs/replication.html, it's not clear to me whether the master database is still accessible for searching during the replication process - do you know? The docs say
> 
> > Searching (on the slave databases) and indexing (on the master database) can continue during synchronisation.
> 
> Not sure if that implies that searching on the master can also continue.

@crowbot: My reading of https://media.readthedocs.org/pdf/getting-started-with-xapian/latest/getting-started-with-xapian.pdf would suggest that they can, if xapian-replicate-server(1) is a "reader" for the purpose of backend operation there.

I certainly have no reason to believe that it would be a writer, but to be honest -- even if it was, that documentation states that "Multiple concurrent readers are supported (in addition to the writer)." (under the section "Concurrent access")

Issue #3328, opened 2016-06-17
[ERROR] request#show (Encoding::ConverterNotFoundError) "code converter not found (UTF-7 to UTF-8)


garethrees:
Error on IPV. They're on 0.24.1.4, which includes https://github.com/mysociety/alaveteli/commit/abfb6eb34831dfac3dbb1086ebb08ab9380c1b30 – the only commit I found RE UTF-7. 

Related? https://github.com/mysociety/alaveteli/issues/2863

```
An Encoding::ConverterNotFoundError occurred in request#show:

code converter not found (UTF-7 to UTF-8)
lib/mail_handler/backends/mail_extensions.rb:51:in `encode'


-------------------------------
Request:
-------------------------------

* URL : http://www.infoprovsechny.cz/request/zlodejna_o2?nocache=incoming-8463
* IP address: FILTERED
* Parameters: {"nocache"=>"incoming-8463", "controller"=>"request", "action"=>"show", "url_title"=>"zlodejna_o2"}
* Rails root: /FILTERED/www.infoprovsechny.cz/releases/20160520113907
* Timestamp : 2016-06-17 09:50:31 +0200

-------------------------------
Backtrace:
-------------------------------

lib/mail_handler/backends/mail_extensions.rb:51:in `encode'
lib/mail_handler/backends/mail_extensions.rb:51:in `b_value_decode'
mail (2.5.4) lib/mail/encodings.rb:245:in `b_value_decode'
mail (2.5.4) lib/mail/encodings.rb:135:in `block (3 levels) in value_decode'
mail (2.5.4) lib/mail/encodings.rb:128:in `map'
mail (2.5.4) lib/mail/encodings.rb:128:in `block (2 levels) in value_decode'
mail (2.5.4) lib/mail/encodings.rb:123:in `map'
mail (2.5.4) lib/mail/encodings.rb:123:in `block in value_decode'
mail (2.5.4) lib/mail/encodings.rb:122:in `map'
mail (2.5.4) lib/mail/encodings.rb:122:in `value_decode'
mail (2.5.4) lib/mail/encodings.rb:101:in `decode_encode'
mail (2.5.4) lib/mail/fields/unstructured_field.rb:73:in `do_decode'
mail (2.5.4) lib/mail/fields/unstructured_field.rb:55:in `decoded'
mail (2.5.4) lib/mail/fields/unstructured_field.rb:59:in `default'
mail (2.5.4) lib/mail/field.rb:167:in `method_missing'
[...]

PR #3329, opened 2016-06-17
Handle UTF-7 Subject

garethrees:
Fixes https://github.com/mysociety/alaveteli/issues/3328

### DIFF
--- a/lib/mail_handler/backends/mail_extensions.rb
+++ b/lib/mail_handler/backends/mail_extensions.rb
@@ -38,18 +38,22 @@ module Mail
   class Ruby19
 
     def self.b_value_decode(str)
-      match = str.match(/\=\?(.+)?\?[Bb]\?(.+)?\?\=/m)
+      match = str.match(/\=\?(.+)?\?[Bb]\?(.*)\?\=/m)
       if match
         charset = match[1]
         str = Ruby19.decode_base64(match[2])
-        # Rescue an ArgumentError arising from an unknown encoding.
-        begin
-          str.force_encoding(pick_encoding(charset))
-        rescue ArgumentError
-        end
+        # The commented line below is the actual implementation in a217776
+        # (https://git.io/vozPG) but we don't have a `charset_encoder` object
+        # available, so revert to the behaviour of the default
+        # Mail::Ruby19::StrictCharsetEncoder#encode (https://git.io/vozXb).
+        # str = charset_encoder.encode(str, charset)
+        str.force_encoding(Mail::Ruby19.pick_encoding(charset))
       end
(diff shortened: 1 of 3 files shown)


garethrees:
Ah sorry, too eager!

garethrees:
So the spec that's failing is:

```ruby
  it 'should not be confused by subject lines with malformed UTF-8 at the end' do
    # The base64 subject line was generated with:
    #   printf "hello\360" | base64
    # ... and wrapping the result in '=?UTF-8?B?' and '?='
    mail = get_fixture_mail('subject-bad-utf-8-trailing-base64.email')
    expect(mail.subject).to eq('hello')
    # [snip]…
  end
```

If we inspect the decoded value after running it through `normalize_string_to_utf8` it's `helloð`:

```diff
diff --git a/lib/mail_handler/backends/mail_extensions.rb b/lib/mail_handler/backends/mail_extensions.rb
index f6db780..22e8f30 100644
--- a/lib/mail_handler/backends/mail_extensions.rb
+++ b/lib/mail_handler/backends/mail_extensions.rb
@@ -49,6 +49,7 @@ module Mail
         end
       end
       decoded = normalize_string_to_utf8(str)
+      puts decoded
       decoded.valid_encoding? ? decoded : decoded.encode("utf-16le", :invalid => :replace, :replace => "").encode("utf-8")
     end
```

```
Randomized with seed 38496
helloð
F

Failures:

  1) when creating a mail object from raw data should not be confused by subject lines with malformed UTF-8 at the end
     Failure/Error: expect(mail.subject).to eq('hello')

       expected: "hello"
            got: "hello?"

       (compared using ==)
     # ./spec/lib/mail_handler/mail_handler_spec.rb:52:in `block (2 levels) in <top (required)>'

Finished in 0.33247 seconds (files took 27.33 seconds to load)
1 example, 1 failure
```

Given that we can apparently convert it correctly, I'm not sure why the spec is expecting `"hello?"`.

lib/mail_handler/backends/mail_extensions.rb
```diff
@@ -38,18 +38,21 @@ def encoded
...
-        # Rescue an ArgumentError arising from an unknown encoding.
-        begin
-          str.force_encoding(pick_encoding(charset))
-        rescue ArgumentError
-        end
+        # The line below is the actual implementation in a217776
+        # (https://git.io/vozPG) but we don't have a `charset_encoder` object
+        # available, so I've reverted to the previous implementation.
+        # str = charset_encoder.encode(str, charset)
+        str.force_encoding(pick_encoding(charset))
```
garethrees: Wonder whether this should still rescue an `ArgumentError` here.

PR #3342, opened 2016-06-21
Postfix delivery status parser

crowbot:


garethrees:
Looks good, but there's a bit more work to do here (if the intention was to have the end-to-end process working by merging this PR).

At the moment we only have the ability to get logs for a single `OutgoingMessage` when we're using [Exim](https://github.com/mysociety/alaveteli/blob/develop/app/models/outgoing_message.rb#L234-L262).

Does Postfix usually have multiple log lines per message? (e.g. a line for `sent` and then a line for `delivered`?). If so, it would be worth adding some specs to [`OutgoingMessage#delivery_status`](https://github.com/mysociety/alaveteli/blob/develop/spec/models/outgoing_message_spec.rb#L1066-L1130) to check that it can pluck out the correct status given a full set of log lines for the various stats a message can be in.

I've also noticed that my implementation of equality on the log line classes is a bit flawed. They should check `self.class == other.class` as well as `to_s`. Once done, [this spec](https://github.com/mysociety/alaveteli/blob/develop/spec/models/mail_server_log_spec.rb#L270-L292) should fail – need to add a stub on `AlaveteliConfiguration#mta_log_type` to assert we get the correct log type back.

spec/models/outgoing_message_spec.rb
```diff
@@ -808,322 +808,658 @@
...
+
+            load_mail_server_logs <<-EOF.strip_heredoc
+            Jun 15 16:02:40 host postfix/qmgr[5216]: 1A6A236F4E08: removed
+            Jun 15 16:02:40 host postfix/smtp[12120]: 1A6A236F4E08: to=<#{ body_email }>, relay=example.com[0.0.0.0]:25, delay=1.2, delays=0.12/0.01/0.96/0.09, dsn=2.0.0, status=sent (250 2.0.0 Ok: queued as B7F313A054)
+            Jun 15 16:02:39 host postfix/qmgr[5216]: 1A6A236F4E08: from=<#{ request_email }>, size=1499, nrcpt=1 (queue active)
+            Jun 15 16:02:39 host postfix/cleanup[12118]: 1A6A236F4E08: message-id=<#{ smtp_message_id }>
+            Jun 15 16:02:39 host postfix/pickup[31710]: 1A6A236F4E08: uid=1003 from=<#{ request_email }>
+            EOF
+
+            # Resend the message without importing exim logs for it, simulating a
```
garethrees: s/exim/postfix/

spec/models/outgoing_message_spec.rb
```diff
@@ -808,322 +808,658 @@
...
+          2016-04-06 12:01:08 [14933] 1anlCt-0003sm-LG ** foi@example.net F=<request-326806-hk82iwn7@localhost>: all relevant MX records point to non-existent hosts
+          2016-04-06 12:01:08 [14933] 1anlCt-0003sm-LG ** foi@example.net F=<request-326806-hk82iwn7@localhost>: all relevant MX records point to non-existent hosts
+          2016-04-06 12:01:08 [14935] 1anlCu-0003st-1p <= <> R=1anlCt-0003sm-LG U=Debian-exim P=local S=2934 T="Mail delivery failed: returning message to sender" from <> for request-326806-hk82iwn7@localhost
+          2016-04-22 13:13:03 [24970] 1atZxH-0006Uk-KF <= request-326806-hk82iwn7@localhost U=alaveteli P=local S=1923 id=ogm-531356+571a154f7b7c5-2a7e@localhost T="Freedom of Information request - Some Information" from <request-326806-hk82iwn7@localhost> for foi@example.net foi@example.net
+          2016-04-22 13:24:41 [29720] 1atZxH-0006Uk-KF => foi@example.net F=<request-326806-hk82iwn7@localhost> P=<re
```
garethrees: I'd add a spec for normal message delivery, and possibly a redelivered message if we have an example.

spec/support/email_helpers.rb
```diff
@@ -29,8 +29,14 @@ def parse_all_incoming_messages
...
-  log.split("\n").each_with_index do |line, index|
-    MailServerLog.create_exim_log_line(line, batch, index + 1)
+  mta_log_type = AlaveteliConfiguration.mta_log_type.to_sym
+  io_stream = StringIO.new(log)
+  case mta_log_type
+  when :exim
+    MailServerLog.load_exim_log_data(io_stream, batch)
+  when :postfix
+    MailServerLog.load_postfix_log_data(io_stream, batch)
+  else raise "Unexpected MTA type: #{ mta_log_type }"
```
garethrees: Newline after `else`.

garethrees:
Looks good! Couple of minor points, and we probably want to fix the [log line equality issue](https://github.com/mysociety/alaveteli/pull/3342#issuecomment-227706583) but +1 👍  🎉

Issue #3291, opened 2016-06-02
Refactor notices and messages


wrightmartin:
When working on https://github.com/mysociety/whatdotheyknow-theme/pull/318 I found it hard to follow the HTML and CSS for the various notices and messages. 

We should refactor this to ensure it makes logical sense and shares terminology and code in a thoughtful way

wrightmartin:
**Notes from investigation**

There are 5 types of message that share code at the moment
- General pop-up message
- Country notice
- Flash notice
- Flash error
- Link box

Functionally messages and notices are identical. The Link box is not. It should be separated from this code as it's not similar at all, except sharing a javascript close button

wrightmartin:
Approach wise I plan to add JS and CSS for the new style notices and move, General pop-up message, Country notice, Flash notice and Flash error to use this. 

I'll keep the link box as is, using the legacy code and open a ticket to look at redesigning this feature

wrightmartin:
I *think* this is ready, again I may have missed some notices or messages that come via less-travelled paths

PR #3346, opened 2016-06-22
Refactored alerts to be more clear

wrightmartin:
Fixes https://github.com/mysociety/alaveteli/issues/3291

- Using BEM syntax, so reusers can adapt, modify and build on them
- Using common terminology to help reusers and themers comprehend them better
- Build the country message entirely in JS (it relies on this to work anyway) to remove the needed to empty divs in the DOM and simplify styling


app/assets/javascripts/general.js
```diff
@@ -1,31 +1,46 @@
 $(document).ready(function() {
  // flash message for people coming from other countries
- if(window.location.search.substring(1).search("country_name") == -1) {
-    if (!$.cookie('has_seen_country_message')) {
-  $.ajax({
-    url: "/country_message",
-        dataType: 'html',
-        success: function(country_message){
+ var htmlWrapperFront = [
```
wrightmartin: my JavaScript is *very* rusty. This is likely bad code, feel free to rip it apart

garethrees:
Does this fix https://github.com/mysociety/alaveteli/pull/3346, or is there more that needs doing?

wrightmartin:
@garethrees it fixes #3291, I believe. Unless there are notices and other messages I haven't been able to reproduce yet

garethrees:
I think this needs a bit of a rethink. It looks like you were intending to change the [flash messages](https://github.com/mysociety/alaveteli/blob/develop/app/views/layouts/default.html.erb#L101-L106) (which I think is what we should call them, not alerts or popups) but you've actually changed the "[popup](https://github.com/mysociety/alaveteli/blob/develop/app/views/layouts/default.html.erb#L79-L87)". I actually like the name of that, as its an element that can be closed (which is what you associate with a popup).

You can set a popup by adding some content to [`app/views/general/_popup_banner.html.erb`](https://github.com/mysociety/alaveteli/blob/develop/app/views/general/_popup_banner.html.erb).

wrightmartin:
Here's the notes from our chat about this

---
**Flash messages (rails terminology)**
- Undo the rename from flash to alerts
- Telling people something’s happened (error, success, warning, validation)
- Flash partials - boolean driven flash message that uses a partial to render

**Popups**
- Showing you're visiting the wrong country
- Survey CTA, Donate CTA
- Dismissible + cookie setting

**Form validation message**

- errorExplanation

lizconlan:
rebased in hope of fixing travis fail

lizconlan:
I think the `alaveteli` file was included in error

app/assets/javascripts/general.js
lizconlan: ha ha - like mine's any better!

lizconlan:
@wrightmartin code looks good to me, apart from the accidental file include (easily fixed). Can I assume that you've done the browserstack thing or should I? (either's fine!)

app/assets/javascripts/general.js
```diff
@@ -1,31 +1,46 @@
...
-   $.cookie('seen_foi2', 1, { expires: 7, path: '/' });
-   return false;
-   });
+      }
+    })
+  }
+}
+
+ // popups
+$('#locality-popup .js-popup__close').click(function() {
```
lizconlan: This doesn't work as the block that creates `$('#locality-popup .js-popup__close')` is async (I think) and may still be running by the time you get here, meaning there's no object to handle the click event for. It fails silently, breaking the close button 😢 

Moving this inside the `success:` block, just after the closing `}` of `if (country_message)` seems to fix it.

alaveteli
```diff
@@ -0,0 +1 @@
+/home/vagrant/alaveteli
```
lizconlan: this file included in error?

lizconlan:
@wrightmartin I rebased it BTW, just so that the tests would pass - I recommend a git pull before you start work on it (sorry)

alaveteli
wrightmartin: ah, yes - looks like something that shouldn't be here at all

wrightmartin:
The weird styling is due to alavetelitheme's styles and alaveteli core's styles conflicting. I'll make sure this can't happen when other themes are using the updated core

wrightmartin:
Thanks for the feedback liz, I've made all the changes you recommended

Issue #3281, opened 2016-05-31
Improve pagespeed


garethrees:
Should be aiming for greens in the [pagespeed test](https://developers.google.com/speed/pagespeed/insights/).

| Site | Mobile | Desktop |
|-------|----------|------------|
| [AskTheEU](https://developers.google.com/speed/pagespeed/insights/?url=asktheeu.org&tab=mobile) | 64/100 | 82/100 |
| [WhatDoTheyKnow](https://developers.google.com/speed/pagespeed/insights/?url=https%3A%2F%2Fwww.whatdotheyknow.com%2F) | 70/100 | 84/100 |
| [RightToKnow](https://developers.google.com/speed/pagespeed/insights/?url=https%3A%2F%2Fwww.righttoknow.org.au%2F) | 64/100 | 77/100 |
| [Imamo pravo znati](https://developers.google.com/speed/pagespeed/insights/?url=https%3A%2F%2Fimamopravoznati.org%2F) | 67/100 | 82/100 |
| [Informace pro vsechny](https://developers.google.com/speed/pagespeed/insights/?url=http%3A%2F%2Fwww.infoprovsechny.cz%2F) | 66/100 | 80/100 |
| [KiMitTud](https://developers.google.com/speed/pagespeed/insights/?url=http%3A%2F%2Fkimittud.atlatszo.hu%2F) | 77/100 | 88/100 |
| [FYI NZ](https://developers.google.com/speed/pagespeed/insights/?url=https%3A%2F%2Ffyi.org.nz%2F) | **55**/100 | 73/100 |

KiMitTud is the only site that gets a green tick (on desktop). They all score well on mobile UX other than IPV because of the old theme.

wrightmartin:
Suggestions from Google PageSpeed are

- Remove Render-Blocking JavaScript
- Optimise CSS delivery

### Render-Blocking JavaScript
JavaScript from core Alaveteli is all loaded in application.js at the top of the file they are...

- jQuery
  - jqueryujs
  - jqueryui-datepicker
  - jqueryui-position
  - jquery.cookie
- general
- ba-throttle-debounce

**jqueryujs**
unobtrusive scripting support

**jqueryui-datepicker**
Used on request search page to select a date

 **jqueryui-position**
Used to position the ‘link_to_this’ box (soon to be redesigned)

**jquery.cookie**
[...]

PR #3348, opened 2016-06-24
Improve page speed

wrightmartin:
Connects to #3281

Some small improvements to alaveteli core to improve the experience for mobile users, and those on slow connections


lizconlan:
👍

PR #3353, opened 2016-06-30
Fix updated at changes on rejection

crowbot:


app/models/info_request.rb
```diff
@@ -1446,9 +1446,10 @@ def accept_incoming?(email, raw_email_data)
       end
 
     will_be_rejected = (response_rejector && response_rejection) ? true : false
-
     if will_be_rejected && response_rejection.reject(response_rejector.reason)
-      self.increment!(:rejected_incoming_count)
+      # update without changing the updated_at field
+      self.update_column(:rejected_incoming_count,
+        self.increment(:rejected_incoming_count).rejected_incoming_count)
```
garethrees: I don't think there's a need to use `increment` now; its a bit confusing in this context.

```ruby
update_column(:rejected_incoming_count, rejected_incoming_count.next)
```

Could use `+= 1` instead of [`next`](http://ruby-doc.org/core-1.9.3/Integer.html#method-i-next) – either way :)

For some reason I needed `self.rejected_incoming_count` for the tests to pass when trying `+= 1` as it was returning `nil` instaed of `0`. `next` seems to be fine, even though `nil` doesn't respond to `next`.

Issue #3335, opened 2016-06-20
Use new relic to look at performance slowdown for WDTK


crowbot:
To be done after rollout of #624

garethrees:
`RequestController#show` is causing the most pain:

![screen shot 2016-06-29 at 17 13 36](https://cloud.githubusercontent.com/assets/282788/16459926/dec351b0-3e1c-11e6-967d-3185c5e8e471.png)

The sidebar is taking up the most time (~200ms):

![screen shot 2016-06-29 at 17 14 47](https://cloud.githubusercontent.com/assets/282788/16459970/0af91d32-3e1d-11e6-927d-8255d082c650.png)

In development mode, the overall time is 384ms, but the none of the details underneath seem to account for this time:

![screen shot 2016-06-29 at 17 16 24](https://cloud.githubusercontent.com/assets/282788/16460025/561b9b0a-3e1d-11e6-95d4-673fe544c148.png)

Either way, I expect this is due to the similar requests. We're caching the fragment for a day but we're still getting a load of writes, meaning that we're performing the calculation a fair amount:

```sh
$ sudo cat applogs/production.log | g -E "Read\sfragment\sviews\/request\/similar" | wc -l
136200
$ sudo cat applogs/production.log | g -E "Write\sfragment\sviews\/request\/similar" | wc -l
94714
```

garethrees:
Some fixes made in https://github.com/mysociety/alaveteli/pull/3354. Will need to have them deployed for a day to see if we've made a difference and to decide on next actions.

garethrees:
Very minor improvement:

**Before (0.65 Apdex):**

![screen shot 2016-07-11 at 17 52 54](https://cloud.githubusercontent.com/assets/282788/16913071/1bc3cc80-4ce0-11e6-8d30-f0e308d233f7.png)


**After (0.67 Apdex):**


![screen shot 2016-07-11 at 17 56 52](https://cloud.githubusercontent.com/assets/282788/16913076/203f6a12-4ce0-11e6-9e96-74c132678174.png)

garethrees:
Suggest we dig in to two things a little more:

- Rendering the `request/_request_listing_short_via_event.html.erb` partial
  - Are we doing anything _inside_ it that can be quicker
  - Can we cache these, given that the same sidebar listing may appear elsewhere. That way, we might be able to prevent loading the record from the search results
- `InfoRequest#similar_requests`
  - Anything we can do here?
[...]

PR #3354, opened 2016-06-30
[#3335] request show performance

garethrees:
First fixes for #3335 

Also adds tracking for https://github.com/mysociety/alaveteli-experiments/issues/87


lizconlan:
👍 looks promising, looking forward to seeing what sort of results we get

Issue #2510, opened 2015-05-30
Better error messages when importing from CSV file


samsmith:
in the admin interface, for loading authorities from a CSV file, move the errors box to the top (above the successful lines), and make it clear that nothing imports until the errors in that box are fixed.

Maybe a nice red colour to signify an error

garethrees:
Closing as pretty similar to https://github.com/mysociety/alaveteli/issues/2109

Issue #2109, opened 2015-01-21
catch CSV format errors nicely on CSV upload


davewhiteland:
Currently, CSV upload blows up and pings out a HTTP 500 if there's a wayward double-quote in there; would be kind to absorb it and report it as a problem with a line number.

I believe this means wrapping https://github.com/mysociety/alaveteli/blob/f0bbeb4abf4bf07e5cfb46668f39bbff72ed7210/app/models/public_body.rb#L473 with a `begin-rescue-end` with something like:

``` ruby
rescue CSV::IllegalFormatError
  errors.push "error: line #{line.to_s}: CSV format error"
```

Extra points if the error message pushed out "(maybe need to double-up quotes?)" _if_ the row contains quotation marks because that is the most common cause of this blowing up. Note that I think it's reasonable to anticipate quotes being present, because the `notes` column may contain HTML.

garethrees:
I expect something similar caused the recent errors @kingqueen3065 was having when uploading a CSV file.

```
An ActiveRecord::StatementInvalid occurred in admin_public_body#import_csv:

PG::NotNullViolation: ERROR: null value in column "first_letter" violates not-null constraint
: INSERT INTO "public_bodies" ("api_key", "created_at", "disclosure_log", "first_letter", "home_page", "info_requests_count", "info_requests_not_held_count", "info_requests_overdue_count", "info_requests_successful_count", "info_requests_visible_classified_count", "last_edit_comment", "last_edit_editor", "name", "notes", "publication_scheme", "request_email", "short_name", "updated_at", "url_name", "version") VALUES ($1, $2, $3, $4, $5, $6, $7, $8, $9, $10, $11, $12, $13, $14, $15, $16, $17, $18, $19, $20) RETURNING "id"
vendor/bundle/ruby/1.9.1/gems/activerecord-3.2.21/lib/active_record/connection_adapters/postgresql_adapter.rb:1176:in `get_last_result'


-------------------------------
Request:
-------------------------------

* URL : https://www.whatdotheyknow.com/admin/bodies/import_csv
* IP address: [REDACTED]
[...]

Issue #506, opened 2012-06-08
Only one tag allowed in tag box on CSV upload


RichardTaylor:
Only one tag can be specified in CSV upload form box "Optional: Tag to add entries to / alter entries for:"  at:

https://www.whatdotheyknow.com/admin/bodies/import_csv

It would be nice if this supported multiple, space separated, tags. 

With #60 having improved the CSV upload itsself to accept multiple tags this isn't very important at all. 

Simply improving the form's help text to explain that one tag can be put in the form field, but anything more involved should be done in the CSV would help, the issue is surprise / unexpected behaviour as much as a lack of functionality.

HelenWDTK:
This has been fixed. Adding multiple tags is possible.

PR #3355, opened 2016-06-30
Improve on-page CSV upload docs

### LINKED ISSUE #3325: Improve documentation relating to the CSV upload page
Suggestions:

* Link from the CSV upload form to the documentation at: http://alaveteli.org/docs/running/admin_manual/#creating-changing-and-uploading-public-authority-data 
* Add an example with a comma in the name to the example given on the form itsself, and a line with the use of ""  around entries. 
* The Alaveteli documentation and / or the text on the CSV upload form should explain what the "import tag" is. Presumably the "import tag" isn't a tag, or is some kind of special tag (it doesn't appear on the body admin page in the tag field).

lizconlan:
The second commit (2500c3e) relies on my understanding of the phrase "import tag" being correct. See https://github.com/mysociety/alaveteli/issues/3325#issuecomment-229682964

Closes #3325

### DIFF
--- a/app/views/admin_public_body/import_csv.html.erb
+++ b/app/views/admin_public_body/import_csv.html.erb
@@ -2,17 +2,6 @@
 
 <h1><%=@title%></h1>
 
-<br>
-<div class="row">
-  <div class="span8">
-    <div class="alert alert-info">
-      <p>Full documentation on uploading public authority data can be found <%= link_to ' on alaveteli.org',
-        'http://alaveteli.org/docs/running/admin_manual/#creating-changing-and-uploading-public-authority-data', :target => '_blank' %>.
-      </p>
-    </div>
-  </div>
-</div>
-
 <% if not @notes.empty? %>
   <pre id="notice"><%=@notes %></pre>
 <% end %>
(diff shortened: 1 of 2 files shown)


garethrees:
Some minor tweaks. I've also suggested a layout improvement in https://github.com/mysociety/alaveteli/pull/3361. Feel free to cherry-pick it in if you think it will benefit.

garethrees:
Updated with fixed commits and nicer layout

admin-authority-csv-upload

garethrees:
See also https://github.com/mysociety/alaveteli/pull/3361#issuecomment-234576946

garethrees:
> Is the request_email column really required as the text on the form says?

I've just verified that it's _not_ required, at least when creating a new authority. Looks like that text was [introduced](https://github.com/mysociety/alaveteli/commit/5a501861810fca5c9cea46b5787b671ccff4131e/#diff-927ed4b53ac48bb51fc90b5604333c97R26)) very early on in this feature's development and has stuck ever since. Need to have a better poke around to see if there are cases where it _is_ required.

> Also isn't the description of the input box for the tag unclear…
>
>…So why not just use latter description from the documentation on the form? The wording as it is means I'm feeling unsure if there's some function I'm missing.

I've probably made this a bit clearer since https://github.com/mysociety/alaveteli/issues/3361 (see https://github.com/mysociety/alaveteli/pull/3355#issuecomment-233683056), but I think naming the form label to "Optional import tag" to match the documentation would be better.

> I note there's also #506 which is related.

Yeah, there doesn't seem to be a really good reason to limit this to one tag, but we'll tackle that separately I think.

> I thought there was also a ticket somewhere about the fact the page you get after submitting the form prints the whole list of bodies - which is excessive - but that's going off the topic of the form itself.

https://github.com/mysociety/alaveteli/issues/2510 & https://github.com/mysociety/alaveteli/issues/2109 I think. Will tackle separately.

garethrees:
Ah, so the `request_email` field _is_ required, but only when:
- The authority has a `short_name`
- The CSV contains a `short_name` field

:/

lizconlan:
:+1: much clearer 🍰

PR #3361, opened 2016-07-05
[#3325] Improve layout of Public Body CSV Import

garethrees:
Addition to #3355

Moves the help section below the form to make the actual input options clearer.

Move the "Standard tags" to a sidebar next to the help as the list can
get pretty long (see WDTK).

Minor rewording of the link to alaveteli.org documentation to be more
inline with Admin Censor Rules inline help.

screen shot 2016-07-05 at 11 48 32

### DIFF
--- a/app/views/admin_public_body/import_csv.html.erb
+++ b/app/views/admin_public_body/import_csv.html.erb
@@ -2,17 +2,6 @@
 
 <h1><%=@title%></h1>
 
-<br>
-<div class="row">
-  <div class="span8">
-    <div class="alert alert-info">
-      <p>Full documentation on uploading public authority data can be found <%= link_to ' on alaveteli.org',
-        'http://alaveteli.org/docs/running/admin_manual/#creating-changing-and-uploading-public-authority-data', :target => '_blank' %>.
-      </p>
-    </div>
-  </div>
-</div>
-
 <% if not @notes.empty? %>
   <pre id="notice"><%=@notes %></pre>
 <% end %>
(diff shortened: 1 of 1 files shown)


RichardTaylor:
I suspect this might not be the best place for these notes, and I'm happy improvements have been made here,  but anyway:

Is the request_email column really required as the text on the form says? (Why? One can create a public body individually on the system without an email address.)  

I think a more important question for site maintenance is : Is the request_email column required even when updating bodies ? If it isn't the form text and related Alaveteli.org documentation should say so. 

The [documentation](http://alaveteli.org/docs/running/admin_manual/#creating-changing-and-uploading-public-authority-data) says
>If the authority already exists (the name matches an existing authority’s name exactly), a blank entry leaves the existing value for that column unchanged — that is, that item of data on your site will not be changed. This means you only really need to include data you want to update.

So from that I'd expect to be able to update say all have a spreadsheet with a  name and notes field and use the CSV upload form to update all the notes for that set of bodies, even if I didn't have the list of email addresses to hand. 


Also isn't the description of the input box for the tag unclear:

It says

>Tag to add entries to / alter entries for

which the [documentation](http://alaveteli.org/docs/running/admin_manual/#creating-changing-and-uploading-public-authority-data) explains as

>Tag to apply to every body that is imported from your CSV file.

So why not just use latter description from the documentation on the form? The wording as it is means I'm feeling unsure if there's some function I'm missing.

===

I note there's also #506 which is related.  

I thought there was also a ticket somewhere about the fact the page you get after submitting the form prints the whole list of bodies - which is excessive - but that's going off the topic of the form itself. 

===
Personal admission - I made updates to about 30 bodies on WDTK from a spreadsheet provided by a user recently via the individual body edit forms because I wasn't sure how the CSV upload was going to behave (in any case there were name changes to be made which the CSV upload wouldn't cope with - but I could have made the other edits using the CSV I think).   

cc @lizconlan

PR #3368, opened 2016-07-06
Make request#list only respond to html requests

### LINKED ISSUE #2918: [ERROR] request#list (ActionView::MissingTemplate) "Missing template request/list, application/...
Routes with no `.json` endpoint should 404 rather than error.

```
An ActionView::MissingTemplate occurred in request#list:

  Missing template request/list, application/list with {:locale=>[:en], :formats=>[:json], :handlers=>[:erb, :builder, :coffee]}. Searched in:
  * "/APP_ROOT/alaveteli-2015-11-17T16-35-35/lib/themes/whatdotheyknow-theme/lib/views"
  * "/APP_ROOT/alaveteli-2015-11-17T16-35-35/app/views"
  * "/APP_ROOT/shared/vendor/bundle/ruby/1.9.1/gems/foundation-rails-5.2.1.0/app/views"

  vendor/bundle/ruby/1.9.1/gems/actionpack-3.2.21/lib/action_view/path_set.rb:58:in `find'

Request:

  * URL       : https://www.whatdotheyknow.com/list/successful.json
  * IP address: REDACTED
  * Parameters: {"view"=>"successful", "controller"=>"request", "action"=>"list", "format"=>"json"}
  * Rails root: /APP_ROOT/alaveteli-2015-11-17T16-35-35
  * Timestamp : 2015-11-19 15:44:45 +0000
```

lizconlan:
Closes #2918

### DIFF
--- a/app/controllers/request_controller.rb
+++ b/app/controllers/request_controller.rb
@@ -167,6 +167,12 @@ class RequestController < ApplicationController
   end
 
   def list
+    # respond with a 404 without a database lookup if request was not for html
+    if request.format && !request.format.html?
+      respond_to { |format| format.any { head :not_found } }
+      return
+    end
+
     medium_cache
     @view = params[:view]
     @locale = I18n.locale.to_s
--- a/spec/controllers/request_controller_spec.rb
+++ b/spec/controllers/request_controller_spec.rb
@@ -26,6 +26,11 @@ describe RequestController, "when listing recent requests" do
     }.to raise_error(ActiveRecord::RecordNotFound)
   end
 
+  it "returns 404 for non html requests" do
+    get :list, :view => "all", :format => :json
+    expect(response.status).to eq(404)
+  end
+
   it 'should not raise an error for a page param of less than zero, but should treat it as
         a param of 1' do
     expect{ get :list, :view => 'all', :page => "-1" }.not_to raise_error


app/controllers/request_controller.rb
```diff
@@ -167,6 +167,12 @@ def similar
   end
 
   def list
+    # respond with a 404 and do not execute the search if request was not for html
+    if request.format && !request.format.html?
+      respond_to { |format| format.any { head :not_found } }
+      return
+    end
```
garethrees: Wouldn't it be better to respond to the formats we do accept? https://github.com/mysociety/alaveteli/commit/1644cbffa6fff252fca85c407ec55e3dacd38a5f

lizconlan: This has the advantage of returning from the controller early, without any potentially expensive hits on the db, otherwise all the controller code is run then not used for anything.

garethrees: Ah lame!

lizconlan: The comment talking about "search" is probably wrong though - lazy copy and paste from 67740f0 :(

Agree that intent is clearer with `:respond_to`

lizconlan:
very minor tweak to change the line comment to talk about db lookups rather than a "search"

PR #3372, opened 2016-07-08
Update style guide

crowbot:


.ruby-style.yml
```diff
@@ -985,7 +985,7 @@ Metrics/CyclomaticComplexity:
 Metrics/LineLength:
   Description: 'Limit lines to 80 characters.'
   StyleGuide: 'https://github.com/bbatsov/ruby-style-guide#80-character-limits'
-  Enabled: false
+  Enabled: true
```
garethrees: This one might be tricky for views, but lets give it a go!

PR #3390, opened 2016-07-19
[#3382] Add temporary fix for invalid params

garethrees:
Fixes [#3382]

We've had a few examples of spammers submitting invalid params (just
`[:about_me]`, rather than `[:user][:about_me]` causing an undefined
method `[]' for nil:NilClass error.

This just redirects such requests to the user profile page. When we
upgrade to strong params, we can require the `:user` key and (I think)
remove this commit.


lizconlan:
👍

Issue #3137, opened 2016-03-31
Release 0.25


garethrees:
Ticket to track issues and reminders for Release 0.25

crowbot:
Note #3203 and check it before committing to making a release.

lizconlan:
## Checklist
- [x] merge release branch into master
- [x] merge master into develop and wdtk
- [x] check whether alavetelitheme needs updating (IMHO it didn't)
- [x] update and tag whatdotheyknow-theme
- [x] redeploy whatdotheyknow
- [x] merge the outstanding docs PRs
- [x] make "we merged it!" announcement to dev list
- [ ] write coherent blog post, pref with nice header image and feature screenshots

lizconlan:
blog post live!

PR #3391, opened 2016-07-19
[#3386] Convert Fixnum to ActiveSupport::Duration for query

### LINKED ISSUE #3386: Reminders to classify requests being sent three times within an hour
We've a WhatDoTheyKnow user complain about being sent the 

>Was the response you got to your FOI request any good?

chasing email three times. 

I've just experienced it myself today too. In respect of the same request I received a reminder at: 07:34, 07:56, and 08:21

Looking at my emails I can see this happened in relation to a previous request for which I got reminders on the 2nd of June; it doesn't appear to have happened prior to that. 

The latest request in question was that at:
https://www.whatdotheyknow.com/request/chief_constables_council_13_14_j

garethrees:
Fixes #3386 

`InfoRequest.find_old_unclassified` called `#days` on `age_in_days` to
convert it to an `ActiveSupport::Duration` [1]. I think the gist is that
it converts `3` in to `259200` seconds ago, or something like that.

`OLD_AGE_IN_DAYS` already has `#days` applied to it, hence the longer
conditional, rather than calling `#days` in the query interpolation.
Doing so extends the old age:

    InfoRequest::OLD_AGE_IN_DAYS
    # => 1814400
    InfoRequest::OLD_AGE_IN_DAYS.days
    # => 156764160000

This method needs specs, but we don't really have time to do so before making #3137. Here's how I tested that this produces the original query: https://github.com/mysociety/alaveteli/issues/3386#issuecomment-233595718

[1] http://api.rubyonrails.org/v3.2.22/classes/Numeric.html#method-i-days

### DIFF
--- a/app/models/info_request.rb
+++ b/app/models/info_request.rb
@@ -960,7 +960,13 @@ class InfoRequest < ActiveRecord::Base
 
 
   def self.where_old_unclassified(age_in_days=nil)
-    age_in_days ||= OLD_AGE_IN_DAYS
+    age_in_days =
+      if age_in_days
+        age_in_days.days
+      else
+        OLD_AGE_IN_DAYS
+      end
+
     where("awaiting_description = ?
           AND last_public_response_at < ?
           AND url_title != 'holding_pen'


lizconlan:
👍

garethrees:
Merged as 9746c3a16123f95f0ddfdacb75c912de60cd7f74

PR #3392, opened 2016-07-19
Track new requests as Google Analytics virtual pageviews.

### LINKED ISSUE #3360: Make new request goal an event rather than a specific URL
https://github.com/mysociety/alaveteli-experiments/pull/86 suggests that having the goal be a URL which is sharable and visitable by other users is causing false positive goal conversions.

crowbot:
Closes #3360. See also https://github.com/mysociety/whatdotheyknow-theme/pull/329

### DIFF
--- a/app/views/request/show.html.erb
+++ b/app/views/request/show.html.erb
@@ -11,6 +11,13 @@
 <% if flash[:request_sent] %>
   <%= render :partial => 'request_sent',
              :locals => { :info_request => @info_request } %>
+  <% content_for :ga_pageview do %>
+    ga('send', 'pageview', '<%= show_new_request_path(@info_request.url_title) %>');
+  <% end %>
+<% else %>
+  <% content_for :ga_pageview do %>
+    ga('send', 'pageview', '<%= show_request_path(@info_request.url_title) %>');
+  <% end %>
 <% end %>
 
 <% if @info_request.prominence == 'hidden' %>
--- a/app/controllers/request_controller.rb
+++ b/app/controllers/request_controller.rb
@@ -388,7 +388,7 @@ class RequestController < ApplicationController
     end
 
     flash[:request_sent] = true
-    redirect_to show_new_request_path(:url_title => @info_request.url_title)
+    redirect_to show_request_path(:url_title => @info_request.url_title)
   end
 
   # Submitted to the describing state of messages form
--- a/app/views/layouts/default.html.erb
+++ b/app/views/layouts/default.html.erb
@@ -134,7 +134,11 @@
       })(window,document,'script','https://www.google-analytics.com/analytics.js','ga');
 
       ga('create', '<%= AlaveteliConfiguration::ga_code %>', 'auto');
-      ga('send', 'pageview');
+      <% if content_for? :ga_pageview %>
+        <%= content_for :ga_pageview %>
+      <% else %>
+        ga('send', 'pageview');
+      <% end %>
       </script>
     <% end %>
 
--- a/doc/CHANGES.md
+++ b/doc/CHANGES.md
@@ -1,3 +1,7 @@
+# develop
+
+* New requests are now recorded as virtual pageviews in Google Analytics (Louise Crow)
+
 # Version 0.25.0.0
 
 ## Highlighted Features


app/views/request/show.html.erb
```diff
@@ -11,6 +11,10 @@
 <% if flash[:request_sent] %>
   <%= render :partial => 'request_sent',
              :locals => { :info_request => @info_request } %>
+  <% content_for :analytics_events do
+    raw track_analytics_event(AnalyticsEvent::Category::REQUEST_ACTION,
```
lizconlan: It would be good to be know which request the event belongs to so we could capture in as an event label with something like:

```ruby
raw track_analytics_event(AnalyticsEvent::Category::REQUEST_ACTION,
                          AnalyticsEvent::Action::NEW_REQUEST,
                          :label => @info_request.url_title)
```

lizconlan:
Comment about capturing detail inline.

For info - from a quick look at analytics goals, it appears that an Event based goal is a single step rather than a funnel

lizconlan:
At the moment this will still create a duplicate analytics line for `/new` if you navigate to `/new` manually (or click on an old link) as completing a new request and visiting the page sends the same data to Google Analytics.

We can either get around this by altering the virtual pageview to append something like "-event" to the url and edit the funnel to match, or change the view code to something like:

```diff
  <% if flash[:request_sent] %>
    <%= render :partial => 'request_sent',
               :locals => { :info_request => @info_request } %>
    <% content_for :ga_pageview do %>
      ga('send', 'pageview', '<%= show_new_request_path(@info_request.url_title) %>');
    <% end %>
+ <% else %> # log manual hits on /new as standard views
+   <% content_for :ga_pageview do %>
+     ga('send', 'pageview', '<%= show_request_path(@info_request.url_title) %>');
+   <% end %>
  <% end %>
  		  
  <% if @info_request.prominence == 'hidden' %>
```

app/views/request/show.html.erb
```diff
@@ -11,6 +11,13 @@
 <% if flash[:request_sent] %>
   <%= render :partial => 'request_sent',
              :locals => { :info_request => @info_request } %>
+  <% content_for :ga_pageview do %>
+    ga('send', 'pageview', '<%= show_new_request_path(@info_request.url_title) %>');
+  <% end %>
+<% else %> # log manual hits on /new as standard views
```
lizconlan: my bad - that comment is outside the ruby tag /o\

lizconlan:
One last thing - which is entirely my fault - then it's ready to go 👍

equivalentideas:
We recently setup goals on Right To Know using a config exported from WhatDoTheyKnow in March 2016.

We're getting the problem of getting lots of false positive goal completions with the tracking based on the path.

Also, I find the difference between the funnel and the overview pretty confusing :)
screen shot 2016-11-04 at 1 06 36 pm
screen shot 2016-11-04 at 1 06 51 pm

We don't have this commit deployed yet, but this sounds really promising. Have you updated your Goal in GA to use this event? I'm keen to fix this for us, so if you’re getting better results with this solution I’d love to know 😍 

We're tracking Follows as an event this way.
screen shot 2016-11-04 at 1 07 57 pm

I’ve actually [added a few different events to Right To Know](https://github.com/openaustralia/righttoknow/blob/production/assets/javascripts/event_tracking.js). I'd be up for putting in a PR to core with that stuff and expanding it to cover more things we want to track that have the non-distinct path problem (annotations for example).

screen shot 2016-11-04 at 1 14 49 pm

equivalentideas:
>  It just removes the false positives for the page view goal by making sure it's only recorded when a requester is seeing their request for the first time.

Ah cool. I don't understand how that works, but I'll look it up :)

@crowbot What do you think about the idea of extracting the tracking event triggers into a script file, like [event_tracking.js](https://github.com/openaustralia/righttoknow/blob/production/assets/javascripts/event_tracking.js) in Right To Know (That file could definitely be refactored ;-) )? I find this really helpful as you can see everything being tracked in one place and it keeps the view files simpler.

equivalentideas:
>  It just removes the false positives for the page view goal by making sure it's only recorded when a requester is seeing their request for the first time.

@crowbot has this worked? E.g. is the data nice and clean now?

lizconlan:
Hi @equivalentideas, sorry for the delay - I was putting off looking into it as I realised I'd probably fall into playing with analytics for the entire day once I got started!

**TL;DR** - it works! Fiddly details below...

## Recap

Back in May 2016 (to pick a month at random that definitely predates our code changes), our report on WDTK in Behaviour > Site Content > Landing Pages for visitors from a t.co referrer (filtered down to pages matching '/new$') looked like this:

screen shot 2016-12-05 at 23 39 27

With a conversion rate of over 99% but averaging fewer than 2 pages per session with a session duration measurable in seconds, there's something wrong with this picture. Tenative conclusion - we're counting anyone arriving on a request's /new page as a joyous success, even if they immediately close the browser.

And in our Goal Flow, another suspicious thing - particularly noticable from social media referrers (the report just shows t.co and facebook.com) - it's the wrong shape, more people are "completing" the third stage of the process than arrived at the second. A lot more.

screen shot 2016-12-05 at 23 56 28


## Current stats

Last month (November 2016), our the Landing Pages report makes sense again - we've eliminated the single step false positive "conversions" completely.

screen shot 2016-12-05 at 23 57 51

And our Goal Flow looks much more plausible:

screen shot 2016-12-05 at 23 56 51

## The 'secret'

With [this piece of code](https://github.com/mysociety/whatdotheyknow-theme/blob/12609f03dcdb824cb3b1521340b038d37e09015f/lib/views/request/show.html.erb#L11-L21), we're preventing the request-title/new page (aka `show_new_request_path`) from sending a pageview event to GA (it identifies as a standard view instead, which is a better indication of what's really happened) unless we know - from examining the flash message - that a request has been created. It's a bit messy but has the advantage that the Goal didn't need to change as it's still looking for hits on the request-title/new url.

So our Goal description remains:

screen shot 2016-12-06 at 00 19 37

Issue #3394, opened 2016-07-20
Request page: redesign and test social media buttons


wrightmartin:
Part of #3269. (see wireframe for design)

Paging research for options on testing this @jacksonj04

jacksonj04:
My thoughts on testing would be around placement and styling. Unfortunately, Google Analytics Experiments doesn't let us do multivariate testing, so we'd either need to do it one by one, or just have a look on the bit that we're most unsure about.

My gut says the most useful thing to test would be placement of buttons; even a simple slice between the 'info' panel for a request, and at the bottom for when people are done reading would probably be worth doing.

garethrees:
Looks pretty close, but the original is outperforming the new design. Given that we think [social traffic is generally the most engaged](https://github.com/mysociety/alaveteli-experiments/blob/master/experiments/are-some-traffic-sources-better-for-requests.md), I think we should do as much as we can to keep the conversion rate high here and revert back to the original (and maybe try a new variant?).

![screen shot 2016-08-30 at 11 32 32](https://cloud.githubusercontent.com/assets/282788/18085963/7b6c6ae8-6ea5-11e6-8966-4e5c544d02ef.png)

garethrees:
For reference, here's the original:

![screen shot 2016-10-07 at 09 44 55](https://cloud.githubusercontent.com/assets/282788/19183971/c0ac2d60-8c72-11e6-8821-e4c50c227ff5.png)

And this was the variant:

![screen shot 2016-08-05 at 16 24 21](https://cloud.githubusercontent.com/assets/282788/19183980/c769a204-8c72-11e6-8fc3-af712478bbb6.png)

garethrees:
Latest experiment data:

![screen shot 2016-10-07 at 09 46 46](https://cloud.githubusercontent.com/assets/282788/19184026/07133cd0-8c73-11e6-9d6b-386095da2cd7.png)

The variant is down 20% compared to the original with only a 33% chance of outperforming it.

jacksonj04:
Those are much more concrete numbers than we usually get from A/B tests!

PR #3395, opened 2016-07-20
Redesigned 'share this request' 

wrightmartin:
Part of #3394.

this adds a redesigned 'share this request' (previously 'act on what you've learnt') section to the request page.

screen shot 2016-07-20 at 11 16 34

- [x] Decide how we'd like to test this


wrightmartin:
for #3394

app/views/request/_act.html.erb
```diff
@@ -1,46 +1,49 @@
...
+                :onclick => track_analytics_event(
+                  AnalyticsEvent::Category::OUTBOUND,
+                  AnalyticsEvent::Action::FACEBOOK_EXIT,
+                  :label => "Request sidebar") %>
+  </div>
 
-<% if AlaveteliConfiguration::enable_widgets %>
-  <div class="act_link act_link--widget">
-    <%= link_to _("Create a widget for this request"), new_request_widget_path(@info_request) %>
+  <div class="act_link act_link--medium">
```
garethrees: The white Medium icon looks weird on the white background. Is there an alternative we can use?

garethrees:
Assigned to @lizconlan to help with setting up an A/B test of this vs the original

app/views/request/_act.html.erb
wrightmartin: @garethrees there's a black one we use currently, although I've never been happy with that one either. Which do you prefer?

garethrees: Yeah, neither are great. I'll leave it up to you. I just find the white one feels smaller because of the lack of border.

wrightmartin: There is a border but it's very subtle, I'll beef it up

garethrees:
Merged in https://github.com/mysociety/alaveteli/pull/3437

Issue #3269, opened 2016-05-24
Create a more obvious path from someone browsing a request to making their own request


crowbot:
We'd like to experiment to see if this increases our ability to get people browsing the site to make their own requests (https://github.com/mysociety/alaveteli-experiments/issues/33). Our placeholder idea for text for this is 'Make a request like this', but open to other suggestions. In general would be good to think about any revisions we might make to the request page to make it more helpful for someone arriving at the site for the first time. e.g. finding out what the site is, how to use it.

MyfanwyNixon:
Perhaps stating the obvious, but to kick the discussion off:

Where we see:

> [user name] made this Freedom of Information request to [authority name]

we could include a 'what is a Freedom of Information request?' callout or link to the beginner's guide at https://www.mysociety.org/2014/08/15/how-to-make-a-freedom-of-information-request-with-whatdotheyknow/ (without them having to leave the page).

You could also label each part of the correspondence a bit more explicitly: this is the request a member of the public made via this site; this is the automated reply from an authority; here's the actual information they received - not in those exact words, obviously, but a framing device that explains each communication.

People often link directly to the response itself, eg https://www.whatdotheyknow.com/request/331242/response/816738/attach/html/3/Response%20Letter%20001818.pdf.html, which currently includes the header "This is an HTML version of an attachment to the Freedom of Information request 'title'."

That could include a more explicit invitation to 'See this request in context' and also 'What is Freedom of Information?'.

As for your main question, 'make a request like this' seems good and succinct. I suppose another option is including both 'Ask for similar information from a different authority' and 'ask for different information from this authority'.

crowbot:
A related addition would be a more obvious call for browsers to share a request on social media (we currently have a very discrete twitter link in the sidebar).

wrightmartin:
I'm having a very initial look at this now, and my immediate thinking has lead me to look at the request page as a whole - there's a lot to it, and a lot we want people to do here. I've made a quick list and it's already long...

- Read the request
- Understand the FOI process
- Follow this request
- Make a new request
- Share this request 
- Donate (soon)
- Be aware of AskTheEU
- Write to their MP
- Report it if it goes against our guidelines

There's a lot here and maybe we've got a bit of a noise problem. Maybe part of making the path to a new request more obvious is putting it in context of the request page as a whole. What's our priority here and how do we make it the priority for the user?

wrightmartin:
[...]

PR #3397, opened 2016-07-21
Redesigned correspondence section of request page

wrightmartin:
screen shot 2016-07-21 at 11 05 56

Part of #3269

### DIFF
--- a/app/assets/stylesheets/responsive/_request_layout.scss
+++ b/app/assets/stylesheets/responsive/_request_layout.scss
@@ -45,16 +45,41 @@
 
 .correspondence_delivery {
   padding: 1em;
-  margin: 0 -1em;
 
   @include respond-min( $main_menu-mobile_menu_cutoff ){
     padding: 1.2em;
-    margin: 0 -1.2em;
   }
 
   & > :last-child {
     margin-bottom: 0;
   }
+}
+
+.correspondence__header {
+  padding: 0.6em 1.2em;
+}
+
+.correspondence__header__from--with-delivery-status {
+  @include respond-min( $main_menu-mobile_menu_cutoff ){
+    margin-right: 8em;
+  }
+}
+
+.correspondence__header__delivery-status {
+  @include respond-min( $main_menu-mobile_menu_cutoff ){
+    position: absolute;
+    right: 1.2em;
+    top: 0.6em;
+    max-width: 8em;
+  }
+}
+
+.correspondence_text {
+  padding: 1.2em;
+}
+
+.event_actions {
+  padding: 1.2em;
 }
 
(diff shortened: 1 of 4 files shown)


garethrees:
Some quick notes (all based on the screenshot, so ignore if they are handled)

**Questions**

- What happens with [really long bodies](https://www.whatdotheyknow.com/request/transport_management_complaints#incoming-842186)?
- If we can, we [show the From name](https://www.whatdotheyknow.com/request/proposed_housing_development_and#incoming-812825) as well as the authority name in case different FOI officers reply. How are we handling this?

**Suggestions**

- Could we make the dates clickable to jump to the correspondence (like GitHub threads)?
- The user names could be links to the user profile
- I'd make the dates slightly smaller and/or greyer

**Things to check**

- Note to self to check how it looks when logged in as an admin
- Note to self to check how it looks when a request has attachments
- Note to self to check how it looks when you click the delivery status

wrightmartin:
> What happens with really long bodies?

screen shot 2016-07-22 at 10 05 34

> If we can, we show the From name as well as the authority name in case different FOI officers reply. How are we handling this?

I haven't changed how that works, it might not be obvious from the screenshot but it's there as 'The Minister Ministry of Silly Walks'. Maybe I should add a comma?

wrightmartin:
> Could we make the dates clickable to jump to the correspondence (like GitHub threads)? 

yep, that's a great idea

> The user names could be links to the user profile

That was my intention, I couldn't work out how to do that though as it's using ```<%= incoming_message.safe_mail_from %>``` rather than the typical ```user_profile``` style link

> I'd make the dates slightly smaller and/or greyer

Yeah I agree

garethrees:
> > What happens with really long bodies?
>
> I haven't changed how that works, it might not be obvious from the screenshot but it's there as 'The Minister Ministry of Silly Walks'. Maybe I should add a comma?

You'd actually get "Loughborough University Freedom of Information, Loughborough University" (with the added comma mentioned above.

Also, in the case of outgoing correspondence we need to take the delivery status in to account - some users have [really long names](https://www.whatdotheyknow.com/user/tom_of_the_commonly_known_talent/profile).

wrightmartin:
screen shot 2016-07-22 at 10 33 08

garethrees:
@wrightmartin looks good. Made a couple of minor fixes/tweaks.

What are your thoughts on e766a98?

1 thing left – delivery status drop down overflows the correspondence box:

screen shot 2016-07-25 at 12 24 08

With the `box` class re-applied, the delivery status drop down fits in, but breaks the header:

screen shot 2016-07-25 at 12 26 03

PR #3399, opened 2016-07-21
replaced the pop-up driven 'link to this widget'

wrightmartin:
Part of #3269

Now uses a click to copy style method in the footer of each correspondence element

The form uses flexbox for the layout, but includes no fallback layout for browsers that don't support it. Instead the form is slightly more vertical. 

**TODO**
- [ ] JS implementation of the copy to clipboard functionality

screen shot 2016-07-21 at 17 10 12

### DIFF
--- a/app/assets/javascripts/general.js
+++ b/app/assets/javascripts/general.js
@@ -27,21 +27,12 @@ $(document).ready(function() {
    return false;
    });
 
-  // "link to this" widget
-  $('a.link_to_this').click(function() {
-    var box = $('div#link_box');
-    var location = window.location.protocol + "//" + window.location.hostname + $(this).attr('href');
-    box.width(location.length + " em");
-    box.find('input').val(location).attr('size', location.length + " em");
-    box.show();
-    box.position({
-      my: "right center",
-      at: "left bottom",
-      of:  this,
-      collision: "fit" });
-    box.find('input').select();
-    return false;
+  // "link to this" box
+  $('.cplink__button').click(function() {
+    var box = $(this).prev('.cplink__field');
+    box.select();
   });
+
 
      $('.close-button').click(function() { $(this).parent().hide() });
      $('div#variety-filter a').each(function() {
--- a/app/assets/stylesheets/responsive/_popups_layout.scss
+++ b/app/assets/stylesheets/responsive/_popups_layout.scss
@@ -21,9 +21,3 @@
 #other-country-notice{
   display:none;
 }
-
-#link_box {
-  position:absolute;
-  z-index:999;
-  display:none;
-}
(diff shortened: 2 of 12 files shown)


garethrees:
> Instead the form is slightly more vertical.

Got a screenshot of that?

wrightmartin:
part of #3269

wrightmartin:
> Got a screenshot of that?

IE8 
screen shot 2016-07-22 at 09 26 40

IE9
screen shot 2016-07-22 at 09 31 55

IE10
screen shot 2016-07-22 at 09 47 12

garethrees:
Looks good!

Just some ideas on reducing interface text:

screen shot 2016-07-25 at 12 37 10

screen shot 2016-07-25 at 12 37 22

Any thoughts?

wrightmartin:
> Just some ideas on reducing interface text:

On the first example the button doesn't describe the behaviour of the button - it might be confusing, especially as copying to clipboard is an invisible process, it might not be obvious that's whats happened - unless we added something in the UI after you click the button

The second example is better, but there is no call to action - but that might not matter? I'm on the fence does @zarino have a second opinion?

garethrees:
Happy to go with whatever you decide on this one.

zarino:
@gareth @wrightmartin: I assume @gareth is looking to re-use an existing string from elsewhere in the codebase, to avoid the need for a dozen new translations of "Copy URL"?

That said – A button with "copy" in it somewhere does seem more logical than one that says something like "link to this".

My suggestion for the colour of the bikeshed would be to use "link" rather than URL, because only geeks know what a URL is. So the button would read: "Copy link"

zarino:
I assume you've already thought about this, but – do we even *need* a "copy" button? Why not just have a bit of javascript that selects all the text in the `<input>` when you click into it? Avoids the hassle over loading a nasty Flash clipboard plugin.

wrightmartin:
> I assume you've already thought about this, but – do we even need a "copy" button? Why not just have a bit of javascript that selects all the text in the <input> when you click into it? Avoids the hassle over loading a nasty Flash clipboard plugin.

Worth thinking about. The goal here was the make sharing easier, I've seen this pattern used all over the place so re-appropiated it for here. @zarino's solution pushes one step back to the user, however I don't know if that's a bad thing or not yet...

garethrees:
> I assume @gareth is looking to re-use an existing string from elsewhere in the codebase, to avoid the need for a dozen new translations of "Copy URL"?

Wasn't thinking about translations (we do have "Link to this" translated) – I just thought less text =~ better, if we can do it in a way that makes sense.

garethrees:
Here's what I propose we do (872fd0d) in the interests of shipping an improvement quickly:

link-to-this

- Don't bother with JS/Flash copy. We don't have it now so we're not losing anything.
- Use "Link to this" as the button text. It seems like the best wording to convey the intent of the button. The automated selection of the text should be a good enough indicator that the user then needs to do something with that text (i.e. copy it, as they did before)

zarino:
👍 👍 👍

PR #3402, opened 2016-07-25
Request call to action

wrightmartin:
This adds a request CTA to the sidebar of the request page. 

I've reused to frontpage partial to do this, in the process I renamed it, as the 'frontpage' part of the filename no  longer applies"


wrightmartin:
Part of #3269

garethrees:
Looks good – probably want to wait until after the "donation drive" has finished before we deploy this.

garethrees:
Plan is to get this merged and edit out in WDTK until donate stuff has finished

PR #3403, opened 2016-07-25
Redesigned sidebar list of requests

wrightmartin:
This makes the similar requests sidebar on the request page a lot neater, but removing some of the information presented 

Part of #3269

### DIFF
--- a/app/assets/stylesheets/responsive/_lists_style.scss
+++ b/app/assets/stylesheets/responsive/_lists_style.scss
@@ -33,11 +33,7 @@
 }
 
 .request_short_listing {
-  margin-top:1em;
-  h3 {
-    font-size: 1.3em;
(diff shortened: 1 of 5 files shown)


garethrees:
Much nicer. This also changes the results in the "related requests" results when someone is making a new request:

screen shot 2016-07-25 at 12 48 56

I don't think the extra context is particularly valuable there, other than maybe the date?

wrightmartin:
@garethrees yeah the date would be useful there - is there a way in the template to only display it in the typeahead search?

garethrees:
> @garethrees yeah the date would be useful there - is there a way in the template to only display it in the typeahead search?

Made it optional via 40b2509 – may or may not need to add a class to it.

PR #3415, opened 2016-07-29
Destroy raw email record when destroying incoming message.

### LINKED ISSUE #3393: Destroy raw email when destroying associated incoming message
Helps moving from staging to production (when you want to clear out test data).

`IncomingMessage` actually [destroys](https://github.com/mysociety/alaveteli/blob/master/app/models/incoming_message.rb#L64) the raw email file on disk, but doesn't destroy the `RawEmail` database record. An `IncomingMessage` shouldn't be responsible for the `RawEmail` internals.

An `InfoRequest`'s incoming messages are correctly destroyed through the [association](https://github.com/mysociety/alaveteli/blob/master/app/models/info_request.rb#L72), so if an `InfoRequest` is deleted by an admin, everything important in the chain is deleted.

crowbot:
Fixes #3393.

### DIFF
--- a/app/models/incoming_message.rb
+++ b/app/models/incoming_message.rb
@@ -55,7 +55,7 @@ class IncomingMessage < ActiveRecord::Base
   # never really has many info_request_events, but could in theory
   has_many :info_request_events, :dependent => :destroy
 
-  belongs_to :raw_email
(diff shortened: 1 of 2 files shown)


garethrees:
👍

PR #3416, opened 2016-07-29
Destroy associations when destroying user.

### LINKED ISSUE #3407: Destroy related TrackThings when a User is destroyed
In a similar vein to https://github.com/mysociety/alaveteli/issues/3393, when you try to destroy a user with associated track_things you get an IntegrityError. Is there any reason this relationship couldn't be `dependent: destroy`?

crowbot:
Note that destroying users is not available from the admin interface,
so the expected use case is from the command line. Fixes #3407.


lizconlan:
👍

PR #3421, opened 2016-08-02
Don't include original mail in bounce messages

### LINKED ISSUE #2097: Don't include original email with undeliverable notifications
When we send a `Your response to an FOI request was not delivered` message, we include the original email as an attachment.  This could enable spammers to use us as a reflector for malicious attachments by forging the `From:` address.

crowbot:
This could be a vector for distributing spam via backscatter.

Closes #2097


lizconlan:
👍

PR #3442, opened 2016-08-05
2019 trusty install scripts

crowbot:


lizconlan:
yay, it works \o/

But as with jessie, the motd behaviour has changed (but, annoyingly, in a different way) - now, apparently we have to [edit the actual `/etc/motd` file](http://askubuntu.com/questions/538732/no-motd-on-ubuntu-14-04). Although getting the standard motd message **and** the custom one is quite nice.

Suggest changing the Vagrantfile (around line 146) as follows:

```diff
   config.vm.provision :shell, :inline => "ln -s /var/run/motd /etc/motd"
+  elsif ALAVETELI_OS == 'trusty64'
+    config.vm.provision :shell, :inline => "echo '#{ motd }' >> /etc/motd"
  end

lizconlan:
👍 otherwise

garethrees:
Looks like this is ready to merge. GitHub currently showing two WIP commits. First is unrelated to this PR so should be dropped. Second is the adding of the Trusty vagrant VM. I think the Trusty related commits can be squashed together.

PR #3443, opened 2016-08-05
Prevent escaping of plain text email

### LINKED ISSUE #1400: Incorrect entity escaping in plaintext notification emails
Some characters from request titles (and perhaps bodies) appear to be being HTML entity escaped in outgoing plaintext notification emails. For example a user reported that an apostrophe was showing up as "&#x27;".

[Apologies as usual if this is a duplicate - I vaguely recall seeing something about this before but I've searched and can't find any relevant issue]

lizconlan:
Closes #1400

### DIFF
--- a/app/mailers/contact_mailer.rb
+++ b/app/mailers/contact_mailer.rb
@@ -69,7 +69,8 @@ class ContactMailer < ApplicationMailer
                   ),
          :to => contact_from_name_and_email,
          :subject => _('Add authority - {{public_body_name}}',
-                       :public_body_name => @change_request.get_public_body_name))
+                       :public_body_name => @change_request.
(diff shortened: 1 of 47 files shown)


lizconlan:
rails/rails#8235 turned out not to be relevant because the problem is that our call to the translation library is what's causing the entity escaping unless we flag the text as a safe string. The template handler is working as expected then the dynamic text we're passing in is being substituted.

Will split the commit and resubmit the changes.

lizconlan:
(Code the same at this point, have just added the extended commit message at d97c67d)

lizconlan:
I've marked `display_status` as safe as well as it's overridable in themes so it might be affected

PR #3445, opened 2016-08-08
Add more clearly differentiated status icons from Alavetelitheme.

crowbot:


lizconlan:
👍 yay, shiny new icons 🏆

lizconlan:
once this is done, we should be able to remove the iconset from asktheeu-theme (and possibly others)

PR #3450, opened 2016-08-09
Only populate message attributes from non-nil params.

### LINKED ISSUE #3448: Default text for new request has disappeared
No template text is appearing on `https://www.whatdotheyknow.com/new/[body name]`

crowbot:
Otherwise we'll override default values set in initialising the
object. Closes #3448


lizconlan:
looks great 👍 🎉

Issue #3293, opened 2016-06-02
Hide parts of 'things to do with this request' when they aren't relevant


wrightmartin:
Eg. this section should only be visible to the (logged in) creator
![group](https://cloud.githubusercontent.com/assets/2292925/15749708/41747c14-28db-11e6-9e80-9882331c01aa.png)

garethrees:
I assume this was a design decision to help (less computer-literate) request owners understand what they can do next even if they're not logged in.

crowbot:
Yeah, and also people from authorities I think. Maybe there's a more
graceful way of doing that though.

wrightmartin:
Other sites handle this with a less verbose 'are the the owner of this x, log in to see your options'

lizconlan:
I suspect the big slab of options/text can be a bit overwhelming to casual visitors.

Issue #3336, opened 2016-06-20
Plan how to research, redesign and test a redesign of the request page


wrightmartin:
related to #3269

wrightmartin:
# Notes from our call with the team to discuss this

## Suggested approach

1. Split the page into elements and set priority
2. Drop the low priority items completely
3. Design a new page
4. Decide if we implement the new design all at once, or piece meal
5. AB test the new page on a live site to determine success
6. Repeat 3-6 until we're happy

### Priority items
Non-exhaustive

#### Low priority things

- Write to your MP
- Donate
- AskTheEU
- Share/tweet/other social

####  High priority things

- Read the request
- Make a new request

#### Medium priority things

- Follow
- Report a request
- Similar request

####  Logged in actions

- Only show these to logged in users

### Testing
Use WDTK, or find a partner site which has enough usage to run a valid test

garethrees:
Is this *the plan*? Can we consider this ticket done and continue tracking work on this in https://github.com/mysociety/alaveteli/issues/3269?

wrightmartin:
> Is this the plan? Can we consider this ticket done and continue tracking work on this in #3269?

If we all agree, then yep!

garethrees:
Worth noting that we've already made some of these decisions and started [discussing priority](https://github.com/mysociety/alaveteli/issues/3269#issuecomment-233596276).

> Split the page into elements and set priority

https://github.com/mysociety/alaveteli/issues/3269#issuecomment-233596276

> Design a new page

https://github.com/mysociety/alaveteli/issues/3269#issuecomment-233310573

> Decide if we implement the new design all at once, or piece meal
> AB test the new page on a live site to determine success

Not all at once. Probably a small element at a time in quick succession, with A/B tests on contentious items

garethrees:
I think given how much progress has been made already that reviewing this is kind of redundant. Would be better for @lizconlan to jump in on individual tickets (e.g. https://github.com/mysociety/alaveteli/issues/3394 / https://github.com/mysociety/alaveteli/pull/3395)

PR #3451, opened 2016-08-10
Add action menus for request actions

wrightmartin:
Work on https://github.com/mysociety/alaveteli/issues/3269
Fixes https://github.com/mysociety/alaveteli/issues/3293

This replaces the majority of the 'lesser used' actions on the request page with an action menu
### todo
- ~~Hide the role-based action submenus based on permissions~~ Ticketed in #3474


wrightmartin:
Here's what it looks like now...


screen shot 2016-08-12 at 14 35 04

garethrees:
Breaks with user avatar:

screen shot 2016-08-16 at 17 50 39

garethrees:
The actions hover/focus state is pretty subtle – especially when tabbing to it.

garethrees:
Looks a bit messy while the JS loads. Anything we can do about this?

screen shot 2016-08-16 at 17 54 35

What are we going to do for a non-JS fallback?

app/assets/javascripts/general.js
```diff
@@ -104,3 +104,105 @@ $(document).ready(function() {
...
+/*
+ * Dropit v1.1.0
+ * http://dev7studios.com/dropit
+ *
+ * Copyright 2012, Dev7studios
+ * Free to use and abuse under the MIT license.
+ * http://www.opensource.org/licenses/mit-license.php
+ */
+
+;(function($) {
```
garethrees: Alaveteli uses two-space indentation

wrightmartin:
Thanks for the feedback @garethrees 

The latest commits should address it all

app/assets/javascripts/application.js
```diff
@@ -4,6 +4,5 @@
 //= require jquery-ui/datepicker
 //= require jquery-ui/position
 //= require jquery.cookie
-//= require foundation
```
garethrees: This only got added in the previous commit. How come we're removing it here?

app/views/layouts/default.html.erb
```diff
@@ -136,9 +136,6 @@
     <a href="#content" class="show-with-keyboard-focus-only skip-to-link">
       <%= _('Back to content') %>
     </a>
-    <script>
-      $(document).foundation();
```
garethrees: Where do we load foundation now? Is this removal related to the removal of foundation in https://github.com/mysociety/alaveteli/pull/3451/commits/e0eb7e34760b1c4fa46635f1fd83f8b4f2b5264f#r77012948?

app/assets/stylesheets/responsive/_global_style.scss
```diff
@@ -257,9 +257,12 @@ div.pagination {
   &:active,
   &:focus {
     background-color: white;
-    border: 1px solid #ddd;
+    border: 1px solid #bbb;
```
garethrees: I'd say this could be even more obvious:

**Actions focused**

screen shot 2016-08-31 at 16 21 10

**Follow focused**

screen shot 2016-08-31 at 16 21 14

app/views/request/show.html.erb
```diff
@@ -76,65 +76,76 @@
...
-          <%= n_("We're waiting for someone to read a recent response and " \
-                    "update the status accordingly. Perhaps " \
-                    "<strong>you</strong> might like to help out by doing that?",
-                 "We're waiting for someone to read recent responses and " \
-                    "update the status accordingly. Perhaps " \
-                    "<strong>you</strong> might like to help out by doing that?",
+<div class="request-header">
+  <div class="row">
+    <div class="request-status">
+      <p id="request_status" class="request_icon_line icon_<%= @info_request.calculate_status %>">
```
garethrees: Note to self: Use `@status` instead of `@info_request.calculate_status`

app/views/track/_tracking_links_simple.html.erb
```diff
@@ -6,10 +6,10 @@
 
 <% if existing_track %>
   <div class="feed_link feed_link_<%= location %>">
-    <%= link_to _("&ndash; Unfollow"), { :controller => 'track', :action => 'update', :track_id => existing_track.id, :track_medium => "delete", :r => request.fullpath }, :class => "link_button_green track-request-action" %>
+    <%= link_to _("Unfollow"), { :controller => 'track', :action => 'update', :track_id => existing_track.id, :track_medium => "delete", :r => request.fullpath }, :class => "link_button_green track-request-action" %>
   </div>
 <% elsif track_thing %>
   <div class="feed_link feed_link_<%= location %>">
-      <%= link_to _("&plus; Follow"), do_track_path(track_thing), :class => "link_button_green track-request-action" %>
+    <%= link_to _("Follow"), do_track_path(track_thing), :class => "link_button_green track-request-action" %>
```
garethrees: I've removed the +/- because I think they should be CSS icons. It doesn't make sense for `+` and `-` to be in the translation strings.

app/views/request/_sidebar.html.erb
```diff
@@ -2,17 +2,6 @@
...
   </div>
-  <div id="track-request" class="track-request">
-    <h2><%= _('Follow this request') %></h2>
-
-    <p>
-      <%= n_("There is {{count}} person following this request",
-             "There are {{count}} people following this request",
-             @follower_count,
-             :count => @follower_count) %>
-    </p>
```
garethrees: Are we losing "social proof" enticement by dropping this?

app/views/request/_sidebar.html.erb
```diff
@@ -15,44 +15,25 @@
     <%= render :partial => 'track/rss_feed', :locals => { :track_thing => @track_thing, :location => 'sidebar' } %>
   </div>
 
-  <% if @info_request.described_state != "attention_requested" %>
```
garethrees:
```irc
[11:56:12]  <gareth>	 hmm looks like a few cases have been lost in https://github.com/mysociety/alaveteli/pull/3451/commits/6ce397b69ce4a24a4a3d7f48147e9c80b4b71d74
[11:56:18]  <gareth>	 was that intentional?
[12:19:56]  <Martin>	 ah I remember these,
[12:20:28]  <Martin>	 hmm, they're supposed to be there still - I think I thought about moving them when I moved the report button
```

app/assets/stylesheets/responsive/_sidebar_layout.scss
```diff
@@ -22,6 +22,12 @@
   }
 }
 
+.new-request-cta {
```
garethrees: Doesn't look related to this commit

app/views/request/show.html.erb
```diff
@@ -138,7 +138,10 @@
             <%= _('to {{public_body}}',:public_body=>public_body_link(@info_request.public_body)) %>
           <% end %>
         </p>
-        <%= render :partial => 'after_actions' %>
+        <div class="request-header__toolbar-actions">
+          <%= render :partial => 'after_actions' %>
+          <%= render :partial => 'track/tracking_links_simple', :locals => { :track_thing => @track_thing, :own_request => @info_request.user && @info_request.user == @user, :location => 'toolbar ' } %>
```
garethrees: Not sure why we're passing `:own_request` given its not used in the partial

app/views/track/_tracking_links_simple.html.erb
```diff
@@ -0,0 +1,15 @@
+<%
+  existing_track = local_assigns.fetch(:existing_track) do
+    TrackThing.find_existing(@user, track_thing) if @user
+  end
+%>
+
+<% if existing_track %>
```
garethrees: Just get an empty space when its your own request:

screen shot 2016-09-02 at 12 37 11

app/views/request/_sidebar.html.erb
garethrees: I think we need to reconsider what we do with this section. It's already a bit weird:

<table>
  <tr>
    <th>Normal State</th>
    <th>Reported State</th>
  </tr>
  <tr>
    <td>
      
     </td>
      <td>
        
      </td>
    </tr>
</table>

There are a few different states depending on whether the request is hidden to some degree.

Do these make sense when they're detached from the "report" button?

app/views/track/_tracking_links_simple.html.erb
wrightmartin: Is that because it only hides the button and not the container too? If it hid ```  <div class="feed_link feed_link_<%= location %>">``` as well it would collapse in, it would be a bit tricky to make it collapse in any other way

app/assets/stylesheets/responsive/_sidebar_layout.scss
wrightmartin: it is - becuase moving the follow and report boxes to the action menu made the new-request call to action have way too much space above it

app/views/request/_sidebar.html.erb
wrightmartin: We are yeah. As we discussed in the original plan (https://github.com/mysociety/alaveteli/issues/3336#issuecomment-228323347) 'follow' isn't a high priority for us on this page.

We could look into a small label maybe in the action menu that shows how many are following?

app/views/track/_tracking_links_simple.html.erb
garethrees: The [current version](https://github.com/mysociety/alaveteli/blob/develop/app/views/track/_tracking_links.html.erb#L7-L10) accounts for the request being your own. I guess we'd need to do something similar.

app/assets/stylesheets/responsive/_sidebar_layout.scss
garethrees: Ah great

app/views/request/_sidebar.html.erb
garethrees: Okay lets just drop it if we're thinking its not super vital.

garethrees:
https://github.com/mysociety/alaveteli/pull/3451#discussion-diff-77332041 discussion got lost but needs resolving before merge.

app/views/request/_sidebar.html.erb
wrightmartin: I think this can be removed - we could add a more info link next to the report link in the action menu, and this sort of thing can be in the help docs

garethrees: Yeah that's a good point. Really it should be on `/request/:url_title/report/new` where you actually start entering your reasons.

garethrees:
screen shot 2016-09-05 at 09 52 47

Anything we can do with the styles in alaveteli-core to make the action menu work nicely without needing to add additional styles to all themes?

wrightmartin:
> Anything we can do with the styles in alaveteli-core to make the action menu work nicely without needing to add additional styles to all themes?

Tricky this one...

We can add ```class="button"``` to the action menu button and it should use some styles, but as it's an entirely new design element the whole action bar will look unstyled.

I guess we could make it even more unstyled, so that it doesn't have the off brand background colour, and maybe remove some of the button styles etc?

garethrees:
> I guess we could make it even more unstyled, so that it doesn't have the off brand background colour

Yeah, I think that would be a good call. Or maybe a `rgba(0,0,0,0.2)` border?

screen shot 2016-09-05 at 10 30 53

> We can add `class="button"`

Could try that – I guess I'm just keen we do as much as we can to make upgrades to the themes less effort. Your call on this.

garethrees:
Also need to update the PDF styles after the move to the action bar:

screen shot 2016-09-05 at 12 49 38

app/views/request/_sidebar.html.erb
garethrees: Just for reference, here's what we do for people following authorities:

screen shot 2016-09-12 at 11 24 37

wrightmartin: We could reuse that here, space is a bit more limited but I don't think it will be a huge issue

wrightmartin:
^uuuhh, hmmm

wrightmartin:
@garethrees this is now ready for you to take another look

lizconlan:
Have had a quick look at the 2 spec failures...

`./spec/controllers/request_controller_spec.rb:192`

Looks like a fairly simple case of the functionality (and therefore the expected text) having moved from `app/views/request/_sidebar.html.erb` to `app/views/reports/new.html.erb` so that should be a quick fix.

--

`./spec/controllers/request_controller_spec.rb:212`

Could be a bit trickier as it looks as though the expected string "This request has been marked for review by the site administrators, who have not hidden it at this time." has been removed altogether. If this is intentional, we can fix it by removing the spec - needs a closer look.

app/assets/stylesheets/responsive/_global_style.scss
```diff
@@ -319,14 +319,29 @@ div.pagination {
...
+  .action-menu__info-link {
     a {
-      text-decoration: none;
-      color: #333;
-      &:hover,
-      &:active,
-      &:focus {
-        text-decoration: underline;
-      }
+      font-size: 0.625em; //10px
```
lizconlan: this makes the "Help" text uncomfortably small for me 😢 

screen shot 2016-10-18 at 15 07 06

Is that what 10px looks like? 🔍

app/views/request/_after_actions.html.erb
```diff
@@ -49,7 +49,7 @@
         <li class="action-menu__menu__item">
           <ul class="action-menu__menu__submenu anyone_actions">
             <li>
-              <%= link_to _("Report this request"), new_request_report_path(:request_id => @info_request.url_title) %>
+              <%= link_to _("Report this request"), new_request_report_path(:request_id => @info_request.url_title) %><span class="action-menu__info-link"><%= link_to _("Help"), help_about_path %></span>
```
lizconlan: I'm not sure where this should go?

`<%= link_to _("Help"), help_about_path(:anchor => "focused") %>` maybe?

wrightmartin:
> The system doesn't allowed the same request to be reported twice - so as it stands the user gets a confusing door slam after pressing the "Report" button as there's no visual indication that it's already been vetted

In the interest of simplifying the UI: If we don't allow a second report, rather than exposing this edge-case of our logic to the front-end, we should hide the report button, or show the report button in a disabled state with a link to find out why

garethrees:
> we should hide the report button, or show the report button in a disabled state with a link to find out why

+1 to explaining that the request has already been reported

app/views/request/_after_actions.html.erb
```diff
@@ -48,9 +48,17 @@
...
         <li class="action-menu__menu__item">
           <ul class="action-menu__menu__submenu anyone_actions">
-            <li>
-              <%= link_to _("Report this request"), new_request_report_path(:request_id => @info_request.url_title) %><span class="action-menu__info-link"><%= link_to _("Help"), help_about_path %></span>
-            </li>
+
+            <% if @info_request.described_state != "attention_requested" && @info_request.attention_requested %>
+              <%# If the request has already been reported and admins chose not to hide it %>
+              <li>
+                <%=  _("Report this request") %><span class="action-menu__info-link"><a href="#">Unavailable</a></span>
```
wrightmartin: This currently links nowhere - do we need a help page or something ready to go?

lizconlan:
ToDo:

 - hide report link in the dropdown if the request has already been reported
 - restore the old sidebar text where the request was previously reported but cleared by an admin
 - fix failing test (actually that second bit might do that automatically)
 - put it back up for review to check I've got this right

lizconlan:
Aha - the trick here is that a request has both an `attention_requested` property and a `described_state` of "attention_requested". The obvious way of clearing a reported request is to change the latter, in which case `attention_requested` remains set so a check for a mismatch between the 2 reveals that it was previously reported but an admin chose not hide it.

I'm mostly writing this down because I'm tired. Also so that people can argue with me if I've got this wrong. (I expect a certain amount of overlap between those 2 sentences.)

lizconlan:
Ok, so now we have...

## A standard request which can be reported

### It has a "Report this request" menu option

screen shot 2016-11-23 at 20 22 38

## A request which has already been reported

### It shows a message at the top of the page to say it's been reported

screen shot 2016-11-23 at 20 25 00

### It disables the "Report this request" menu option

screen shot 2016-11-23 at 20 29 16

## A request which was reported then an admin chose not to hide it

### It disables the "Report this request" menu option *and* displays additional sidebar text

screen shot 2016-11-23 at 20 31 57

## TBC

* The "Help" and "Unavailable" links that appear next to the "Report this request" menu text currently don't have anything to link to (should be new help page sections?)
* Personally I find the text for those links a bit too small ("yeah, that's just you" is fine but I think this got lost previously - or I missed the reply)

lizconlan:
Suggested html for the about page (looking at WDTK, I would place this between "#useful" and "#how_many".

```rb
<dl>
    <dt id="reporting">What if I see a request I feel to be inappropriate? <a href="#reporting">#</a></dt>
    <dd>
      <p>
        Requests for personal information and vexatious requests are not considered valid for FOI purposes (read more).
      </p>

      <p>
        If you believe a request is not suitable, you can report it for attention by the site administrators.
      </p>
    </dd>

    <dt id="reporting_unavailable">Why won't it let me report some requests? <a href="#reporting_unavailable">#</a></dt>
    <dd>
      <p>
        If a request has already been reported to the site administrators, you
        can't report it a second time - this is to prevent the administrators
        being notified multiple times about the same issue before they've had
        a chance to conduct a review.
      </p>

      <p>
        Where a request has been previously reported but an administrator has
        chosen not to hide it from public view, you can use the form in the
        sidebar of the request page to contact the administrators if you still
        think it should be hidden.
      </p>
    </dd>
  </dl>
```

Not added to the view template as we have changed the templates in core to "Placeholder to be overriden by theme"

garethrees:
The "report this request" stuff is fine (if a little confusing, but c'est la vie).

We appear to have lost some of the styling on the buttons in the action bar.

As a reminder:

> Here's what it looks like now...
>
> screen shot 2016-08-12 at 14 35 04
>
> – https://github.com/mysociety/alaveteli/pull/3451#issuecomment-239447673

And here's what we've got:

screen shot 2016-11-29 at 12 06 44

Looks even worse when you apply alavetelitheme:

screen shot 2016-11-29 at 12 04 29

I've already raised bug reports for the [sidebar](https://github.com/mysociety/alavetelitheme/issues/84) and [status](https://github.com/mysociety/alavetelitheme/issues/83).

Here's what I think needs to get done:

- [ ] Make an inline-list container for items on the right of the action menu
- [ ] Pull the "actions" button closer to the "follow" button as we had before (by putting it in the list just mentioned)
- [ ] [Improve](https://github.com/mysociety/alaveteli/pull/3451#r78348135) the default styling of the follower count
- [ ] Make the "follow" button use the primary button colour when a theme is applied
- [ ] Make the "actions" button use the secondary button colour when a theme is applied

wrightmartin:
> We appear to have lost some of the styling on the buttons in the action bar.

I pulled all the styling right back to basics as I couldn't find a default that worked with any of the themes I was testing with without lots of extra work. 

We don't even include button styles in the core alaveteli theme, so making them look buttony as we did previously went against the convention (see login screen)

screen shot 2016-11-29 at 14 53 16

--- 

I'm not quite sure what broke the status bar and the sidebar - maybe a tricky merge or something? Should be easy to fix though

wrightmartin:
^ just checked my local copy on the latest develop and it has the status bug - it looks like it came from a rebase so not as a result of this work

garethrees:
> ^ just checked my local copy on the latest develop and it has the status bug - it looks like it came from a rebase so not as a result of this work

Sorry yeah they're from other tickets that have already been merged. I was just pointing out that I don't think we need to fix those two things here.

garethrees:
> We don't even include button styles in the core alaveteli theme, so making them look buttony as we did previously went against the convention (see login screen)

Hmm, ISWYM. I still think that the positioning & padding should be correct so that when you look at the whitelabel theme it looks okay.

Sounds like we'll then need to make a new PR against the themes to add the button styling, e.g:

```scss
.action-bar__follow-button {
  .link_button_green {
    @include button-secondary();
  }
}
```

wrightmartin:
Ok I think I've nailed it with the latest changes

**Core**
screen shot 2016-12-01 at 12 29 39

**Theme**
screen shot 2016-12-01 at 12 15 13


**Theme with styling**
screen shot 2016-12-01 at 12 26 51

garethrees:
Looks great! I'll have a quick check and then we can get this merged 🎉 🍺

garethrees:
Oh, very minor comment – I think follow should be green and actions grey? I assume this is just a tweak to the PR we create for the theme?

wrightmartin:
> Oh, very minor comment – I think follow should be green and actions grey? I assume this is just a tweak to the PR we create for the theme?

Yeah just a quick example implementation - could be either way, or different

garethrees:
Ready to go pending build passing

RichardTaylor:
Is this a good place for feedback on the new action menu?

One of WhatDoTheyKnow's most prolific annotators got in touch with us concerned they'd been prevented from adding annotations: "I can't add it as an annotation. System won't let me." ... "am I banned from WDTK or something". This was due to the relevant link moving to the action menu. 

My own view is it's odd / surprising that the action options are now only at the top of the request meaning you've got to scroll back up for the links for things to do next after reading to the bottom of the thread.  Might the new menu be usefully duplicated at the bottom of of the correspondence too?

RichardTaylor:
Further user feedback, apparently from a user who hasn't spotted the new action menu:

>It would appear that parts of this application have been disabled, and it is now impossible to make replies of any kind in reference to any of my requests.

PR #3462, opened 2016-08-18
Remove intro sentence on view public body email page

henare:
You don't need to enter words with recaptcha any more so the sentence
was wrong. Since it's pretty self-explanatory now I've just removed the
sentence altogether.


garethrees:
Cheers!

PR #3467, opened 2016-08-26
Add special cases for Trusty where required.

crowbot:


docs/installing/manual_install.md
```diff
@@ -636,14 +642,28 @@ Create a directory for optional Alaveteli configuration
 Copy the example VirtualHost configuration file. You will need to change all
 occurrences of `www.example.com` to your URL
 
+For Wheezy and Precise:
```
garethrees: We've used bold for OS-specific instructions – I think we should continue to do so.

garethrees:
1 minor comment; 👍  when fixed

PR #3476, opened 2016-09-07
[#3335] Cache message delivery status

lizconlan:
Part of #3335

### DIFF
--- a/app/models/mail_server_log.rb
+++ b/app/models/mail_server_log.rb
@@ -23,6 +23,8 @@ class MailServerLog < ActiveRecord::Base
 
   serialize :delivery_status, DeliveryStatusSerializer
 
+  before_create :calculate_delivery_status
+
   # Load in exim or postfix log file from disk, or update if we already have it
   # Assumes files are named with date, rather than cyclically.
   # Doesn't do anything if file hasn't been modified since it was last loaded.
@@ -218,10 +220,24 @@ class MailServerLog < ActiveRecord::Base
   end
 
   def delivery_status
-    @delivery_status ||= line(:decorate => true).delivery_status
+    # read in the status from the database if it's available
+    if attributes['delivery_status'].present?
+      DeliveryStatusSerializer.load(read_attribute(:delivery_status))
+    else
+      # attempt to parse the status from the log line and store if successful
+      decorated = line(:decorate => true)
+      if decorated && decorated.delivery_status
+        self.delivery_status = decorated.delivery_status
+      end
+    end
   end
 
   private
+
+  def calculate_delivery_status
+    delivery_status
+    true
+  end
 
   def self.create_mail_server_logs(emails, line, order, done)
     emails.each do |email|
--- a/app/models/mail_server_log.rb
+++ b/app/models/mail_server_log.rb
@@ -215,6 +215,10 @@ class MailServerLog < ActiveRecord::Base
     line
   end
 
+  def delivery_status
+    @delivery_status ||= line(:decorate => true).delivery_status
+  end
+
   private
 
   def self.create_mail_server_logs(emails, line, order, done)
--- a/app/models/mail_server_log.rb
+++ b/app/models/mail_server_log.rb
@@ -21,6 +21,8 @@ class MailServerLog < ActiveRecord::Base
   belongs_to :info_request
   belongs_to :mail_server_log_done
 
+  serialize :delivery_status, DeliveryStatusSerializer
+
   # Load in exim or postfix log file from disk, or update if we already have it
   # Assumes files are named with date, rather than cyclically.
   # Doesn't do anything if file hasn't been modified since it was last loaded.
--- a/app/models/mail_server_log/delivery_status_serializer.rb
+++ b/app/models/mail_server_log/delivery_status_serializer.rb
@@ -1,0 +1,47 @@
+class MailServerLog::DeliveryStatusSerializer
+  # Public: Tries to load the cached delivery_status value from the database
+  #
+  # value - The String representation of the delivery status, as held in the
+  #         database (e.g 'sent', 'rejected', 'failed')
+  #
+  # Returns the appropriate DeliveryStatus class depending on which MTA is
+  # in use (so currently a MailServerLog::EximDeliveryStatus or a
+  # MailServerLog::PostfixDeliveryStatus object). Or nil, if the value
+  # is not stored
+  def self.load(value)
+    if value.is_a?(String)
+      delivery_status_class.new(value.to_sym)
+    else
+      value
+    end
+  end
+
+
+  # Public: Casts the delivery_status object to a String so it can be stored in
+  # the database.
+  #
+  # value - a MailServerLog::EximDeliveryStatus or
+  #         MailServerLog::PostfixDeliveryStatus object representing the
+  #         message log line's delivery status
+  #
+  # Returns a String or nil
+  def self.dump(value)
+    return unless value
+    # Both delivery status classes return the MTA status Symbol as a String:
+    value.to_s
+  end
+
+  private
+
+  def self.delivery_status_class
+    mta = AlaveteliConfiguration.mta_log_type.to_sym
+    case mta
+    when :exim
+      MailServerLog::EximDeliveryStatus
+    when :postfix
+      MailServerLog::PostfixDeliveryStatus
+    else
+      raise "Unexpected MTA type: #{ mta }"
+    end
+  end
+end
--- a/db/migrate/20160907144809_add_delivery_status_to_mail_server_logs.rb
+++ b/db/migrate/20160907144809_add_delivery_status_to_mail_server_logs.rb
@@ -1,0 +1,12 @@
+class AddDeliveryStatusToMailServerLogs < ActiveRecord::Migration
+  def up
+    add_column :mail_server_logs, :delivery_status, :string
+    MailServerLog.where(:delivery_status => nil).find_each do |mail_log|
+      mail_log.update_attributes(:delivery_status => mail_log.delivery_status)
+    end
+  end
+
+  def down
+    remove_column :mail_server_logs, :delivery_status
+  end
+end
--- a/spec/models/mail_server_log/delivery_status_serializer_spec.rb
+++ b/spec/models/mail_server_log/delivery_status_serializer_spec.rb
@@ -1,0 +1,64 @@
+# -*- encoding : utf-8 -*-
+require File.expand_path(File.dirname(__FILE__) + '/../../spec_helper')
+
+describe MailServerLog::DeliveryStatusSerializer do
+
+  describe '.dump' do
+
+    it 'returns a String representation of the Exim status' do
+      status = MailServerLog::EximDeliveryStatus.new(:message_arrival)
+      result = MailServerLog::DeliveryStatusSerializer.dump(status)
+      expect(result).to eq 'message_arrival'
+    end
+
+    it 'returns a delivery status for the Postfix status' do
+      status = MailServerLog::PostfixDeliveryStatus.new(:sent)
+      result = MailServerLog::DeliveryStatusSerializer.dump(status)
+      expect(result).to eq 'sent'
+    end
+
+  end
+
+  describe '.load' do
+
+    context 'using the :exim MTA' do
+
+      it 'returns an EximDeliveryStatus object' do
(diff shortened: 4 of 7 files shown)


lizconlan:
ToDo - break the cache if the status is updated (er... from the rake task? there's a rake task, right?)

lizconlan:
>ToDo - break the cache if the status is updated

Wait, not needed because each MailServerLog has one line which won't change (unless there was an error, which is a whole different ball game) and the message status is calculated by looking at the most recent log `mail_server_logs.map(&:delivery_status).compact.last` - thanks for the tip @garethrees!

spec/models/mail_server_log/delivery_status_serializer_spec.rb
```diff
@@ -0,0 +1,64 @@
...
+    end
+
+    context 'using the :postfix MTA' do
+
+      before do
+        allow(AlaveteliConfiguration).to receive(:mta_log_type).
+          and_return('postfix')
+      end
+
+      it 'returns an EximDeliveryStatus object' do
```
garethrees: s/Exim/Postfix/

db/migrate/20160907144809_add_delivery_status_to_mail_server_logs.rb
```diff
@@ -0,0 +1,9 @@
+class AddDeliveryStatusToMailServerLogs < ActiveRecord::Migration
+  def up
+    add_column :mail_server_logs, :delivery_status, :string, :null => true
```
garethrees: Isn't `:null => true` the default? If so, we can drop it from the migration.

spec/models/mail_server_log_spec.rb
```diff
@@ -331,21 +331,56 @@
...
+      it 'does not look at the line text' do
+        expect(log).to_not receive(:line)
+        log.delivery_status
+      end
+
+    end
+
+    context 'there is not a stored value' do
+
+      it 'attempt to parse the line text' do
```
garethrees: s/attempt/attempts/

garethrees: In fact, I'd probably just write `it 'parses the line text' do`.

spec/models/mail_server_log_spec.rb
```diff
@@ -331,21 +331,56 @@
...
+        log = MailServerLog.new(:line => "hi")
+        expect(log).to receive(:line)
+        log.delivery_status
```
garethrees: This is testing the internals a bit much IMO.

I'd give it real data to actually do the work:

```ruby
log = MailServerLog.new(:line => "…<=…")
expect(log.delivery_status).
  to eq(MailServerLog::EximDeliveryStatus.new(:message_arrival))
```

spec/models/mail_server_log_spec.rb
```diff
@@ -331,21 +331,56 @@
...
         it 'returns a delivery status for the log line' do
-          line = "Apr 28 15:53:37 server exim[12105]: 2016-04-28 15:53:37 " \
-               "[12105] 1avnJx-00039F-Hs <= " \
-               "foi+request-331612-13811a2b@example.com U=foi P=local " \
-               "S=1986 id=ogm-538593+572f16e888-166a@example.com " \
-               "T=\"Freedom of Information request - example request\" " \
-               "from <foi+request-331612-13811a2b@example.com> for " \
-               "foi@example.org foi@example.org"
-          log = MailServerLog.new(:line => line)
+          log = MailServerLog.new(:line => line, :order => 1)
```
garethrees: Why is the `:order` attribute relevant here?

spec/models/mail_server_log_spec.rb
```diff
@@ -331,21 +331,56 @@
...
           status = MailServerLog::EximDeliveryStatus.new(:message_arrival)
           expect(log.delivery_status).to eq(status)
         end
 
+        it 'stores the delivery status value' do
+          log = MailServerLog.new(:line => line, :order => 1)
+          status = MailServerLog::EximDeliveryStatus.new(:message_arrival)
+          log.delivery_status
+          expect(log.persisted?).to eq true
```
garethrees: I wouldn't expect the `MailServerLog` to get persisted after calling `#delivery_status` – I can see _why_ its happening by reading the method definition, but I just wouldn't expect it.

app/models/mail_server_log/delivery_status_serializer.rb
```diff
@@ -0,0 +1,26 @@
+class MailServerLog::DeliveryStatusSerializer
+  def self.load(value)
+    return unless value
+    delivery_status_class.new(value.to_sym)
+  end
+
+  def self.dump(value)
```
garethrees: I think the public methods here could do with a little inline documentation about:

- What type of object they take
- What type of object they return

app/models/mail_server_log.rb
```diff
@@ -218,7 +218,19 @@ def line(opts = {})
...
-    @delivery_status ||= line(:decorate => true).delivery_status
+    # read in the status from the database if it's available
+    if attributes['delivery_status'].present?
+      DeliveryStatusSerializer.load(read_attribute(:delivery_status))
+    else
+      # attempt to parse the status from the log line and store if successful
+      decorated = line(:decorate => true)
+      if decorated && decorated.delivery_status
+        self.delivery_status = DeliveryStatusSerializer
+                                 .load(decorated.delivery_status)
```
garethrees: Convention is `.` at the end of lines, not at the start

app/models/mail_server_log.rb
```diff
@@ -218,7 +218,19 @@ def line(opts = {})
...
+        save
```
garethrees: This method is fine, but I don't think we should call save here. Its really confusing that a record could be saved by calling `#delivery_status`. Instead, I think its best to initialize the delivery status just before we save for the first time – that way the serializer should handle writing the value and subsequent loads/saves should just get the record from the DB?

```ruby
before_create do
  delivery_status
end
```

spec/models/mail_server_log_spec.rb
```diff
@@ -354,19 +389,31 @@
...
           status = MailServerLog::PostfixDeliveryStatus.new(:sent)
           expect(log.delivery_status).to eq(status)
         end
 
+        it 'stores the delivery status value' do
+          log = MailServerLog.new(:line => line, :order => 1)
+          status = MailServerLog::PostfixDeliveryStatus.new(:sent)
+          log.delivery_status
+          expect(log.persisted?).to eq true
+        end
```
garethrees: same comments as the exim version https://github.com/mysociety/alaveteli/pull/3476/commits/c4f7c2503333adb1101ad8a58538467586d4e6a4#r78357600

spec/models/outgoing_message_spec.rb
```diff
@@ -1556,6 +1556,7 @@
         before do
           allow(AlaveteliConfiguration).
             to receive(:mta_log_type).and_return('exim')
+          allow_any_instance_of(MailServerLog).to receive(:save)
```
garethrees: Shouldn't need these after removing the call to `save` in `delivery_status`

garethrees:
Looks great. Couple of tweaks but nearly ready to 🚢

db/migrate/20160907144809_add_delivery_status_to_mail_server_logs.rb
lizconlan: good point, think I forced the issue to reset it from where I'd temporarily made it default to a blank string

spec/models/mail_server_log_spec.rb
lizconlan: this may need more detail - I was thinking 'attempts' as it might not find a status value, although it's still parsing it either way

spec/models/mail_server_log_spec.rb
lizconlan: cool, thanks!

spec/models/mail_server_log_spec.rb
lizconlan: a side effect of the `save` operation - it chokes without it, might be able to get rid of this post refactoring

app/models/mail_server_log.rb
lizconlan: ugh, missed one

lizconlan:
as is the tradition, now I have 2 problems - one with tests breaking in 2.3.0 and one theoretical one where only new statuses are going to be cached (um, a rake task for a one-off tidy-up on multiple sites? not sure)

lizconlan:
There are roughly (as it may have gone up since I looked) 2,212,782 MailServerLog records on live so that migration might take a while. Although the lookup is really quick.

app/models/mail_server_log.rb
```diff
@@ -23,7 +23,7 @@ class MailServerLog < ActiveRecord::Base
 
   serialize :delivery_status, DeliveryStatusSerializer
 
-  before_create { |record| record.delivery_status }
```
lizconlan: turns out, this syntax makes Ruby 2.3.0 really unhappy :(

app/models/mail_server_log.rb
```diff
@@ -227,15 +227,18 @@ def delivery_status
       # attempt to parse the status from the log line and store if successful
       decorated = line(:decorate => true)
       if decorated && decorated.delivery_status
-        self.delivery_status = DeliveryStatusSerializer.
-                                 load(decorated.delivery_status)
-        delivery_status
```
lizconlan: and this was accidentally calling itself

lizconlan: and wasn't needed anyway as the result of an assignment is the assigned value so we can just return that. Simplify!

app/models/mail_server_log.rb
garethrees: What was the error? They use this form in a [current example](http://api.rubyonrails.org/classes/ActiveRecord/Callbacks.html):

```ruby

class Firm < ActiveRecord::Base
  # Disables access to the system, for associated clients and people when the firm is destroyed
  before_destroy { |record| Person.where(firm_id: record.id).update_all(access: 'disabled')   }
  before_destroy { |record| Client.where(client_of: record.id).update_all(access: 'disabled') }
end
```

lizconlan: might have been helped along by the confusion between `display_status` being a method and an db attribute

spec/models/mail_server_log/delivery_status_serializer_spec.rb
```diff
@@ -0,0 +1,64 @@
...
+        status = MailServerLog::PostfixDeliveryStatus.new(:sent)
+        result = MailServerLog::DeliveryStatusSerializer.load('sent')
+        expect(result).to eq status
+      end
+
+    end
+
+    it 'raises an error if passed an invalid status string' do
+      expect { MailServerLog::DeliveryStatusSerializer.load('whut') }
+        .to raise_error(ArgumentError, "Invalid MTA status: whut")
```
garethrees: `.` at the end of the preceding line; not at the start

spec/models/mail_server_log/delivery_status_serializer_spec.rb
```diff
@@ -0,0 +1,64 @@
...
+    end
+
+    it 'raises an error if it detects an unfamiliar MTA type' do
+      allow(AlaveteliConfiguration).to receive(:mta_log_type).
+        and_return('imaginary')
+
+      expect { MailServerLog::DeliveryStatusSerializer.load('sent') }
+        .to raise_error(RuntimeError, "Unexpected MTA type: imaginary")
```
garethrees: `.` at the end of the preceding line; not at the start

garethrees:
> Note that the migration will take ~4 mins
>
> – https://github.com/mysociety/alaveteli/pull/3476#pullrequestreview-1512599

Damn. What I didn't account for there was a ~ 10ms DB transaction for each record.

```ruby
Benchmark.measure do
  MailServerLog.where(:delivery_status => nil).find_each do |mail_log|
   mail_log.update_attributes!(:delivery_status => mail_log.delivery_status)
   puts "Updated #{mail_log.id}"
  end
end
#    (0.2ms)  BEGIN
#    (0.5ms)  UPDATE "mail_server_logs" SET "updated_at" = '2016-11-01 15:38:11.901899', "delivery_status" = NULL WHERE "mail_server_logs"."id" = 2917059
#    (0.4ms)  COMMIT
# Updated 2917059
# => 5808.970000 493.830000 6302.800000 (9391.507351)
```

PR #3486, opened 2016-09-14
Install pry

henare:
I like being able to use pry but pry-debugger only works on Ruby 1.9 and
I'm now developing Alaveteli on Ruby 2.0. pry-byebug seems to be the
successor to pry-debugger so I've decided to install that instead of
vanilla pry (which I probably would've been happy with anyway).


lizconlan:
Good choice - that's my preferred gem as well! Unfortunately we've not been able to work out how to make this backwards compatible for 1.9 users or play nice with the CI server without the test suite falling over with `'ruby_23' is not a valid platform` :cry:

Unless we only enable it for ruby_20 and ruby_21 I suppose. Bundler really needs a greater than operator!

garethrees:
From memory the problem is our version of bundler is too old. We've tried a couple of things:

- https://github.com/mysociety/alaveteli/commit/8ba36b06ce08dbff87885c3c6b59b4443fd8285d
- https://github.com/mysociety/alaveteli/commit/f2fad0565fb72ebf4cb1a927927d19f7d4ca225a

Can you install and configure preferred development tools locally? Something like:

```ruby
# ~/.irbrc
#!/usr/bin/env ruby

# Always prefer pry over irb
begin
  require 'pry'
rescue LoadError
  puts "Couldn't find pry. Install globally with `gem install pry` or include it in your bundle"
end

Pry.start || exit rescue LoadError


# ~/.pryrc
#!/usr/bin/env ruby

# Try to include pry-byebug
begin
  require 'pry-byebug'
rescue LoadError
  puts "Couldn't find pry-byebug. Install globally with `gem install pry-byebug` or include it in your bundle"
end
```

garethrees:
> I've got a confession to make - I sometimes like to shove `require "pry"; binding.pry` in my tests while I'm developing them to poke around in the environment 

I do this all the time :)

> However we could just install vanilla pry, no?

Yep! We [already](https://github.com/mysociety/alaveteli/blob/develop/Gemfile#L82) install pry-debugger for 1.9, and it looks like [pry is built against 1.9.3+](https://github.com/pry/pry/blob/master/.travis.yml) so I don't see a problem with adding `gem 'pry'` above the pry-debugger line. We probably want to specify a version [compatible with pry-debugger](https://github.com/mysociety/alaveteli/blob/develop/Gemfile.lock#L234).

henare:
@garethrees great! Thanks for all the tips - I've updated this PR as you've suggested.

garethrees:
Thanks! Merged in 8c1c1b4.

garethrees:
Had to revert this because of build failures. I think its due to some bundler version problems. Reopened in #3497

PR #3497, opened 2016-09-21
Add pry

garethrees:
https://github.com/mysociety/alaveteli/pull/3486


garethrees:
lolwut, now it passes?

garethrees:
Reopening in https://github.com/mysociety/alaveteli/pull/3880

Issue #3503, opened 2016-09-26
Remove suspended users from search index


crowbot:
Currently banned user accounts are returned from a search, perhaps ahead of more relevant results - we could remove them from the search index.

PR #3505, opened 2016-09-26
[#3503] Remove banned users from search index

garethrees:
Fixes https://github.com/mysociety/alaveteli/issues/3503

### DIFF
--- a/lib/tasks/cleanup.rake
+++ b/lib/tasks/cleanup.rake
@@ -57,6 +57,13 @@ namespace :cleanup do
     end
   end
 
+  desc 'Reindex banned users'
+  task :reindex_banned_users => :environment do
+    User.where("ban_text <> ''").find_each do |user|
+      user.xapian_mark_needs_index
+    end
+  end
+
 end
 
 def display_user(user, spam_score)
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -477,7 +477,7 @@ class User < ActiveRecord::Base
   end
 
   def indexed_by_search?
-    email_confirmed
+    email_confirmed && !banned?
   end
 
   def for_admin_column(complete = false)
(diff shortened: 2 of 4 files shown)


lib/tasks/cleanup.rake
```diff
@@ -57,6 +57,13 @@ namespace :cleanup do
     end
   end
 
+  desc 'Reindex banned users'
+  task :reindex_banned_users => :environment do
+    User.where("ban_text <> ''").find_each do |user|
+      user.xapian_mark_needs_index
+    end
+  end
+
```
garethrees: WDTK has 13094 banned users, so this might be too much to cope with in one run, but I think we should give it a go – I assume we just might get a few exceptions as the transaction fails from a locked record in a different process?

crowbot: How about [find_each](http://apidock.com/rails/ActiveRecord/Batches/find_each)?

garethrees: That's what I've used.

The problem is that this just creates records in the `acts_as_xapian_jobs` table. `ActsAsXapian.update_index` is called every 5 mins through a cron job (via a script and rake task), which then tries to process _all_ records in the `acts_as_xapian_jobs` table. I don't know what happens if a job hasn't completed by the time the next one starts.

Maybe the lock file just prevents any new indexes happening until the "big" job has finished? What happens if new records are added to `acts_as_xapian_jobs` while the big job is running  it might never exit?

IDK – thinking out loud a bit here :)

crowbot: Oops, sorry, missed that. Don't think we need to worry about ```acts_as_xapian``` - it will just run for a long time and hold a lock. It gets the jobs at the start so should exit.

garethrees:
This is going to get deployed via #3562

PR #3518, opened 2016-10-04
Fix sample data bug

lizconlan:
Fixes the comment belonging to different requests in the comments file and the info_requests_event file which was causing links based on the comment to be calculated incorrectly (stumbled across this while testing the Rails 4 branch, will fix it properly later just didn't want to lose the fix once I'd worked out what it was)

### DIFF
--- a/spec/fixtures/comments.yml
+++ b/spec/fixtures/comments.yml
@@ -21,10 +21,10 @@ silly_comment:
   user_id: "2"
   created_at: 2008-08-13 01:25:17.486939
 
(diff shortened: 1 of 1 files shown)


lizconlan:
ToDo: tests demonstrating the problem

garethrees:
Don't need tests because it won't occur from a user doing something in the app. Add the ticket description to the commit message

garethrees:
Merged as 712581a4a62c70bd1

PR #3520, opened 2016-10-04
Add an Alaveteli::Features module for feature flags

stevenday:
This adds a new module: Alaveteli::Features, which contains a single method:
`active?(:feature_name, optional_user_instance)` that lets you check if a
given feature is activated (optionally for a given user).

Available feature names are exposed through the Alaveteli::Features::FEATURES
constant and as individual constants on the module. The module is defined and
configured in config/initializers/features.rb

Underneath this uses https://github.com/fetlife/rollout to manage
activating/deactivating features, but with an adapter class so that we can
store feature config in the database rather than in Redis. This required a
new model to store these configs in.

I've refactored the ENABLE_ANNOTATIONS flag as an example of using this to
control a feature.

I'm not 100% sure about this, I'd be open to some critique, starters for ten:
- Is this better than just using new config flags?
- Should the module live in `lib/` rather than in the initializer?
- Should it be AlaveteliFeatures or just Features to be terser?
- Should it wrap the rollout object like I do now, or just expose it directly?

NB: PRed against the rails-4 branch so that the diff is sane, obviously it'll
be a while before this can be merged.


app/views/followups/new.html.erb
```diff
@@ -46,7 +46,7 @@
         <%= _('When you receive the paper response, please help others find ' \
                  'out what it says:') %>
         <ul>
-          <% if AlaveteliConfiguration.enable_annotations %>
+          <% if Alaveteli::Features.active?(Alaveteli::Features::ANNOTATIONS) %>
```
garethrees: Would be nice to supply a helper method so that in a view we could use something like:

```erb
<% if feature_active?(:annotations) %>
```

config/initializers/features.rb
```diff
@@ -0,0 +1,38 @@
+# Initialize our store of feature flags using rollout
+module Alaveteli
+  module Features
+    # Rollout assumes you'll use Redis for storing key:value pairs that
+    # describe features and who can use them. For now we're storing this in an
+    # ActiveRecord model instead for simplicity, so we configure Rollout to
+    # use that instead.
+    @@rollout = Rollout.new(RolloutKeyValueStore.new)
```
garethrees: Not sure whether using a "class" variable has the same problems as when using them in a class. Probably want to  make this a singleton-style module/gem in lib anyway, e.g. https://github.com/mysociety/alaveteli/blob/develop/lib/alaveteli_geoip.rb

spec/factories/rollout_key_value_stores.rb
```diff
@@ -0,0 +1,6 @@
+FactoryGirl.define do
+  factory :rollout_key_value_store do
+    key "key"
+    value "value"
```
garethrees: Would be worth making these a bit more realistic?

I'm assuming that `key` is the feature name `annotations` and value is some sort of flag (`true`?). I expect there's a good reason, but why isn't value a boolean?

garethrees:
Actually, I wonder whether we could drop the database requirement altogether?

Could we do something where we add an initializer in a theme which configures the app at boot time?

```ruby
Alaveteli::Features.enable(:embargoed_requests, :batch_requests)
Alaveteli::Features.disable(:annotations)
```

Another question – how do we enable "Alaveteli Pro"? Do we say "if you want Pro, enable X,Y,Z", or I guess we could do something like:

```ruby
# lib/alaveteli/features.rb
module Alaveteli
  module Features
    ALAVETELI_PRO = [:embargoed_requests, :batch_requests]
    # ...
  end
end

# THEME/config/initializers/features.rb
Alaveteli::Features.enable(Alaveteli::Features::ALAVETELI_PRO)
```

Might also be worth thinking about non-boolean features, e.g. price points.

stevenday:
Closing in favour of #3524

PR #3524, opened 2016-10-07
Add a feature flipping gem and use it to control annotations

stevenday:
Replaces #3520 after feedback and a better search of available third-party
libs for feature flipping.

Changes over #3520:
- Uses Flipper instead of Rollout, this has a better api for storing feature
  flags and managing where they're stored, with a much more
  persistence-agnostic approach.
- Pushes most of the new code into a local unbuilt gem which uses ideas from
  rails engines to provide a simple install generator that gets everything set
  up.
- Adds a helper method to hide the feature library away from Alaveteli code
- Feature library interaction is now tested (in the gem)

### DIFF
--- a/gems/alaveteli_features/spec/helpers/feature_enabled_spec.rb
+++ b/gems/alaveteli_features/spec/helpers/feature_enabled_spec.rb
@@ -1,0 +1,62 @@
+require 'spec_helper'
+require 'flipper/adapters/memory'
+require 'alaveteli_features/helpers'
+
+describe AlaveteliFeatures::Helpers do
+  let(:instance) { Class.new { include AlaveteliFeatures::Helpers }.new }
+  let(:test_backend) { Flipper.new(Flipper::Adapters::Memory.new) }
+  let(:user_class) do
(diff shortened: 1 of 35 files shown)


gems/alaveteli_features/spec/helpers/feature_enabled_spec.rb
```diff
@@ -0,0 +1,62 @@
...
+    end
+  end
+
+  before do
+    AlaveteliFeatures.backend = test_backend
+    # Seems to be the only way to make sure we don't register a group twice
+    begin
+      AlaveteliFeatures.backend.group(:admins)
+    rescue Flipper::GroupNotRegistered
+      Flipper.register :admins do |actor|
```
crowbot: What do we need the ```:admins``` group for? Doesn't seem to do anything at the moment.

stevenday:
I've rebased this to keep it up to date, but should this just be closed @crowbot (I can't remember if you merged everything into the alaveteli-pro-develop branch already or not?)

stevenday:
Closing this too then

PR #3534, opened 2016-10-12
Include all requests except hidden requests in the total count on public body statistics page

### LINKED ISSUE #3532: Total number of requests for public bodies on /body_statistics does not match the rest of the site
While the page says “Unclassified or hidden requests are not counted.”, the totals chart shows unclassified and hidden requests. This was very confusing when we switched the feature on because the numbers didn't match with what we saw on the site.

The tests [suggest this is intentional](https://github.com/mysociety/alaveteli/blob/4395a80856870831c34fd91969e07a18ebfefd0b/spec/models/public_body_spec.rb#L1218-L1221) but it seems wrong to me. Is there any reason for not changing this to match the other charts?

henare:
Fixes #3532.

### DIFF
--- a/app/models/public_body.rb
+++ b/app/models/public_body.rb
@@ -599,12 +599,12 @@ class PublicBody < ActiveRecord::Base
     # lowest) number of requests, but only returning data for those
     # with at least 'minimum_requests' requests.
     def self.get_request_totals(n, highest, minimum_requests)
-      ordering = "info_requests_count"
+      ordering = "info_requests_visible_count"
       ordering += " DESC" if highest
-      where_clause = where_clause_for_stats minimum_requests, 'info_requests_count'
+      where_clause = where_clause_for_stats minimum_requests, 'info_requests_visible_count'
       public_bodies = PublicBody.order(ordering).where(where_clause).limit(n)
       public_bodies.reverse! if highest
-      y_values = public_bodies.map { |pb| pb.info_requests_count }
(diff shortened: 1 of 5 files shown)


app/models/public_body.rb
```diff
@@ -599,12 +599,12 @@ def self.where_clause_for_stats(minimum_requests, total_column)
     # lowest) number of requests, but only returning data for those
     # with at least 'minimum_requests' requests.
     def self.get_request_totals(n, highest, minimum_requests)
-      ordering = "info_requests_count"
+      ordering = "info_requests_visible_count"
       ordering += " DESC" if highest
-      where_clause = where_clause_for_stats minimum_requests, 'info_requests_count'
+      where_clause = where_clause_for_stats minimum_requests, 'info_requests_visible_count'
```
lizconlan: ah, yes - good spot!

spec/spec_helper.rb
```diff
@@ -48,6 +48,7 @@
 
   config.include Capybara::DSL, :type => :request
   config.include Delorean
+  config.include FactoryGirl::Syntax::Methods
```
lizconlan: I think this is one of those things that makes it easier for us devs but harder for anyone coming to the codebase for the first time so I'm going to edit it out during the rebase/merge - sorry!

lizconlan:
Closed in favour of #3605

PR #3535, opened 2016-10-12
Remove drb argument from rspec

henare:
It was used by spork, which we don't use any more.


garethrees:
Merged in f4797c7 :parasol_on_ground:

PR #3549, opened 2016-10-17
Use CanCanCan for authorization

stevenday:
This swaps out the custom `lib/ability.rb` class for [CanCanCan](https://github.com/CanCanCommunity/cancancan). The main changes necessary for this were:
- Replacing `lib/ability.rb` with `app/models/ability.rb`
- Replacing usage of `Ability::can_update_request_state`, `Ability::can_view_with_prominence` `IncomingMessage.user_can_view?` and `OutgoingMessage.user_can_view?` with their new CanCanCan equivalents `can?(:update_request_state, request)` and `can?(:read, thing)`.
- Migrating specs for those methods into one place (`spec/models/ability_spec.rb`) and adding specs to test every salient permutation of visibility in requests.
- Updating `InfoRequest.make_zip_cache_path` to not use `Ability::can_view_with_prominence` when figuring out the appropriate name for the user's zip file (it wasn't possible to use CanCanCan for this, so I un-factored the code to call the primitives it would have called before)
- Changing a few tests that used simple `doubles` to use `mock_model` instead (so that CanCanCan recognised the class of the request in question).

I also added some new specs for `make_zip_cache_path` since it wasn't being tested very directly before, and separated some specs which were making multiple assertions in a single example into separate examples (because when they failed, they caused me some confusion due to the doubling up).

Finally, I fixed a few tests, factories or comments that caused me issues when I made other changes, but that weren't really related to the actual changes.

There's are a couple of places that still do authorisation stuff that I haven't moved. Firstly, the `all_can_view?` methods in Incoming/OutgoingMessage and InfoRequest. These are hard to do with CanCanCan (it's geared around can this user do "x" to this "thing") and the alternative (inspecting CanCanCan's internal storage essentially) is much more confusing, so I've left these as they are for now.

Likewise, the `User.can_view_hidden?` method is left alone - outside of our custom CanCanCan ability it's only used in one place (the zip file cache path making) where, as I mentioned before, CanCanCan isn't well suited.


spec/controllers/request_controller_spec.rb
```diff
@@ -809,21 +809,25 @@ def expect_hidden(hidden_template)
       expect_hidden('hidden')
     end
 
-    it "should show request to requester and admin if logged in" do
+    it "should show not request if logged in but not the requester" do
```
crowbot: Typo here I think "show not" for "not show".

spec/models/ability_spec.rb
```diff
@@ -0,0 +1,152 @@
...
+
+  context 'if the prominence is hidden' do
+    before do
+      resource.prominence = 'hidden'
+    end
+
+    it 'should return true for an admin user' do
+      expect(admin_ability).to be_able_to(:read, resource)
+    end
+
```
crowbot: Would be good to have a spec demonstrating that it returns `false` for the owner.

spec/models/ability_spec.rb
```diff
@@ -0,0 +1,152 @@
...
+    it 'should return false for a non-admin user' do
+      expect(other_user_ability).not_to be_able_to(:read, resource)
+    end
+  end
+
+  context 'if the prominence is requester_only' do
+    before do
+      resource.prominence = 'requester_only'
+    end
+
```
crowbot: A spec demonstrating that it returns `true` for admin would help make clear the nature of `requester_only`

spec/models/ability_spec.rb
```diff
@@ -0,0 +1,152 @@
...
+    it_behaves_like "a class with message prominence"
+  end
+
+  describe "updating request state of InfoRequests" do
+    context "given an old and unclassified request" do
+      let(:request) { FactoryGirl.create(:old_unclassified_request) }
+
+      before do
+        allow(request).to receive(:is_old_unclassified?).and_return(true)
```
crowbot: Not needed since 82ea7ffe1941c9e98c46650d1b60b0b20f27544f?

spec/models/ability_spec.rb
```diff
@@ -0,0 +1,152 @@
...
+          let(:ability) { Ability.new(FactoryGirl.create(:user)) }
+
+          it "should return false" do
+            expect(ability).not_to be_able_to(:update_request_state, request)
+          end
+        end
+      end
+    end
+  end
+end
```
crowbot: Looks like this file has has its end newline removed

spec/factories/info_requests.rb
```diff
@@ -73,8 +73,18 @@
...
     factory :old_unclassified_request do
       after(:create) do |info_request, evaluator|
-        incoming_message = FactoryGirl.create(:incoming_message, :info_request => info_request)
-        info_request.log_event("response", {:incoming_message_id => incoming_message.id})
+        incoming_message = FactoryGirl.create(
+          :incoming_message,
+          :info_request => info_request,
+          :created_at => Time.now - 31.days
+        )
+        FactoryGirl.create(
```
crowbot: There's a failing spec at the moment, which can be fixed by explicitly assigning this event to the InfoRequest:

```
        info_request.info_request_events = [
          FactoryGirl.create(
            :info_request_event,
            :info_request => info_request,
            :event_type => "response",
            :incoming_message_id => incoming_message.id,
            :created_at => Time.now - 31.days
          )
        ]
```

spec/models/ability_spec.rb
stevenday: Fixed in 172ba9d

spec/models/ability_spec.rb
stevenday: Fixed in b935a33

spec/controllers/request_controller_spec.rb
stevenday: Fixed in 9628e36

spec/models/ability_spec.rb
stevenday: Fixed in debf75c

stevenday:
@crowbot - thanks! I think I've fixed up all of those issues (relevant fixup commits noted in each comment). I'm getting one spec error which seems unrelated (it's `/spec/lib/alaveteli_geoip_spec.rb:91`) do you get this too, or is this due to something I've mis-configured in my dev vm?

stevenday:
Just FYI, I rebased this over the latest rails-4 and fixupped the fixups

stevenday:
I've rebased this to keep it up to date, but should this just be closed @crowbot (I can't remember if you merged everything into the alaveteli-pro-develop branch already or not?)

stevenday:
There's probably no point keeping this open against rails 4 then, I'll close it.

PR #3550, opened 2016-10-18
Attempt to load the Rake tasks from the current theme dir

lizconlan:
Allows rake tasks in the theme's `lib/tasks` directory to be run from the app's rake command. (Ignores the theme `Rakefile`)

Connects to mysociety/whatdotheyknow-theme#344

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -9,3 +9,14 @@ if Rails.env == 'test'
 end
 # Make sure the the acts_as_xapian tasks are also loaded:
 Dir[Rails.root.join('lib','acts_as_xapian','tasks','*.rake')].each { |file| load(file) }
+
+# Attempt to add in Rake tasks from the theme, ignores exceptions to avoid making
+# Rake unavailable because of theme script errors or THEME_URLS not being set
+begin
+  theme_name = File.basename(AlaveteliConfiguration::theme_urls.first, '.git')
+  theme_rake_path = File.join(Rails.root,"lib","themes", theme_name, "lib", "tasks", "*.rake")
+
+  Dir[theme_rake_path].each { |file| load(file) }
+rescue
+  #ignore
+end


Rakefile
```diff
@@ -9,3 +9,14 @@ if Rails.env == 'test'
...
+
+# Attempt to add in Rake tasks from the theme, ignores exceptions to avoid making
+# Rake unavailable because of theme script errors or THEME_URLS not being set
+begin
+  theme_name = File.basename(AlaveteliConfiguration::theme_urls.first, '.git')
+  theme_rake_path = File.join(Rails.root,"lib","themes", theme_name, "lib", "tasks", "*.rake")
+
+  Dir[theme_rake_path].each { |file| load(file) }
+rescue
+  #ignore
```
garethrees: Why not something like?

```ruby
Dir[theme_rake_path].each { |file| load(file) if File.exist?(file) }
```

lizconlan: Presumably it would return an empty array if the files didn't exist - I'd been thinking more of the case where the theme path didn't exist (or was broken). Although it does also prevent errors in the theme tasks from causing errors which is a good and a bad thing.

lizconlan: (you can guess how I found that out)

garethrees: > Presumably it would return an empty array if the files didn't exist

I don't immediately see a problem with that, but might be missing something

> I'd been thinking more of the case where the theme path didn't exist (or was broken). 

`File.exist?` should handle that case

> Although it does also prevent errors in the theme tasks from causing errors which is a good and a bad thing.

I think its better to surface errors than let them fail silently – sounds like it would be confusing to debug if you were expecting to see your theme task but it wasn't appearing. Did you have a case in mind where it would be good to ignore a theme task exception?

lizconlan:
closing as moving the rake task for mysociety/whatdotheyknow-theme#344 into core

PR #3551, opened 2016-10-18
Allow the word 'to' to be translated

lizconlan:
Does this look ok? Will mean regenerating the translations (I have had the French translation confirmed as "à" so I can add it for Belgium)

### DIFF
--- a/app/views/public_body/view_email.html.erb
+++ b/app/views/public_body/view_email.html.erb
@@ -51,6 +51,6 @@
     <% else %>
       <%= link_to _("Make a new FOI request"), new_request_to_body_path(:url_name => @public_body.url_name)%>
     <% end %>
-    to <%= h(@public_body.name) %>
+    <%= _("to") %> <%= h(@public_body.name) %>
   </strong>
 </div>
--- a/config/initializers/alaveteli.rb
+++ b/config/initializers/alaveteli.rb
@@ -11,7 +11,7 @@ load "debug_helpers.rb"
 load "util.rb"
 
 # Application version
-ALAVETELI_VERSION = '0.25.0.12'
+ALAVETELI_VERSION = '0.25.0.14'
 
 # Add new inflection rules using the following format
 # (all these examples are active by default):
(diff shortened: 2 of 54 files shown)


lizconlan:
**Note to self**: (if approved) download current translations (and push), then regenerate, edit, re-download

garethrees:
+1, but it would be good to make an improved fix (where we don't "build" the sentence) to develop while this is in our minds.

garethrees:
p.s. you want to hotfix to master; not develop :)

app/views/public_body/view_email.html.erb
```diff
@@ -47,10 +47,10 @@
 <div id="stepwise_make_request_view_email" class="stepwise_make_request_view_email">
   <strong>
     <% if @public_body.eir_only? %>
-      <%= link_to _("Make a new EIR request"), new_request_to_body_path(:url_name => @public_body.url_name)%>
+      <%= link_to _("Make a new EIR request to "), new_request_to_body_path(:url_name => @public_body.url_name)%>
     <% else %>
-      <%= link_to _("Make a new FOI request"), new_request_to_body_path(:url_name => @public_body.url_name)%>
+      <%= link_to _("Make a new FOI request to "), new_request_to_body_path(:url_name => @public_body.url_name)%>
```
garethrees: The previous version was better, as it only introduced a new string. This will _change_ strings, but won't solve the problem of joining partial strings.

lizconlan: yeah, pushed in error while trying to debug the connection - sorry (reverted)

lizconlan:
Have pushed a nicer version of the translation string BUT the meh version can be deployed as-is and I can cherry-pick 2606432 to another branch if there's not enough time to switch it around this afternoon

lizconlan:
note to self: dump d0edc36 and e902a54 and reinstate (tx push) the previous Transifex versions if we end up going with the 2606432 version of the translation

lizconlan:
Executive decision - horrible/lazy way for now, new fix later (possibly tomorrow)

lizconlan:
The updated version of the translation fix can now be found in the branch https://github.com/mysociety/alaveteli/tree/hotfix/0.25.0.15

PR #3555, opened 2016-10-20
Created research export rake

ajparsons:
Currently is dumping csvs files in 'exports'  - there might be a better location than that.

Connects to #3734

### DIFF
--- a/lib/tasks/export.rake
+++ b/lib/tasks/export.rake
@@ -133,6 +133,8 @@ end
 desc 'exports all non-personal information to export folder'
 task :research_export => :environment do
 
+  cut_off_date = Date.today
+
   csv_export(PublicBodyCategory)
   csv_export(PublicBodyHeading)
   csv_export(PublicBodyCategoryLink)
@@ -144,7 +146,7 @@ task :research_export => :environment do
 
   #export public body information
   csv_export( PublicBody,
-              nil,
+              PublicBody.where("created_at < ?", cut_off_date),
               ["id",
               "short_name",
               "created_at",
@@ -160,7 +162,8 @@ task :research_export => :environment do
 
   #export non-personal user fields
   csv_export( User,
-              nil,
+              User.where(ban_text: '').
+                where("updated_at < ?", cut_off_date),
               ["id",
               "name",
               "info_requests_count",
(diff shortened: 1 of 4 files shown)


lizconlan:
@ajparsons would you mind re-running the rake task to check I haven't broken anything? (I'll fix it if I have!)

ajparsons:
@lizconlan All changes seem fine (and are neater) - still works fine on test data. Any way of testing it on live data? Test data isn't good at testing the scale of the export.

lizconlan:
@ajparsons foolish question (to which I could probably find the answer some other way) has it been run against live data before? (also, I can almost certainly run it against live data to see what happens)

lib/tasks/export.rake
```diff
@@ -0,0 +1,200 @@
...
+# Exports a model
+#
+# query    - a query used to limit the export to matching records
+# header   - used to restrict exported columns
+# override - pass in lambdas to modify a given column based on values in the row
+#
+# Returns a String
+def csv_export(model, query=nil, header=nil, override={})
+  # set query and header to default values unless supplied
+  query = model.all unless query
```
garethrees: @lizconlan probably want to use `find_each` here rather than trying to load everything in to memory all at once?

lizconlan: @garethrees good point - thank you!

lizconlan: oh, wait - this breaks for `InfoRequestBatchPublicBody` as it doesn't have a primary key

lib/tasks/export.rake
```diff
@@ -59,7 +75,7 @@ end
 # Returns a String
 def csv_export(model, query=nil, header=nil, override={})
   # set query and header to default values unless supplied
-  query = model.all unless query
+  query = model unless query
```
garethrees: Could do:

```ruby
query ||= model
```

lizconlan: ah, yes - much neater!

lizconlan: ...I'll apply the same to the next line as well

ajparsons:
FYI I added a gender detector because it seemed better to do that at this point. I added a gem to do this, this change anything process-wise?

lizconlan:
> I added a gem to do this, this change anything process-wise?

I think we've decided to tentatively allow it and review if it complicates things

lizconlan:
ToDo:
 - [ ] squash/rebase
 - [ ] add date limits to queries to make sure it's not open-ended

PR #3594, opened 2016-11-01
Add a data structure (and some helpers) for being a pro user

### LINKED ISSUE #94: acts_as_xapian leaves open files hanging around
For each query, a new xapian connection is opened, but they are never explicitly closed, leading to errors in the log like

```
Couldn't open /data/vhost/www.whatdotheyknow.com/alaveteli/app/models/../../vendor/plugins/acts_as_xapian/lib/../xapiandbs/production/termlst.baseB: Too many open files
```

There are comments like this in the current `acts_as_xapian` code:

```
   # XXX we need to reopen the database each time, so Xapian gets changes to it.
    # Hopefully in later version of Xapian it will autodetect this, and this can
    # be commented back in again.
    # return unless @@db.nil
```

As part of this, should also review forks of `acts_as_xapian` to see if already fixed elsewhere.

stevenday:
This adds a new model ProAccount which is used to denote pro users and can
store their settings/preferences in the future. I've also added a basic
helper to the User model to test for pro users and a scope to find them
quickly.

@crowbot - is this sufficient for a first stab at #94 or would you like me
to take it further into logging on as a pro and ending up at the dashboard
(that feels like a further ticket or two to me)?

Closes #94

### DIFF
--- a/app/models/pro_account.rb
+++ b/app/models/pro_account.rb
@@ -1,0 +1,15 @@
+# -*- encoding : utf-8 -*-
+# == Schema Information
+#
+# Table name: pro_accounts
+#
+#  id                       :integer          not null, primary key
+#  user_id                  :integer          not null
+#  default_embargo_duration :string(255)
+#  created_at               :datetime         not null
+#  updated_at               :datetime         not null
+#
+
+class ProAccount < ActiveRecord::Base
+  belongs_to :user
+end
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -70,6 +70,8 @@ class User < ActiveRecord::Base
            :dependent => :destroy
   has_many :request_classifications,
            :dependent => :destroy
+  has_one :pro_account,
+          :dependent => :destroy
 
   validates_presence_of :email, :message => _("Please enter your email address")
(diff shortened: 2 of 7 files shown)


Gemfile.lock
```diff
@@ -49,7 +49,7 @@ PATH
       flipper
       flipper-active_record
       mime-types (< 3.0.0)
-      rails (~> 4.0.13)
+      rails (~> 3.2.22)
```
crowbot: I've made an equivalent commit in alaveteli-pro-develop, so I don't _think_ this needs to be here.

app/models/pro_account.rb
```diff
@@ -0,0 +1,14 @@
+# == Schema Information
```
crowbot: Can you add an encoding line to the top of all the new files? Should say ```# -*- encoding : utf-8 -*-```. If you're adding a lot of files, worth using the ```magic_encoding``` gem to do it.

stevenday: I've done this in 718b4f2

stevenday:
This was merged into alaveteli-pro-develop in bdd7386f17f82810a6e4187780f85d421133f13d

PR #3598, opened 2016-11-01
[#3566] Add sorting to admin users list

### LINKED ISSUE #3566: Sort option to list users by signup date in the admin interface

garethrees:
Fixes #3566 

screen shot 2016-11-01 at 17 30 32

### DIFF
--- a/app/assets/stylesheets/admin.scss
+++ b/app/assets/stylesheets/admin.scss
@@ -131,6 +131,11 @@ body.admin {
     }
   }
 
+  /* Users */
+  .sort-order {
+    margin-bottom: 30px;
+  }
+
   /* Censor Rules */
   .censor-rule-list td.text,
   .censor-rule-list td.replacement {
--- a/app/controllers/admin_user_controller.rb
+++ b/app/controllers/admin_user_controller.rb
@@ -17,13 +17,23 @@ class AdminUserController < AdminController
 
   def index
     @query = params[:query]
-    if @query
-      users = User.where(["lower(name) like lower('%'||?||'%') or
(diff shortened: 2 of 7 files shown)


garethrees:
Broke out the sorting to a button group and humanised the options:

screen shot 2016-11-23 at 12 11 45

garethrees:
> 'Name A-Z' and 'Name Z-A' might be even more readable.

@stevenday suggested this, but we came to the conclusion that there are complications for some international installs.

> `<steve>` so TIL that in languages chinese you can sort by the number of strokes in the characters of the name, and this is how they decided the order that countries would walk out in the bejing olympics
> `<steve>`  *languages like chines
> `<gareth>` wow
> `<steve>` good luck representing that gareth
> `<steve>` Name ascending / Name descending is probably the best


We don't do [internationalised sorting](http://alaveteli.org/docs/developers/i18n/#internationalised-sorting) on this page, and admins would probably "get" what A-Z / Z-A mean, so maybe its a better call.

garethrees:
I think we should just go with what's here now.

The arrows would probably be the best choice, but it would be much more fiddly to do (css classes, fallbacks etc) so I think any improvements can be made if we start using this sorting pattern more widely through the admin interface.

stevenday:
You could just do unicode up/down arrows: ↑ ↓

lizconlan:
👍 on the new approach, I think this is a lot nicer than the search filter method and allows us (or at least gives us the option) to hide the more technical language.

garethrees:
Done

screen shot 2016-11-24 at 09 56 09

PR #3601, opened 2016-11-01
embargo

crowbot:
Fixes https://github.com/mysociety/alaveteli-professional/issues/93

To do:
- [x] Sort out approach, consistency with cancan and `MessageProminence`
- [x] Early review
- [x] Inconsistency in `visible` scope vs. `public?` (previously `all_can_view?`)
- [x] Specs (move from InfoRequest if appropriate)
- [x] Deprecate methods
- [x] `public` scope is overwriting existing method
- [x] Can view

Separate PRs for
- [ ] 404s for embargoed request URLs.
- [ ] Separate admin ability to view embargoed requests


app/models/info_request/prominence.rb
```diff
@@ -0,0 +1,11 @@
+# -*- encoding : utf-8 -*-
+class InfoRequest
+  module Prominence
+
+    VALUES = [ 'normal',
```
crowbot: Different set of values (see https://github.com/mysociety/alaveteli/pull/3601#issuecomment-258457816)

app/models/info_request.rb
```diff
@@ -83,6 +83,8 @@ class InfoRequest < ActiveRecord::Base
   has_tag_string
 
   scope :visible, Prominence::VisibleQuery.new
+  scope :embargoed, Prominence::EmbargoedQuery.new
+  scope :unembargoed, Prominence::UnembargoedQuery.new
```
stevenday: Or just `:public` - we have a `public?` method so it makes sense to me to make that a matching scope too

stevenday:
Totally in agreement with @garethrees - this approach looks great (as does decorating the prominence).

> What do we want a request for an embargoed request page to return? Hidden requests get a 403 but I think a 404 might be more appropriate

Yep, absolutely a 404, I guess (like github) it should be a custom 404 page with a prompt that you might need to login

> Does this make sense with respect to cancan or should cancan be handling more of it?

I think cancan is exactly the right place for the use of this data, but it doesn't really replace this lower level stuff. (renaming stuff like `all_can_view` to `public` is a good example of separating those concerns)

I shied away from using all of cancan's features in my first pull request for it, but as well as defining the `can?` helper, it also has a whole set of before_filters/actions for checking authorization for a resource. https://github.com/CanCanCommunity/cancancan/wiki/authorizing-controller-actions has some more info, but it seems like a good fit for `authorize!` or maybe even using `load_and_authorize_resource` in the controllers? If you like, this can be extracted right up to the ApplicationController and become very generic, but we can also start using it in a more adhoc way, which might be easier. We already have some helper methods like `check_prominence` floating around, replacing them with cancan would seem like a good step?

spec/mailers/request_mailer_spec.rb
```diff
@@ -283,6 +283,16 @@ def sent_alert_params(request, type)
       expect{ send_alerts }.to raise_error(expected_message)
     end
 
+    context 'if the request is embargoed' do
+
+        it 'does not send the reminder' do
+        old_request.create_embargo(:publish_at => Time.now + 3.days)
```
stevenday: Seems to be some odd indentation in this section

stevenday:
This looks great, one minor comment, and the only other thing is that there don't seem to be any specs for the PublicQuery or SearchableQuery classes?

PR #3602, opened 2016-11-02
[#3572] Add AlaveteliSpamTermChecker

### LINKED ISSUE #3572: Make list of spam regexes overridable in some easy way
https://github.com/mysociety/alaveteli/commit/acb8b058c053e14de2056252030e8f2813af75ab

garethrees:
Make spam term checking configurable.

You can either use the defaults
(`AlaveteliSpamTermChecker::DEFAULT_SPAM_TERMS`) or configure your own:

    AlaveteliSpamTermChecker.default_spam_terms = [
      /a/,
      'b', # Strings get converted to a Regexp
      /c/
    ]

Fixes #3572


lib/alaveteli_spam_term_checker.rb
```diff
@@ -0,0 +1,83 @@
...
+    /FULL .O?N?_?L?I?N?E? ?\.?MOVIE/i,
+    /\[Full 20x5\]/i,
+    /Putlocker/i,
+    /se?\d+ep?\d+/i,
+    /\[free\]/i,
+    /vodlocker/i,
+    /online free full/i,
+    /streaming 2016/i,
+    /films?-?hd/i,
+    /full online/i,
```
garethrees: Not sure about this as a default – could easily mention "a full online transcript" or something like that.

lizconlan: yeah, possibly

lib/alaveteli_spam_term_checker.rb
```diff
@@ -0,0 +1,83 @@
...
+    /Watch.*?Movie Online/i,
+    /Gratuitement.*?TELECHARGER/i,
+    /full album/i,
+    /album complet/i,
+    /Album.*?Gratuit/i,
+    /Free.Download/i,
+    /\{FR\}/i,
+    /\[Album\]/i,
+    /watch.*?online free/i
+  ].freeze
```
lizconlan: ideally we need some way of storing these so that we don't inadvertantly give away a free tutorial on avoiding detection (notes for future revisions)

lib/alaveteli_spam_term_checker.rb
```diff
@@ -0,0 +1,83 @@
...
+    /Completa.*?Ver Online Gratis/i,
+    /Assistir.*?Completa Film Portuguese/i,
+    /\[MP3\]/i,
+    /Watch.*?Movie Online/i,
+    /Gratuitement.*?TELECHARGER/i,
+    /full album/i,
+    /album complet/i,
+    /Album.*?Gratuit/i,
+    /Free.Download/i,
+    /\{FR\}/i,
```
lizconlan: might that be potentially problematic as well?

lizconlan: ...or will the enclosing {} be distinctive enough?

lib/alaveteli_spam_term_checker.rb
garethrees: The defaults can be set from a config file if required. Something like:

```ruby
# THEME/config/initializers/spam_terms.rb
AlaveteliSpamTermChecker.default_spam_terms =
  File.read('config/spam_terms.txt').split("\n")

# or…
AlaveteliSpamTermChecker.default_spam_terms =
  YAML.load('config/spam_terms.yml')[:spam_terms]
```

lib/alaveteli_spam_term_checker.rb
garethrees: Just noting that I kept this one – I think its pretty unlikely that `{FR}` will be a legit thing

Issue #3458, opened 2016-08-15
Help people avoid making a blank search for an authority


equivalentideas:
The most common search (1 in 20) on [Right To Know’s Select Authority page](https://www.righttoknow.org.au/select_authority) is [a blank search](https://www.righttoknow.org.au/select_authority?utf8=%E2%9C%93&query=&bodies=1&commit=Search).

Are the other sites seeing this as well? Why is this happening?

Would it be helpful to users to prevent a reload and suggest they enter a term if searching is there intention (using the simple `required` attribute on the field)? e.g.
![screen shot 2016-08-15 at 4 44 11 pm](https://cloud.githubusercontent.com/assets/1239550/17657531/8ce73568-6307-11e6-9dc2-713aaf932a30.png)

zarino:
Adding a `required` attribute will certainly prevent blank searches.

I also wonder whether there should be a `<label>` above the input, telling you what to type? (In fact, the `<label>` is already there, but hidden – I wonder why?)

garethrees:
> Adding a `required` attribute will certainly prevent blank searches.

+1

> I also wonder whether there should be a `<label>` above the input, telling you what to type? (In fact, the `<label>` is already there, but hidden – I wonder why?)

I guess the placeholder covers that:

![screen shot 2016-10-11 at 11 17 16](https://cloud.githubusercontent.com/assets/282788/19267030/64e10458-8fa4-11e6-8dfc-d1b8315757a9.png)

zarino:
@garethrees: I wouldn’t be surprised if a significant chunk of visitors skip right over that “First, type in…” sentence, and then either mistake the placeholder text for a pre-filled input (the text colour is fairly dark) and click "Search" because they want to query the Department of Health, or mistake the placeholder text as exactly what it appears, just an example of something to search for, and mistakenly click "Search" expecting to be shown a thing that will let them begin a search.

equivalentideas:
Would a question to frame this initial step be a good label? e.g. “Who would you like to request documents from?”, “Who is your request to?”, etc. Then maybe move the explanation text (“by law they have to respond”) underneath the input. Is that text important for users?

garethrees:
I don't have any evidence to support this, but I prefer that we _tell_ the user what to do ("Find an authority") rather than asking them a question that might not translate so well to other languages.


The quickest win here would be to add the `required` attribute.


For extra points we could improve the prominence of the search form, but this is a bit better in the newer Alaveteli theme.

[...]

PR #3608, opened 2016-11-07
[#3458] Add required attribute to select authority form to prevent blank searches

garethrees:
Fixes https://github.com/mysociety/alaveteli/issues/3458

screen shot 2016-11-07 at 12 15 14


lizconlan:
Ah, previous required field issue was over on WDTK: https://github.com/mysociety/whatdotheyknow-theme/pull/323

PR #3610, opened 2016-11-08
Deal with Alaveteli Professional logins

stevenday:
This adds a few things to deal with pro user's logging in.

Firstly, it adds a config setting and a feature flag so that we can use the
features to turn on alaveteli pro.

Secondly, it adds a basic namespaced controller setup for alavetel pro, with
a single route for the dashboard.

Next, it overrides the homepage to redirect to the dashboard if the user is a
pro.

Finally, it provides a hook in the signin process that can inspect the
PostRedirect in use and alter it if needed for pro users. This is left as a
bit of a dead end because the main example (after a request is made without
logging in) doesn't have a pro-equivalent place to send the user to yet.

There's a couple of questions here:
- I have a failing test because of guarding the routes with a feature-check,
  should we do this kind of route guarding, and if so, how can I make that
  spec pass? Maybe there's something I can do with tagging the examples and
  rspec's config to always enable alaveteli pro before loading the routes?
- If not/as well as that, should I guard all the controller actions with a
  feature test too?

This is pull requested onto the pro-status branch because it needs the work
from there and I didn't want to confuse the diff.


app/controllers/alaveteli_pro/base_controller.rb
```diff
@@ -0,0 +1,36 @@
...
+          }
+        )
+      end
+      return true
+    end
+    return false
+  end
+
+
+end
```
crowbot: Missing end line here

spec/controllers/alaveteli_pro/base_controller_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+        session[:user_id] = user.id
+      end
+
+      it "doesn't redirect anywhere" do
+        get :index
+        expect(@response.status).to be 200
+      end
+    end
+  end
+end
```
crowbot: Missing end line here

app/controllers/alaveteli_pro/dashboard_controller.rb
```diff
@@ -0,0 +1,12 @@
...
+# Dashboard controller, for pro user dashboards.
+#
+# Copyright (c) 2008 UK Citizens Online Democracy. All rights reserved.
+# Email: hello@mysociety.org; WWW: http://www.mysociety.org/
+
+class AlaveteliPro::DashboardController < AlaveteliPro::BaseController
+  def index
+    @user = current_user
+  end
+end
```
crowbot: Missing end line here

app/views/alaveteli_pro/dashboard/index.html.erb
```diff
@@ -0,0 +1,2 @@
+<h1>Your dashboard</h1>
+<p>Welcome to your dashboard <%= @user.name %></p>
```
crowbot: Missing end line here

spec/controllers/alaveteli_pro/dashboard_controller_spec.rb
```diff
@@ -0,0 +1,22 @@
...
+      get :index
+      expect(@response.status).to be 200
+    end
+
+    it "sets @user" do
+      get :index
+      expect(assigns[:user]).to eq user
+    end
+  end
+end
```
crowbot: Missing end line here

stevenday:
@crowbot - I've changed things around as you requested in 4c806e1 - does that seem a better way of doing the pro redirect override?

I've also had another go at the routing issue I mentioned. I"ll need to rearrange 84d21a4 in the history so that it comes before the commit that uses it but this seems like a nicer way to do the conditional route declaration. As a side-effect, it's evaluated at run time so there are no more test issues with routes being declared before we can mock the feature setting.

stevenday:
I've rebased and added the extra "allow_overrides" flag as suggested, but I didn't quite understand part of your comment:

> When it's called from the pro override function, that can explicitly set the override flag to false.

The "pro override" now happens inside `do_post_redirect`, it doesn't call `do_post_redirect` separately, so as far as I can see, nothing needs to override the flag at the moment?.

PR #3612, opened 2016-11-14
Don't strip email in controller.

crowbot:
`User.find_by_email` now strips and nil-checks.


garethrees:
Looks good to me providing the build passes 👍 👍

PR #3614, opened 2016-11-16
Add a CanCan ability for accessing alaveteli pro

stevenday:
The title change is hopefully fairly simple and self explanatory, but there's
an aspect to the testing that I'd appreciate some other eyes on @crowbot /
@garethrees

Do the methods in AlaveteliFeatures::SpecHelpers seem like a good way to
enable/disable feature flags in specs?

In django there's a decorator for tests that lets you override settings, which
is kinda what I was going for here, but I'm not sure if there's a cleaner way
to do it?

### DIFF
--- a/app/models/ability.rb
+++ b/app/models/ability.rb
@@ -1,5 +1,6 @@
 class Ability
   include CanCan::Ability
+  include AlaveteliFeatures::Helpers
 
   def initialize(user)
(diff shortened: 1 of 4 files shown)


garethrees:
> Do the methods in `AlaveteliFeatures::SpecHelpers` seem like a good way to
enable/disable feature flags in specs?

I really like it!

PR #3617, opened 2016-11-17
Issues/alaveteli pro 93/404 embargoed requests

crowbot:
- [ ] Do we want to hide embargoed requests from the API? At the moment, an API user with a valid key is assumed to have all the powers of the public authority.

### DIFF
--- a/app/controllers/comment_controller.rb
+++ b/app/controllers/comment_controller.rb
@@ -92,7 +92,8 @@ class CommentController < ApplicationController
 
   def find_info_request
     if params[:type] == 'request'
-      @info_request = InfoRequest.find_by_url_title!(params[:url_title])
+      @info_request =
+        InfoRequest.not_embargoed.find_by_url_title!(params[:url_title])
     else
       raise "Unknown type #{ params[:type] }"
     end
--- a/app/controllers/followups_controller.rb
+++ b/app/controllers/followups_controller.rb
@@ -165,7 +165,7 @@ class FollowupsController < ApplicationController
   end
 
   def set_info_request
-    @info_request = InfoRequest.find(params[:request_id].to_i)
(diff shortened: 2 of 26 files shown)


stevenday:
A couple of comments and questions:

- There's a few "Handle embargo case" commits that could have better names
- http://betterspecs.org suggests to use names like `describe "#new"` when describing methods

I grepped through the code looking for unscoped calls to `InfoRequest`
- `HelpController#unhappy` and `HelpController#contact` still do unscoped lookups, is that ok? It feels like we should be guarding those with a CanCan access check if so?
- `ReportsController#new` and `#create` too
- `TrackController#track_request` too 
- `WidgetVotesController#find_info_request` and `WidgetController#find_info_request` as well
- The stats rake task counts embargoed requests (and some other stuff it probably shouldn't)

RE: the api, I agree that they should be shown through any endpoints where you need the authority's api key to access it

garethrees:
> http://betterspecs.org suggests to use names like `describe "#new"` when describing methods

They don't explicitly mention it, but they do show [an example](http://betterspecs.org/#sharedexamples) of using the HTTP verb, but with the URL path. Personally I prefer the `describe` to contain the method name, because we do use routes that don't align with the the method name.

_Some_ actions respond to different HTTP verbs, so I find it clearer to also add that to the spec description, e.g. for the case of:

```ruby
describe 'GET #new' do
  # …
end

describe 'POST #new' do
  # …
end
```

garethrees:
Yep, that's my preferred convention too.

stevenday:
Yep, sorry for the confusion, that's what I was suggesting really, just using the `#new` to show that it's a method, I'm happy with the verb too.

stevenday:
This all looks good to me 🚢

PR #3618, opened 2016-11-17
Exclude banned users from stats and graphs

### LINKED ISSUE #3548: Graphs and stats rake tasks should exclude or separate banned users.
At the moment, you can't really see trends in usage if there's a significant number of spam accounts being banned.

lizconlan:
Fixes #3548

### DIFF
--- a/spec/models/user_spec.rb
+++ b/spec/models/user_spec.rb
@@ -874,6 +874,15 @@ describe User do
 
   end
 
+  describe '.not_banned' do
+
+    it 'should not return banned users' do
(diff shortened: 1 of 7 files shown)


spec/models/user_spec.rb
```diff
@@ -874,6 +874,15 @@
 
   end
 
+  describe '.not_banned' do
+
+    it 'should not return banned users' do
+      user = FactoryGirl.build(:user, :ban_text => 'banned')
```
garethrees: It wouldn't return this anyway, given its not persisted.

garethrees:
👍

PR #3619, opened 2016-11-17
Controller & Actions for Pro request process

stevenday:
So far I have:
- A draft request model
- A new request form
- A create action
- A "preview" page
- An edit form
- An update action

Next is the big question, what happens when we "send" the request.

You can see that I've tried to simplify the data model a bit by moving the
message body into the Draft, rather than creating an OutgoingMessage alongside
it. My working assumption was that next I'd add a method to InfoRequest called
`new_from_draft` or something similar. This could do the transfer of data,
make an outgoing message an so on. Alternatively, this could happen in a new
controller (we'll need one anyway, so it's maybe a moot point).

My main question is what happens with any validation errors. Right now both
the message and the request do some, so we'd need a way to combine that and
feed it back to the draft I guess?

After that, there are a few outstanding issues with stuff I took from the
current new request form, like how we display the `law_used_short` text etc.

### DIFF
--- a/app/controllers/alaveteli_pro/draft_info_requests_controller.rb
+++ b/app/controllers/alaveteli_pro/draft_info_requests_controller.rb
@@ -1,0 +1,46 @@
+# -*- encoding : utf-8 -*-
+# app/controllers/alaveteli_pro/draft_info_requests_controller.rb
+# Controller for draft info requests
+#
+# Copyright (c) 2008 UK Citizens Online Democracy. All rights reserved.
+# Email: hello@mysociety.org; WWW: http://www.mysociety.org/
+
+class AlaveteliPro::DraftInfoRequestsController < AlaveteliPro::BaseController
+  def create
+    @draft = current_user.draft_info_requests.create(draft_params)
+    redirect_after_create_or_update
+  end
+
+  def update
+    @draft = current_user.draft_info_requests.find(params[:id])
+    @draft.update_attributes(draft_params)
+    redirect_after_create_or_update
+  end
+
+  private
+
+  def redirect_after_create_or_update
+    if params[:preview]
+      redirect_to preview_new_alaveteli_pro_info_request_path(draft_id: @draft.id)
+    else
+      redirect_to new_alaveteli_pro_info_request_path(draft_id: @draft.id),
+                  notice: _("Your draft has been saved!")
+    end
+  end
+
+  def draft_params
+    info_request_params.merge(outgoing_message_params.merge(embargo_params))
+  end
+
+  def info_request_params
+    params.require(:info_request).permit(:title, :public_body_id)
+  end
+
+  def outgoing_message_params
+    params.require(:outgoing_message).permit(:body)
+  end
+
+  def embargo_params
+    params.require(:embargo).permit(:embargo_duration)
+  end
+end
--- a/app/controllers/alaveteli_pro/info_requests_controller.rb
+++ b/app/controllers/alaveteli_pro/info_requests_controller.rb
@@ -1,0 +1,96 @@
+# -*- encoding : utf-8 -*-
+# app/controllers/alaveteli_pro/info_requests_controller.rb
+# Controller for info requests
+#
+# Copyright (c) 2008 UK Citizens Online Democracy. All rights reserved.
+# Email: hello@mysociety.org; WWW: http://www.mysociety.org/
+
+class AlaveteliPro::InfoRequestsController < AlaveteliPro::BaseController
+  before_filter :set_draft
+  before_filter :load_data_from_draft, only: [:preview, :create]
+
+  def new
+    if @draft_info_request
+      load_data_from_draft
+    else
+      create_initial_objects
+    end
+  end
+
+  def preview
+    if all_models_valid?
+      render "preview"
+    else
+      show_errors
+    end
+  end
+
+  def create
+    if all_models_valid?
+      @info_request.save # Saves @outgoing_message too
+      @embargo.save if @embargo.present?
+      send_initial_message(@outgoing_message)
+      destroy_draft
+      redirect_to show_request_path(url_title: @info_request.url_title)
+    else
+      show_errors
+    end
+  end
+
+  private
+
+  def show_errors
+    # There'll be a duplicate error if the outgoing_message is invalid, so
+    # delete it
+    @info_request.errors.delete(:outgoing_messages)
+    render "new"
+  end
+
+
+  def all_models_valid?
+    @info_request.valid? && @outgoing_message.valid? && \
+    (@embargo.nil? || @embargo.present? && @embargo.valid?)
+  end
+
+  def set_draft
+    if params[:draft_id]
+      @draft_info_request = current_user.draft_info_requests.find(
+        params[:draft_id]
+      )
+    end
+  end
+
+  def load_data_from_draft
+    @info_request = InfoRequest.from_draft(@draft_info_request)
+    @outgoing_message = @info_request.outgoing_messages.first
+    @embargo = @info_request.embargo
+  end
+
+  def create_initial_objects
+    @draft_info_request = DraftInfoRequest.new
+    @info_request = InfoRequest.new
+    @outgoing_message = OutgoingMessage.new(info_request: @info_request)
+    # TODO: set duration based on current user's account settings
+    @embargo = Embargo.new(info_request: @info_request)
+  end
+
+  def destroy_draft
+    if params[:draft_id]
+      current_user.draft_info_requests.destroy(params[:draft_id])
+    end
+  end
+
+  def send_initial_message(outgoing_message)
+    if outgoing_message.sendable?
+      mail_message = OutgoingMailer.initial_request(
+        outgoing_message.info_request,
+        outgoing_message
+      ).deliver
+
+      outgoing_message.record_email_delivery(
+        mail_message.to_addrs.join(', '),
+        mail_message.message_id
+      )
+    end
+  end
+end
--- a/app/models/draft_info_request.rb
+++ b/app/models/draft_info_request.rb
@@ -1,0 +1,23 @@
+# == Schema Information
+# Schema version: 20161128095350
+#
+# Table name: draft_info_requests
+#
+#  id               :integer          not null, primary key
+#  title            :string(255)
+#  user_id          :integer
+#  public_body_id   :integer
+#  body             :text
+#  embargo_duration :string(255)
+#  created_at       :datetime         not null
+#  updated_at       :datetime         not null
+#
+
+class DraftInfoRequest < ActiveRecord::Base
+  validates_presence_of :user
+
+  belongs_to :user
+  belongs_to :public_body
+
+  strip_attributes
+end
--- a/app/models/embargo.rb
+++ b/app/models/embargo.rb
@@ -1,16 +1,44 @@
 # -*- encoding : utf-8 -*-
 # == Schema Information
+# Schema version: 20161128095350
 #
-# Table name: embargos
+# Table name: embargoes
 #
-#  id              :integer          not null, primary key
-#  info_request_id :integer          not null
-#  publish_at      :datetime         not null
-#  created_at      :datetime         not null
-#  updated_at      :datetime         not null
+#  id               :integer          not null, primary key
+#  info_request_id  :integer
+#  publish_at       :datetime         not null
+#  created_at       :datetime         not null
+#  updated_at       :datetime         not null
+#  embargo_duration :string(255)
 #
 
 class Embargo < ActiveRecord::Base
   belongs_to :info_request
   validates_presence_of :info_request
+  validates_presence_of :publish_at
+  validates_inclusion_of :embargo_duration,
+                         in: lambda { |e| e.allowed_durations },
+                         allow_nil: true
+  after_initialize :set_publish_at_from_duration
+
+  DURATIONS = {
+    "3_months" => Proc.new { 3.months },
+    "6_months" => Proc.new { 6.months },
+    "12_months" => Proc.new { 12.months }
+  }.freeze
+
+  def allowed_durations
+    DURATIONS.keys
+  end
+
+  def duration_as_duration
+    DURATIONS[self.embargo_duration].call
+  end
+
+  def set_publish_at_from_duration
+    unless self.publish_at.present? || self.embargo_duration.blank?
+      self.publish_at = Time.zone.today + duration_as_duration
+    end
+  end
+
 end
--- a/app/models/embargo.rb
+++ b/app/models/embargo.rb
@@ -27,12 +27,22 @@ class Embargo < ActiveRecord::Base
     "12_months" => Proc.new { 12.months }
   }.freeze
 
+  DURATION_LABELS = {
+    "3_months" => _("3 Months"),
+    "6_months" => _("6 Months"),
+    "12_months" => _("12 Months")
+  }.freeze
+
   def allowed_durations
     DURATIONS.keys
   end
 
   def duration_as_duration
     DURATIONS[self.embargo_duration].call
+  end
+
+  def duration_label
+    DURATION_LABELS[self.embargo_duration]
   end
 
   def set_publish_at_from_duration
--- a/app/models/info_request.rb
+++ b/app/models/info_request.rb
@@ -946,6 +946,24 @@ class InfoRequest < ActiveRecord::Base
     info_request
   end
 
+  def self.from_draft(draft)
+    info_request = new(:title => draft.title,
+                       :user => draft.user,
+                       :public_body => draft.public_body)
+    info_request.outgoing_messages.new(:body => draft.body,
+                                       :status => 'ready',
+                                       :message_type => 'initial_request',
+                                       :what_doing => 'normal_sort',
+                                       :info_request => info_request)
+    if draft.embargo_duration
+      info_request.embargo = Embargo.new(
+        :embargo_duration => draft.embargo_duration,
+        :info_request => info_request
+      )
+    end
+    info_request
+  end
+
   def self.hash_from_id(id)
     Digest::SHA1.hexdigest(id.to_s + AlaveteliConfiguration::incoming_email_secret)[0,8]
   end
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -45,6 +45,9 @@ class User < ActiveRecord::Base
   has_many :info_requests,
            :order => 'created_at desc',
            :dependent => :destroy
+  has_many :draft_info_requests,
+           :order => 'created_at desc',
+           :dependent => :destroy
   has_many :user_info_request_sent_alerts,
            :dependent => :destroy
   has_many :post_redirects,
--- a/spec/integration/alaveteli_pro/request_creation_spec.rb
+++ b/spec/integration/alaveteli_pro/request_creation_spec.rb
@@ -1,0 +1,197 @@
+# -*- encoding : utf-8 -*-
+require File.expand_path(File.dirname(__FILE__) + '/../../spec_helper')
+require File.expand_path(File.dirname(__FILE__) + '/../alaveteli_dsl')
+
+describe "creating requests in alaveteli_pro" do
+  context "when writing a new request from scratch" do
+    let!(:public_body) { FactoryGirl.create(:public_body) }
+    let!(:pro_user) { FactoryGirl.create(:pro_user) }
+    let!(:pro_user_session) { login(pro_user) }
+
+    it "allows us to save a draft" do
+      using_pro_session(pro_user_session) do
+        # New request form
+        visit new_alaveteli_pro_info_request_path
+        expect(page).to have_content "Make a request"
+        select public_body.name, from: "To:"
+        fill_in "Summary:", with: "Does the pro request form work?"
+        fill_in "Your request:", with: "A very short letter."
+        select "3 Months", from: "Set embargo:"
+        click_button "Save draft"
+
+        # Redirected back to new request form
+        drafts = DraftInfoRequest.where(title: "Does the pro request form work?")
+        expect(drafts).to exist
+        draft = drafts.first
+        expect(draft.body).to eq "A very short letter."
+        expect(draft.embargo_duration).to eq "3_months"
+
+        expect(page).to have_content("Your draft has been saved!")
+        expect(page).to have_content("This request will be embargoed " \
+                                     "until #{Time.zone.today + 3.months}")
+
+        # The page should pre-fill the form with data from the draft
+        expect(page).to have_select("To:", selected: public_body.name)
+        expect(page).to have_field("Summary:",
+                                   with: "Does the pro request form work?")
+        expect(page).to have_field("Your request:",
+                                   with: "A very short letter.")
+        expect(page).to have_select("Set embargo:", selected: "3 Months")
+      end
+    end
+
+    it "allows us to preview the request" do
+      using_pro_session(pro_user_session) do
+        # New request form
+        visit new_alaveteli_pro_info_request_path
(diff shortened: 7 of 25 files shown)


stevenday:
I've now added a really basic version of the publishing process to do what I've described above. It's in the controller for now, as it cuts across three models so it seems odd to put it any one of them and worse to spread it across all three. I haven't extensively tested the validation/errors side of things and that's the area I think will need some more thought. Any ideas how to make that generic/future-proof? I guess there's something meta I can do looping over all the errors and checking which ones match a property of the draft object?

app/controllers/alaveteli_pro/draft_info_requests_controller.rb
```diff
@@ -0,0 +1,121 @@
...
+    if @draft_info_request.save
+      redirect_to action: :show, id: @draft_info_request.id
+    else
+      render "new"
+    end
+  end
+
+  def update
+    @draft_info_request = DraftInfoRequest.find(params[:id])
+    @draft_info_request.update_attributes(draft_info_request_params)
```
garethrees: update_attributes calls save.

You want:

```ruby
if @draft_info_request.update_attributes(draft_info_request_params)
  # ...
```

app/controllers/alaveteli_pro/draft_info_requests_controller.rb
```diff
@@ -0,0 +1,121 @@
...
+    @draft_info_request = DraftInfoRequest.new
+  end
+
+  def edit
+    @draft_info_request = DraftInfoRequest.find(params[:id])
+  end
+
+  def create
+    @draft_info_request = DraftInfoRequest.new(draft_info_request_params)
+    @draft_info_request.user = current_user
```
garethrees: Instead of this you can (and should) do:

```ruby
@draft_info_request = current_user.draft_info_requests.new(draft_info_request_params)
```

app/controllers/alaveteli_pro/draft_info_requests_controller.rb
```diff
@@ -0,0 +1,121 @@
...
+    else
+      render "new"
+    end
+  end
+
+  def show
+    @draft_info_request = DraftInfoRequest.find(params[:id])
+  end
+
+  def publish
```
garethrees: I think we should try to stick to the default RESTful actions for new controllers.

I'd make this something like `DraftPublicationsController#create`.

app/controllers/alaveteli_pro/draft_info_requests_controller.rb
```diff
@@ -0,0 +1,121 @@
...
+    @draft_info_request = DraftInfoRequest.find(params[:id])
+    build_info_request_and_outgoing_message(@draft_info_request)
```
garethrees: I'd also add `InfoRequest.from_draft` that initializes a new `InfoRequest` from the draft:
 rather than keeping this logic in the controller:

```ruby
class InfoRequest
  def self.from_draft(draft)
    request = new(:title => draft.title,
                  :user => draft.user,
                  :public_body => draft.public_body)
    request.outgoing_messages.new(:body => draft.body,
                                  :status => 'ready',
                                  :message_type => 'initial_request',
                                  :what_doing => 'normal_sort')
    request
  end
```

app/controllers/alaveteli_pro/draft_info_requests_controller.rb
```diff
@@ -0,0 +1,121 @@
+# -*- encoding : utf-8 -*-
+# app/controllers/alaveteli_pro/request_controller.rb
+# Dashboard controller, for pro user dashboards.
```
garethrees: copy/paste error?

app/controllers/alaveteli_pro/draft_info_requests_controller.rb
```diff
@@ -0,0 +1,121 @@
...
+# Copyright (c) 2008 UK Citizens Online Democracy. All rights reserved.
+# Email: hello@mysociety.org; WWW: http://www.mysociety.org/
+
+class AlaveteliPro::DraftInfoRequestsController < AlaveteliPro::BaseController
+  def new
+    @draft_info_request = DraftInfoRequest.new
+  end
+
+  def edit
+    @draft_info_request = DraftInfoRequest.find(params[:id])
```
garethrees: Should scope to `current_user`, otherwise you can just bash in ID numbers to the URL and load other users' drafts:

```ruby
@draft_info_request = current_user.draft_info_requests.find(params[:id])
```

Should do this for all actions

stevenday:
@garethrees, @crowbot thanks for the comments! A few thoughts in return after I've worked through your suggestions:

> We could put the validations that apply to both draft requests and requests in a concern or custom validator that applies to both

> I'm wondering whether a Draft should also initialize an internal InfoRequest so that methods like law_used can be passed over to that.

In either of these cases it's complicated a bit because I've merged an OutgoingMessage and an InfoRequest into one model, so we need some of the validation from both of them. I tried extracting the validation into two concerns, but OutgoingMessage's validation is pretty complicated because of the use of the templates, which makes it quite hard to extract. The alternative of keeping an internal model for each seems more do-able though, I'll flesh that out. (This also gets us easier access to the law_used stuff, so that's good)

> I think we should try to stick to the default RESTful actions for new controllers.
> I'd make this something like DraftPublicationsController#create.

> In terms of the conversion of a draft to a request, it feels to me like this is, as discussed above, really the creation of an InfoRequest using a DraftInfoRequest. How about we use a create action on the RequestController to do this, passing the id of the draft?

I think I prefer the latter suggestion here, as `DraftPublicationsController#create` seems a bit contrived? Keeping it inside the pro namespace seems like the best idea for now, so that we don't mix up the draft request functionality until it's meant to be available for the public too.

garethrees:
I think a [sketch of the UI flow](https://signalvnoise.com/posts/1926-a-shorthand-for-designing-ui-flows) would help figure out how we want to do this.

Maybe we don't ever want to go straight from a saved draft to a published InfoRequest. I'm thinking now that editing a draft would actually just be hitting `RequestController#new` with some pre-filled data like http://alaveteli.org/docs/developers/api/#starting-new-requests-programmatically.

If we did this, then we could eliminate `edit`, `show`, `publish` and possibly `update` from the `DraftInfoRequestsController`.

stevenday:
I've just pushed a load of commits with some updates to this. I've implemented the `InfoRequest#from_draft` method and @garethrees internal request/message object, with some method shadowing to make the titles work and make sure that the draft keeps it's internal objects up to date. I've also pushed the InfoRequest creation to a new controller. I took some ideas from @crowbot validation, making it possible to choose whether the draft validates its internal objects or not. There's still the downside of copying across errors, but it seems like we're stuck in a similar situation regardless because otherwise we're copying across (or including) *some* validation but not all.

Re: the UI flow, here's a gif of how it works currently:


As you can see, I've added a new "Save Draft" button, with the intention being that we'd also expand this into some auto-saving JS eventually. I've slightly renamed the Preview button to make it clearer that it's kind of a "next step" so the validation is less of a shock, but I wonder whether this step is even necessary for Pros? You don't "preview" an email, so should you preview this?

If this is generally good, then I'll start writing some specs for all the new things and tidy up the commits.

garethrees:
> Re: the UI flow, here's a gif of how it works currently:

Looking solely at the gif, its not clear why we need any of the `DraftInfoRequestsController` actions other than `#create`.

Why do we need a separate form etc? Seems to be a lot of duplication (and complexity in `DraftInfoRequest`) that we might be able to avoid.

stevenday:
> Why do we need a separate form etc? Seems to be a lot of duplication (and complexity in DraftInfoRequest) that we might be able to avoid.

So this is just the basic version where it looks a lot like the current request form, but soon we'll be adding embargo options (which need to do some funky stuff to set account/project defaults etc), project filing and the js autocomplete authority picker, which will differentiate it quite a bit from the current request flow.

The idea of a draft request is twofold: firstly so that we can actually let people build up requests without having to send them immediately and secondly so that we can deal with the problem of a non-pro writing a request anonymously and then logging in - because we need to send them somewhere to set the embargo, project etc. 

> Looking solely at the gif, its not clear why we need any of the DraftInfoRequestsController actions other than #create and #index.

I'm not sure I follow here, I think this way of doing it (with separate controller actions for new, edit, create, update etc) is clearer and more idiomatic.. 

I admit that the model is a bit tricky now, I'd be happy to have another go at extracting the validation stuff from existing models if that seemed preferable?

stevenday:
Here's a UI flow diagram as I see it at the mo:

img_0139

garethrees:
Here's what I was thinking we could do:

draft-creation-flow

The top line is the current "make a request process" as it is now.

The new addition would be an option to click "save as draft" in step 2. This would post the parameters to `InfoRequestDraftsController#create` and save the title, body and public body ID.

When the user wants to continue with their draft, they'd visit `InfoRequestDraftsController#index` to find the draft and click "Edit draft". This would actually be a link to `RequestController#new` with the ["API" parameters](http://alaveteli.org/docs/developers/api/#starting-new-requests-programmatically) pre-set from the persisted draft.

They could then continue through the process as normal.

stevenday:
@garethrees - aha, gotcha. 

Hmm, the downside of this for me is that it forces you to make a valid request (title, body and public_body) before you can save it as a draft. I like the idea of being able to save something straight away, even if it's just a title or a selected public body.

As I said above, we're probably going to need a new `InfoRequestController#create` and `#new` for the new pro options anyway, so does this make as much sense then?

garethrees:
> Hmm, the downside of this for me is that it forces you to make a valid request (title, body and public_body) before you can save it as a draft. I like the idea of being able to save something straight away, even if it's just a title or a selected public body.

Not necessarily – when you POST to `DraftsController#create` you'd still be creating an `InfoRequestDraft` – which could allow all attributes to be blank. It won't matter if they're blank when you use the API params, as they just won't pre-fill the blank fields.

> As I said above, we're probably going to need a new `InfoRequestController#create` and `#new` for the new pro options anyway, so does this make as much sense then?

Yeah, this comment…

> So this is just the basic version where it looks a lot like the current request form, but soon we'll be adding embargo options (which need to do some funky stuff to set account/project defaults etc), project filing and the js autocomplete authority picker, which will differentiate it quite a bit from the current request flow.

…made me think that actually, what this PR is discussing is almost that – a new request process for pro.

In which case, yes, I'd be thinking about making `AlaveteliPro::RequestController` for whatever actions need it, and keep the drafting pretty simple as I illustrated.

I think the only complication is if you wanted to store the embargo with the draft, but tbh I'm not sure how valuable that is. I think you're more likely to want to set it at the point of actually creating the request?

garethrees:
> I think the only complication is

Oh, you'd require a public body, but it looks like you were validating that existed already.

stevenday:
> Not necessarily – when you POST to DraftsController#create you'd still be creating an InfoRequestDraft – which could allow all attributes to be blank. It won't matter if they're blank when you use the API params, as they just won't pre-fill the blank fields.

OK, currently clicking the preview button does the validation though, rather than validating after you click send, so we'd have to change that. I think that would be a (minor) step backwards in UX terms.

> Oh, you'd require a public body, but it looks like you were validating that existed already.

It isn't required on the draft model so it shouldn't be validated until you try to send it, I just dropped a `<select>` in the html to pave the way for the autocomplete version and didn't give it a blank default `<option>`.


I'm still not totally sold on the value of re-using the new form to edit drafts. 

I don't think we're going to want to use the existing form anyway because of the amount of new stuff we're adding (only the title and body fields will remain the same), so what difference does it make whether we're creating/editing a draft and then POSTing to create an InfoRequest, or vice-versa?

(I'm not sure how re-editing a draft would work in your proposal, but presumably it would need an #update or a more complicated #create to deal with it? Either way it doesn't seem to get us much less or simpler code than what we have currently.)

app/models/draft_info_request.rb
```diff
@@ -0,0 +1,80 @@
...
+
+  def law_used_human(key)
+    info_request.law_used_human(key)
+  end
+
+  def info_request
+    if @info_request.nil?
+      @info_request = InfoRequest.from_draft(self)
+    end
+    @info_request
```
garethrees: You can make this much simpler:

```ruby
  def info_request
    @info_request ||= InfoRequest.from_draft(self)
  end
```

garethrees:
> OK, currently clicking the preview button does the validation though, rather than validating after you click send, so we'd have to change that. I think that would be a (minor) step backwards in UX terms.

I was imagining you save the draft before going through to the preview, but ISWYM.

I was mostly thinking of this workflow being like how Sparrow works:

sparrow-draft-ui

> (I'm not sure how re-editing a draft would work in your proposal, but presumably it would need an #update or a more complicated #create to deal with it? Either way it doesn't seem to get us much less or simpler code than what we have currently.)

Yeah, that might be a bit trickier. With this workflow I'd maybe make the "edit" button hit `GET /new/:body_url_name?draft_id=4`, and then set the API params from the draft record. That way you could use the ID when you update the draft, or delete the draft once you've clicked "send". But yeah, a little more complex.

> I don't think we're going to want to use the existing form anyway because of the amount of new stuff we're adding (only the title and body fields will remain the same), so what difference does it make whether we're creating/editing a draft and then POSTing to create an InfoRequest, or vice-versa?

Here are my main thoughts, but I'm not dead against what you've got here.

- What I've suggested could be applied to Alaveteli core with minimum effort, and isn't tied to the pro-only request process
- The less actions in `DraftsController`, the less a user needs to think about. Why should editing a "draft" be different to editing a "real" request.
- The composition of the current `InfoRequestDraft` is quite complicated. I'm not 100% convinced a draft needs to know much more than title, body and public body – all of which are pretty optional.

stevenday:
ok, here goes again :)

I've moved the action into an `InfoRequestController` as suggested by @crowbot. This has three actions, `#new` for the creating/editing form, `#preview` for previewing the request and `#create` for actually making it. This has allowed me to remove all the fancy validation stuff from `DraftInfoRequest`, and actually get rid of the `DraftInfoRequestController` altogether.

The icky part is that `InfoRequestController#preview` has to know about creating/updating drafts, because the `#new` form can only post to one place. I don't think it's turned out too horribly, but I'd appreciate it if you can point out anything obvious I've missed there which would make it nicer? One upside is that I think it makes previewing stuff much simpler, because we can save the draft and then just recreate the form from that. As a side-effect, if you forget about your request whilst previewing, it'll still be there as a draft. Again, once the `#new` form auto-saves too (or saves via an ajax post at least) this will be a bit clearer.

Perhaps I should formalise the relationship between draft and info_request with a `has_one` relationship? (Possibly doing away with destroying the draft after sending too?)

stevenday:
I've also just pushed the embargo side of this, it's a pretty simple addition in the same vein as the way outgoing messages are treated, with a field to cache the important stuff on the draft, and then a little bit of controller work to create/validate an embargo object in the various actions and views.

app/controllers/alaveteli_pro/info_requests_controller.rb
```diff
@@ -0,0 +1,123 @@
...
+      info_request_params,
+      outgoing_message_params,
+      current_user
+    )
+    @outgoing_message = @info_request.outgoing_messages.first
+  end
+
+  def build_embargo
+    @embargo = Embargo.new(embargo_params)
+    @embargo.info_request = @info_request
```
crowbot: These two lines could be 

```
@info_request.build_embargo(embargo_params)
```

app/models/draft_info_request.rb
```diff
@@ -0,0 +1,21 @@
...
+# Table name: draft_info_requests
+#
+#  id             :integer          not null, primary key
+#  title          :string(255)
+#  user_id        :integer
+#  public_body_id :integer
+#  body           :text
+#  created_at     :datetime         not null
+#  updated_at     :datetime         not null
+#  publish_at     :datetime
```
crowbot: I don't think we want to store a date here, otherwise if you pick up your draft the next day, the embargo form won't pick up the saved value - I think we want to store it somehow as 3mos/6mos/12mos or whatever our options are on the draft so that we can repopulate the form. Also, conceptually, you don't really want the clock to start until you send the request.

app/models/info_request.rb
```diff
@@ -946,6 +946,19 @@ def self.create_from_attributes(info_request_atts, outgoing_message_atts, user=n
...
+  def self.from_draft(draft)
+    info_request = new(:title => draft.title,
+                       :user => draft.user,
+                       :public_body => draft.public_body)
+    info_request.outgoing_messages.new(:body => draft.body,
+                                       :status => 'ready',
+                                       :message_type => 'initial_request',
+                                       :what_doing => 'normal_sort',
+                                       :info_request => info_request)
+    Embargo.new(:publish_at => draft.publish_at, info_request => info_request)
```
crowbot: Missing colon on the info_request key here.

stevenday:
> Ah! Perhaps we can have the best of both worlds - the 'new' form can be submitted to the draft controller 'create' or 'update' action (depending on whether there's a draft id in the form), but if the params include 'preview' ie. the preview button was used to submit the form, then those actions redirects to the 'preview' action of the InfoRequestController, not the 'new' action. The preview controller creates an InfoRequest etc from the draft and validates them, then continues as it does now.

I've just updated the code to match this description as best I can, and re-worked the embargo setting to use a text duration that only gets turned into a `publish_at` date on the Embargo when it's saved. I've also rebased everything and tidied it up as if this was the plan all along.

I'm leaving embargo extending/altering and the authority autocomplete until later, as this is getting quite big already and I don't want to drag it on any longer than necessary.

@crowbot - this is hopefully ready for a final review now

app/models/user.rb
```diff
@@ -32,6 +33,7 @@
 #  request_classifications_count     :integer          default(0), not null
 #  public_body_change_requests_count :integer          default(0), not null
 #  info_request_batches_count        :integer          default(0), not null
+#  pro                               :boolean          default(FALSE), not null
```
crowbot: I don't think this should be here - possibly an artefact of your database having has migrations run that we didn't end up using. `pro?` on the user model is determined by the existence of an associated `pro_account` model.

app/models/draft_info_request.rb
```diff
@@ -0,0 +1,29 @@
...
+#  updated_at     :datetime         not null
+#
+
+class DraftInfoRequest < ActiveRecord::Base
+  validates_presence_of :user
+
+  belongs_to :user
+  belongs_to :public_body
+
+  before_save :strip_empty_duration
```
crowbot: I think you could just use `strip_attributes!` here as in [`OutgoingMessage`](https://github.com/mysociety/alaveteli/blob/develop/app/models/outgoing_message.rb#L64) (without the `:empty => true` bit).

app/models/draft_info_request.rb
```diff
@@ -0,0 +1,29 @@
...
+# Schema version: 20161128095350
+#
+# Table name: draft_info_requests
+#
+#  id             :integer          not null, primary key
+#  title          :string(255)
+#  user_id        :integer
+#  public_body_id :integer
+#  body           :text
+#  duration       :string(255)
```
crowbot: Might be worth renaming this and the `duration` field on `embargo` to `embargo_duration` to make this one a bit more explicit about what it is.

spec/models/draft_info_request_spec.rb
```diff
@@ -0,0 +1,26 @@
+require 'spec_helper'
+
+describe DraftInfoRequest do
+  it "belongs to a public body" do
+    expect(FactoryGirl.create(:draft_info_request).public_body).not_to be nil
```
crowbot: You could use
```
 expect(FactoryGirl.create(:draft_info_request).public_body).to be_a(PublicBody)
```
 in this and the following specs to make them a bit more specific.

app/models/embargo.rb
```diff
@@ -1,16 +1,56 @@
...
+  validates_inclusion_of :duration,
+                         in: lambda { |e| e.allowed_durations },
+                         allow_nil: true
+  after_initialize :set_publish_at_from_duration
+
+  DURATIONS = {
+    "3_months" => Proc.new { 3.months },
+    "6_months" => Proc.new { 6.months },
+    "12_months" => Proc.new { 12.months },
+    "indefinitely" => Proc.new { 1000.years }
```
crowbot: We could leave out the indefinite option for now, in the interests of simplicity.

app/controllers/alaveteli_pro/info_requests_controller.rb
```diff
@@ -0,0 +1,96 @@
...
+  end
+
+  private
+
+  def all_models_valid?
+    @info_request.valid? && @outgoing_message.valid? && \
+    (@embargo.nil? || @embargo.present? && @embargo.valid?)
+  end
+
+  def set_publish_at_options
```
crowbot: This seems like it's maybe more suited to being a function in a helper, as it's about how to display things.

app/views/alaveteli_pro/info_requests/preview.html
```diff
@@ -0,0 +1,40 @@
+<% @title = _("Preview new {{law_used_short}} request to '{{public_body_name}}", :law_used_short => h(@info_request.law_used_human(:short)), :public_body_name => h(@info_request.public_body.name)) %>
```
crowbot: Could do with some line wrapping here for readability.

spec/integration/alaveteli_pro/request_creation_spec.rb
```diff
@@ -0,0 +1,209 @@
...
+          visit new_alaveteli_pro_info_request_path
+          expect(page).to have_content "Make a request"
+          select public_body.name, from: "To:"
+          fill_in "Summary:", with: "Does the pro request form work?"
+          fill_in "Your request:", with: "A very short letter."
+          select "3 Months", from: "Set embargo:"
+          click_button "Save draft"
+
+          # Redirected back to new request form
+          draft = DraftInfoRequest.where(title: "Does the pro request form work?")
```
crowbot: This is nitpicky, I know, but it'd be clearer to have the variable here be `drafts` as it's an array of drafts, and then set `draft` below when you pick the first one. Ditto with `info_request` below.

spec/integration/alaveteli_pro/request_creation_spec.rb
```diff
@@ -0,0 +1,209 @@
...
+                                     with: "A very short letter.")
+          expect(page).to have_select("Set embargo:", selected: "3 Months")
+        end
+      end
+    end
+
+    it "allows us to preview the request" do
+      with_feature_enabled(:alaveteli_pro) do
+        using_session(pro_user_session) do
+          # New request form
```
crowbot: Might be worth factoring this section out into `alaveteli_dsl.rb` as a reusable bit.

app/models/user.rb
stevenday: Oops, thanks for spotting that, I've dropped the annotation commit

spec/models/draft_info_request_spec.rb
stevenday: Done in 0e1880e

app/models/embargo.rb
stevenday: Done in 93b3eb4

stevenday: and c3f2823

app/controllers/alaveteli_pro/info_requests_controller.rb
stevenday: Done in 1d8e80d

spec/integration/alaveteli_pro/request_creation_spec.rb
stevenday: done in f9ca9dc

spec/integration/alaveteli_pro/request_creation_spec.rb
stevenday: Done in c5cf2d6

PR #3621, opened 2016-11-24
[Pro] WIP dashboard request list

crowbot:
Dev tweaks

- [x] Better/shorter status descriptions
- [x] Implement draft list
- [x] Pagination
- [x] Applying filters - how does non-js form submit? 
- [x] Change 'unknown status' to 'other'
- [x] Embargoed requests won't be in the text index - currently doing ilike on title (include authority and request body)
- [ ] Non javascript way of submitting action menus for each request. 
- [x] Give some indication of what phase and search (if any) is being used to filter the requests
- [x] Proper specs

Design tweaks
- [x] Visual separation of drafts in the project nav  - we're going to treat them as separate from requests
- [x] Make more compact vertically
- [x] Make work (at least minimally) on mobile
- [x] Add indication of embargo to request list
- [x] Get rid of duplicate filter widget
- [x] Filter and search reset option
- [x] Icons/colour to help phase and status stand out at a glance
- [x] More obvious submission for search
- [x] Blank slate design for when there are no requests yet


wrightmartin:
Here's what the blank slate will look like (copy is model's own)

screen shot 2016-12-02 at 11 57 46

wrightmartin:
Tidied up the requests list, made it fit better and added mobile styles

Large screens:

screen shot 2016-12-02 at 15 01 41

Med-small screens
screen shot 2016-12-02 at 15 01 54

Tiny screens
screen shot 2016-12-02 at 15 02 01

wrightmartin:
Embargo status display in request list

screen shot 2016-12-05 at 14 04 32

wrightmartin:
Search bar input method 

screen shot 2016-12-07 at 10 19 18

wrightmartin:
Icons on dashboard

screen shot 2016-12-08 at 09 54 51

wrightmartin:
^any ideas why the number next to 'all requests' doesn't add up right? Should be 4 there?

stevenday:
I've been clicking through this today and noticed a few things:
- The sort dropdown text gets slightly obscured when on the core theme, not sure if this is font size issue or what, but it happens on both mobile and desktop view:
screen shot 2016-12-14 at 11 26 08
- Embargoed requests don't link to anywhere, I presume that's intentional because it's waiting on #3643, but they could link to the normal request page for now and it would Just Work when we merge that
- Normal requests link to urls with IDs in, rather than the nice slugs, this will just result in an extra redirect, so it'll work, but it would be nice to avoid that
- On the onboarding view, the "start a new request" CTA looks like it's missing a link style in the core theme:
screen shot 2016-12-14 at 11 51 16


Very minor things
- The icons for the status in the request lists don't line up too well when the status breaks over two lines:
screen shot 2016-12-14 at 11 41 06
- Is it just my eyes, or do the embargo padlocks also need to come down a pixel or two to line up with the title text?
- Do we need retina versions of the icons?

wrightmartin:
- [ ] Retina icons
- [ ] Check styles between themed and non-themed alaveteli
- [ ] Bug in onboarding styles in core
- [ ] Icon spacing
- [ ] Cut-off text in labels

wrightmartin:
Ok, all updated with the above fixes

There are some more styles I'd like to remove from the core and put into WDTK/other themes, but suggest that work is done separately, as I'm sure there will be many more of these styles we can do all at the same time.

stevenday:
These fixups look good 👍

Issue #3412, opened 2016-07-28
Release 0.26


garethrees:

garethrees:
Made a milestone to track a couple of things I'd like to get in to 0.26 https://github.com/mysociety/alaveteli/milestone/47

PR #3624, opened 2016-11-29
Remove depreactions

garethrees:
For #3412


garethrees:
oops branch name. oh well.

PR #3630, opened 2016-12-01
[Pro] Add autocomplete to the authority select

stevenday:
This adds the selectize.js based autocomplete from the prototype to the pro
info request form. I've reworked the functionality slightly to make it
simpler and more robust:

- Authorities are loaded directly from the select options. Since we're
  providing a select box as a no-js fallback anyway, it seemed the easiest
  way to get that data into the js. Now there's no need for a new controller,
  any form of searching in Ruby, or any ajax stuff in the js.
- All the js and css is loaded from pro specific manifests, which are only
  included in the main application if pro is enabled.

This last bit is probably the most contentious as it makes the manifest an erb
file (more detail on the commit). An alternative would be to always build and
include the stuff from the pro manifests separately, e.g. adding a conditional
to the base layout. Any preferences?


app/assets/javascripts/alaveteli_pro/alaveteli_pro.js
```diff
@@ -1 +1,3 @@
-// ...
\ No newline at end of file
+// ...
+//= require alaveteli_pro/selectize
+//= require alaveteli_pro/authority_select
```
crowbot: Probably also want to include `authority_select.js`  in this commit.

app/assets/javascripts/alaveteli_pro/authority_select.js
```diff
@@ -0,0 +1,24 @@
...
+      valueField: 'id',
+      labelField: 'name',
+      searchField: 'name',
+      options: [],
+      create: false,
+      maxItems: 1,
+      render: {
+        option: function(body, escape) {
+          // No need to use escape because data is trusted (from our DB)
+          var html = '<div class="recipient-result">';
```
crowbot: Ultimately, we might want to put the HTML for this in a partial template, that gets rendered and hidden, and then reused for each item in the js with new values inserted into it.

Issue #3615, opened 2016-11-16
default_letter API param doesn't set default letter


garethrees:
https://www.whatdotheyknow.com/new/liverpool_city_council?title=test&default_letter=hi should pit "hi" between the salutation and signoff.

[Works on RTK](https://www.righttoknow.org.au/new/dibp?title=test&default_letter=hi) (currently on a 0.25 release) so might be in develop or missing something in whatdotheyknow-theme.

garethrees:
Also broken on http://demo.alaveteli.org/new/tgq?title=test&default_letter=hi

lizconlan:
It is a version thing, it works in 0.25.0.3 (per RTK), does not work in the latest build. Narrowing the search

lizconlan:
Huh, works in every 0.25.0.x branch using the "none" theme - but breaks on some deploys; broken on develop.

lizconlan:
going to try to work out what changed in develop and work out what's going wrong from there

lizconlan:
can get demo to break locally by mirroring deployed settings - alaveteli commit: 1c2a0206, theme commit: b26b78c0

lizconlan:
## Core code bug

This seems to [break for the first time here](https://github.com/mysociety/alaveteli/commit/62d8d482facf41c47f84afff87c33f4b39f319b9#diff-1fa1a855bf4fb5aaec94a56340bcc062) when we move to creating the `OutgoingMessage` directly as part of introducing strong params.

A [partial fix is then introduced](https://github.com/mysociety/alaveteli/commit/5d47531f85fa38a30050d47a5d91a772ac155ea3#diff-1fa1a855bf4fb5aaec94a56340bcc062) to avoid inadvertently overwriting default data with null values.

While #3627 is a useful workaround, the core problem appears to be that some values (including `default_letter`) are only evaluated when new `OutgoingMessage` objects are initialised (although `set_signature_name` may have its own workaround for that). A more comprehensive fix would address that as well to prevent future misunderstandings and potential problems with theme code.

lizconlan:
Just dragging this back for review for [the new alavetelitheme PR](https://github.com/mysociety/alavetelitheme/issues/87) as I think it demonstrates how to solve the theme override part of the puzzle.

garethrees:
Done

PR #3631, opened 2016-12-01
Updating OutgoingMessage.default_letter should update default body text

lizconlan:
Connects to #3615


garethrees:
Looks good pending build passing again.

PR #3633, opened 2016-12-01
[Pro] Issues/alaveteli pro 60/pro navbar

crowbot:
- [ ] OK to show user their email? 
- [ ] First name is just done by split
- [x] What links should pro user get? I've added their profile
- [ ] Need to intercept all new request links for pro users? 
- [x] Needs to use the feature flags? 
- [x] Take out mySociety header - when? 
- [x] What should happen on mobile?
- [x] Is whitelabel version ok? It requires js (underlining in `pro-pill`)
- [ ] Styles for alavetelitheme? 

To do
- [x] Fill in or remove account url
- [ ] Profile photo being resized in browser - can do on server side? 
- [x] Remove `greeting` and other obsolete css


app/views/general/_responsive_topnav.html.erb
```diff
@@ -1,38 +1,8 @@
-<div id="topnav" class="topnav">
+<div id="topnav" class="topnav <%= @pro_area ? 'pro' : '' %>">
```
stevenday: Separately, @wrightmartin has added an `in_pro_area` class to the body too, it seems like that would serve for all of the cases you have here, so perhaps we should standardise on that (naming aside)?

stevenday:
> Need to intercept all new request links for pro users?

I've ticketed this as https://github.com/mysociety/alaveteli-professional/issues/112

> Needs to use the feature flags?

If I'm reading the code correctly, the main things we're adding are styles and a body class. The body class bit could be wrapped in a `feature_enabled?` guard, but the styles we'll need to sort out as part of https://github.com/mysociety/alaveteli-professional/issues/104 so there's probably nothing to do now.

One minor thought from clicking around, should clicking anywhere else off the pro user's menu close it too?

stevenday:
> Is whitelabel version ok? It requires js (underlining in pro-pill)

Remit used a snazzy non-js version that might work here: https://github.com/mysociety/remit/blob/master/app/assets/stylesheets/_houdini.scss, basically you add a hidden radio button and a label for it instead of a link. Clicking the label allows you to show/hide the element immediately after it in the DOM, so if you can rejig the html for the user menu to work like that, it shouldn't need JS. It loses the slide down animation, but maybe @wrightmartin can work some css animation magic, or we can hack that back in with javascript as an enhancement.

stevenday:
@crowbot - yeah, that's fine. All looks good 🚢

Issue #3629, opened 2016-12-01
Delayed status seems to have red (rather than yellow) status bar


garethrees:
In develop:

![screen shot 2016-12-01 at 12 04 27](https://cloud.githubusercontent.com/assets/282788/20793344/7581717a-b7be-11e6-9c0a-dae54a5b0a51.png)

garethrees:
Might be my test data, so will check this

PR #3460, opened 2016-08-15
Redesign request status messages

wrightmartin:
This neatens up and redesigns the way request status appears around alaveteli.

Unfortunately, request states are too complex to adhere to the desired design #3269 

screen shot 2016-08-15 at 17 06 12
screen shot 2016-08-15 at 17 06 21

PR #3639, opened 2016-12-06
Match delayed status background with icon colour

garethrees:
Fixes https://github.com/mysociety/alaveteli/issues/3629

Delayed should be "pending" – not "failure" – given the icon is orange.

Also added line breaks to shared style selectors for clarity.

screen shot 2016-12-06 at 11 23 05

### DIFF
--- a/app/assets/stylesheets/responsive/_lists_style.scss
+++ b/app/assets/stylesheets/responsive/_lists_style.scss
@@ -82,7 +82,8 @@ $status-pending: #f4a140;
   }
 }
 
-.icon_waiting_response,.icon_waiting_clarification {
+.icon_waiting_response,
+.icon_waiting_clarification {
   background-image:image-url('status/waiting.png');
   background-color: $status-pending;
   background-size: 22px 22px;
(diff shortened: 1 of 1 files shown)


garethrees:
Also added a commit to dim the background colours as per https://github.com/mysociety/alaveteli/pull/3460#issuecomment-244033998.

screen shot 2016-12-06 at 12 30 28
screen shot 2016-12-06 at 12 30 32
screen shot 2016-12-06 at 12 30 36

lizconlan:
Possibly a partial fix - the new background colour makes the icons stand out but with a theme applied, I'm still getting coloured text and I'm not sure if the contrast is sufficient.

screen shot 2016-12-06 at 15 31 22

garethrees:
I _think_ that was the intention. @wrightmartin can you remember what was planned here?

lizconlan:
odd question/suggestion - at the risk of turning this into the tedious bit of an eye test... is the second one clearer?

screen shot 2016-12-06 at 15 31 22

screen shot 2016-12-06 at 15 50 09

garethrees:
Looks as good as identical to me

lizconlan:
spoiler: 70% vs 75% grey

lizconlan:
(comments about text don't really belong here, moving them to the ticket)

PR #3640, opened 2016-12-06
Alaveteli pro develop

crowbot:


app/assets/javascripts/request-attachments.js
```diff
@@ -0,0 +1,32 @@
+(function($) {
+  $(function () {
+    $('.correspondence .attachments').each(function() {
+      var limit = 3;
+      var link_collapsed_text = 'Show all attachments';
```
stevenday: Just a note that some un-translateable text has slipped in here, not sure if it was always there or has been messed up by moving this into a file?

crowbot: This should be fixed in https://github.com/mysociety/alaveteli/issues/3662, which is allocated to the current sprint for the core Alaveteli team.

app/models/info_request.rb
```diff
@@ -127,26 +146,9 @@ class InfoRequest < ActiveRecord::Base
...
-      'attention_requested',
-      'vexatious',
-      'not_foi'
-    ]
-    if @@custom_states_loaded
-      states += InfoRequest.theme_extra_states
-    end
-    states
+    warn %q([DEPRECATION] InfoRequest.enumerate_states will be removed in
+    0.28. It has been replaced by InfoRequest::States#all).squish
```
garethrees: s/States/State/ by the looks of it

crowbot: Good spot, fixed in fd8b652.

PR #3641, opened 2016-12-06
[Pro] Issues/alaveteli pro 40/dashboard

crowbot:
- [x] Where should we be linking the embargoes 'to-do' item to? 
- [x] 'Update statuses' should be singular when there's only one request to update.
- [x] Annotate models
- [x] Populate and update new fields? 
- [x] Filter for 'embargo_expiring' in request list
- [x] Filter for 'overdue', 'very_overdue' in request list
- [ ] Better icons for 'overdue' and 'very overdue' in request list - TBD in https://github.com/mysociety/alaveteli-professional/issues/120
- [x] Time zones if storing dates? 
- [x] Release notes

I've run into a couple of questions in attempting to implement the 'To-do' list part of this page. 

First, where to link to when there are multiple embargoes expiring this week. I've had a chat with @wrightmartin about this and we think the dashbaord request list with either a hidden or explicit 'embargo' filter set. 

Second, how to implement the 'there are n overdue requests' to-do item. We don't currently store an explicit field for each request that specifies its due date - we calculate this [on-the-fly](https://github.com/mysociety/alaveteli/blob/develop/app/models/info_request.rb#L717) where we need it. We could hack the query here - as the [overdue alert sending code does](https://github.com/mysociety/alaveteli/blob/develop/app/mailers/request_mailer.rb#L272) by getting the user's requests which might be overdue and then calculating if they are, but that seems not ideal. Additionally I suspect the overdue alert sending code of being a bit of a resource hog, so it may not make sense to keep using that approach. 

I _think_ we can save the `initial_request_last_sent_at`,  `date_response_required_by` and `date_very_overdue_after` values for a request to the database, as shown in this PR so far, which would make it simple to define some scopes that would return overdue or very overdue requests. The trickier bit would be to make sure that we update them when the request is resent. I'm thinking maybe we could do this via `log_event`. We then also have to populate the fields for all existing requests which might take some time. So I'm a bit torn between thinking this would be a useful improvement for core and thinking that it's overkill at this point.


stevenday:
> First, where to link to when there are multiple embargoes expiring this week. I've had a chat with @wrightmartin about this and we think the dashbaord request list with either a hidden or explicit 'embargo' filter set.

Yeah, I think there will need to be a page which lists them somehow. A nice js enhancement would be to just drop them down as a list from the todo item too?

> Second, how to implement the 'there are n overdue requests' to-do item. 

If you don't think it's much more work, I think I'd be inclined to follow this through now that you've done some of the work. Presumably updating the fields could also happen in a before_save callback? That way they'd always be populated whenever you save the request, and updating the existing ones would simply be a matter of looping over them and re-saving? If the query in the alert sending stuff is run regularly, it won't be much worse to run it once more to set up the field will it? Or is it the calculations themselves, rather than the looping over requests which is intensive?

stevenday:
A couple of thoughts whilst poking around with this that may or may not be helpful:

- Could we wrap up the TODOs as some kind of object (another ActiveModel thing perhaps?) that we then just list using [Rails' nice collection rendering](http://guides.rubyonrails.org/layouts_and_rendering.html#rendering-collections)? Maybe these should even be full models, it would make the listing page much easier, but it might add some overhead to keep track of when one is "done"?
- Assuming we keep things simple and don't save TODOs for now, it feels like we could probably use a helper or a method squirrelled away somewhere to produce the list of todos for a user, rather than having that logic in the view?
- Is the plan for the activity feed to use the existing info request events? It seems like some things are covered by it, but some would need to be added. In particular, stuff like a request becoming overdue or an embargo expiring - how will we add those? I guess we need something that can run via cron to build them up regularly.

spec/models/info_request_event_spec.rb
```diff
@@ -665,4 +665,78 @@
...
+
+  describe '#resets_due_dates?' do
+
+    it 'returns true if the event is a sending of the request' do
+      info_request_event = FactoryGirl.create(:sent_event)
+      expect(info_request_event.resets_due_dates?).to be true
+    end
+
+    it 'returns true if the event is a clarification' do
+      info_request_event = FactoryGirl.create(:sent_event)
```
crowbot: Nope, good spot. I've also added some comments to the `is_clarification?` method.

app/models/info_request/state/awaiting_response_query.rb
```diff
@@ -7,8 +7,10 @@ def initialize(relation = InfoRequest)
       end
 
       def call
-        @relation.where(described_state: 'waiting_response',
+        @relation
+          .where(described_state: 'waiting_response',
                         awaiting_description: false)
+            .where("date_response_required_by > ?", Time.now.to_date)
```
crowbot: Yeah, I think you're right (although it won't have too much effect unless the system time (which is what is returned by Time.now) is not UTC. I've fixed up these occurrences and done a separate [PR](https://github.com/mysociety/alaveteli/pull/3670) to be more consistent in the codebase as a whole.

stevenday:
This looks great now 👍

PR #3643, opened 2016-12-09
AlaveteliPro request in progress page

stevenday:
This mainly updates the `request/show.html.erb` view so that it does some
different stuff for pro requests:
 - Removes the existing sidebar and shows a new pro-specific one with an
   embargo extending form and a status update form.
 - Edits the action menu to remove things that pros won't need
 - Removes the prompting describe state form for pros

It also makes a few other changes to make this possible:
- Adds an EmbargoExtension model to deal with extending embargoes
- Adds a controller + action to extend embargoes
- Adds a controller + action to update statuses for pro requests
- Adds a new route for viewing requests in a pro context through the existing
  controller
- Redirects all pro users to that new route for their own requests
- Uses CanCanCan to authorise stuff in the new controllers and provides a
  default action for when there's an authorisation error

Initially I was going to make a whole new controller + view because there was
so much going on in the existing one, but after I started down that route, it
seemed worse to duplicate loads of stuff than to make the existing thing a
little bit more complicated. Instead, I did a lot of work to simplify the show
view and the related partials and controller action, so that I could add new
pro-related stuff without making it worse overall.

This will have to be viewed in the default theme because there's been a lot of
changes to the show view in develop which haven't made it into wdtk-theme yet,
so I've only been working with the core theme.

Some stuff I'm not totally sure about/haven't finished yet
- What happens when you do stuff in the action menu like annotating a request
  I think it should just work to redirect you back to the current page afterwards,
  which keeps things simple for now. The stub integration specs are there to
  remind me to do this.
- How we limit/present the options for the request status? I was kinda
  expecting to see logic for this in the model, should I add it (and refactor
  the existing describe_state partial to use it too?)
- Continuity checking to make sure all the pro stuff uses the new pro request
  route

Screenshots to follow


app/controllers/request_controller.rb
```diff
@@ -1085,4 +1086,20 @@ def redirect_numeric_id_to_url_title
       end
     end
   end
+
+  def redirect_pro_users
+    # Pro users should see their own requests in the pro page, so that if
+    # other site functions send them to a request page, they end up back in
+    # the pro area
+    if feature_enabled?(:alaveteli_pro) && \
```
crowbot: I have a general question, which is whether we want to show the pro view of a request to the request owner once it's published? Do we not want to show them the public request page once it's published?

stevenday: Hmm, interesting. I'd been thinking that since it was your request, you'd want to see it in the same place/livery all the time. However, you're probably right. I guess we need to do a little bit of redirecting both ways, so that an embargoed request goes to the pro page, and then a public one goes to the public version, no matter which url you hit them at.

app/controllers/alaveteli_pro/embargo_extensions_controller.rb
```diff
@@ -0,0 +1,32 @@
+# -*- encoding : utf-8 -*-
+# app/controllers/alaveteli_pro/embargo_extensions_controller.rb
+# Controller for embargo extensions
+#
+# Copyright (c) 2008 UK Citizens Online Democracy. All rights reserved.
+# Email: hello@mysociety.org; WWW: http://www.mysociety.org/
+
+class AlaveteliPro::EmbargoExtensionsController < AlaveteliPro::BaseController
+  def create
```
stevenday: I felt like this was more rails' "golden path" because of the simpler form builders, validation and so on. It also keeps a record of extensions to a particular embargo, which I thought might be valuable in the future, even if just for usage tracking. 

I'm not wed to it though, happy to turn it into an update instead if you'd rather..

app/views/request/_request_meta_tags.html.erb
```diff
@@ -0,0 +1,7 @@
+<% if @info_request.prominence(:decorate => true).is_public? %>
+  <% content_for :open_graph_tags do %>
+    <meta property="article:published_time" content="<%= @info_request.created_at.iso8601 %>" />
+    <meta property="article:modified_time" content="<%= @info_request.updated_at.iso8601 %>" />
+    <meta property="og:description" content="<%= @info_request.initial_request_text %>" />
+  <% end %>
+<% end %>
```
stevenday: Done, sorry!

crowbot: Just done mine too ;)

app/helpers/info_request_helper.rb
```diff
@@ -13,6 +13,23 @@ def status_text(status, opts = {})
     end
   end
 
+  def awaiting_description_text(info_request,
+                                is_owning_user,
```
stevenday: Done in 37014e46e9f18e054b7502cbf07af6c090a71462

spec/helpers/info_request_helper_spec.rb
```diff
@@ -409,4 +409,127 @@
 
   end
 
+  describe '#awaiting_description_text' do
+    let(:info_request) { FactoryGirl.create(:info_request) }
+
+    shared_examples_for 'when we can\'t ask the user to update the status' do
```
stevenday: Done in e0c383d

app/views/request/_request_status.html.erb
```diff
@@ -0,0 +1,33 @@
+<%# Partial view that displays a line of text displaying the current status
+    of a request %>
+<% if @info_request.awaiting_description %>
```
stevenday: Done in 0ace4c0a8eb6fb97df476ad04c414c100695a8d0

app/views/alaveteli_pro/info_requests/new.html.erb
```diff
@@ -62,16 +62,15 @@
...
           <p>
             <label class="form_label" for="outgoing_message_body">
-              <%= _('Your request:') %></label>
+              <%= _('Your request') %></label>
             <%= o.text_area :body, :rows => 20, :cols => 60 %>
           </p>
         <% end %>
 
         <div class="form_button">
-          <button><%= _("Save draft") %></button>
```
stevenday: Sorry, it's not explained but these need to be buttons so that they both submit the form, but we can pass a name and value with the preview one (it doesn't work if they're submit input unfortunately). Can we style the <button> tags the same here, or is that breaking some other semantics?

Issue #3651, opened 2016-12-19
Write a script to check and ban users by domain


crowbot:
We have several email domains associated with new accounts created daily on WDTK that are apparently not being used for anything. It would be useful to have a script that accepts a domain and returns whether a request, comment or subscription has ever been made by an account at that domain, how many accounts overall have been created, and gives an option to ban all accounts from that domain (and perhaps prevent new ones signing up).

lizconlan:
With my usual mild guilt of writing something that might only work for Postgres, the key to this would seem to be this style of thing:

```
SELECT substring(email, position('@' in email)+1) AS domain, 
            COUNT(id) AS count
FROM users
WHERE admin_level = 'none'
GROUP BY domain
ORDER BY count DESC;
```

lizconlan:
```
User.where(:admin_level => 'none').
     where("email like ?", "%@#{domain}").
     count
```

garethrees:
Marking as current sprint as this needs to get run to be completed

lizconlan:
Merged to develop

lizconlan:
Initial report for discussion here (currently internal only), not as many candidates for a blanket ban as I expected but a few domains to consider treating as a factor in detecting new spam signups.

https://docs.google.com/a/mysociety.org/document/d/1FqEKBcc5jvHHyzJaroYv20roN6rBbgQeNe32527IMNk/

I ended up generating the list of all the domains then looking at any up that weren't immediate familiar (or were thought to be spam sources) where there were more than 200 signups* (roughly the top 30, plus anything that looked particularly odd).

\*the 2 domains with the greatest number of signups are comfortably into 5 figures

lizconlan:
Done, new ticket #3752 opened for followup action

PR #3654, opened 2016-12-20
Add some rake tasks for dealing with users on a per domain basis

lizconlan:
Connects to #3651 

Contains 3 hopefully useful rake tasks...

## top20

Produces a top 20 (or fewer) table listing number of signups per domain based on email address
 
e.g.

```
" domain            |  count "
"------------------ | -------"
" gmail.com         |  332"
" hotmail.com       |  93"
" hotmail.co.uk     |  57"
" yahoo.co.uk       |  51"
" yahoo.com         |  50"
" webgarden.com     |  41"
" outlook.com       |  31"
```

Accepts optional ENV var `START_DATE` to limit scope (date range is inclusive)

## stats

Prints the total number of users and the percentage of spammers and dormant accounts for the required domain

e.g.

```
"total users: 5"
"     spam %: 40% (2)"
"  dormant %: 40% (2)"
```

Requires the ENV var `DOMAIN` (e.g. 'hotmail.com')
Accepts optional ENV var `START_DATE` to limit scope (date range is inclusive)

## ban_users

Runs the stats task then asks if you want to go ahead and ban all those users. Be warned, it does not ask if you're sure.

Requires the ENV var `DOMAIN` (e.g. 'hotmail.com')
Accepts optional ENV var `START_DATE` to limit scope (date range is inclusive)

### DIFF
--- a/lib/tasks/email_domain_actions.rake
+++ b/lib/tasks/email_domain_actions.rake
@@ -1,0 +1,144 @@
+# -*- encoding : utf-8 -*-
+namespace :email_domain_actions do
+
+  desc "Lists top 20 most used email domains"
+  task :top20 => :environment do
+    # the limit of 20 is fairly arbitrary but I imagine this being more
+    # annoying than useful otherwise
+
+    from = ENV["START_DATE"]
+
+    sql = if from
+      "SELECT substring(email, position('@' in email)+1) AS domain, " \
+      "COUNT(id) AS count " \
+      "FROM users " \
+      "WHERE admin_level = 'none' AND created_at >= '#{from}' " \
+      "GROUP BY domain " \
+      "ORDER BY count DESC " \
+      "LIMIT 20"
+    else
+      "SELECT substring(email, position('@' in email)+1) AS domain, " \
+      "COUNT(id) AS count " \
+      "FROM users " \
+      "WHERE admin_level = 'none' " \
+      "GROUP BY domain " \
+      "ORDER BY count DESC " \
+      "LIMIT 20"
+    end
+    results = User.connection.select_all(sql)
+
+    column1_width = results.map { |x| x["domain"].length }.sort.last
+
+    p "Since #{from}..." if from
+
+    p " domain ".ljust(column1_width + 2, " ") + " | " + " count "
+    p "--------".ljust(column1_width + 2, "-") + " | " + "-------"
+
+    results.each do |result|
+      p " #{result["domain"].ljust(column1_width, " ")}  |  #{result["count"]}"
+    end
+
+  end
+
+  desc "Lists per domain stats"
+  task :stats => :environment do
+    raise "must supply a DOMAIN value" unless ENV["DOMAIN"]
+    domain = ENV["DOMAIN"]
+    from = ENV["START_DATE"]
+
+    total_users = if from
+      User.where("email like ?", "%@#{domain}").
+           where(:admin_level => 'none').
+           where("created_at >= ?", from).
+           count
+    else
+      User.where("email like ?", "%@#{domain}").
+           where(:admin_level => 'none').
+           count
+    end
+
+    spammers = if from
+      User.where("email like ?", "%@#{domain}").
+           where("ban_text != ''").
+           where("created_at >= ?", from).
+           count
+    else
+      User.where("email like ?", "%@#{domain}").
+           where("ban_text != ''").
+           count
+    end
+
+    spam_percent = if total_users == 0
+      0
+    else
+      (spammers.to_f / total_users * 100).round
+    end
+
+    # When we have Rails 4 across the board, we get to say "where.not" and
+    # we rewrite this using the ORM
+    # example code here: http://stackoverflow.com/a/23389130), until then...
+    #
+    # Reminder - check that the returned ids in the subquery does not include
+    # null values otherwise this will unexpectedly return 0
+    # (see http://stackoverflow.com/a/19528722) this should not be a thing but
+    # is happening on WDTK with the info_requests table for some reason
+    sql = "SELECT count(*) FROM users " \
+          "WHERE id NOT IN ( " \
+          "  SELECT DISTINCT user_id FROM info_requests " \
+          "  WHERE user_id IS NOT NULL " \
+          ") AND id NOT IN ( " \
+          "  SELECT DISTINCT tracking_user_id FROM track_things " \
+          "  WHERE tracking_user_id IS NOT NULL " \
+          ") AND id NOT IN ( " \
+          "  SELECT DISTINCT user_id FROM comments " \
+          "  WHERE user_id IS NOT NULL " \
+          ") AND email like '%@#{domain}'"
+    sql += "AND created_at >= '#{from}'" if from
+    dormant = User.connection.select_all(sql).first["count"]
+
+    dormant_percent = if total_users == 0
+      0
+    else
+      (dormant.to_f / total_users * 100).round
+    end
+
+    p "Since #{from}..." if from
+    p "total users: #{total_users}"
+    p "     spam %: #{spam_percent}% (#{spammers})"
+    p "  dormant %: #{dormant_percent}% (#{dormant})"
+  end
+
+  desc "Bans all users for a specific domain"
+  task :ban_users => :environment do
+    raise "must supply a DOMAIN value" unless ENV["DOMAIN"]
+    domain = ENV["DOMAIN"]
+    from = ENV["START_DATE"]
+
+    Rake.application.invoke_task("email_domain_actions:stats")
+
+    p ""
+
+    message = "Do you want to ban all the users for #{domain}"
+    message += " created on or after #{from}" if from
+    message += "(y/N)"
+    p message
+    input = STDIN.gets.strip
+
+    if input.downcase == "y"
+      count = if from
+        User.where("email like ?", "%@#{domain}").
+             where(:admin_level => 'none').
+             where("created_at >= ?", from).
+             update_all(:ban_text => "Banned for use of #{domain} email")
+      else
+        User.where("email like ?", "%@#{domain}").
+             where(:admin_level => 'none').
+             update_all(:ban_text => "Banned for use of #{domain} email")
+    end
+      p "#{count} accounts banned"
+    else
+      p "No action taken"
+    end
+  end
+
+end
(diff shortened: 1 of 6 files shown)


lib/tasks/email_domain_actions.rake
```diff
@@ -0,0 +1,144 @@
+# -*- encoding : utf-8 -*-
+namespace :email_domain_actions do
+
+  desc "Lists top 20 most used email domains"
+  task :top20 => :environment do
+    # the limit of 20 is fairly arbitrary but I imagine this being more
+    # annoying than useful otherwise
```
garethrees: I don't think there should be a limit. I If I want to limit it I can just run `rake email_domain_actions:count_by_domain | head -n 20`

lib/tasks/email_domain_actions.rake
```diff
@@ -0,0 +1,144 @@
...
+
+    from = ENV["START_DATE"]
+
+    sql = if from
```
garethrees: It would be nice to see stuff like this in a lib class with methods that just return data, so that they can be tested (particularly valuable if we're about to ban a domain's worth of users :)). Don't care about testing the output formatting though.

lib/tasks/email_domain_actions.rake
```diff
@@ -0,0 +1,144 @@
...
+           where(:admin_level => 'none').
+           where("created_at >= ?", from).
+           count
+    else
+      User.where("email like ?", "%@#{domain}").
+           where(:admin_level => 'none').
+           count
+    end
+
+    spammers = if from
```
garethrees: I don't think we should use the term spammers here (in var name or output) – users could be banned for other reasons. When looking at the PR description I had thought that this was using the spam scorer.

lib/tasks/email_domain_actions.rake
```diff
@@ -0,0 +1,144 @@
...
+    end
+
+    # When we have Rails 4 across the board, we get to say "where.not" and
+    # we rewrite this using the ORM
+    # example code here: http://stackoverflow.com/a/23389130), until then...
+    #
+    # Reminder - check that the returned ids in the subquery does not include
+    # null values otherwise this will unexpectedly return 0
+    # (see http://stackoverflow.com/a/19528722) this should not be a thing but
+    # is happening on WDTK with the info_requests table for some reason
```
garethrees: As above, lets extract this to a tested class so that we can be sure

lib/tasks/email_domain_actions.rake
```diff
@@ -0,0 +1,144 @@
...
+
+    dormant_percent = if total_users == 0
+      0
+    else
+      (dormant.to_f / total_users * 100).round
+    end
+
+    p "Since #{from}..." if from
+    p "total users: #{total_users}"
+    p "     spam %: #{spam_percent}% (#{spammers})"
```
garethrees: I found the formatting a bit difficult to understand at first. I think it would be clearer for the banned (née spam) and dormant lines to show the percentage in brackets:

```
total users: 1000
     banned: 10 (1%)
    dormant: 800 (80%)
```

lib/tasks/email_domain_actions.rake
```diff
@@ -0,0 +1,144 @@
...
+    p "  dormant %: #{dormant_percent}% (#{dormant})"
+  end
+
+  desc "Bans all users for a specific domain"
+  task :ban_users => :environment do
+    raise "must supply a DOMAIN value" unless ENV["DOMAIN"]
+    domain = ENV["DOMAIN"]
+    from = ENV["START_DATE"]
+
+    Rake.application.invoke_task("email_domain_actions:stats")
```
garethrees: Worth confirming that this retains the `ENV` (I haven't checked and I assume it does, but looks like it could be a gotcha)

lib/tasks/email_domain_actions.rake
```diff
@@ -0,0 +1,144 @@
...
+
+    p ""
+
+    message = "Do you want to ban all the users for #{domain}"
+    message += " created on or after #{from}" if from
+    message += "(y/N)"
```
garethrees: Usually its `Y/n` to make it harder to perform a destructive action

lib/tasks/email_domain_actions.rake
```diff
@@ -0,0 +1,144 @@
...
+      count = if from
+        User.where("email like ?", "%@#{domain}").
+             where(:admin_level => 'none').
+             where("created_at >= ?", from).
+             update_all(:ban_text => "Banned for use of #{domain} email")
+      else
+        User.where("email like ?", "%@#{domain}").
+             where(:admin_level => 'none').
+             update_all(:ban_text => "Banned for use of #{domain} email")
+    end
```
garethrees: bad indent

lib/tasks/email_domain_actions.rake
garethrees: Would be nice to use heredocs too

lib/tasks/email_domain_actions.rake
```diff
@@ -0,0 +1,144 @@
...
+  end
+
+  desc "Lists per domain stats"
+  task :stats => :environment do
+    raise "must supply a DOMAIN value" unless ENV["DOMAIN"]
+    domain = ENV["DOMAIN"]
+    from = ENV["START_DATE"]
+
+    total_users = if from
+      User.where("email like ?", "%@#{domain}").
```
garethrees: Nice to use capitalisation for sql keywords

lib/tasks/email_domain_actions.rake
```diff
@@ -0,0 +1,144 @@
...
+
+    if input.downcase == "y"
+      count = if from
+        User.where("email like ?", "%@#{domain}").
+             where(:admin_level => 'none').
+             where("created_at >= ?", from).
+             update_all(:ban_text => "Banned for use of #{domain} email")
+      else
+        User.where("email like ?", "%@#{domain}").
+             where(:admin_level => 'none').
```
garethrees: I think we should exclude already banned users from this query, in case they were banned for a different reason.

lib/tasks/email_domain_actions.rake
```diff
@@ -0,0 +1,144 @@
+# -*- encoding : utf-8 -*-
+namespace :email_domain_actions do
```
garethrees: I think a nicer namespacing structure might be:

```
users
  count_per_domain
  stats_by_domain
  ban_by_domain
```

lizconlan:
>  I think extracting in to a/some tested class(/es) would be worth it, especially as we'll be banning multiple users on a fairly regular basis.

Strongly agree. First pass was rushed in case it needed to be used (or cherry picked for useful SQL queries) over the break

lib/tasks/email_domain_actions.rake
lizconlan: It's a confirm prompt - probably badly phrased - so the destructive action is `Y` in this case

garethrees: Yeah – usually the destructive should be `Y` (rather than `y`) IYSWIM

lizconlan: Wouldn't that imply that Y is the default response?

lib/tasks/email_domain_actions.rake
lizconlan: calling the lib class `UserStats` for now - happy to rename it later but just `Users` felt too close to  key model name

garethrees: +1

spec/lib/user_stats_spec.rb
```diff
@@ -86,4 +86,41 @@
...
+                                   :email => "banned@example.com",
+                                   :ban_text => "Banned")
+
+      Delorean.back_to_the_present
+      @user2 = FactoryGirl.create(:user, :email => "newbie@example.com")
+      @admin = FactoryGirl.create(:admin_user, :email => "admin@example")
+    end
+
+    it "returns a list of eligible users" do
+      expect(UserStats.unbanned_by_domain("example.com").count).to eq(2)
```
garethrees: Probably want to check they're the right users, not just any two?

```ruby
.to match_array([@user1, @user2])
```

lib/user_stats.rb
```diff
@@ -27,4 +27,30 @@ def self.list_user_domains(start_date=nil, limit=nil)
...
   end
 
+  # Returns the number of domant users for the given domain
+  # (A dormant user is one with no requests, tracks or comments)
+  def self.count_dormant_users(domain, start_date=nil)
+    # When we have Rails 4 across the board, we get to say "where.not" and
+    # rewrite this using the ORM
+    # example code here: http://stackoverflow.com/a/23389130), until then...
+    #
+    # Reminder - check that the returned ids in the subquery does not include
```
lizconlan: As far as it can be, reworded as the note for future maintainers that I think my brain intended to send

lizconlan:
Changed the specs for `.count_dormant_users` to build up example data because 1) super unclear and 2) it wasn't demonstrating what it was supposed to (see point 1)

lib/user_stats.rb
```diff
@@ -34,10 +34,10 @@ def self.count_dormant_users(domain, start_date=nil)
...
     # example code here: http://stackoverflow.com/a/23389130), until then...
     #
-    # Reminder - check that the returned ids in the subquery does not include
-    # null values otherwise this will unexpectedly return 0
-    # (see http://stackoverflow.com/a/19528722) this should not be a thing but
-    # is happening on WDTK with the info_requests table for some reason
+    # If extending this query, remember to check that the returned ids in the
+    # subquery do not include null values otherwise it will unexpectedly return 0
+    # (see http://stackoverflow.com/a/19528722) this behaviour has been
+    # observed with the info_requests table
```
garethrees: I'm really not clear on this comment.

Is it explaining why L45 `WHERE user_id IS NOT NULL` (and L48 and L51) is in place, rather than only using a `SELECT`?

lizconlan: Yes as it seems like an odd thing to do and the `JOIN table ON ... WHERE` syntax is pretty horrible for this particular query IIRC. Could be better phrased for clarity

lib/tasks/users.rake
```diff
@@ -5,7 +5,11 @@ namespace :users do
   task :count_per_domain => :environment do
     from = ENV["START_DATE"]
 
-    results = UserStats.list_user_domains(from)
+    results = if from
+      UserStats.list_user_domains(:start_date => from)
+    else
+      UserStats.list_user_domains
+    end
```
lizconlan: hmm, yes - need to fix that

garethrees:
Looks good. Don't forget changelog notes.

PR #3655, opened 2016-12-22
Render status text from ruby.

crowbot:


stevenday:
This looks like a no-brainer: 👍

PR #3656, opened 2016-12-22
Refactor status update options into InfoRequest::State

stevenday:
This takes the status options and the logic for displaying them out of the
`_describe_state.html.erb` and `_other_describe_state.html.erb` and places
them into the new InfoRequest::State module instead. Then this is used to
update the status options in the pro sidebar as well, so that the options
presented there are logical and consistent with what normal users can update
(but with different labels).

I've added classes to enable us to hook in some styling from @wrightmartin's
design for this form, but I don't have the right icons available yet so I
haven't actually connected them up.

This is based on a weird merge branch until #3643 and #3621 are both merged
into alaveteli-pro-develop.


stevenday:
One thing to say, I haven't put very much thought into the text of the pro versions of these labels, so any comments there would be much appreciated.

app/models/info_request/state/calculator.rb
```diff
@@ -34,6 +34,73 @@ def phase(cached_value_ok=false)
           end
         end
       end
+
+      def transitions(opts = {})
+        cached_value_ok = opts.fetch(:cached_value_ok, false)
+        state = @info_request.calculate_status(cached_value_ok)
+        if complete_states(include_admin_states: true).include?(state)
```
crowbot: We might want to retain the behaviour just for the admin states (which a user can't put a request into). However, we'd probably just want to remove the option to update the status of a request from the action menu in that case.

stevenday:
@crowbot - thanks for the comments, and fair point about changing the status back if you do it by accident. 

I've updated the status options as suggested (so that only `vexatious` or `not_foi` responses return no options for new states), and explicitly turned off the status updating action and hidden the form if the request has no possible states. I've also fixed up the other minor issues.

PR #3657, opened 2016-12-23
[Pro] Issues/alaveteli pro 40/dashboard activity

crowbot:
- [x] Tests
- [x] Add example cron
- [x] Can we backload overdue and very overdue events for requests that were overdue but aren't anymore? 
- [ ] Should embargo `publish_at` be stored as a date rather than a `datetime`? Should we expire them all at midnight? 
- [ ] Should the cron stuff be integrated with `RequestMailer.alert_overdue_requests`?

This PR also closes https://github.com/mysociety/alaveteli-professional/issues/113


spec/models/activity_list/item_spec.rb
```diff
@@ -0,0 +1,17 @@
+# -*- encoding : utf-8 -*-
+require 'spec_helper'
+
+describe ActivityList::Item do
+
+  describe '.new' do
+    it 'requires a user, page and per_page arguments' do
```
stevenday: Copy/paste error? I think this should just be checking for an event?

spec/models/activity_list/new_response_spec.rb
```diff
@@ -0,0 +1,35 @@
+# -*- encoding : utf-8 -*-
+require 'spec_helper'
+
+describe ActivityList::NewResponse do
+  include Rails.application.routes.url_helpers
+
+  before do
+    @event = FactoryGirl.create(:response_event)
```
stevenday: Any reason these aren't in a `let` or `let!` instead?

stevenday: (same goes for all the Item specs)

lib/tasks/temp.rake
```diff
@@ -16,6 +16,44 @@ namespace :temp do
     end
   end
 
+  def backload_overdue(event_type, verbose)
+    overdue_alerts = UserInfoRequestSentAlert.
```
stevenday: Is there any need to assign to a variable like this and loop over it at the same time? It doesn't seem like the assigned var is used afterwards, and it makes the loop harder to see.

lib/tasks/temp.rake
```diff
@@ -16,6 +16,44 @@ namespace :temp do
     end
   end
 
+  def backload_overdue(event_type, verbose)
```
stevenday: Reading through this in more detail, aside from the ability to write some output per event, is there any functional difference between what this is doing and just calling `InfoRequest.log_overdue_events` or `InfoRequest.log_very_overdue_events`?

stevenday:
Generally, I can see why you're unhappy about faking the dates with the events, but I can't think of a better solution either, aside from stopping using created_at and instead having an event_date column or similar instead?. 

Regarding the questions:

> Should embargo publish_at be stored as a date rather than a datetime? Should we expire them all at midnight?

I initially thought about storing it as a date, but it does leave it less explicit (say, for database spelunking in the event of bug) as to when it expires, I think I'd prefer it as a datetime set to midnight on the date the user selected. Expiring them all at midnight was what I thought was best.

> Should the cron stuff be integrated with RequestMailer.alert_overdue_requests?

What would be the potential benefit? Personally I'd assume it's better to have things in smaller chunks, so that (e.g.) we could spin off alerting onto a separate app server or something, or run them at different intervals.

lib/tasks/temp.rake
stevenday: I get now that it's based on alerts being sent, rather than inspecting the list of events to see if the request is overdue directly (it caught me out because locally I hadn't sent the alerts, so I wasn't getting any items for overdue requests). Presumably this is intended to be more performant? If so, is it a big difference, because it seems clearer to call the method on the model rather than rely on the alerts having been sent?

spec/models/info_request_event_spec.rb
```diff
@@ -65,57 +65,46 @@
...
-                                                :info_request => info_request)
-      expect(info_request_event.indexed_by_search?).to be_truthy
+    it 'returns a truthy value for an incoming message that is indexed by search' do
+      incoming_message = FactoryGirl.create(:incoming_message)
+      response_event = FactoryGirl.build(:response_event,
+                                         :incoming_message => incoming_message)
+      expect(response_event.indexed_by_search?).to be_truthy
     end
 
     it 'should return a falsey value for an outgoing message that is not indexed by search' do
```
crowbot: Fixed in 229308b.

spec/models/info_request_spec.rb
```diff
@@ -3096,4 +3106,151 @@ def email_and_raw_email(opts = {})
...
 
+  describe '.log_overdue_events' do
+
+    let(:info_request){ FactoryGirl.create(:info_request) }
+
+    context 'when an InfoRequest is not overdue' do
+
+      it 'does not create an event' do
+        time_travel_to(Date.parse('2014-12-31')){ info_request }
+        time_travel_to(Date.parse('2015-01-15')) do
```
crowbot: Much nicer - fixed in ecd5971

app/models/activity_list/embargo_expiry.rb
```diff
@@ -0,0 +1,13 @@
+module ActivityList
+  class EmbargoExpiry < Item
+
+    def description
+      N_('The embargo for your request to {{public_body_name}} "{{info_request_title}}" has ended and is now live.')
```
crowbot: Sounds good, fixed in 50dcfa2.

spec/models/activity_list/item_spec.rb
crowbot: Indeed. Fixed in 7d30acd.

spec/models/activity_list/new_response_spec.rb
crowbot: TBH, I forgot about `let!` 😮 Fixed in aedb478.

lib/tasks/temp.rake
crowbot: It's not performance that was the issue - I did it this way because I couldn't think of another way to find requests that _had_ been overdue at one point, but aren't overdue now. `InfoRequest.log_overdue_events` and `InfoRequest.log_very_overdue_events` only insert `InfoRequestEvents` for `InfoRequests` that are currently overdue (or very overdue).

lib/tasks/temp.rake
crowbot: Good spot, fixed in 3e315a7.

spec/models/info_request_spec.rb
crowbot: Ah, sorry, scatch that - I did it using concrete dates so that the working day calculations would be predictable - otherwise the results would vary depending on what day of the week you run the tests on. Reverting ecd5971.

lib/tasks/temp.rake
stevenday: Aha, I see - ok, that makes sense, I hadn't thought about the need for that.

stevenday:
>  I've pushed a compromise fix in af815e7 that records the time the event was actually created in it's params. I'm a bit loath to add more big changes now to the stack we have for rollout at one time. What do you think?

Yep, this is a good compromise for now

>  the created_at timestamp of the InfoRequestEvent will reflect the actual time the embargo object was destroyed and thus the request became public, but it will be as close to midnight as possible. Sound good?

Perfect, I don't think anyone is going to mind if it's a few minutes later

> It would be to have less duplicated logic - I'm a bit aware that with this PR we're now looking for overdue requests in several places. The downside would be messing with code that's currently working and generally increasing the size of the changes we're rolling out with alaveteli-pro-develop - maybe it's one to come back to later?

Agreed, it would be nice to tidy this stuff up into a single place (maybe somewhere like the scopes module for InfoRequest I guess) but it doesn't have to be perfect now. Maybe something to just ticket, referring to the issues with the cron's lookup and backfilling historical events?

Overall: 🚢

PR #3658, opened 2016-12-23
[Pro] Redirect pros to their request form from the non-pro process

stevenday:
Redirects the select_authority page (so that most links to make a request
work automatically for pros) and then the actual request form (in case anyone
has bookmarked the url or somehow gets direct linked to it). In the latter
case I've made sure that the selected public body is carried through too.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -340,11 +340,22 @@ class ApplicationController < ActionController::Base
   # if they start making a request, then we realise they're a pro when they
   # log in, so we want to send them into the pro system
   def override_post_redirect_for_pro(uri, post_redirect, user)
-    case uri
-    when "/new"
-      # TODO: Send this to the new 'draft' controller action for pros
-      # instead. I guess we could just update the uri, but we might need to
-      # tweak the saved params too.
+    # We could have a locale in the url, or we could not, e.g. /en/new or /new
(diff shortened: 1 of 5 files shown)


stevenday:
@crowbot - I wasn't originally planning to include that too, but it makes sense, so I've added it now.

When you log in as a pro it now sends you on, saving a draft in the process. Here's what it looks like afterwards:

screen shot 2017-01-04 at 15 56 50

I'm not sure about the message though, it's a bit long. Is there any other precedent in Alaveteli for showing a custom message in a situation like this, or should I stick with the flash?

app/controllers/application_controller.rb
```diff
@@ -340,11 +340,22 @@ def add_post_redirect_param_to_uri(uri)
   # if they start making a request, then we realise they're a pro when they
   # log in, so we want to send them into the pro system
   def override_post_redirect_for_pro(uri, post_redirect, user)
-    case uri
-    when "/new"
-      # TODO: Send this to the new 'draft' controller action for pros
-      # instead. I guess we could just update the uri, but we might need to
-      # tweak the saved params too.
+    # We could have a locale in the url, or we could not, e.g. /en/new or /new
+    if uri =~ /(\/[a-z]{2})?\/new/
```
crowbot: Locales can be longer than two characters - they can be two chars, a hyphen and then another two chars.

PR #3659, opened 2016-12-31
[Pro] Enable pro users to do things to embargoed requests

stevenday:
embargoed request. This PR adjusts those where a pro user *should* be able to
do something to one of their embargoed requests so that they can.

Todo:
- [x] Sending a followup / writing a reply
- [x] Downloading as a zip file
- [x] Adding an annotation
- [x] Requesting an internal review

Questions:
- [x] Follow ups currently require you to be logged in, but have special logic
      for if you are not the request owner (rather than just 404/403ing the
      request to follow up). Why is this? (It could simplify the code a fair
      bit if we removed it and I can't see an obvious UX reason over just
      hiding the follow up action from non-owners completely).
- [x] A lot of what I've added feels like authorization stuff that could be in cancancan instead, but
       because everything is doing that slightly differently, it's hard to separate/generalise. 
       For example, we  could just have a `can(:followup, request)` or `can(:annotate, request)`. The 
       main issue is that each view reacts slightly differently to failures - sometimes it's a 403 page, 
       other times a custom message, or a 404 for embargoed requests, etc. Does anyone have any 
       ideas on how solve that? (It's probably a separate thing that we need to consider as an 
       improvement/overhaul to core I guess.

### DIFF
--- a/app/views/alaveteli_pro/comment/_suggestions.html.erb
+++ b/app/views/alaveteli_pro/comment/_suggestions.html.erb
@@ -1,0 +1,34 @@
+<p>
+  <%= _("Annotations allow you to add extra notes on a request, which will " \
+          "eventually be made public. For example:")%>
+</p>
+
+<ul>
+  <% if [ 'successful', 'partially_successful' ].include?(@info_request.described_state) %>
+    <li><%= _(' <strong>Summarising</strong> the content of any information returned. ')%></li>
+    <li><%= _(" Say how you\'ve <strong>used the information</strong>, with " \
+                "links if possible.")%> </li>
+    <li><%= _("<strong>Thank</strong> the public authority.") %></li>
+    <li> <%= _("Point to <strong>related information</strong>, campaigns or " \
+                 "forums which may be useful.")%></li>
+  <% else %>
+    <li><%= _("Provide extra context on the request and your reasons for " \
+                "making it.") %></li>
+  <% end %>
+
+  <% if [ 'gone_postal' ].include?(@info_request.described_state) %>
+    <li> <%= _("A <strong>summary</strong> of the response if you have " \
+                  "received it by post. ")%></li>
+  <% end %>
+
+  <% if [ 'error_message' ].include?(@info_request.described_state) %>
+    <li> <%= _("You know what caused the error, and can <strong>suggest a " \
+                  "solution</strong>, such as a working email address.")%> </li>
+  <% end %>
+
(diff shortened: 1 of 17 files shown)


stevenday:
> The comment and followup views both have text that says that your contributions are going to be public - I think we probably need to change that for an embargoed request to say that they'll be public when the request is published.

Yep, I missed this, I'll sort it out now.

> On a related question, do we want to use the AP header when you're looking at an embargoed request or performing followup actions on one? I think if the idea is that it shows when what you're doing is private, then we probably should.

It should already show on the request page, I think it's just my error not changing the variable name that the controller sets to turn the header on. I've made mysociety/alaveteli-professional#117 to sort this out. I'll add the same for the other relevant pages too.

stevenday:
I've added some text variations to the comment and followup views now, they look like this for pros:

Comments:
screen shot 2017-01-05 at 17 58 09

Follow ups:
screen shot 2017-01-05 at 17 21 57

Replies:
screen shot 2017-01-05 at 17 57 16

stevenday:
FYI: This needs a matching PR in WDTK-theme: mysociety/whatdotheyknow-theme#371 in order to see the new replies/followups bit, as this overrides the followup form view.

stevenday:
@crowbot - this should be ready for another review now. I've fixed up your comments, and fixed mysociety/alaveteli-professional#117 at the same time, so all the headers should be the right colour throughout.

stevenday:
I've now added a pro-specific list of suggestions for the comment content. I tried to re-use the bullet points that were still appropriate, but one of the big ones we lost was having something to say when the request hasn't completed successfully (either because it's not had a response, or because it's unsuccessful in various forms). This is used as a fallback in the existing template to make sure that there's always something shown, so I added a new generic point which could probably be improved.

app/views/alaveteli_pro/comment/_suggestions.html.erb
```diff
@@ -0,0 +1,39 @@
...
+                "links if possible.")%> </li>
+    <li><%= _("<strong>Thank</strong> the public authority.") %></li>
+  <% end %>
+
+  <% if not [ 'successful', 'partially_successful' ].include?(@info_request.described_state) %>
+    <li><%= _("Provide extra context on the request and your reasons for " \
+                "making it.") %></li>
+  <% end %>
+
+  <% if [ 'successful', 'partially_successful' ].include?(@info_request.described_state) %>
```
crowbot: Looks like this could be merged with the two previous `if` conditionss as an `if..else`

PR #3660, opened 2016-12-31
[Pro] Check that scheduled emails are sent for embargoed requests

stevenday:
I decided that some new specs were probably the easiest way to be sure of this

Todo:

- [x] comment alerts (are sent)
- [x] new response reminder alerts (are not sent)
- [x] not clarified alerts (are sent)
- [x] overdue request alerts (are sent)
- [x] tracked thing alerts (are not sent)

So all but one of the alerts that goes to the requestor are still sent. It feels like we should be consistent here, perhaps sending reminders about new responses for now, until we have an alternative alerting system in place? (I've added a spec for this, which is currently failing)

### DIFF
--- a/spec/mailers/request_mailer_spec.rb
+++ b/spec/mailers/request_mailer_spec.rb
@@ -285,10 +285,13 @@ describe RequestMailer do
 
     context 'if the request is embargoed' do
 
-      it 'does not send the reminder' do
+      it 'sends the reminder' do
         old_request.create_embargo(:publish_at => Time.zone.now + 3.days)
-        expect(RequestMailer).not_to receive(:new_response_reminder_alert)
         send_alerts
+        deliveries = ActionMailer::Base.deliveries
+        mail = deliveries[0]
+        expect(mail.body).to match(/#{old_request.title}/)
+        expect(mail.body).to match(/Letting everyone know whether you got the information/)
       end
 
     end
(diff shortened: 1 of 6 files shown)


stevenday:
@crowbot - I've fixed up all of those things now, I think: 
- response reminders are now sent (and tested for that, I realised there was an existing test for the inverse of this, so I removed my new one and just changed the existing one)
- Everything should use the block form of time_travel_to, and just travel to the future rather than setting weird dates on the data to trigger alerts
- I've tidied up the other code issues from things you pointed out

spec/mailers/request_mailer_spec.rb
```diff
@@ -620,6 +637,20 @@ def kitten_mails
         end
       end
 
+      it "sends very overdue alerts for embargoed requests" do
+        info_request = FactoryGirl.create(:embargoed_request)
+
+        time_travel_to(31.days.from_now) do
+          RequestMailer.alert_overdue_requests
```
crowbot: I think this should be `alert_very_overdue_requests` and should go further forward in time.

stevenday: `alert_overdue_requests` sends both kinds of overdue alert here, but you're right about the date, I've changed that and added something to the spec to that we're sure it's sending the right kind of overdue email

PR #3664, opened 2017-01-06
Put mysociety-docs-theme config into Sass scope

zarino:
Update for compatibility with mysociety-docs-theme which now expects config variables to be in scope when global.scss is compiled (mysociety/mysociety-docs-theme#23).

For simplicity, we now just import mysociety-docs-theme/global.scss straight into alaveteli-org.scss, rather than producing two separate CSS files - hence the change to Gruntfile.js, README.md, and the two default.html layouts.


garethrees:
Looks good! Not going to deploy on Friday 13th just before the end of the day, so will do so Monday morning! 🍰

PR #3666, opened 2017-01-06
[Pro] Fill in the authority salutation when one is selected

stevenday:
Closes mysociety/alaveteli-professional#96

### DIFF
--- a/app/assets/javascripts/alaveteli_pro/authority_select.js
+++ b/app/assets/javascripts/alaveteli_pro/authority_select.js
@@ -1,6 +1,8 @@
 (function($){
   $(function(){
     var $select = $('.js-authority-select');
+    var $message = $('.js-outgoing-message-body');
+    var authorityName = '';
     $select.selectize({
       valueField: 'id',
       labelField: 'name',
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/alaveteli_pro/authority_select.js
```diff
@@ -18,6 +20,18 @@
           html += '</div>';
           return html;
         }
+      },
+      onChange: function(value) {
+        var oldAuthorityName = authorityName;
+        var oldSalutation = 'Dear ' + oldAuthorityName + ',';
```
crowbot: We don't necessarily need to make this a blocker now, but when we come to translate this, we're going to need to find a way for the salutation pattern to come from the application, rather than being hardcoded in English in the javascript. This needs to allow for the possibility that the word order is going to be different in different languages too. We could write a localised string to the page in a `data-` attribute with a known dummy text where the authority should go and then replace that, perhaps.

crowbot: I've created a separate ticket for that issue https://github.com/mysociety/alaveteli-professional/issues/133

PR #3675, opened 2017-01-12
Hotfix/0.26.0.2

lizconlan:
PR created to check whether test is actually failing or my test box is broken

(Update - aw, my test box is broken / yay, _only_ my test box is broken)


garethrees:
Don't forget the changelog notes :)

lizconlan:
> Don't forget the changelog notes :)

And the version bump!

Literally just working around my test box throwing errors which I suspect are memory related or some other local quirk

Issue #3130, opened 2016-03-30
Batch request placeholder not replaced


schlos:
Placeholder [Authority name] is not replaced with autority name when sending group requests.

Examples:
http://imamopravoznati.org/info_request_batch/3
http://imamopravoznati.org/info_request_batch/6

garethrees:
Reported by Sweden e.g. https://www.fragastaten.se/sv/info_request_batch/1

garethrees:
The code is fine – the problem for Sweden is that they don't have translations at their [current release version (0.23.0.3)](https://github.com/mysociety/alaveteli/blob/0.23.0.3/locale/sv/app.po#L707-L714).

The easiest answer is to upgrade a minor version (to 0.23.0.4), where [translations exist](https://github.com/mysociety/alaveteli/blob/0.23.0.4/locale/sv/app.po#L716-L723) for these strings. These aren't the current translations, but better than nothing. The current translations were introduced in [0.24.0.7](https://github.com/mysociety/alaveteli/blob/0.24.0.7/locale/sv/app.po#L695-L702).

PR #3681, opened 2017-01-18
[Pro] Use translated strings for authority salutation

stevenday:
This technically fixes mysociety/alaveteli-professional#133, however, it's a
bit of naive hack at the moment because to do this properly requires changing
how `OutgoingMessage` templates initial requests:

Right now, `OutgoingMessage::Template::InitialRequest` seems to require a public
body name in the list of replacements passed into it, so I've altered the
method that provides one to give a default string if there's no body.

This is a bit of a bug at the moment really, as the template looks like it's
trying to enforce the presence of a body name, but it's actually happy with
nil data being passed so long as the key is present (which gives a slightly
dud salutation in the message).

There are lots of methods involved in templating this message, but I *think*
the best way forward is to copy what the batch request template does - set a
placeholder salutation in the template and then replace it in the message
initialisation if there's a body present. This will allow us to combine some
aspects of the batch template and the initial request template (maybe even
into one), and then to refer to one authoritative place for the salutation to
pass to the JS.

It's a bit of a larger change though, so I wanted to double check if that
seemed sane before continuining. @garethrees, @crowbot, any thoughts?

Closes mysociety/alaveteli-professional#133

### DIFF
--- a/app/assets/javascripts/alaveteli_pro/authority_select.js
+++ b/app/assets/javascripts/alaveteli_pro/authority_select.js
@@ -2,7 +2,23 @@
   $(function(){
     var $select = $('.js-authority-select');
     var $message = $('.js-outgoing-message-body');
-    var authorityName = '';
+    var defaultAuthorityName = $message.data('salutation-body-name');
+    var currentAuthorityName = defaultAuthorityName;
+    var salutationTemplate = $message.data('salutation-template');
+
+    var updateSalutation = function updateSalutation(value) {
+      var oldAuthorityName = currentAuthorityName;
+      var oldSalutation = salutationTemplate.replace(defaultAuthorityName, oldAuthorityName);
+      var oldMessage = $message.val();
+
+      var newAuthorityName = $select.find('option:selected').text();
+      var newSalutation = salutationTemplate.replace(defaultAuthorityName, newAuthorityName);
+      var newMessage = oldMessage.replace(oldSalutation, newSalutation);
+
+      $message.val(newMessage);
+      currentAuthorityName = newAuthorityName;
+    };
+
     $select.selectize({
       valueField: 'id',
       labelField: 'name',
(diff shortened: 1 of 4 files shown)


garethrees:
I think I'd do this – is it the same as what you're suggesting?

```diff
diff --git a/app/models/outgoing_message/template/initial_request.rb b/app/models/outgoing_message/template/initial_request.rb
index 4587dce..104ab76 100644
--- a/app/models/outgoing_message/template/initial_request.rb
+++ b/app/models/outgoing_message/template/initial_request.rb
@@ -2,13 +2,21 @@
 class OutgoingMessage
   module Template
     class InitialRequest
+      def self.placeholder_salutation
+        _('Dear [Authority name],')
+      end
+
       def body(opts = {})
         assert_required_keys(opts, :public_body_name)
         template_string(opts)
       end

       def salutation(replacements = {})
-        _("Dear {{public_body_name}},", replacements)
+        if replacements[:public_body_name]
+          _("Dear {{public_body_name}},", replacements)
+        else
+          self.class.placeholder_salutation
+        end
       end

       def letter(replacements = {})
```

> This is a bit of a bug at the moment really, as the template looks like it's
> trying to enforce the presence of a body name, but it's actually happy with
> nil data being passed so long as the key is present (which gives a slightly
> dud salutation in the message).

Yeah, I think it should have checked keys and values, but hey, not doing that allows the above!

stevenday:
@garethrees - yeah, basically, but trying to reduce the duplication between that and the batch template process a little too (for that the Template class sets a placeholder salutation and the OutgoingMessage replaces it).

garethrees:
> the Template class sets a placeholder salutation and the OutgoingMessage replaces it

This was one of the things I wasn't keen on when extracting the templates, but had to leave it in for compatibility. While the `OutgoingMessage` has the method to replace it, it actually gets called in [`InfoRequestBatch`](https://github.com/mysociety/alaveteli/blob/5a6e0a8/app/models/info_request_batch.rb#L63) so we'd then have to do this substitution for normal messages.

My ideal was basically to be able to create a new `Template` instance with whatever data the template needs so that you get a `String` that's going to get sent to the authority.

> trying to reduce the duplication between that and the batch template process a little

I originally extracted the templates as prep for adding support for external ombudsmen, so I was preferring duplication for now as I don't know how things will change when we introduce that.

I think I'd favour minimal changes around this at this point, but its been a while since I've got my head in to it and don't fully understand what changes pro has made to the end-to-end process. I'm having a look at a related area (https://github.com/mysociety/alaveteli/issues/3130) this sprint (probably this afternoon) so I'll get my head back in to it then. Happy to look at spiked code in the meantime.

stevenday:
@crowbot - I've updated this PR to match roughly what @garethrees suggested in [his comment](https://github.com/mysociety/alaveteli/pull/3681#issuecomment-273738487) - basically making it work for our use case by adding a default salutation in the template class, but leaving any further refactoring/cleanup for now. This should be good for a review now.

PR #3682, opened 2017-01-19
[Pro] Send requests for admin attention to a separate email address

stevenday:
This adds a new contact address/name for pro-specific contact, and then uses
this in the requires_admin mailer to send requests for admin attention on
embargoed requests to that address instead of the normal one.

Questions:
- [x] Should this new address be used for any requests from a pro user,
      whether they're embargoed or not?

Closes mysociety/alaveteli-professional#116

### DIFF
--- a/app/mailers/request_mailer.rb
+++ b/app/mailers/request_mailer.rb
@@ -6,6 +6,8 @@
 # Email: hello@mysociety.org; WWW: http://www.mysociety.org/
 
 class RequestMailer < ApplicationMailer
+  include AlaveteliFeatures::Helpers
+
   # Used when an FOI officer uploads a response from their web browser - this is
   # the "fake" email used to store in the same format in the database as if they
   # had emailed it.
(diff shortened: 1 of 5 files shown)


config/general.yml-example
```diff
@@ -972,3 +972,30 @@ NEW_REQUEST_RECAPTCHA: false
 #
 # ---
 ENABLE_ALAVETELI_PRO: false
+
+# Contact email address for Alaveteli Professional.
+# Requests made through Alaveteli Professional may be embargoed, meaning that
+# the user has an expectation that they're private, and this can include
+# being private from some of the site's administration team. Even if this is
+# not the case, you may wish to redirect pro support email away from the usual
+# email address for. These details allow you to specify a different address.
```
crowbot: Looks like there's a missing word after `email address for.`

app/mailers/request_mailer.rb
```diff
@@ -55,6 +55,11 @@ def stopped_responses(info_request, email, raw_email_data)
   # An FOI response is outside the scope of the system, and needs admin attention
   def requires_admin(info_request, set_by = nil, message = "")
     user = set_by || info_request.user
+    if info_request.embargo
```
crowbot: You could assign at the beginning of this statement to remove a bit of duplication e.g. 

```
to = if info_request.embargo
  pro_contact_from_name_and_email
else
  contact_from_name_and_email
end
```

stevenday:
@crowbot - I've made those suggested fixes and changed it so that it sends emails based on the user being a pro rather than the request being embargoed (I rebased so that I could rename the commit appropriately). Want to double check that before I merge?

PR #3684, opened 2017-01-19
[Pro] Redirect contact emails to/from pro users

stevenday:
This is for the main bit of mysociety/alaveteli-professional#75, redirecting
support mail from pro users to the new pro contact address. I've also changed
the reverse, for when admins send mail *to* pro users, but I'm not 100% sure
if that's desirable? Will we use that feature to contact them or will that
just be for the current admins, in which case they wouldn't see the replies..
see mysociety/alaveteli-professional#138 for related questions about what the
pro contact address is actually going to be @.

### DIFF
--- a/app/controllers/help_controller.rb
+++ b/app/controllers/help_controller.rb
@@ -28,6 +28,9 @@ class HelpController < ApplicationController
 
   def contact
     @contact_email = AlaveteliConfiguration::contact_email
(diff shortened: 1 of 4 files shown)


stevenday:
I'm closing this in favour of merging all the pro-email stuff into one new PR

garethrees:
> I'm closing this in favour of merging all the pro-email stuff into one new PR

https://github.com/mysociety/alaveteli/pull/3706

PR #3687, opened 2017-01-19
Prevent info requests with titles such as "re"

### LINKED ISSUE #3653: Request with subject "re" allowed through
Alaveteli's filtering that catches titles all in lower case etc. failed to prevent [this request](https://www.whatdotheyknow.com/request/re_31) from being sent with the title subject "re".

lizconlan:
* Drops the string length limit for mixed case to be applied
* Adds a minimum length requirement of ~~4~~ 3 characters
* Allows mixed case exemptions for:
  * all upper case, single word strings of less than 6 chars (on the grounds that they're probably an acronym)
  * a multiword string starting with a number 

Fixes #3653

### DIFF
--- a/app/models/info_request.rb
+++ b/app/models/info_request.rb
@@ -1603,12 +1603,22 @@ class InfoRequest < ActiveRecord::Base
   end
 
   def title_formatting
-    if title && !MySociety::Validate.uses_mixed_capitals(title, 10)
+    return unless title
+    unless MySociety::Validate.uses_mixed_capitals(title, 1) ||
+      title_starts_with_number || title_is_acronym(6)
       errors.add(:title, _('Please write the summary using a mixture of capital and lower case letters. This makes it easier for others to read.'))
     end
-    if title && title =~ /^(FOI|Freedom of Information)\s*requests?$/i
+    if title =~ /^(FOI|Freedom of Information)\s*requests?$/i
       errors.add(:title, _('Please describe more what the request is about in the subject. There is no need to say it is an FOI request, we add that on anyway.'))
     end
+  end
+
+  def title_is_acronym(max_length)
+    title.upcase == title && title.length <= max_length && !title.include?(" ")
+  end
+
+  def title_starts_with_number
+    title.include?(" ") && title.split(" ").first =~ /^\d+$/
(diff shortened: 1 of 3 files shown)


garethrees:
Looks good. Pushed a slight wording tweak as a suggestion, but after adding it noticed that there are some existing examples that use "Please describe more…". Not super precious about it either way. Don't forget a changelog entry :)

garethrees:
> Not super precious

I do think we should use "summary" over "subject" though, given the form field title.

lizconlan:
@garethrees hmm, I think I picked up "subject" from `title_summary` that might need a tweak at some point

The difference in the wording is probably slightly nicer for the user (particularly if they're hitting a sequence of errors in good faith) but slightly tougher on the translators. Sorry translators, new text reads better!

lizconlan:
last reviewed @ 017f5b6

lizconlan:
last reviewed @ 017f5b6

garethrees:
One tiny comment, but good to go after that tweak (no need for re-review).

Also needs a changelog entry.

app/models/info_request.rb
```diff
@@ -1456,14 +1456,24 @@ def applicable_law
...
       errors.add(:title, _('Please write the summary using a mixture of capital and lower case letters. This makes it easier for others to read.'))
     end
-    if title && title =~ /^(FOI|Freedom of Information)\s*requests?$/i
+    if title =~ /^(FOI|Freedom of Information)\s*requests?$/i
       errors.add(:title, _('Please describe more what the request is about in the subject. There is no need to say it is an FOI request, we add that on anyway.'))
     end
   end
 
+  def title_is_acronym(length)
+    title.upcase == title && title.length < length && !title.include?(" ")
```
lizconlan: sure, have changed the param name to `max_length` for better readability

lizconlan: although maybe I should have rewritten it as 

```
def title_is_acronym(allow_shorter_than)
```

and kept the original less than to match 

```
def self.uses_mixed_capitals(s, allow_shorter_than = 20)
```

from commonlib/validate

Left for a future revision

PR #3688, opened 2017-01-20
[Pro] Use beginning_of_day (a DateTime) not today (a Date).

crowbot:
Converting back from a Date to a DateTime on saving causes time zone
errors as Dates are not time zone aware.

### DIFF
--- a/spec/factories/embargos.rb
+++ b/spec/factories/embargos.rb
@@ -18,7 +18,7 @@ FactoryGirl.define do
     embargo_duration "3_months"
 
     factory :expiring_embargo do
-      publish_at Time.now + 3.days
+      publish_at Time.zone.now + 3.days
     end
   end
 end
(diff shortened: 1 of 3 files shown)


spec/factories/embargos.rb
```diff
@@ -14,7 +14,7 @@
 FactoryGirl.define do
   factory :embargo do
     info_request
-    publish_at Time.zone.today + 3.months
+    publish_at (Time.zone.now + 3.months).beginning_of_day
```
stevenday: For continuity with the model, should this be beginning of day then + 3 months, rather than + 3 months then beginning of day? (I can't put my finger on an example when this would matter, but it feels like there'd be a leap second/year loophole somewhere).

crowbot: Good call, have fixed.

PR #3689, opened 2017-01-20
[Pro] Separate pro locale files

crowbot:


stevenday:
The only thing I could find that this will miss is the flash message in the new post redirect case for pros: https://github.com/mysociety/alaveteli/blob/alaveteli-pro-develop/app/controllers/application_controller.rb#L353

Perhaps an argument to move the text for that into the controller, and set a param to trigger it in application_controller instead?

stevenday:
> I think I've covered that in this PR by moving the method to a module in the alaveteli_pro namespace that gets included by application controller.

Aha, yep I see that now (sorry, I found it easier to look through the alaveteli-pro-develop diff to try to spot things based on how you described this PR yesterday, wasn't looking at this PR directly). All looks good then!

stevenday:
Running this locally all seems fine, though I'm surprised at how few new strings there are!

stevenday:
One final minor thing, should this run the find_pro task and commit the .pot file too?

PR #3690, opened 2017-01-23
Force visually-hidden elements to act as block-level elements

### LINKED ISSUE #3646: Statistics page is very long
The body significantly extends beyond the footer. https://www.whatdotheyknow.com/statistics

lizconlan:
Fixes #3646 

(This should probably be a hotfix when it grow up so we can roll it out to existing sites that have the new statistics page - although we could fix it in themes for now)


lizconlan:
@wrightmartin does this look sane or could it break something?

wrightmartin:
@lizconlan can't see a problem, it's a 👍  from me

PR #3693, opened 2017-01-24
[Pro] Remind users about their exipiring embargoes

stevenday:
This adds a mailer and appropriate script/cron job to send reminder emails
about embargoes that are expiring.

Closes mysociety/alaveteli-pro#119


stevenday:
NB: If this is merged after #3689 we'll have to move the mailers etc into an alaveteli_pro folder, and add /mailers to the list of dirs we check for files to translate.

app/models/user_info_request_sent_alert.rb
```diff
@@ -29,7 +29,8 @@ class UserInfoRequestSentAlert < ActiveRecord::Base
     # recent response
     'not_clarified_1', # reminder that user has to explain part of the
     # request
-    'comment_1' # tell user that info request has a new comment
+    'comment_1', # tell user that info request has a new comment
+    'embargo_expiring' # tell user that their embargo is expiring
```
crowbot: To match existing practice, this should probably have a `_1` suffix to reflect the fact that it is the first reminder.

spec/mailers/embargo_mailer_spec.rb
```diff
@@ -0,0 +1,133 @@
...
+  let!(:not_embargoed) { FactoryGirl.create(:info_request) }
+
+  describe '.alert_expiring' do
+    it 'only sends one email per user' do
+      EmbargoMailer.alert_expiring
+      mails = ActionMailer::Base.deliveries
+      expect(mails.size).to eq 2
+      # We don't know the order, so find each mail in this slightly awkward
+      # way. These would probably get an IndexError if they didn't exist
+      first_mail = mails.select{ |mail| mail.to == [pro_user.email] }[0]
```
crowbot: You could use `detect` here to get the first matching object or `nil` and avoid the index error.

app/mailers/embargo_mailer.rb
```diff
@@ -0,0 +1,64 @@
...
+# Alerts relating to embargoes.
+#
+# Copyright (c) 2017 UK Citizens Online Democracy. All rights reserved.
+# Email: hello@mysociety.org; WWW: http://www.mysociety.org/
+
+class EmbargoMailer < ApplicationMailer
+  def self.alert_expiring
+    InfoRequest.embargo_expiring.group_by(&:user).each do |user, info_requests|
+      info_requests.reject! do |info_request|
+        alert_event_id = info_request.last_event_forming_initial_request.id
```
crowbot: In the overdue alerts, the last event forming the initial request is recorded on the sent alert record because if it changes it means that the deadline changes for the request, and so it could become overdue again. That wouldn't affect the expiry of an embargo though, so I think we need to record some other identifier. I think the behaviour we want is that if an embargo is alerted on here, and then subsequently extended and gets close to expiry again, the user gets notified. So I don't think we can use the embargo id as it will not change when the embargo is extended. I guess we could create an `embargo_added` `InfoRequestEvent` on embargo creation and key off that? Not very elegant though.

stevenday: Aha, ok, I think I misunderstood the event ID as not being important (hence just putting something in the field) because I didn't think about extensions to the embargo then expiring again.

It feels like an "embargo extended" event is really what we're after (as it might actually be a useful thing to print in a timeline for example), but we already save the extensions in a separate table, so perhaps for this purpose we could just use their ids instead of an event? (Perhaps with a special 0 id, for the initial embargo or something?).

stevenday:
So thinking about this further with @crowbot, it needs to be an id of an actual event we're using for uniqueness here, so we'll need to create a new kind of event and do a few more things to make this possible:

- Create an EmbargoDateSetEvent (e.g. an embargo date was set to something new) which covers both initial embargoes, potential new embargoes after a request is created and extending embargoes.
- Create these events whenever the publish_at date is set or changed on an Embargo.
- Add a method to InfoRequest to get the most recent of these events
- Use the ids of these events in the alerting code in this PR
- Add some specs to test that alerts are sent correctly for extended embargoes as well as initial ones

spec/models/info_request_spec.rb
```diff
@@ -3327,4 +3327,31 @@ def email_and_raw_email(opts = {})
...
+      end
+
+    end
+
+    context 'if embargos have been set' do
+      let(:embargo){ FactoryGirl.create(:embargo) }
+      let(:embargo_extension){ FactoryGirl.create(:embargo_extension) }
+
+      it 'returns the last "set_embargo" event' do
+        embargo.extend(embargo_extension)
```
stevenday: Might be good to check we get the right event before we extend the embargo here too, just for completeness?

PR #3694, opened 2017-01-25
[#3645] Report spam/vexatious messages

### LINKED ISSUE #3645: Add link "report annotation / message as spam"
A user has suggested adding a button / link by which users can bring spam annotations to the attention of administrators. I think this would also be useful for request responses.

lizconlan:
Fixes #3645 

Up for early/initial "is this doing the right thing?" review before I get too far into extending the behaviour to followup messages (currently just for comments). (Will probably do that in a branch off of this one so as not to mix up the commits)

### DIFF
--- a/app/assets/stylesheets/responsive/_request_layout.scss
+++ b/app/assets/stylesheets/responsive/_request_layout.scss
@@ -318,6 +318,19 @@ input.cplink__field {
   line-height: 1em;
 }
 
+a.cplink__button {
+  padding: 0.5em;
+  &:hover,
+  &:active,
+  &:focus {
+    border-left: 0;
+    color: #fff;
+  }
+  &:visited {
+    color: #000;
+  }
+}
+
 #reports_new {
     form {
       max-width: 38em;
--- a/app/controllers/admin_comment_controller.rb
+++ b/app/controllers/admin_comment_controller.rb
@@ -29,15 +29,20 @@ class AdminCommentController < AdminController
   def update
     old_body = @comment.body
     old_visible = @comment.visible
+    old_attention = @comment.attention_requested
 
     if @comment.update_attributes(comment_params)
-      @comment.info_request.log_event("edit_comment",
-                                      { :comment_id => @comment.id,
-                                        :editor => admin_current_user,
-                                        :old_body => old_body,
-                                        :body => @comment.body,
-                                        :old_visible => old_visible,
-                                        :visible => @comment.visible })
+      @comment.
+        info_request.
+        log_event("edit_comment",
+                  { :comment_id => @comment.id,
+                    :editor => admin_current_user,
+                    :old_body => old_body,
+                    :body => @comment.body,
+                    :old_visible => old_visible,
+                    :visible => @comment.visible,
+                    :old_attention_requested => old_attention,
+                    :attention_requested => @comment.attention_requested })
       flash[:notice] = 'Comment successfully updated.'
       redirect_to admin_request_url(@comment.info_request)
     else
@@ -49,7 +54,7 @@ class AdminCommentController < AdminController
 
   def comment_params
     if params[:comment]
-      params.require(:comment).permit(:body, :visible)
+      params.require(:comment).permit(:body, :visible, :attention_requested)
     else
       {}
     end
--- a/app/controllers/admin_comment_controller.rb
+++ b/app/controllers/admin_comment_controller.rb
@@ -27,14 +27,19 @@ class AdminCommentController < AdminController
   end
 
   def update
-    old_body = @comment.body
+    old_body = @comment.body.dup
     old_visible = @comment.visible
     old_attention = @comment.attention_requested
 
     if @comment.update_attributes(comment_params)
+      update_type = if comment_hidden?(old_visible, old_body)
+        "hide_comment"
+      else
+        "edit_comment"
+      end
       @comment.
         info_request.
-          log_event("edit_comment",
+          log_event(update_type,
                     { :comment_id => @comment.id,
                       :editor => admin_current_user,
                       :old_body => old_body,
@@ -62,6 +67,10 @@ class AdminCommentController < AdminController
 
   def set_comment
     @comment = Comment.find(params[:id])
+  end
+
+  def comment_hidden?(old_visibility, old_body)
+    !@comment.visible && old_visibility && old_body == @comment.body
   end
 
 end
--- a/app/controllers/admin_comment_controller.rb
+++ b/app/controllers/admin_comment_controller.rb
@@ -34,15 +34,15 @@ class AdminCommentController < AdminController
     if @comment.update_attributes(comment_params)
       @comment.
         info_request.
-        log_event("edit_comment",
-                  { :comment_id => @comment.id,
-                    :editor => admin_current_user,
-                    :old_body => old_body,
-                    :body => @comment.body,
-                    :old_visible => old_visible,
-                    :visible => @comment.visible,
-                    :old_attention_requested => old_attention,
-                    :attention_requested => @comment.attention_requested })
+          log_event("edit_comment",
+                    { :comment_id => @comment.id,
+                      :editor => admin_current_user,
+                      :old_body => old_body,
+                      :body => @comment.body,
+                      :old_visible => old_visible,
+                      :visible => @comment.visible,
+                      :old_attention_requested => old_attention,
+                      :attention_requested => @comment.attention_requested })
       flash[:notice] = 'Comment successfully updated.'
       redirect_to admin_request_url(@comment.info_request)
     else
--- a/app/controllers/reports_controller.rb
+++ b/app/controllers/reports_controller.rb
@@ -5,13 +5,7 @@ class ReportsController < ApplicationController
 
   def create
     @reason = params[:reason]
-    @message = if @comment
-      "#{params[:message]}\n\nThe user wishes to draw attention to the " \
-      "comment: #{comment_url(@comment)} " \
-      "\nadmin: #{edit_admin_comment_path(@comment)}"
-    else
-      params[:message]
-    end
+    @message = params[:message] || ""
 
     if @reason.empty?
       flash[:error] = _("Please choose a reason")
--- a/app/controllers/reports_controller.rb
+++ b/app/controllers/reports_controller.rb
@@ -22,8 +22,7 @@ class ReportsController < ApplicationController
     elsif @info_request.attention_requested
       flash[:notice] = _("This request has already been reported for administrator attention")
     else
-      reportable = @comment || @info_request
-      reportable.report!(@reason, @message, @user)
+      @reportable.report!(@reason, @message, @user)
       flash[:notice] = if @comment
         _("This annotation has been reported for administrator attention")
       else
@@ -55,8 +54,8 @@ class ReportsController < ApplicationController
                       not_embargoed.
                       find_by_url_title!(params[:request_id])
     @comment = @info_request.comments.where(:id => params[:comment_id]).first
-    reportable = @comment || @info_request
-    @report_reasons = reportable.report_reasons
+    @reportable = @comment || @info_request
+    @report_reasons = @reportable.report_reasons
   end
 
 end
--- a/app/controllers/reports_controller.rb
+++ b/app/controllers/reports_controller.rb
@@ -5,7 +5,13 @@ class ReportsController < ApplicationController
 
   def create
     @reason = params[:reason]
-    @message = params[:message]
+    @message = if @comment
+      "#{params[:message]}\n\nThe user wishes to draw attention to the " \
+      "comment: #{comment_url(@comment)}"
+    else
+      params[:message]
+    end
+
     if @reason.empty?
       flash[:error] = _("Please choose a reason")
       render "new"
--- a/app/controllers/reports_controller.rb
+++ b/app/controllers/reports_controller.rb
@@ -34,10 +34,11 @@ class ReportsController < ApplicationController
   end
 
   def new
-    @page_title = if @comment
-      "Report annotation on request: #{@info_request.title}"
+    @title = if @comment
+      _("Report annotation on request: {{title}}",
+        :title => @info_request.title)
     else
-      "Report request: #{@info_request.title}"
+      _("Report request: {{title}}", :title => @info_request.title)
     end
     if authenticated?(
       :web => _("To report this request"),
--- a/app/controllers/reports_controller.rb
+++ b/app/controllers/reports_controller.rb
@@ -7,7 +7,8 @@ class ReportsController < ApplicationController
     @reason = params[:reason]
     @message = if @comment
       "#{params[:message]}\n\nThe user wishes to draw attention to the " \
-      "comment: #{comment_url(@comment)}"
+      "comment: #{comment_url(@comment)} " \
+      "\nadmin: #{edit_admin_comment_path(@comment)}"
     else
       params[:message]
     end
--- a/app/controllers/reports_controller.rb
+++ b/app/controllers/reports_controller.rb
@@ -30,9 +30,10 @@ class ReportsController < ApplicationController
 
   def new
     if authenticated?(
-        :web => _("To report this request"),
-        :email => _("Then you can report the request '{{title}}'", :title => @info_request.title),
-      :email_subject => _("Report an offensive or unsuitable request"))
+      :web => _("To report this request"),
+      :email => _("Then you can report the request '{{title}}'", :title => @info_request.title),
+      :email_subject => _("Report an offensive or unsuitable request"),
+      :comment_id => params[:comment_id])
     end
   end
 
--- a/app/controllers/reports_controller.rb
+++ b/app/controllers/reports_controller.rb
@@ -24,12 +24,21 @@ class ReportsController < ApplicationController
     else
       reportable = @comment || @info_request
       reportable.report!(@reason, @message, @user)
-      flash[:notice] = _("This request has been reported for administrator attention")
+      flash[:notice] = if @comment
+        _("This annotation has been reported for administrator attention")
+      else
+        _("This request has been reported for administrator attention")
+      end
     end
     redirect_to request_url(@info_request)
   end
 
   def new
+    @page_title = if @comment
+      "Report annotation on request: #{@info_request.title}"
+    else
+      "Report request: #{@info_request.title}"
+    end
     if authenticated?(
       :web => _("To report this request"),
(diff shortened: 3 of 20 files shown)


lizconlan:
Might be worth trying to extract this into a behaviour that can be applied to any type of message but trying to avoid the premature optimisation trap

spec/controllers/reports_controller_spec.rb
```diff
@@ -92,9 +92,41 @@
...
+
+      let(:comment) do
+        FactoryGirl.create(:comment, :info_request => info_request,
+                                     :body => "This is a vexations comment")
+      end
+
+      it "ignores the comment id if it does not belong to the request" do
+        new_request = FactoryGirl.create(:info_request)
+        new_comment = FactoryGirl.create(:comment, :info_request => new_request,
+                                                   :body => "Some comment")
```
garethrees: There are a few cases of this – no need to introduce extra context thats not relevant to the given spec.

app/controllers/reports_controller.rb
```diff
@@ -2,8 +2,21 @@
 class ReportsController < ApplicationController
   def create
     @info_request = InfoRequest.find_by_url_title!(params[:request_id])
+    if params[:comment_id]
```
garethrees: The `comment_id` param isn't persisted across a log in:

- log out
- go to request page
- click report comment
- go through log in
- Send report
- Report does not contain link to comment

garethrees: Also, once _anything_ has been reported once, you can no longer report further problems because of the `#attention_requested` attribute. Open to options about what you think might be a good way around this…

spec/controllers/reports_controller_spec.rb
```diff
@@ -79,14 +79,11 @@
 end
 
 describe ReportsController, "#new_report_request" do
-  let(:info_request) { mock_model(InfoRequest, :url_title => "foo") }
-  before :each do
-    expect(InfoRequest).to receive(:find_by_url_title!).with("foo").and_return(info_request)
-  end
+  let(:info_request) { FactoryGirl.create(:info_request, :title => "foo") }
```
lizconlan: apparently not, removing

app/controllers/reports_controller.rb
lizconlan: 2 broad approaches to consider...

1) "just" send an email about the comment (or message) to ask admins to look (but how will this be reflected admin interface?)
2) make comments reportable, possibly via Concerns so that we can extend it to other objects more readily, to store their own data about having been reported, again will require something new for the admin interface to be genuinely useful

spec/controllers/reports_controller_spec.rb
lizconlan: trying to just keep the `:info_request` param where the relationship is part of the test

app/controllers/reports_controller.rb
```diff
@@ -23,6 +23,11 @@ def create
 
   def new
     @info_request = InfoRequest.find_by_url_title!(params[:request_id])
+    if params[:comment_id]
+      if @info_request.comments.map(&:id).include?(params[:comment_id].to_i)
```
lizconlan: Good spot, thanks - had overlooked that a non-matching id would safely return `nil` here

lizconlan:
[This comment](#discussion_r99614586) is still relevant

lizconlan:
After a bit more thought...

What would be nicer here is if the `Comment` itself were reportable and could keep track of complaints against it without interfering with the parent request, then we could also have something like this in the Summary section of the admin interface:

review comments

(A relatively lightweight change as the link for each report would just link through to the existing comment admin page.)

This opens up the possibility of refactoring the reporting logic into a Reportable behaviour/concern so that we can apply it to Followups in the future, but that feels like it should be a separate piece of work.

lizconlan:
Still to do:
- [ ] retain the `comment_id` through a login
- [ ] allow comments to be reported without preventing future reporting of the request itself

lizconlan:
Rebased before continuing. Nothing new added at this point aside from incorporating latest changes from `develop`

lizconlan:
Expected command line output:

```
exporting to: exports/PublicBodyCategory-16-02-2017.csv
exporting to: exports/PublicBodyHeading-16-02-2017.csv
exporting to: exports/PublicBodyCategoryLink-16-02-2017.csv
exporting to: exports/PublicBodyCategoryTranslation-16-02-2017.csv
exporting to: exports/PublicBodyHeadingTranslation-16-02-2017.csv
exporting to: exports/InfoRequestBatch-16-02-2017.csv
exporting to: exports/InfoRequestBatchPublicBody-16-02-2017.csv
exporting to: exports/HasTagStringTag-16-02-2017.csv
exporting to: exports/PublicBody-16-02-2017.csv
exporting to: exports/User-16-02-2017.csv
Error processing data:
invalid byte sequence in UTF-8
[4, "\x92s Social", 0, 0, 0, 0, 0]
Error processing data:
invalid byte sequence in UTF-8
[7, "\x92s Social", 1, 0, 0, 0, 0]
exporting to: exports/InfoRequest-16-02-2017.csv
exporting to: exports/IncomingMessage-16-02-2017.csv
exporting to: exports/OutgoingMessage-16-02-2017.csv
exporting to: exports/FoiAttachment-16-02-2017.csv
```

lizconlan:
`Comment` now has its own reporting capability with its own `report!` method and `attention_requested` flag (the latter could be slightly problematic as it has a subtly different function to its `InfoRequest` equivalent as it doesn't persist to indicate that the object has been reported at some time in the past, but this seemed simpler than either inventing different terminology or recreating `described_state`).

New section in the admin summary for reported comments:

screen shot 2017-02-17 at 12 02 27
 
And a new dropdown for comments to allow the `attention_requested` flag to be cleared:

screen shot 2017-02-17 at 12 39 24

I'm not entirely happy with the workflow there as updating the comment causes a redirect "back" to the parent request edit page which feels jarring but I'm not sure what a good fix for that would look like. Also the messaging for the user is about reporting the request rather than the comment which ... is sort of true? An alternative would be to fully separate out a report/comment controller but that could result in a lot of code duplication.

spec/models/comment_spec.rb
```diff
@@ -67,4 +67,33 @@
...
+    it 'sets attention_requested to true' do
+      comment.report!("Vexatious comment", "Comment is bad, please hide", user)
+      expect(comment.attention_requested).to eq(true)
+    end
+
+    it 'sends a message a message to admins' do
+      message = double(RequestMailer)
+      allow(RequestMailer).to receive(:requires_admin).and_return(message)
+      expect(message).to receive(:deliver)
+      comment.report!("Vexatious comment", "Comment is bad, please hide", user)
```
garethrees: Would it be better to pull out the last mail from ActionMailer::Deliveries and check content? See ["Don't mock what you don't own"](http://felixclack.com/post/51541743233/dont-mock-what-you-dont-own). Since we're not passing the request mailer in to the subject, we don't control it.

spec/models/comment_spec.rb
```diff
@@ -67,4 +67,33 @@
...
+    it 'prepends the reason to the message before sending' do
+      message = double(RequestMailer)
+      allow(message).to receive(:deliver)
+
+      expected = "Reason: Vexatious comment\n\nComment is bad, please hide"
+      expect(RequestMailer).to receive(:requires_admin).
+        with(comment.info_request, user, expected).
+        and_return(message)
+
+      comment.report!("Vexatious comment", "Comment is bad, please hide", user)
```
garethrees: Same comment as https://github.com/mysociety/alaveteli/pull/3694/commits/5e129149e3f8dab03a58bb1a70b4f43eb7c40ccb#r102201345

spec/controllers/reports_controller_spec.rb
```diff
@@ -87,6 +87,61 @@
...
+      before do
+        session[:user_id] = user.id
+      end
+
+      let(:comment) do
+        FactoryGirl.create(:comment, :info_request => info_request)
+      end
+
+      it "should mark the comment as having been reported" do
+        expect(comment.attention_requested).to eq(false)
```
garethrees: This is redundant – you can explicitly pass `:attention_requested => false` to the factory if its worth illustrating the default state of the subject. Same applies for all cases of this.

spec/controllers/reports_controller_spec.rb
```diff
@@ -87,6 +87,61 @@
...
+        expect(comment.attention_requested).to eq(true)
+      end
+
+      it "should not mark the parent request as having been reported" do
+        expect(info_request.attention_requested).to eq(false)
+
+        post :create, :request_id => info_request.url_title,
+                      :comment_id => comment.id,
+                      :reason => "my reason"
+        expect(response)
```
garethrees: You've already got an assert for the redirect in the spec above.

Personally, I'd have a spec for each:

```ruby
it "marks the comment as having been reported" do
it "does not mark the parent request as having been reported" do
it "sends an email alerting admins to the report" do
it "redirects to the parent request" do
it "informs the user that the comment has been reported" do
```

spec/controllers/reports_controller_spec.rb
```diff
@@ -87,6 +87,61 @@
...
+                      :reason => "my reason",
+                      :message => "It's just not"
+        deliveries = ActionMailer::Base.deliveries
+        expect(deliveries.size).to eq(1)
+        mail = deliveries[0]
+        expect(mail.subject).to match(/requires admin/)
+        expect(mail.header['Reply-To'].to_s).to include(user.email)
+        expect(mail.body).to include(user.name)
+        expect(mail.body)
+          .to include("Reason: my reason\n\nIt's just not")
```
garethrees: Should also include the admin comment URL?

app/models/info_request_event.rb
```diff
@@ -34,6 +34,7 @@ class InfoRequestEvent < ActiveRecord::Base
     'edit', # title etc. edited (in admin interface)
     'edit_outgoing', # outgoing message edited (in admin interface)
     'edit_comment', # comment edited (in admin interface)
+    'report_comment', # comment reported as spam by user
```
garethrees: Not necessarily spam – `# comment reported for admin attention by user`?

spec/models/comment_spec.rb
```diff
@@ -95,5 +95,14 @@
       comment.report!("Vexatious comment", "Comment is bad, please hide", user)
     end
 
+    it 'logs the report_comment event' do
+      comment.report!("Vexatious comment", "Comment is bad, please hide", user)
+
+      comment.reload
+      most_recent_event = comment.info_request_events.last
+      expect(most_recent_event.event_type).to eq('report_comment')
+      expect(most_recent_event.comment_id).to eq(comment.id)
```
garethrees: This is implicit in calling `comment.info_request_events.last`?

I'd probably do this to be more explicit:

```ruby
it 'logs the report_comment event' do
  comment.info_request_events.where(:event_type => 'report_comment').destroy_all
  comment.report!("Vexatious comment", "Comment is bad, please hide", user)
  comment.reload!
  most_recent_event = comment.info_request_events.last
  expect(most_recent_event.event_type).to eq('report_comment')
end
```

app/controllers/admin_comment_controller.rb
```diff
@@ -29,15 +29,20 @@ def edit
   def update
```
garethrees: This commit seems to be doing much more than the message suggests…

app/helpers/admin_helper.rb
```diff
@@ -35,6 +35,13 @@ def user_both_links(user)
       link_to(h(user.name), admin_user_path(user), :title => "view full details")
   end
 
+  def comment_both_links(comment)
```
garethrees: No spec?

app/views/admin_general/index.html.erb
```diff
@@ -110,6 +110,40 @@
...
+          </span>
+          <%= chevron_right %>
+          Review comments reported by users as requiring administrator
+          attention
+        </a>
+      </div>
+      <div id="attention-comments" class="accordion-body collapse">
+        <table class="table table-striped table-condensed">
+          <tbody>
+            <% for @comment in @attention_comments %>
```
garethrees: `each` instead of `for`

app/views/admin_general/index.html.erb
```diff
@@ -110,6 +110,40 @@
...
+              <tr>
+                <td class="link">
+                  <%= comment_both_links(@comment) %>
+                </td>
+                <td class="span2">
+                  <%= simple_date(@comment.info_request_events.
```
garethrees: Could do with `Comment#last_report` here so that this could be:

```erb
<%= simple_date(@comment.last_report.created_at) %>
```

app/controllers/admin_comment_controller.rb
```diff
@@ -29,15 +29,20 @@ def edit
...
 
     if @comment.update_attributes(comment_params)
-      @comment.info_request.log_event("edit_comment",
-                                      { :comment_id => @comment.id,
-                                        :editor => admin_current_user,
-                                        :old_body => old_body,
-                                        :body => @comment.body,
-                                        :old_visible => old_visible,
-                                        :visible => @comment.visible })
+      @comment.
```
garethrees: No specs?

app/controllers/reports_controller.rb
```diff
@@ -24,12 +24,21 @@ def create
...
+        _("This annotation has been reported for administrator attention")
+      else
+        _("This request has been reported for administrator attention")
+      end
     end
     redirect_to request_url(@info_request)
   end
 
   def new
+    @page_title = if @comment
```
garethrees: `@title` is the convention, given it sets the `<title>` tag in the layout.

app/controllers/reports_controller.rb
```diff
@@ -24,12 +24,21 @@ def create
...
+      "Report annotation on request: #{@info_request.title}"
+    else
+      "Report request: #{@info_request.title}"
```
garethrees: These need to be translatable.

spec/views/reports/new.erb_spec.rb
```diff
@@ -28,4 +33,20 @@
       expect(rendered).to have_content("This request has already been reported")
     end
   end
+
+  context "reporting a comment" do
+    let(:comment) do
+      FactoryGirl.create(:comment, :info_request => info_request)
```
garethrees: Probably fine with `#build`?

app/controllers/reports_controller.rb
```diff
@@ -24,12 +24,21 @@ def create
     else
       reportable = @comment || @info_request
       reportable.report!(@reason, @message, @user)
-      flash[:notice] = _("This request has been reported for administrator attention")
+      flash[:notice] = if @comment
```
garethrees: No specs?

app/controllers/reports_controller.rb
```diff
@@ -23,13 +23,22 @@ def create
...
   end
+
+  private
+
+  def set_request_data
+    @info_request = InfoRequest.
+                      not_embargoed.
+                      find_by_url_title!(params[:request_id])
+    @comment = @info_request.comments.where(:id => params[:comment_id]).first
+    reportable = @comment || @info_request
```
lizconlan: I think it's needed in create in case there's an error in the form (i.e. if `@reason` is empty) and it displays the `new` template

lizconlan: ...and also now an instance variable to address what you said, not what I misread

app/views/comment/_single_comment.html.erb
```diff
@@ -25,10 +25,12 @@
     <p class="event_actions">
       <% if comment.id %>
         <% if @user && @user.admin_page_links? %>
-          <%= link_to "Admin", edit_admin_comment_path(comment) %>
+          <%= link_to "Admin", edit_admin_comment_path(comment) %> |
         <% end %>
 
-        <!-- | <%= link_to _('Report abuse'), comment_path(comment) %> -->
+        <%= link_to _('Report abuse'),
```
lizconlan: ah, css fiddling required to get it to look right as by default it does this:

screen shot 2017-02-22 at 10 05 08

garethrees: I just literally copied the link to this button code, duped it and pasted it next to it with "report" as the text. Will probably need to make the classes applied to it a bit more generic.

app/assets/stylesheets/responsive/_request_layout.scss
```diff
@@ -318,6 +318,19 @@ input.cplink__field {
   line-height: 1em;
 }
 
+a.cplink__button {
+  padding: 0.5em;
+  &:hover,
+  &:active,
+  &:focus {
+    border-left: 0;
```
lizconlan: that doesn't look quite right as it causes the border to appear cropped **but** the only alternative I could come up with caused the button to move sideways on rollover

spec/models/comment_spec.rb
```diff
@@ -67,6 +67,17 @@
 
   end
 
+  describe 'report_reasons' do
+
+    let(:comment) { FactoryGirl.build(:comment) }
+
+    it 'returns an array of strings' do
+      expect(comment.report_reasons).to be_an(Array)
+      expect(comment.report_reasons).to all(be_a(String))
```
garethrees: Surely its better to assert on what its actually returning?

app/views/admin_general/index.html.erb
lizconlan: spot the lazy copy and paste from the section above

app/controllers/admin_comment_controller.rb
lizconlan: I'll leave this here as a reminder for now (otherwise the history gets too messy to deal with), tempted to split the commit into what that says and adding the admin functionality afterwards

app/views/admin_general/index.html.erb
garethrees: or indeed, `#last_reported_at`

spec/models/comment_spec.rb
lizconlan: probably, was playing with this rather than have a "test" that a hard-coded array returns the same hard-coded array

garethrees: Actually, you're right. The content doesn't actually matter here (in that, there's no validation to say that it _needs_ to be one of those exact strings). I'd drop the expectation of it being an `Array`, because `.to all()` makes the same assertion.

lizconlan:
ToDo:
- [ ] quick read through to see if it make sense
- [ ] consider rewording that last commit message
- [ ] screenshots of Events on the comment editing screen

lizconlan:
Events section of the Edit annotations page:

screen shot 2017-02-27 at 10 35 04

screen shot 2017-02-27 at 10 35 16

screen shot 2017-02-27 at 11 45 41

PR #3709, opened 2017-02-02
[pro] Design polish

wrightmartin:

PR #3713, opened 2017-02-03
[Pro] Improve pro status picker

stevenday:
This add some really basic css and js to make the pro status picker work nicer

@wrightmartin, I think this is ready for some more styling polish and perhaps
some WDTK-specific theming now.

For mysociety/alaveteli-professional#120


stevenday:
From looking at the other changes in #3709 - there's a couple of things that should probably change:
- button class on the update button for non-js users?
- should the styles like under /responsive/alaveteli_pro instead?

@crowbot - should I be using @info_request.described_state here or @info_request.calculate_status?

stevenday:
@crowbot  - ok, I've updated this quite a bit, hence squashing everything down into a new commit.

Instead of the js I had before, and so that it fits in with @wrightmartin's work using houdini for the embargo, it now uses houdini to do all the dropdown stuff, and only uses JS to actually submit the form when you pick a new status. I've also moved the css into alaveteli_pro as requested.

I think what I'm explicitly waiting for @wrightmartin to do is add icons to each status option - they should have classes already to make that possible.

stevenday:
@wrightmartin - I've rebased this and fixed the merge conflict, as well as adding a current status line like we discussed. However, there are couple of css issues (including one I've just added) that I need some help with:

- [ ] The icons in the dropdown form don't quite line up with the text in the labels
- [ ] The new icon I added to the current status line doesn't line up either
- [ ] (Maybe related to the first thing) the hover states for the labels of each option in the dropdown cause the icon to be obscured in the row above

I'm also wondering whether increasing the font size of the form labels was such a good idea now, as with the icon they end up taking up three lines in some places. Perhaps we should knock it back down to `0.85em`?

stevenday:
I fixed the css stuff myself (with a lot of help from css-tricks). It now looks like:

screen shot 2017-02-22 at 12 49 39

Which I think is ok, but will wait for @crowbot or @wrightmartin to confirm before merging...

wrightmartin:
👍

Issue #3676, opened 2017-01-13
Using double quotes in the display name part of an email address breaks mail sending


lizconlan:
Double quotes were added to the translation string used to create the address line[1] for one of the sites and it stopped new messages being sent. (This took us a while to work out as we updated the translations as part of a version upgrade and assumed that something in the code was broken, then we thought it might be something to do with UTF-8 handling which, for once, it wasn't.)

Left to its own devices, `InfoRequest#incoming_name_and_email` calls `MailHandler.address_from_name_and_email` which turns `"FOI request at \"Test Authority\" stuff"` into[2]

```rb
"\"FOI request at \\\"Test Authority\\\" stuff\""
````

(which doesn't work)

as `MailHandler.address_from_name_and_email` is one of ours, stripping out the extra escape characters is as "simple" as changing the last line as follows:

```diff
- address.to_s
+ address.to_s.delete('\\')
```

...which gets us...

```rb
"\"FOI request at \"Test Authority\" stuff\""
```

...which distressingly doesn't work either (although the entirely artificial `'"FOI request at "Test Authority" stuff'"` does). At this point I gave up and changed the translation file.

Oh and don't call `address.to_s` more than once (or call `.dup` on it) otherwise something odd happens and accented characters get turned into escape sequences e.g. "Autorità" becomes "Autorit\xC3\xA0" - no idea what's going on there so easiest just to avoid it[3].

It's possible this is related to https://github.com/mikel/mail/pull/1023 but I'm tired and can't be certain (note, we are running on mail 2.5.4 while we're still on Rails 3.2.x so maybe not).

[1] currently (for version 0.26.0.0 and up) `{{law_used}} requests at {{public_body}`
[2] the complete string is `"\"FOI request at \\\"Test Authority\\\" stuff\"<test@example.com>"` but as I only showed one part of what was  being passed in, the full string seemed unexpected but I've included here for clarity about where the string wrapping occurs
[3] appears to have been fixed in a later version of the mail gem

garethrees:
Action: Add translator notes for this translation string

garethrees:
For searchability, this is the exception that came through:

```
An ArgumentError occurred in request#new:

An SMTP To address is required to send a message. Set the message smtp_envelope_to, to, cc, or bcc address.
vendor/bundle/ruby/1.9.1/gems/mail-2.5.4/lib/mail/check_delivery_params.rb:9:in `check_delivery_params'
```

lizconlan:
[...]

PR #3717, opened 2017-02-06
Add translators' note for string which can't contain double quotes

crowbot:
This was added to app.po in 63055c52, but that is erased when updating translation files from Transifex.
This should persist.


garethrees:
Great! Thanks! Fixes https://github.com/mysociety/alaveteli/issues/3676

PR #3718, opened 2017-02-06
[Pro] Fix missing states in Phase calculations

stevenday:
This adds the overdue versions of the waiting_response status to the phase
calculator.

Fixes mysociety/alaveteli-professional#149


stevenday:
@wrightmartin - how's this? I didn't realise we have separate phases for overdue and very overdue, so they should have got their own logic (in your original screenshot it was those that were getting the wrong icons and labels I think).

PR #3724, opened 2017-02-07
Better Postgres collation support check

### LINKED ISSUE #3683: Make `DatabaseCollation.supports?` check database encoding
The `DatabaseCollation.supports?` method only checks whether a named collation is available to the current database, it does not check whether it is compatible with the database encoding.

This is currently preventing us from using the it_IT locale as this does exist in the database but is only available for UTF8 encoding (our database uses SQL_ASCII). If we try to use the incompatible location after having checked its availability, we get an error back from Postgres which prevents `PublicBody#list` from rendering:

```
ERROR:  collation "it_IT" for encoding "SQL_ASCII" does not exist
```

(Background - the lookup on `pg_collation` is intended to be a multi-part key which includes the encoding - ref: https://www.postgresql.org/docs/9.2/static/catalog-pg-collation.html)

Fixes https://github.com/mysociety/dirittodisapere-theme/issues/106

lizconlan:
Adds a check to for whether the current database encoding is compatible with the requested named collation when asking whether it is available for use.

Closes #3683

### DIFF
--- a/doc/CHANGES.md
+++ b/doc/CHANGES.md
@@ -11,6 +11,8 @@
   and dashboard for managing to-do items and requests. This is functionality
   being piloted in the UK and is not yet recommended for use in other locales
   (Steve Day, Martin Wright, Louise Crow)
+* Fix bug that meant a Postgres collation that was not compatible with the local
+  database encoding could be chosen (Liz Conlan)
 
 ## Upgrade Notes
 * Please update any overriden templates and theme code that reference times and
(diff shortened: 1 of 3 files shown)


garethrees:
Also, changelog plz! 👍

lizconlan:
> Main thing was just making filter_collations a little clearer by explicitly passing the same arguments as the intended SQL.

Cheers, wasn't entirely happy with that tbh 👍

lizconlan:
merged as hotfix/0.26.0.9 and hotfix/0.27.0.4

Issue #3102, opened 2016-03-17
Admin interface interferes with "download zip file" for non-admins


kingqueen3065:
When admins  attempt to print a request page, the CSS on the Alaveteli ribbon goes wrong, resulting in ugliness:
![ss 2016-03-17 at 12 33 58](https://cloud.githubusercontent.com/assets/11923618/13832927/a6eae226-ebd8-11e5-9984-977d2e6c9836.png)
No problem: I can log out and print it without the ribbon. 
If, however, I select to "Download a zip file of all correspondence" this creates a problem for anybody else using that feature. It appears that Alaveteli uses a PDF printer, thus resulting in the ugliness in the correspondence.pdf. Alaveteli then caches the created PDF, so anybody else subsequently using the "Download a zip file..." feature for the same request also has the ugliness. There doesn't seem to be a way of undoing it either.
An unimportant bug of limited impact in somewhat obscure circumstances.
Cheers
Doug

PR #3732, opened 2017-02-09
Hide admin nav in PDF download

garethrees:
Before / After:

screen shot 2017-02-09 at 11 04 18

EDIT: ^ Actually that's not quite true, in that there's a bug in core (https://github.com/mysociety/alaveteli/issues/3733) which is affecting layout that I'm looking in to. Those are from wdtk-theme.

Fixes https://github.com/mysociety/alaveteli/issues/3733
Fixes https://github.com/mysociety/alaveteli/issues/3102

### DIFF
--- a/app/assets/stylesheets/responsive/_print_layout.scss
+++ b/app/assets/stylesheets/responsive/_print_layout.scss
@@ -1,7 +1,6 @@
 div#content, div#left_column, div.entirebody div#wrapper {
   width: 100%;
   margin: 0;
-  float: none;
 }
 
 div#content {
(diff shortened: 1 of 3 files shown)


app/assets/stylesheets/responsive/_print_layout.scss
```diff
@@ -1,7 +1,6 @@
 div#content, div#left_column, div.entirebody div#wrapper {
   width: 100%;
   margin: 0;
-  float: none;
```
lizconlan: that makes sense in theory but as it's breaking things, taking it out seems fair 👍

PR #3735, opened 2017-02-09
New mail log parsing

garethrees:
RE: https://github.com/mysociety/sysadmin/issues/799

Measures to handle WDTK move to using a smarthost for outgoing mail.

This PR gets us to the point of being able to switch over to `/var/log/all_servers_mail.log*`. When we change the config flag, we'll also need to create `MailServerLogDone` records with the same filename and `last_stat` as the most recent `all_servers_mail.log` files so that we don't try to ingest logs that we've effectively already got.

When we switch over, we'll end up importing some duplicate lines.

```sh
$ ls /var/log/exim4/exim-mainlog-* | sort | tail -3
/var/log/exim4/exim-mainlog-20170208.gz
/var/log/exim4/exim-mainlog-20170209.gz
/var/log/exim4/exim-mainlog-20170210                 # This is the last file imported
$ ls /var/log/all_servers_mail.log* | sort | tail -3
/var/log/all_servers_mail.log.20170207.gz
/var/log/all_servers_mail.log.20170208.gz
/var/log/all_servers_mail.log.20170209               # This has the lines in exim-mainlog-20170210 and more
```

We're going to have to import the older logs anyway to account for old requests that only have the "sent to smarthost" delivery status.

I think its going to be easier to just import everything, and then remove duplicates:

```ruby
switch_date = Date.parse('…')

MailServerLog.
  where('created_at >= ?', switch_date).
  order('id DESC').
  find_each do |log|
    if MailServerLog.where(:info_request_id => log.info_request_id,
                           :line => strip_syslog_prefix(log.line)).any?
      log.destroy!
    end
end

def strip_syslog_prefix(line)
  prefix_regexp = /\A(.*?\s)\d{4}-\d{2}-\d{2}\s\d{2}:\d{2}:\d{2}.*\z/
  match = line.match(prefix_regexp).try(:[], 1)
  if match
    line.gsub(match, '')
  else
    line
  end
end
```

```sh
# Here are some existing lines:
2017-02-10 00:12:21 [27739] 1cbyp3-0007DP-JK <= track@whatdotheyknow.com U=alaveteli P=local S=42746 id=589d056585275_7f72854fe429646b@titan.mail T="Your WhatDoTheyKnow email alert" from <track@whatdotheyknow.com> for user@example.com user@example.com
2017-02-10 00:12:22 [27742] 1cbyp3-0007DP-JK => user@example.com <user@example.com> F=<track@whatdotheyknow.com> P=<track@whatdotheyknow.com> R=send_to_smarthost T=remote_smtp S=43680 H=smarthost.example.org [127.0.0.1]:25 X=TLS1.2:DHE_RSA_AES_128_CBC_SHA1:128 CV=no DN="CN=mx0.mysociety.org" C="250 OK id=1cbyp4-0000Ec-0E" QT=1s DT=1s

# In all_servers_mail.log they'll get imported like this:
Feb 10 00:12:21 appserver exim[27739]: 2017-02-10 00:12:21 [27739] 1cbyp3-0007DP-JK <= track@whatdotheyknow.com U=alaveteli P=local S=42746 id=589d056585275_7f72854fe429646b@titan.mail T="Your WhatDoTheyKnow email alert" from <track@whatdotheyknow.com> for user@example.com user@example.com
Feb 10 00:12:22 appserver exim[27742]: 2017-02-10 00:12:22 [27742] 1cbyp3-0007DP-JK => user@example.com <user@example.com> F=<track@whatdotheyknow.com> P=<track@whatdotheyknow.com> R=send_to_smarthost T=remote_smtp S=43680 H=smarthost.example.org [127.0.0.1]:25 X=TLS1.2:DHE_RSA_AES_128_CBC_SHA1:128 CV=no DN="CN=mx0.mysociety.org" C="250 OK id=1cbyp4-0000Ec-0E" QT=1s DT=1s

# Once we strip the syslog prefix, we can then compare against existing
# identical lines and delete where necessary
```

### DIFF
--- a/app/models/outgoing_message.rb
+++ b/app/models/outgoing_message.rb
@@ -448,11 +448,31 @@ class OutgoingMessage < ActiveRecord::Base
   end
 
   def exim_mail_server_logs
-    mta_ids.flat_map do |mta_id|
+    logs = mta_ids.flat_map do |mta_id|
       info_request.
         mail_server_logs.
           where('line ILIKE :mta_id', mta_id: "%#{ mta_id }%")
     end
+
+    smarthost_logs = logs.flat_map do |log|
+      line = log.line(:decorate => true)
+      if line.delivery_status.delivered?
+        match = line.to_s.match(/C=".*?id=(?<message_id>\w+-\w+-\w+).*"/)
+        match[:message_id] if match
+      end
+    end
+
+    smarthost_logs.compact!
+
+    more_logs = smarthost_logs.flat_map do |mta_id|
+      info_request.
+        mail_server_logs.
+          where('line ILIKE :mta_id', mta_id: "%#{ mta_id }%")
+    end
+
+    # Need to call #uniq because the more_logs query pulls out the initial
+    # delivery line
+    (logs + more_logs).uniq
   end
 
   def postfix_mta_ids
--- a/app/controllers/outgoing_messages/delivery_statuses_controller.rb
+++ b/app/controllers/outgoing_messages/delivery_statuses_controller.rb
@@ -12,7 +12,7 @@ class OutgoingMessages::DeliveryStatusesController < ApplicationController
 
     if @show_mail_server_logs
       @mail_server_logs = @outgoing_message.mail_server_logs.map do |log|
-        log.line(:redact_idhash => !@user.super?)
+        log.line(:redact => !@user.super?)
       end
     end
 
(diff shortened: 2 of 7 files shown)


app/models/outgoing_message.rb
```diff
@@ -448,11 +448,31 @@ def exim_mta_ids
...
 
   def exim_mail_server_logs
-    mta_ids.flat_map do |mta_id|
+    logs = mta_ids.flat_map do |mta_id|
       info_request.
         mail_server_logs.
           where('line ILIKE :mta_id', mta_id: "%#{ mta_id }%")
     end
+
+    more_mta_ids = logs.flat_map do |log|
```
crowbot: Might be worth being a bit more specific than `more` - `smarthost` or `remote` maybe, to give a clearer indication of what we're doing here.

PR #3744, opened 2017-02-13
Add some more style checks

garethrees:


stevenday:
Is that all? ;) I don't see any problem with these, just wondering if we could turn on even more? (Is there a general plan/process for making the code more consistent?)

garethrees:
> Is that all? ;) I don't see any problem with these, just wondering if we could turn on even more? (Is there a general plan/process for making the code more consistent?)

We initially started with an "everything-on" approach but found it was too difficult with the current code quality, so we decided to try to ease things in.

We've been meaning to do a handful per sprint so that we can ease them in, but haven't done it for a while. I've added "Add 5 hound checks per sprint" to the top of our sprint doc so that we'll hopefully get moving on this a bit faster.

PR #3746, opened 2017-02-13
[Pro] Update pro authority selecting js to search via xapian

stevenday:
Rather than dumping all the bodies into a select box, this turns the To field
back into a basic text box, then uses selectize's ability to load data from a
remote source via ajax to query xapian for results instead.

For non-js users, the same text box can be used to submit a query into the
existing select_authority page, presenting that in a pro livery and allowing
pagination/searching etc to take place before redirecting the user back into
the pro form with the body selected.

Closes mysociety/alaveteli-professional#159

### DIFF
--- a/app/assets/javascripts/alaveteli_pro/authority_select.js
+++ b/app/assets/javascripts/alaveteli_pro/authority_select.js
@@ -27,7 +27,13 @@
       var oldSalutation = salutationTemplate.replace(defaultAuthorityName, oldAuthorityName);
       var oldMessage = $message.val();
 
-      var newAuthorityName = $item.text();
+      // The selected item contains the name of the authority and the remove
+      // button (which is a link), so this gets only the text node from the
+      // DOM, ignoring the remove button. A naive .text() would return the
+      // multiplication symbol in that button too.
+      var newAuthorityName = $item.contents().filter(function(){
+        return this.nodeType === Node.TEXT_NODE;
+      })[0].nodeValue;
       var newSalutation = salutationTemplate.replace(defaultAuthorityName, newAuthorityName);
       var newMessage = oldMessage.replace(oldSalutation, newSalutation);
 
@@ -68,7 +74,8 @@
             searchUrl + '/' + encodeURIComponent(query),
             callback
           ).fail(callback);
-      }
+      },
+      plugins: ['remove_button']
     });
   });
 })(window.jQuery);
--- a/app/assets/javascripts/alaveteli_pro/authority_select.js
+++ b/app/assets/javascripts/alaveteli_pro/authority_select.js
@@ -1,13 +1,19 @@
 (function($){
   $(function(){
     var $select = $('.js-authority-select');
+    var $publicBodyId = $('.js-public-body-id');
+    var $submit = $('.js-authority-select-submit');
     var $message = $('.js-outgoing-message-body');
     var searchUrl = $select.data('search-url');
     var initialOptions = [];
     var initialAuthority;
-    if($select.data('initial-authority').id) {
+    if($select.data('initial-authority')) {
       initialAuthority = $select.data('initial-authority');
       initialOptions = [initialAuthority];
+      // Selectize looks for values from the value field (id in this case) to
+      // see if it should show something selected, but the html currently has
+      // the name, not the id.
+      $select.val(initialAuthority.id)
     }
     var defaultAuthorityName = $message.data('salutation-body-name');
     var salutationTemplate = $message.data('salutation-template');
@@ -17,7 +23,7 @@
       currentAuthorityName = initialAuthority.name;
     }
 
-    var updateSalutation = function updateSalutation(value, $item) {
+    var updateSalutation = function updateSalutation($item) {
       var oldAuthorityName = currentAuthorityName;
       var oldSalutation = salutationTemplate.replace(defaultAuthorityName, oldAuthorityName);
       var oldMessage = $message.val();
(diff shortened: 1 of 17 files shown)


stevenday:
This needs a matching branch in WDTK-theme to be complete: https://github.com/mysociety/whatdotheyknow-theme/pull/381

spec/controllers/alaveteli_pro/public_bodies_controller_spec.rb
```diff
@@ -0,0 +1,38 @@
+require 'spec_helper'
+
+RSpec.describe AlaveteliPro::PublicBodiesController do
+
+  describe "#search" do
+    # Note: these specs rely on the global fixture data loaded by spec_helper
```
crowbot: I don't think this has to use fixtures, as long as it updates the xapian index. How about: 

```diff
-     # Note: these specs rely on the global fixture data loaded by spec_helper
-     # because they need public bodies to be indexed in Xapian
     let!(:pro_user) { FactoryGirl.create(:pro_user) }
+    let!(:body){ FactoryGirl.create(:public_body, :name => 'example') }
 
     before do
       session[:user_id] = pro_user.id
+      update_xapian_index
     end
 
     it "returns json" do
       with_feature_enabled :alaveteli_pro do
-        get :search, query: "department"
+        get :search, query: body.name
         expect(response.content_type).to eq("application/json")
       end
     end
 
     it "returns bodies which match the search query" do
       with_feature_enabled :alaveteli_pro do
-        get :search, query: "Department for Humpadinking"
+        get :search, query: body.name
         results = JSON.parse(response.body)
-        expect(results[0]['name']).to eq("Department for Humpadinking")
+        expect(results[0]['name']).to eq(body.name)
       end
     end
 
     it "returns a whitelisted set of properties for each body" do
       with_feature_enabled :alaveteli_pro do
-        get :search, query: "Department for Humpadinking"
+        get :search, query: body.name
         results = JSON.parse(response.body)
         expected_keys = %w{id name notes info_requests_visible_count weight}
         expect(results[0].keys).to match_array(expected_keys)
```
If we do end up using the fixtures, then we need to add `get_fixtures_xapian_index` to the `before` block to make sure that, no matter what order tests are run in, the xapian index has the fixtures in it before these tests are run.

app/controllers/alaveteli_pro/public_bodies_controller.rb
```diff
@@ -0,0 +1,24 @@
+class AlaveteliPro::PublicBodiesController < AlaveteliPro::BaseController
+  include ActionView::Helpers::TextHelper
+  include ActionView::Helpers::TagHelper
+
+  def search
```
crowbot: Let's keep this here for now, but in the longer term, it seems like it might be simpler to have it returned if `PublicBodyController.search_ahead` is called with a json format param - would keep all the typeahead logic in one place.

app/views/request/select_authority.html.erb
```diff
@@ -21,7 +21,7 @@
 <h1><%= _('Find an authority') %></h1>
 
 <div id="authority_selection" class="authority_selection">
-  <%= form_tag select_authority_path, { :id => 'search_form', :method => 'get' } do %>
+  <%= form_tag "", { :id => 'search_form', :method => 'get' } do %>
```
crowbot: Would be good to add an explanatory note that this form is used from multiple URLs. Otherwise, I can imagine it being changed back by accident.

stevenday:
@crowbot - thanks! I've pushed fixes for most of those things:

- controller spec now uses factories instead in c55ea10
- explanatory comment added about the empty form action in e29b3d9
- potential pro livery viewable fixed in e29b3d9
- flash of submit button fixed in 059c29a (along with inline style attr)
- difficulty removing things hopefully helped by 7262e0e?
- Make a request tab is now highlighted on pro too in b1c520e (I think this is actually the desired behaviour?)

I've also opened https://github.com/mysociety/alaveteli-professional/issues/172 for the batch request link issue and https://github.com/mysociety/alaveteli-professional/issues/174 for the search typeahead

spec/controllers/alaveteli_pro/public_bodies_controller_spec.rb
```diff
@@ -6,29 +6,31 @@
     # Note: these specs rely on the global fixture data loaded by spec_helper
     # because they need public bodies to be indexed in Xapian
```
crowbot: We can drop this note now.

app/assets/javascripts/alaveteli_pro/authority_select.js
```diff
@@ -27,7 +27,13 @@
       var oldSalutation = salutationTemplate.replace(defaultAuthorityName, oldAuthorityName);
       var oldMessage = $message.val();
 
-      var newAuthorityName = $item.text();
+      // The selected item contains the name of the authority and the remove
+      // button (which is a link), so this gets only the text node from the
+      // DOM, ignoring the remove button. A naive .text() would return the
+      // multiplication symbol in that button too.
+      var newAuthorityName = $item.contents().filter(function(){
+        return this.nodeType === 3;
```
crowbot: Any chance we can use `Node.TEXT_NODE` instead of `3` to be clearer what we're doing here?

app/views/general/_nav_items.html.erb
```diff
@@ -1,4 +1,4 @@
-<li class="<%= 'selected' if params[:controller] == 'request' and ['new', 'select_authority'].include?(params[:action]) %>">
+<li class="<%= 'selected' if (params[:controller] == 'request' and ['new', 'select_authority'].include?(params[:action])) or params[:controller] = 'alaveteli_pro/request' and ['new', 'preview', 'edit'].include?(params[:action]) %>">
```
crowbot: There's a typo in this line - `params[:controller]` is being assigned to with a single `=` rather than compared with `==`.

PR #3759, opened 2017-02-16
[pro] Use human-readable dates

crowbot:


stevenday:
screen shot 2017-02-16 at 16 01 32

stevenday:
Sorry, I thought I'd added a link!

Each column is in here: https://github.com/mysociety/alaveteli/blob/develop/app/assets/stylesheets/responsive/alaveteli_pro/_requests_layout.scss#L132-L198 so maybe we can steal a few percent from the title column and add it to the dates?

wrightmartin:
Now I've seen this I'm not sure it's a good idea on the table-esque view - have we made it harder to compare them?

We definitely still need it when we output a date in other places though.

stevenday:
:+1:  Yep, this is fine now, merge away!

PR #3760, opened 2017-02-16
Don't specify status update form location.

crowbot:


stevenday:
I just kicked off a restart of travis here because this looks ok to me, and it was only one out of four builds that failed.

stevenday:
👍  Yep, merge away!

PR #3761, opened 2017-02-16
[Pro] Append user's name to initial request template

stevenday:
Closes mysociety/alaveteli-professional#145

### DIFF
--- a/app/controllers/alaveteli_pro/info_requests_controller.rb
+++ b/app/controllers/alaveteli_pro/info_requests_controller.rb
@@ -101,6 +101,7 @@ class AlaveteliPro::InfoRequestsController < AlaveteliPro::BaseController
     @draft_info_request = DraftInfoRequest.new(public_body: @public_body)
     @info_request = InfoRequest.new(public_body: @public_body)
     @outgoing_message = OutgoingMessage.new(info_request: @info_request)
+    @outgoing_message.set_signature_name(current_user.name)
     # TODO: set duration based on current user's account settings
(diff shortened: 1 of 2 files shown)


app/controllers/alaveteli_pro/info_requests_controller.rb
```diff
@@ -101,6 +101,7 @@ def create_initial_objects
     @draft_info_request = DraftInfoRequest.new(public_body: @public_body)
     @info_request = InfoRequest.new(public_body: @public_body)
     @outgoing_message = OutgoingMessage.new(info_request: @info_request)
+    @outgoing_message.set_signature_name(current_user.name) if current_user.present?
```
crowbot: Elsewhere in the controller, we've just assumed `current_user` is present, so we could probably do the same here for simplicity's sake.

spec/integration/alaveteli_pro/request_creation_spec.rb
```diff
@@ -156,13 +156,26 @@
...
-                                     'faithfully," signature'
         expect(page).to have_content "Please select an authority"
       end
     end
 
+    it "fills out the body with a template message" do
+      using_pro_session(pro_user_session) do
+        # New request form
+        visit new_alaveteli_pro_info_request_path
+        expect(page).to have_content <<-HERE
```
crowbot: We've got an almost-convention of using `EOF` as the end marker, so might be nice to switch to that.

stevenday:
@crowbot - thanks, I've pushed a fixup, will just wait for travis to double-check before I squash and merge

Issue #2968, opened 2015-12-21
Upgrade to Rails 4.0


garethrees:
> As per our maintenance policy, the release of Rails 5.0 will mean that bug fixes will only apply to 5.0.x, regular security issues to 5.0.x and 4.2.x, and severe security issues also to 5.0.x and 4.2.x (but when 5.1 drops, to 5.1.x, 5.0.x, and 4.2.x). **This means 4.1.x and below will essentially be unsupported!** Ruby 2.2.2+ is now also the only supported version of Ruby for Rails.
> 
> – http://weblog.rubyonrails.org/2015/12/18/Rails-5-0-beta1/
## Resources
- http://edgeguides.rubyonrails.org/upgrading_ruby_on_rails.html
- http://railsdiff.org/3.2.22.2/4.0.0
- http://www.ready4rails.net/gemfile_check
- http://www.rails-upgrade-checklist.com/#4.0
## Pre-upgrade steps (to be merged as PRs before upgrade)
- [x] Upgrade gems to latest version compatible with Rails 3.2
- [x] Replace `attr_accessible` calls with use of `strong_params` gem (remove `mass_assignment_sanitizer` option from code.
- [x] Ensure all `:scope` macros use procs or lambdas
- [x] Remove all use of the old-style hash based finder API.
- [x] [Remove dynamic finders](http://edgeguides.rubyonrails.org/upgrading_ruby_on_rails.html#upgrading-from-rails-3-2-to-rails-4-0-active-record)
- [x] Replace `:confirm` on form helpers with data attributes (need to check this works in Rails 3)
- [x] Remove clashing named routes 
- [x] `match` routes to specify the request method or use `resources` method instead
## Upgrade steps for Rails 4.0
- [x] [Remove assets line from gemfile](http://edgeguides.rubyonrails.org/upgrading_ruby_on_rails.html#upgrading-from-rails-3-2-to-rails-4-0-gemfile)
- [x] Replace `ActiveRecord::Fixtures` with `ActiveRecord::FixtureSet`
- [x] Update code which uses `ActiveModel::Validations::ConfirmationValidator` to check  `:#{attribute}_confirmation` instead of `attribute` for errors.
- [x] Add gem 'dalli' to Gemfile - it's the new default memcached client.
- [x] Replace `ActionController::Routing` with `ActionDispatch::Routing`
- [x] Change `config.assets.compress` option to `config.assets.js_compressor`
[...]

PR #3768, opened 2017-02-20
[#2968] Rails 4

garethrees:
Work on #2968


garethrees:
Current failures are all pro-related.

```sh
# expected AlaveteliPro::ActivityList::RequestSent to be a kind of AlaveteliPro::ActivityList::RequestResent
rspec ./spec/models/alaveteli_pro/activity_list/list_spec.rb:53 # AlaveteliPro::ActivityList::List#current_items returns an array of items representing the current page of events

# string to symbol keys
rspec ./spec/controllers/request_controller_spec.rb:420 # RequestController when showing one request should set @state_transitions for the request

# ActionView::Template::Error: First argument in form cannot contain nil or be empty
rspec ./spec/views/comment/new.html.erb_spec.rb:13 # comment/new.html.erb when the request is embargoed says the comment will be public when the embargo expires
rspec ./spec/views/comment/new.html.erb_spec.rb:21 # comment/new.html.erb when the request is embargoed renders the professional comment suggestions
rspec ./spec/views/comment/new.html.erb_spec.rb:34 # comment/new.html.erb when the request is not embargoed says the comment will be public
rspec ./spec/views/comment/new.html.erb_spec.rb:40 # comment/new.html.erb when the request is not embargoed renders the normal comment suggestions

# No route matches {:action=>"new", :controller=>"alaveteli_pro/request", :draft_id=>"299", :info_request=>{"title"=>"Why is your quango called Geraldine?", "public_body_id"=>"9", "tag_string"=>""}, :locale=>"en", :outgoing_message=>{"body"=>"This is a silly letter. It is too short to be interesting.", "what_doing"=>"normal_sort"}, :post_redirect=>"1", :post_redirect_user=>nil, :preview=>"0", :submit=>"Send request", :submitted_new_request=>"1", :utf8=>"✓"}</h2>
rspec ./spec/integration/alaveteli_pro/request_creation_spec.rb:197 # creating requests in alaveteli_pro when writing a new request from scratch redirects to the pro page if the user starts the normal process
```

garethrees:
Commits to re-check after merge freeze on `develop`:


– [ ] Replace find_all_by_ with where                               b2ecd9c
– [ ] Replace find_or_create_by_ with find_or_create_by()           a3ea794
– [ ] Update password_confirmation errors                           d010760
– [ ] Remove deprecated association options                         03f290c
– [ ] Add missing HTTP method on match routes                       54b8390
– [ ] Explicitly expect partial rendered                            e47d062
– [ ] Remove deprecated find_by_                                    a3ac828
– [ ] Update scope to use where instead of conditions               c504bf0
– [ ] Use Arel instead of finder options                            d51b41f
– [ ] Fix ordering of anchor attributes                             4fb514f
– [ ] Use be_within when comparing dates                            e4201f2
– [ ] Keep hyphenated version of locales in i18n.available_locales  65922d9
– [ ] Remove invalid add_index option                               e3d7119
– [ ] Remove manual drop index statements                           2da9e79
– [ ] Remove add_index                                              ef981b9
– [ ] Remove deprecated Relation#all                                b5a970e
– [ ] Remove deprecated finder conditions                           94b63be
– [ ] Add explicit join references                                  3f4614e
– [ ] Remove deprecated disable_with button option                  ff8582f
– [ ] Upgrade will_paginate to remove deprecated finders            1e4072f
– [ ] Add missing erb tags                                          1c07cda
– [ ] Make scss @imports relative to base directory                 c226a5d
– [ ] Replace deprecated :confirm key                               70b7dd8

app/views/request/describe_notices/_error_message.html.erb
```diff
@@ -1 +1 @@
- _("Thank you! We'll look into what happened and try and fix it up.")
\ No newline at end of file
+<%= _("Thank you! We'll look into what happened and try and fix it up.") %>
```
garethrees: I think we should drop this from the rails upgrade so that we avoid creating translations

PR #3769, opened 2017-02-20
[pro] Fix issues with signing in and out as a pro

stevenday:
Pro logins could end up trying to redirect to the "Make a request" page for
pros when the PostRedirect wasn't really for that page, because the regex that
tried to match urls wasn't specific enough.

Similarly, post redirects could end up getting saved that caused problems when
a pro user logged out, because signing out of a pro page set a redirect path
back in to the pro area, which an anonymous user couldn't follow.

For mysociety/alaveteli-professional#196
For mysociety/alaveteli-professional#134

### DIFF
--- a/lib/alaveteli_pro/post_redirect_handler.rb
+++ b/lib/alaveteli_pro/post_redirect_handler.rb
@@ -6,7 +6,7 @@ module AlaveteliPro
     # log in, so we want to send them into the pro system
     def override_post_redirect_for_pro(uri, post_redirect, user)
       # We could have a locale in the url, or we could not, e.g. /en/new or /new
-      if uri =~ /(\/[a-z]{2})?\/new/
+      if uri =~ /^(\/[a-z]{2})?\/new$/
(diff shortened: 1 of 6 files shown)


lib/alaveteli_pro/post_redirect_handler.rb
```diff
@@ -6,7 +6,7 @@ module PostRedirectHandler
     # log in, so we want to send them into the pro system
     def override_post_redirect_for_pro(uri, post_redirect, user)
       # We could have a locale in the url, or we could not, e.g. /en/new or /new
-      if uri =~ /(\/[a-z]{2})?\/new/
+      if uri =~ /^(\/[a-z]{2})?\/new$/
```
crowbot: It'd be nice to have a spec for this that demonstrates that it no longer tries to redirect urls like `/new/body name]`, which seemed to be the cause of the error in this case.

stevenday:
@crowbot - I've pushed some new specs, how's that?

PR #3771, opened 2017-02-20
[Pro] - WIP - Stop contact form raising a 404 for pros with an embargoed last_request_id

stevenday:
I've got two approaches here which I'd like some feedback on @crowbot / @garethrees

The main issue is that when you visit the help form the controller searches
for a `last_request_id` in your cookies and adds this to the context if it
can find it. In f43d64533065272974412c0ab1d51f88baea738b @crowbot limited this
lookup to non-embargoed requests, but now we let pros view requests too
(through the same controller as other requests). Therefore, when you try to
send a help message as a pro who last viewed an embargoed request, you get a 404.

My first thought was to make this lookup a bit smarter:

21a0965dcd1b0de139dc56eefae0fafa1ecd0e28 and ad032e4bf377aeea30151006e5b040b126cd8ac8
add a new scope to InfoRequest that looks up any request the pro user can access,
embargoed or not.

However, this felt like it was stepping on the toes of CanCanCan a bit, so I
had a second go in c5b8d40e81f60f56eb1a51cfaac7e4cc70e1c0c5 to do things a different way.

Which would you prefer?

For mysociety/alaveteli-professional#207

### DIFF
--- a/app/controllers/help_controller.rb
+++ b/app/controllers/help_controller.rb
@@ -46,12 +46,10 @@ class HelpController < ApplicationController
     # request = InfoRequest.find_by(id: cookies["last_request_id"].to_i)
     @last_request = request if can?(:read, request)
 
-    last_body_id = cookies["last_body_id"].to_i
-    if last_body_id > 0
-      @last_body = PublicBody.find(last_body_id)
-    else
-      @last_body = nil
-    end
+    # Rails 3
+    @last_body = PublicBody.find_by_id(cookies["last_body_id"].to_i)
+    # Rails 4
+    # @last_body = PublicBody.find_by(id: cookies["last_body_id"].to_i)
 
     # submit form
     if params[:submitted_contact_form]
--- a/app/controllers/help_controller.rb
+++ b/app/controllers/help_controller.rb
@@ -40,12 +40,12 @@ class HelpController < ApplicationController
     end
 
     # look up link to request/body
-    last_request_id = cookies["last_request_id"].to_i
-    if last_request_id > 0
-      @last_request = InfoRequest.not_embargoed.find(last_request_id)
-    else
-      @last_request = nil
-    end
+    # Rails 3
+    request = InfoRequest.find_by_id(cookies["last_request_id"].to_i)
+    # Rails 4
+    # request = InfoRequest.find_by(id: cookies["last_request_id"].to_i)
+    @last_request = request if can?(:read, request)
+
     last_body_id = cookies["last_body_id"].to_i
     if last_body_id > 0
       @last_body = PublicBody.find(last_body_id)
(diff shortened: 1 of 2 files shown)


garethrees:
Definitely the CanCanCan version.

I think you can reduce all that down quite a bit:

```diff
diff --git a/app/controllers/help_controller.rb b/app/controllers/help_controller.rb
index bc11d66..4713a0c 100644
--- a/app/controllers/help_controller.rb
+++ b/app/controllers/help_controller.rb
@@ -40,17 +40,12 @@ class HelpController < ApplicationController
     end

     # look up link to request/body
-    last_request_id = cookies["last_request_id"].to_i
-    if last_request_id > 0
-      @last_request = InfoRequest.find(last_request_id)
-      # Check @last_request is accessible to the user, just in case we ever
-      # expose it to them somehow.
-      if cannot?(:read, @last_request)
-        @last_request = nil
-      end
-    else
-      @last_request = nil
-    end
+    # Rails 3
+    request = InfoRequest.find_by_id(cookies["last_request_id"].to_i)
+    # Rails 4
+    # request = InfoRequest.find_by(id: cookies["last_request_id"].to_i)
+    @last_request = request if can?(:read, request)
+
     last_body_id = cookies["last_body_id"].to_i
     if last_body_id > 0
       @last_body = PublicBody.find(last_body_id)
```

stevenday:
> EDIT: Do we actually want to do this? If admins can't see the request, then there's not much point setting it

In this case, the admins will be me & @crowbot - so we will be able to see it :)

garethrees:
> In this case, the admins will be me & @crowbot - so we will be able to see it :)

Ah of course, its going to `PRO_CONTACT_EMAIL`.

stevenday:
Thanks both, I've just pushed a revised version with the preferred option implemented properly along the lines of @garethrees suggestion, as well as some specs to cover it. 

The main change aside from the auth checking is that now it just ignores any missing requests/bodies, rather than 404ing, this doesn't lose us anything from the security perspective I don't think, but does remove an unnecessary error which might stop people sending their help message, so it seems like a good thing?

I also updated the way we look up bodies to match this whilst I was in the code.

app/controllers/help_controller.rb
```diff
@@ -46,12 +46,10 @@ def contact
...
-    last_body_id = cookies["last_body_id"].to_i
-    if last_body_id > 0
-      @last_body = PublicBody.find(last_body_id)
-    else
-      @last_body = nil
-    end
+    # Rails 3
+    @last_body = PublicBody.find_by_id(cookies["last_body_id"].to_i)
+    # Rails 4
+    # @last_body = PublicBody.find_by_id(id: cookies["last_body_id"].to_i)
```
garethrees: Typo here:

```diff
- # @last_body = PublicBody.find_by_id(id: cookies["last_body_id"].to_i)
+ # @last_body = PublicBody.find_by(id: cookies["last_body_id"].to_i)
```

Issue #3734, opened 2017-02-09
Run Alex's data export script


lizconlan:

lizconlan:
The error trap code is working, just needs some refinements so we don't have to re-run the entire thing

lizconlan:
ToDo: get files to Alex

lizconlan:
data download in progress \o/

PR #3756, opened 2017-02-16
Make export task skip lines with encoding problems

lizconlan:
Otherwise it raises an error which aborts the script run

Connects to #3734

PR #3777, opened 2017-02-22
Rescue all the research_export errors and only output redacted data

lizconlan:
So that we can rescue any Exception to ensure the run finishes
(with a view to examining the error output afterwards)

At least 2 examples of bad practice in here - worth considering refactor once we've proved whether it's useful

Connects to #3734

### DIFF
--- a/lib/tasks/export.rake
+++ b/lib/tasks/export.rake
@@ -85,16 +85,6 @@ def find_each_record(model)
   end
 end
 
-def append_csv_data(csv, data)
-  csv << data
-  # skip over data that causes "invalid byte sequence in UTF-8" exception
-rescue ArgumentError => err
-  puts "Error processing data:"
-  puts err.message
-  puts data.inspect
-end
-
-
 # Exports a model
 #
 # query    - a query used to limit the export to matching records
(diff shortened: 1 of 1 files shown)


lizconlan:
@garethrees I was trying to avoid begin/end blocks https://github.com/mysociety/alaveteli/blob/develop/.ruby-style.yml#L86-L89

lizconlan:
`csv << line` would need a begin/rescue/end as well

garethrees:
> @garethrees I was trying to avoid begin/end blocks https://github.com/mysociety/alaveteli/blob/develop/.ruby-style.yml#L86-L89

[`BEGIN`](https://github.com/bbatsov/ruby-style-guide#no-BEGIN-blocks) is not the same as [`begin`](https://github.com/bbatsov/ruby-style-guide#begin-implicit) (TIL!)

> `csv << line` would need a begin/rescue/end as well (both points of failure have been observed in production)

What error is happening at the point of appending to the CSV?

lizconlan:
> What error is happening at the point of appending to the CSV?

That will be our old friend `ArgumentError: invalid byte sequence in UTF-8` from https://github.com/mysociety/alaveteli/pull/3756#issuecomment-280321833

lizconlan:
> BEGIN is not the same as begin (TIL!)

I have learned something here today!

\*wanders off to fix code accordingly\*

lizconlan:
To recreate error conditions:

1) Simulate something going wrong in a lambda function

```diff
57:  def case_insensitive_user_censor(text, user)
58:    if user && text
59:      text.gsub(/#{user.name}/i, "<REQUESTER>")
60:    else
61:      text
62:    end
+      raise "fake error"
63:  end
```

2) Simulate a UTF-8 error in a stored attribute
```diff
57:  def case_insensitive_user_censor(text, user)
58:    if user && text
59:      text.gsub(/#{user.name}/i, "<REQUESTER>")
60:    else
61:      text
62:    end
+      "\x96"
63:  end
```

PR #3778, opened 2017-02-22
[Pro] Render authority search results in rails instead of JS

stevenday:
Because we need to translate body names and other text in our search results,
as well as pluralise things like request counts. This moves the HTML rendering
for each search result into a Rails view, passing it to the javascript in the
existing JSON we return.

On my dev box, this is noticeably slower than the previous version. Can anyone
shed any light on why that might be? Looking at the times reported by the
dev server in the console, there's nothing obvious (all the db/xapian stuff is
unchanged and the new partial view rendering seems to run very quickly).

For mysociety/alaveteli-professional#155

### DIFF
--- a/app/assets/javascripts/alaveteli_pro/authority_select.js
+++ b/app/assets/javascripts/alaveteli_pro/authority_select.js
@@ -54,15 +54,7 @@
       create: false,
       maxItems: 1,
       render: {
-        option: function(body, escape) {
-          // No need to use escape because data is trusted (from our DB)
-          var html = '<div class="recipient-result">';
-          html += '<h4 class="name">' + body.name + '</h4>';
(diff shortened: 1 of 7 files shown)


stevenday:
I've added a loading spinner so that there's some indication that the select box is loading something, which seems like a good idea regardless, but still none the wiser why this is so much slower. Maybe it won't be in production with cached templates?

stevenday:
Weird - maybe it's a vagrant thing disk access thing with the template rendering? (Would make sense if it's actually loading the template files each time I guess). 

Yeah, lets deploy it and see how it works for real.

stevenday:
Yep - I agree, this definitely feels like a better approach in general

PR #3779, opened 2017-02-22
[Pro] Javascript up the embargo extension form

stevenday:
Submits the form automatically on selection of an extension, and updates the
displayed expiry text immediately.

Also fixes a missed usage of the old unfriendly expiry dates in a flash message

For mysociety/alaveteli-professional#120


stevenday:
I'd missed the matching specs for the embargo flash change - have pushed a fixup now.

PR #3780, opened 2017-02-23
Improve export task

lizconlan:
Allow `CUTOFF_DATE` and `MODELS`to be passed in to the research_export task to manually specify a cutoff date (rather than just allow it to pick the current date at the start of the run) and to allow reports for one or more specified models to be produced rather than needing to complete the entire run each time.

(I've tested the new features locally and they seem to work)

example usage:
* `bundle exec rake export:research_export MODELS="OutgoingMessage,FoiAttachment"`
* `bundle exec rake export:research_export MODELS="OutgoingMessage" CUTOFF_DATE="2008-01-01"`

Connects to #3734

### DIFF
--- a/lib/tasks/export.rake
+++ b/lib/tasks/export.rake
@@ -92,7 +92,8 @@ end
 # override - pass in lambdas to modify a given column based on values in the row
 #
 # Returns a String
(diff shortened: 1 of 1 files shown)


lizconlan:
@garethrees ta!

PR #3784, opened 2017-02-27
[Pro] Rename 'Embargo' in the user interface to 'privacy'

wrightmartin:
- Embargo becomes 'privacy period'

- Embargo (speaking of the feature) becomes 'privacy'

Does not change any of the logic, where it is still known as embargo.

(this does not update any tests - as we've not decided this should be the change yet)

### DIFF
--- a/app/assets/stylesheets/responsive/_global_style.scss
+++ b/app/assets/stylesheets/responsive/_global_style.scss
@@ -153,6 +153,12 @@ label.form_label {
 
 .form_item_note,.form_note {
   font-size:0.875em;
+  line-height: 1.5em;
+}
+
+.form_item_note--important {
+  font-size: 1.142857143em;
+  line-height: 1.5em;
 }
 
 p + .form_item_note {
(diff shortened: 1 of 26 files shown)


stevenday:
I think this is a good start, but we should go a bit further than just swapping "an embargo" for "a privacy period" - we're still thinking of it as an object that they're creating, but that's an implementation detail, the user just wants to keep their request private.

Maybe to do that, we need to make some of the language more specific. Here's a few attempts to simplify it by just focussing on the "keeping it private" and "making it public" options:

```diff
- Your privacy period has been extended! It will now expire on {{expiry_date}}.
+ Your request will now be private on WhatDoTheyKnow until {{expiry_date}}.
```

```diff
- The privacy period for your request to {{public_body_name}} "{{info_request_title}}" has ended so the request is now public.
+ Your request to {{public_body_name}} "{{info_request_title}}" is now public on WhatDoTheyKnow.
```

```diff
- Requests with expiring privacy periods
+ Requests that will be made public soon
```

```diff
- {{count}} privacy periods are ending this week
+ {{count}} requests will be made public this week
```

```diff
- Extend or approve these privacy periods
+ Publish these requests or keep them private for longer
```

stevenday:
I've rebased this so that it's not conflicted and pushed a version using my ideas from https://github.com/mysociety/alaveteli-professional/issues/197#issuecomment-282696308 as a comparison to @wrightmartin's

PR #3785, opened 2017-02-27
[WIP][Pro] improved nav

crowbot:


stevenday:
I think I prefer this to the other options. My only (minor) comment would be that the "My public requests" item in the little user menu is superfluous for Pro's - they already have a better way of viewing their own requests right there in the main menu (and I think we've already talked about a filter option there for public/private requests, so there would be no need for this other page of just public stuff at all).

app/views/alaveteli_pro/general/_log_in_bar_links.html.erb
```diff
@@ -1,5 +1,3 @@
-<li><%= link_to _("My dashboard"), alaveteli_pro_dashboard_path %></li>
-<li><%= link_to _("My requests"), alaveteli_pro_info_requests_path %></li>
-<li><%= link_to _("My profile"), show_user_profile_path(:url_name => @user.url_name)
+<li><%= link_to _("My profile"), show_user_path(:url_name => @user.url_name)
```
stevenday: This swap of `show_user_profile_path` for `show_user_path` - could it be made everywhere, or is there a reason it should be different for pros? (It seems to end up a the same place?)

app/helpers/application_helper.rb
```diff
@@ -151,4 +151,12 @@ def event_description(event)
   def can_ask_the_eu?(code)
     WorldFOIWebsites.can_ask_the_eu?(code)
   end
+
+  def controller?(*controller)
```
stevenday: Maybe these params should be *controller**s** and *action**s** just to be totally clear?

app/views/alaveteli_pro/general/_log_in_bar_links.html.erb
crowbot: `show_user_path` also shows your public FOI requests, which `show_profile_path` doesn't. It's also the default place where other users will get sent to see info about you. I think I'm still a bit hung up on the idea that journalists will want to keep track of how they, and their requests, are viewed publicly.

PR #3786, opened 2017-02-27
[Pro] User roles

crowbot:
- [x] need to check ability to change roles in admin user update action
- [x] seems to be creating extra roles from admin user update action 
- [x] need corresponding WDTK PR? https://github.com/mysociety/whatdotheyknow-theme/pull/386
- [x] create ticket to remove `pro_account` and `admin_level` in Release 0.30 https://github.com/mysociety/alaveteli/issues/3805
- [ ] need to run `rake db:seed` - add to release notes, run when deploying to WDTK
- [x] specs


stevenday:
When I use the admin, I'm getting more checkboxes than I expect in the roles:

screen shot 2017-03-01 at 09 59 22

(I logged in as an admin, made them a pro, then went to edit them to remove it again and got this).

The last one is trying to reference a non-existent role id (2).

I'm trying to wrap my head around the _keep fields and so on, but I was wondering why you didn't just use the [normal Rails form helper methods](http://api.rubyonrails.org/classes/ActionView/Helpers/FormHelper.html#method-i-check_box) for them? IIRC, they already do something clever with hidden fields so that an unchecked checkbox sends through a value that you could use to remove a role?

app/models/role.rb
```diff
@@ -0,0 +1,26 @@
...
+  validates :resource_type,
+            :inclusion => { :in => Rolify.resource_types },
+            :allow_nil => true
+
+  scopify
+  attr_accessor :_keep
+  validates :name,
+            :inclusion => { :in => lambda { |role| role.allowed_roles } }
+
+  ALLOWED_ROLES = ['admin', 'pro_user'].freeze
```
crowbot: I was thinking of the set as `admin`, `pro_admin` and `pro_user`. We could have `admin`, `pro` and `pro_admin` instead?

app/models/user.rb
```diff
@@ -411,17 +416,23 @@ def get_undescribed_requests
...
   # Does this user have extraordinary powers?
   def super?
-    admin_level == 'super'
+    warn %q([DEPRECATION] User#super? will be removed in 0.30.
+          It has been replaced by User#is_admin?).squish
+    is_admin?
+  end
+
+  def can_admin_roles
+    roles.map{ |role| Role.grants_and_revokes(role.name.to_sym) }.flatten.compact.uniq
```
garethrees: Should be able to use `flat_map` here

db/migrate/20170227140831_rolify_create_roles.rb
```diff
@@ -0,0 +1,19 @@
+class RolifyCreateRoles < ActiveRecord::Migration
+  def change
+    create_table(:roles) do |t|
+      t.string :name
+      t.references :resource, :polymorphic => true
+
+      t.timestamps
```
garethrees: This is going to want an argument of `null: false` for Rails 4

spec/fixtures/users_roles.yml
```diff
@@ -0,0 +1,3 @@
+pro:
```
crowbot: Yup - fixed in 522010d032d8b1c91179c735893b7a5dc8d57a7e

spec/controllers/admin_user_controller_spec.rb
```diff
@@ -41,17 +47,80 @@
...
+
+  it 'does not set a role the setter cannot grant and revoke' do
+    user = FactoryGirl.create(:user)
+    pro_role = Role.where(:name => 'pro').first
+    expect(user.is_pro?).to be false
+    post :update, { :id => user.id,
+                    :admin_user => { :name => user.name,
+                                     :ban_text => user.ban_text,
+                                     :about_me => user.about_me,
+                                      :role_ids => [pro_role.id],
```
crowbot: Good spot, fixed in 96e6113148c727c965eada4bcdea8a6d476aa3a6

PR #3790, opened 2017-02-28
Allow POST-ing to the select_authorities action.

### LINKED ISSUE #3789: Batch request interface doesn't work without JS
It needs to `POST` a form to the `select_authorities` action, but I removed the ability to `POST` to that action in 416ad02a

crowbot:
This is used in the absence of javascript. Fixes #3789


lizconlan:
👍

PR #3791, opened 2017-02-28
Rails 4 - fixup `.includes` join references

lizconlan:
Connects to #3768

### DIFF
--- a/lib/tasks/stats.rake
+++ b/lib/tasks/stats.rake
@@ -151,7 +151,8 @@ namespace :stats do
                     public_body_ids]
 
       request_count =
-        InfoRequest.includes(:public_bodies).where(conditions).count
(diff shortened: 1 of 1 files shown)


lib/tasks/stats.rake
```diff
@@ -151,7 +151,8 @@ namespace :stats do
                     public_body_ids]
 
       request_count =
-        InfoRequest.includes(:public_bodies).where(conditions).count
+        InfoRequest.includes(:public_body).where(conditions).
+          references(:public_bodies).count
```
lizconlan: it works without `references` because the conditions have hardcoded table names but it feels weirder without it

lizconlan: it doesn't work with `.includes(:public_bodies)` because the relation name is `:public_body`

PR #3802, opened 2017-03-02
[Pro] Basic batch request authority selecting

stevenday:
Here's a first stab at the js and non-js versions of the batch authority list
building UI.

(Note, this includes 3228-remove-test-unit because it's needed for the specs
to pass at the moment.)

To test it out, go to http://10.10.10.30:3000/alaveteli_pro/batch_request_authority_searches/new
(it's not linked to from anywhere else in the pro system yet).

You can add/remove and search for authorities via a usual form submission and
page reload process, and when you have javascript you can also do it via lots
of ajax requests. Here's a brief gif:

My main worry at the moment is that this is very slow for me locally. It takes
a really long time to perform the search in particular, which makes it kinda
useless for typeahead style searching. I've made a few performance tweaks, but
it seems to mainly spend time rendering each result and I'm not sure why (as
with the other search, the reported times don't really capture the real world
times). Maybe again this is only an issue in development?

I think otherwise this is ready for some initial review feedback on the
approach, and design input when @zarino or @wrightmartin have time.

For mysociety/alaveteli-professional#215


spec/factories/users.rb
```diff
@@ -32,6 +33,7 @@
 #  request_classifications_count     :integer          default(0), not null
 #  public_body_change_requests_count :integer          default(0), not null
 #  info_request_batches_count        :integer          default(0), not null
+#  draft_info_request_batches_count  :integer          default(0), not null
```
crowbot: Why do we need a counter here?

spec/controllers/alaveteli_pro/batch_request_authority_searches_controller_spec.rb
```diff
@@ -0,0 +1,96 @@
+# -*- encoding : utf-8 -*-
+require 'spec_helper'
+
+shared_examples_for "creating a search" do
+  it "performs a search" do
+    puts assigns[:search].results.inspect
```
crowbot: This should probably be removed now.

stevenday:
I've merged all of @zarino's work into this now too, so it can be reviewed as a whole.

For starters, I've noticed a couple of issues for @zarino:

Damn those long names

screen shot 2017-03-17 at 15 48 44

And double damn those millions of pages of results!

screen shot 2017-03-17 at 15 48 51

stevenday:
@louise - I'm expecting the same two specs here to fail that failed last time. I'm a bit stuck as to why they are - it's another situation where they both pass perfectly fine if run on their own, and even better, only one fails if you just run the single spec file, but in a run of more things, both fail. 

It's an issue with expecting search results from xapian, so I presume I'm doing something wrong with `update_xapian_index`. If you have any ideas on how to fix this, that would be much appreciated.

zarino:
> For starters, I've noticed a couple of issues for @zarino:

The first one is a flexbox bug – not sure why it's happening – will investigate.

There isn't really a quick design fix for the second one. Can we limit the number of pagination buttons that are generated? Or could we drop the numbers altogether and just have "Next" and "Prev"? Numbered pagination is really only useful when you think people will want to jump to a particular point in a very long list – but in our case, the ordering of the list items is so random, I doubt anybody will want to do anything other than click "Next".

louise:
@stevenday I don't think I'm meant to be included in this conversation! when you use @ louise it notifies me .. Thanks

spec/factories/users.rb
stevenday: Good point, I've removed that in d384073a15bb8d8888004411833a827f8db69c90

stevenday:
@crowbot - I've fixed up the last remaining issues here, including the pagination, and added some integration specs, so I think with your spec changes and @zarino's fixes too, this is ready for a final look over?

FYI @zarino - here's what I went for with the pagination:

screen shot 2017-03-21 at 11 02 02

PR #3807, opened 2017-03-03
Prevent re-edited comment from being erased

lizconlan:
#2968 – Prevents the original comment being overwritten with a new blank one when moving from Preview back to New


lizconlan:
TODO - specs!

garethrees:
👍

PR #3808, opened 2017-03-03
[Pro] Add the :pro_admin role, only this role can view embargoed requests

crowbot:
This PR should remove the admin and viewing of embargoed requests from regular admins. I've left the admin abilities out of the `feature_enabled?` call for pro in `Ability`, and haven't added `feature_enabled?` checks to the admin controllers as a hedge against pro being deactivated on an install and that producing unexpected results. Not convinced that's the right thing to do though. 

One place where we can't do much about preventing non pro-admins seeing pro material is in the misdelivered responses list in the admin index - we don't know what request these responses were intended for, so can't hide pro ones. The only thing I think we could do there is make the whole list only visible to pro admins but that does seem a bit overkill.

### DIFF
--- a/app/controllers/admin_comment_controller.rb
+++ b/app/controllers/admin_comment_controller.rb
@@ -28,9 +28,15 @@ class AdminCommentController < AdminController
   end
 
   def edit
+    if cannot? :admin, @comment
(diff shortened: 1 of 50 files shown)


stevenday:
👍

PR #3810, opened 2017-03-06
[#2968] Monkeypatch ActiveRecord to fix counter_cache bug

### LINKED ISSUE #3811: [#2968] Rails 4 `counter_cache` bug
There's a bug in Rails 4's counter cache implementation that (I think) got fixed in Rails 4 but not backported. We hit it when trying to change the association of a record e.g. assigning a request to a new user and finding the counter_cache values have been updated twice.

Sample effect:

```
Expected:
user_a.info_requests_count = 0
user_b.info_requests_count = 1

Got:
user_a.info_requests_count = -1
user_b.info_requests_count = 2
```

Connects to #2968

lizconlan:
Rails 4 introduces a bug that causes customised counter_cache code to be
fired twice when updating an assocation, as described here:
https://github.com/rails/rails/issues/10865#issuecomment-21373642

This fix adapted for Rails 4.0 from:
https://github.com/rails/rails/pull/14849/files

Fixes #3811 for #2968

### DIFF
--- a/config/initializers/alaveteli.rb
+++ b/config/initializers/alaveteli.rb
@@ -64,6 +64,7 @@ require 'alaveteli_pro/post_redirect_handler'
 require 'user_stats'
 require 'belongs_to_destroy_with_fk_constraint'
 require 'patch_tag_helper'
+require 'patch_counter_cache'
 
 AlaveteliLocalization.set_locales(AlaveteliConfiguration::available_locales,
                                   AlaveteliConfiguration::default_locale)
--- a/lib/patch_counter_cache.rb
+++ b/lib/patch_counter_cache.rb
@@ -1,0 +1,167 @@
+# -*- encoding : utf-8 -*-
+# Rails 4 introduces a bug that causes customised counter_cache code to be
+# fired twice as described here:
+# https://github.com/rails/rails/issues/10865#issuecomment-21373642
+#
+# This is a 4.0-specific fix adapted from https://github.com/rails/rails/pull/14849/files
+
+module ActiveRecord
+
+  module Associations
+
+    class BelongsToAssociation < SingularAssociation #:nodoc:
+
+      # this is will need to changed when upgrading to Rails 4.1 - the key thing
+      # is not to call update_counters(record)
+      def replace(record)
+        raise_on_type_mismatch!(record) if record
+
+        replace_keys(record)
+        set_inverse_instance(record)
+
+        @updated = true if record
+
+        self.target = record
+      end
+
+      def decrement_previous_counters # :nodoc:
+        with_cache_name { |name| decrement_previous_counter name }
+      end
+
+      # To be overrided by subclasses
+      def changed?
+        owner.attribute_changed?(reflection.foreign_key)
+      end
+
+      private
+
+      def decrement_previous_counter(counter_cache_name)
+        if scope = previous_target_scope
+          scope.decrement_counter(counter_cache_name)
+        end
+      end
+
+      def previous_target_scope
+        if klass = previous_klass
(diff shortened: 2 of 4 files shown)


config/initializers/alaveteli.rb
```diff
@@ -63,6 +63,7 @@
 require 'alaveteli_pro/post_redirect_handler'
 require 'user_stats'
 require 'belongs_to_destroy_with_fk_constraint'
+require 'patch_counter_cache'
```
garethrees: +1 on the fix – looks good.

spec/models/user_spec.rb
```diff
@@ -1087,4 +1087,25 @@
     end
   end
 
+  describe ".info_requests_count" do
```
garethrees: This passes against develop.

```sh
git fetch origin
git checkout develop
git checkout origin/patch-rails4-counter-cache-bug -- spec/models/user_spec.rb
bundle exec rspec spec/models/user_spec.rb
# …
128 examples, 0 failures
```

I think a spec that more closely follows [the failing example](https://gist.github.com/jimjh/6057340#file-bug_10865-rb-L77) would be better.

lizconlan: ah, that was more a sanity check that the underlying functionality didn't get broken by the patch

spec/models/alaveteli_pro/embargo_extension_spec.rb
```diff
@@ -23,7 +23,7 @@
   end
 
   it "requires an embargo" do
-    embargo_extension.embargo = nil
+    embargo_extension.embargo_id = nil
```
lizconlan: I need to work out what's going on there - something in the way the patch works means that the relation isn't invalidated by assigning nil as the object, it needs the database lookup to break, it must have introduced an extra lookup

app/models/info_request.rb
```diff
@@ -1453,6 +1453,8 @@ def self.find_in_state(state)
   end
 
   def move_to_public_body(destination_public_body, opts = {})
+    return unless destination_public_body
+
```
lizconlan: oddly, the expected "fail" condition is `nil` - there are tests that deliberately pass `nil` as the first argument

if it waits to return, the new functionality in the patch triggers an error by trying to call methods on `Nil` rather than dropping through with a null response

config/initializers/alaveteli.rb
lizconlan: :cake:

spec/models/user_spec.rb
lizconlan: (could replace it with something nicer)

lizconlan:
The "conflict" is because I merged in the security patch fix which also adds a `require` line to `config/initializers/alaveteli.rb`

spec/models/alaveteli_pro/embargo_extension_spec.rb
garethrees: Yes, we should have a good understanding of what's happening here before this gets merged.

The [documentation](http://guides.rubyonrails.org/v4.0.13/active_record_validations.html#presence) says that…

> If you want to be sure that an association is present, you'll need to test whether the associated object itself is present, and not the foreign key used to map the association.

That is [not what `EmbargoExtension` is doing](https://github.com/mysociety/alaveteli/blob/develop/app/models/alaveteli_pro/embargo_extension.rb#L14), so I expect that has an effect on this.

You should also check that the behaviour is correct with other similar records that do not have an incorrect validation when the patch is applied. It looks like there are more cases of this incorrect validation use, so I expect they'll need fixing too.

lizconlan:
closed in favour or #3822

PR #3812, opened 2017-03-07
[#2968] Include security fix for CVE-2016-6316

lizconlan:
The affected code is in a different file in 4.0 to 4.2 but it's still `ActionView::Helpers::TagHelper` so it mostly works as-is. Needed to patch `ERB::Util` as well to include the `unwrapped_html_escape` method and `HTML_ESCAPE_REGEXP`

Connects to #2968


garethrees:
Looks good. I've made an improvement to this in https://github.com/mysociety/alaveteli/pull/3815 by pulling the specs in.

Issue #3809, opened 2017-03-06
[#2968] Rails 4 Performance


garethrees:
I've created a separate ticket to contain discussion around Rails 4 (#2968) performance.

Firstly, some info for reproduction.

```sh
#
# Rails 3
#
git checkout master
bundle

# Fresh database
RAILS_ENV=production be rake db:drop && RAILS_ENV=production be rake db:create && RAILS_ENV=production be rake db:migrate && RAILS_ENV=production script/load-sample-data

# Reindex
RAILS_ENV=production script/rebuild-xapian-index

# Clear assets
rm -rf tmp/cache/assets/* && rm -rf public/assets/*

# Generate assets
RAILS_ENV=production be rake assets:clean && RAILS_ENV=production be rake assets:precompile

# Start server
RAILS_ENV=production bundle exec rails s

#
# Rails 4
#
git checkout develop
bundle

# Fresh database
RAILS_ENV=production be rake db:drop && RAILS_ENV=production be rake db:create && RAILS_ENV=production be rake db:migrate && RAILS_ENV=production script/load-sample-data

# Reindex
RAILS_ENV=production script/rebuild-xapian-index

# Clear assets
rm -rf tmp/cache/assets/* && rm -rf public/assets/*

# Generate assets
RAILS_ENV=production be rake assets:clean && RAILS_ENV=production be rake assets:precompile

# Link static assets
RAILS_ENV=production be rake assets:link_non_digest

# Allow serving of static assets by Rails server as we don't have apache configured
cat << EOF | git apply -
diff --git a/config/environments/production.rb b/config/environments/production.rb
index d6094b1..bc9aecf 100644
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -21,7 +21,7 @@ Alaveteli::Application.configure do
   # config.action_dispatch.rack_cache = true

   # Disable Rails's static asset server (Apache or nginx will already do this).
-  config.serve_static_assets = false
+  config.serve_static_assets = true

   # Compress JavaScripts and CSS.
   config.assets.js_compressor = :uglifier
EOF


# Start server
RAILS_ENV=production bundle exec rails s
```

Turn off HTTP Cache (example shows Firefox):

![screen shot 2017-03-06 at 12 49 57](https://cloud.githubusercontent.com/assets/282788/23610924/9d747fb2-026c-11e7-8409-7e8b8a4a25c1.png)

Get the page:

[...]

PR #3818, opened 2017-03-08
[#2968] Allow assets.debug config to be controlled by ENV

garethrees:
Related to #2968 and #3809.

Assets are served as separate files in the order they are specified in the
manifest file when `assets.debug` is `true`.

This commit sets the default to `false` for better performance.

To enable, simply restart the Rails server with `ASSETS_DEBUG=1`.

See more at
http://guides.rubyonrails.org/v4.0/asset_pipeline.html#turning-debugging-off


lizconlan:
@stevenday the other example that springs to mind is optionally turning coverage back on for a spec run - https://github.com/mysociety/alaveteli/blob/release/0.27.1.0/spec/spec_helper.rb#L8

PR #3822, opened 2017-03-13
Patch rails4 counter cache bug gr

garethrees:


app/models/info_request.rb
```diff
@@ -1475,14 +1475,20 @@ def move_to_public_body(destination_public_body, opts = {})
 
       public_body
     end
+
+    # HACK: Manually reset counter caches
+    # https://github.com/rails/rails/issues/10865
+    old_body.class.reset_counters(old_body.id, :info_requests)
+    public_body.class.reset_counters(public_body.id, :info_requests)
```
lizconlan: there are multiple counters for `PublicBody`

```
#  info_requests_successful_count         :integer
#  info_requests_not_held_count           :integer
#  info_requests_overdue_count            :integer
#  info_requests_visible_classified_count :integer
#  info_requests_visible_count            :integer
```

...and it looks like those need to be updated separately as this fails 😢 

```rb
      it "decrements the old authority's info_requests_visible_count" do
        request = FactoryGirl.create(:info_request)
        old_body = request.public_body
        new_body = FactoryGirl.create(:public_body)
        user = FactoryGirl.create(:user)

        expect { request.move_to_public_body(new_body, :editor => user) }.
          to change { old_body.reload.info_requests_visible_count }.from(1).to(0)
      end
```

PR #3824, opened 2017-03-14
Allow NFS to be used in vagrant if you specify an env variable

stevenday:
`export ALAVETELI_VAGRANT_USE_NFS=true` will turn on vagrant's NFS support for
shared folders, which I find a noticeable speedup over the original settings.


Vagrantfile
```diff
@@ -88,7 +89,11 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
...
 
-  config.vm.synced_folder ".", "/home/vagrant/alaveteli", :owner => "vagrant", :group => "vagrant"
+  if ALAVETELI_USE_NFS
+    config.vm.synced_folder ".", "/home/vagrant/alaveteli", nfs: true
+  else
+    config.vm.synced_folder ".", "/home/vagrant/alaveteli", :owner => "vagrant", :group => "vagrant"
+  end
 
   if File.directory?(ALAVETELI_THEMES_DIR)
     config.vm.synced_folder ALAVETELI_THEMES_DIR,
```
garethrees: Probably want to use NFS for the theme dir share too?

stevenday: Oh I missed that, yep, will do

stevenday:
@garethrees - I've brought this back up to date and squashed a fixup from your comments last time. Assuming travis is happy, do you want to give it another quick look before we merge?

PR #3832, opened 2017-03-15
Set the more info link on the new internal review page using a config param

lizconlan:
Connects to mysociety/belgium-theme#42


lizconlan:
Withdrawing in favour of a non-hotfix version

PR #3838, opened 2017-03-17
Hotfix/0.28.0.2

crowbot:


db/migrate/20170316170248_edit_info_request_batch_index.rb
```diff
@@ -0,0 +1,7 @@
+# -*- encoding : utf-8 -*-
+class EditInfoRequestBatchIndex < ActiveRecord::Migration
+  def change
+    remove_index :info_request_batches, [:user_id, :body, :title]
```
garethrees: When I ran `rake db:rollback` I got an error:

```
$ be rake db:rollback
== 20170316170248 EditInfoRequestBatchIndex: reverting ========================
rake aborted!
StandardError: An error has occurred, this and all later migrations canceled:

remove_index is only reversible if given a :column option.
```

PR #3840, opened 2017-03-17
[Pro] Issues/alaveteli pro 217/mail from pop

crowbot:
# To do
- [x] To test in vagrant (on wheezy):
* set `PRODUCTION_MAILER_RETRIEVAL_METHOD` to `pop` in `config/general.yml-example`. 
* Run `vagrant up`. Once the box is created, in a rails console run
`AlaveteliFeatures.backend.enable(:accept_mail_from_anywhere)`.
* Pipe mail to the `alaveteli-incoming` mailbox, which should be polled and imported into the app - e.g. `mail -s 'Test incoming polling' alaveteli-incoming < spec/fixtures/files/incoming-request-plain.email`
- [ ] Roll out in the stages described [below](https://github.com/mysociety/alaveteli/pull/3840#issuecomment-313149663)
- [ ] Will need documentation for new params, and changes to manual install process https://github.com/mysociety/alaveteli-professional/issues/314
- [ ] Will need to keep an eye on memory bloat as this process is going to stick around, unlike the proceses in passive mail handling.

# Future work
Allow secure setup of dovecot when install script run in production mode, or add proper instructions for setting up Alaveteli to work with gmail and test that this works.


stevenday:
This looks good to me! One thought I had earlier about testing this is whether we could do something similar to [what github do](https://github.com/github/scientist) with changing large features like this - run the two sets of code side-by-side for a while? I guess that depends on whether it's possible to put the mail in a mailbox *and* pipe it into the existing script, and on how much of an impact the actual mail receipt is likely to be the cause of the issue vs the new bits.

Regarding switching over - presumably right now when we deploy a site, mail handling is stopped and then restarted, so deploying this for real would just stop that and then never restart it? I guess we'd have to actually `vhost stop` the site, remove the config for mail handling, and then deploy it again? Maybe I'm missing something about why that's harder than it sounds though - would the mails received in the meantime end up somewhere inaccessible? Maybe we'd have to change exim's config first, so that they pile up in the new mailbox before the switchover...

stevenday:
> Or were you thinking about running both methods for all emails and handling the duplication somehow in Alaveteli?

Yeah that was more my line of thinking - we could wrap the actual `RequestMailer.receive(raw_email)` call in some kind of feature flag and then we're testing everything except that line against all of the possible emails the site would see.

stevenday:
> In terms of the to-do list above, I was thinking more about documenting how to activate the mail polling daemon for reusers.

Ah right ok, sorry, I misunderstood.

db/migrate/20170414140927_create_incoming_message_error.rb
```diff
@@ -0,0 +1,10 @@
+# -*- encoding : utf-8 -*-
+class CreateIncomingMessageError < ActiveRecord::Migration
+  def change
+    create_table :incoming_message_errors do |t|
+      t.timestamps null: false
+      t.text :unique_id, null: false
+      t.datetime :retry_at
```
garethrees: I'd be inclined to store the exception message/stack trace here to give a bit of context around what the error was. I'd always have to look up the difference between a `retry_at` existing and not existing, so having the error alongside would be a good sanity check. Since we also rescue `StandardError`, something unexpected might end up here, so the extra context would be good in that case.

db/migrate/20170414140927_create_incoming_message_error.rb
```diff
@@ -0,0 +1,10 @@
+# -*- encoding : utf-8 -*-
+class CreateIncomingMessageError < ActiveRecord::Migration
+  def change
+    create_table :incoming_message_errors do |t|
+      t.timestamps null: false
+      t.text :unique_id, null: false
```
garethrees: Not `string`?

lib/alaveteli_mail_poller.rb
```diff
@@ -0,0 +1,102 @@
...
+      unique_id = popmail.unique_id
+      if retrieve?(unique_id)
+        raw_email = popmail.pop
+        RequestMailer.receive(raw_email)
+        received = true
+        popmail.delete
+      end
+    rescue Net::POPError, StandardError => error
+      if send_exception_notifications?
+        ExceptionNotifier.notify_exception(error, :data => { :mail => raw_email })
```
garethrees: Does `raw_email` contain the `unique_id`? If not, we should also send it so that we can easily match exception notifications to the `IncomingMessageError` records.

lib/alaveteli_mail_poller.rb
```diff
@@ -0,0 +1,102 @@
...
+    end
+    found_mail
+  end
+
+  def self.poll_for_incoming_loop
+    # Run poll_for_incoming in an endless loop, sleeping when there is
+    # nothing to do
+    while true
+      sleep_seconds = 1
+      while !poll_for_incoming
```
garethrees: Need to call `new` here for a poller instance.

db/migrate/20170414140927_create_incoming_message_error.rb
crowbot: I was hedging against unknown length, but I see that that's actually defined in the [RFC](https://www.ietf.org/rfc/rfc1939.txt) and fairly short, so switched in 571f1c88717e689760ba62a5212b1e99b2b0c13e

lib/alaveteli_mail_poller.rb
crowbot: Good point. Fixed in  346fc7f.

stevenday:
> Turn on accept_mail_from_poller universally (can we easily do this so that it will be on for new users?) and remove the domain email config that sends messages to the mailin script.

Just a quick comment on this to say yep, calling `AlaveteliFeatures.backend.enable(:accept_mail_from_poller)` will set it universally, so it will override any of the `enable_actor` calls we've made earlier

app/controllers/application_controller.rb
```diff
@@ -158,11 +158,6 @@ def clear_session_credentials
     session[:ttl] = nil
   end
 
-  def send_exception_notifications?
-    !AlaveteliConfiguration.exception_notifications_from.blank? &&
```
garethrees: There's a lot going on in this commit – the commit message should explain how this works at a high level.

app/mailers/request_mailer.rb
```diff
@@ -229,10 +229,12 @@ def comment_on_alert_plural(info_request, count, earliest_unalerted_comment)
   # actual original mail sent by the authority in the admin interface (so
   # can check that attachment decoding failures are problems in the message,
   # not in our code). ]
-  def self.receive(raw_email)
```
garethrees: The commit message should describe the outline of what's happening in this commit.

* What do the 4 sources do?
* How do the feature flags affect what happens?

lib/alaveteli_mail_poller.rb
```diff
@@ -35,9 +35,11 @@ def poll_for_incoming
   def self.poll_for_incoming_loop
     if AlaveteliConfiguration.production_mailer_retriever_method == 'pop'
       poller = new
+      Rails.logger.info "Starting #{ poller } polling loop"
```
garethrees: These do log to `log/development.log`.

I guess on reflection that spreading the log messages over the app log and the poller log isn't ideal, but by using `puts` you have no control over the log level. You can set up a new logger for the class that logs to `stdout`, which might be worth doing:

```ruby
  logger = ActiveSupport::Logger.new(STDOUT)
  logger.formatter = Rails.application.config.log_formatter
  logger = ActiveSupport::TaggedLogging.new(logger)
```

Anyway, this was just a suggested improvement as it was difficult to see what was happening in a development context. I'll make PRs for suggested commits in future.

Feel free to drop this and/or a47258099c30a76f6c12b2f7ab8f1520565d8558.

PR #3848, opened 2017-03-21
[#3645] Report comments

### LINKED ISSUE #3645: Add link "report annotation / message as spam"

garethrees:
Fixes #3645 

Supersedes https://github.com/mysociety/alaveteli/pull/3694

@lizconlan I've reworked #3694 a bit. The main problem was that some of the original commits changed unrelated code in the same commit, and then that was compounded by the fixup commits doing similar and not applying cleanly. This made it really tricky to review, so I decided to go through the code and make new commits as I went.

I've left fixups in place where I could to illustrate changes, but worth checking through everything.

The final diff was good. I've made a few minor changes along the way, so would be good to get your comments on:

- Make the actions in `ReportsController` 404 if you pass an invalid `comment_id`. This seemed to make the behaviour more predictable (rather than e.g. ignoring the comment param, but then trying to update anyway).
- Renamed `Comment#last_reported_time` to `Comment#last_reported_at`. I don't really know why the `_at` convention wasn't used, and why there was a conversion to a `DateTime` when we never do that anywhere else.
- Tweaked the CSS for adding the "report" button
- Split up the filters in `ReportsController` for a bit more clarity
- Couple of minor code style tweaks
- Used "Annotation" for the user-facing report reasons
- Couple of additional specs

### DIFF
--- a/spec/controllers/reports_controller_spec.rb
+++ b/spec/controllers/reports_controller_spec.rb
@@ -94,6 +94,44 @@ describe ReportsController do
       end
 
     end
+
+   context "when reporting a comment (logged in)" do
+      before do
+        session[:user_id] = user.id
+      end
+
+      let(:comment) do
+        FactoryGirl.create(:comment, :info_request => info_request,
+                                     :attention_requested => false)
+      end
+
+      it "finds the expected request" do
+        post :create, :request_id => info_request.url_title,
+                      :comment_id => comment.id,
+                      :reason => "my reason"
+
+        expect(assigns(:info_request)).to eq(info_request)
+      end
+
+      it "finds the expected comment" do
+        post :create, :request_id => info_request.url_title,
+                      :comment_id => comment.id,
+                      :reason => "my reason"
+
+        expect(assigns(:comment)).to eq(comment)
+      end
+
+      it "returns a 404 if the comment does not belong to the request" do
+        new_comment = FactoryGirl.create(:comment)
+        expect {
+          post :create, :request_id => info_request.url_title,
+                        :comment_id => new_comment.id,
+                        :reason => "my reason"
+        }.to raise_error(ActiveRecord::RecordNotFound)
+      end
+
+    end
+
   end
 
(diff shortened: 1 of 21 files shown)


lizconlan:
> Renamed Comment#last_reported_time to Comment#last_reported_at. I don't really know why the _at convention wasn't used, and why there was a conversion to a DateTime when we never do that anywhere else.

That would be adapted from the style used in `InfoRequest` when returning a calculated value based on a logged event:

https://github.com/mysociety/alaveteli/blob/0.28.0.0/app/models/info_request.rb#L1511-L1518

as opposed to the _at convention for automatic timestamps

garethrees:
Ah! Wow, has been a while since I've seen [that code](https://github.com/mysociety/alaveteli/commit/44be6f9014728c83adbc582baaba29c231365643).

That's slightly different, in that the attribute is only set after [an unconventional `find`](https://github.com/mysociety/alaveteli/blob/0.28.0.0/app/models/info_request.rb#L1449-L1453). The naming was used to match the existing attribute. I think I should have called it `last_event_created_at` as looking back at at it, `last_event_time` wasn't getting called directly.

In any case, the call to `to_datetime` was redundant here. It was added to 44be6f9 because the value from `attributes` was returning a `String` (it actually gets cast to a `Time` in Rails 4).

Its always worth mentioning things that might cause a reviewer to question code in the commit message.

spec/controllers/reports_controller_spec.rb
```diff
@@ -358,6 +358,14 @@
         expect(response).to render_template("new")
       end
 
+      it "copies the comment id into a hidden form field" do
+        get :new, :request_id => info_request.url_title,
+                  :comment_id => comment.id
+        expect(response.body).
```
lizconlan: next commit fixes this spec by adding the missing  `render_views` so going to discard the commit message about the failure

PR #3849, opened 2017-03-22
Improve research export task

### LINKED ISSUE #3801: Improvements to the research_export task
- [x] make sure info request batch body is censored - the body property here is being imported without running through the censor lambda.
- [x] the full body name is missing from public body export, only returning short name and url

### LINKED ISSUE #3803: Research export doesn't pick up IncomingMessage content
....in approx 2/3rds of cases - especially for older requests - the `cached_main_body_text_unfolded` field is blank so we should consider making a call out to the model's `get_main_body_text_unfolded` method where this is the case (plus applying censor rules)

### LINKED ISSUE #3852: Decide on and implement policy for dealing with embargoed requests in research export
Should they be included in the export? If not, we need to remove them.

lizconlan:
* Removes the InfoRequest body from the report as it can be retrieved via the matching InfoRequest content (by matching its `id` field to  `info_request_batch_id` in the InfoRequest data) instead (and there's currently no means of applying censor rules - that I can see - to the body data)
* Applies the cutoff date to the InfoRequest batch table
* Includes authority names
* Calls the correct methods on IncomingMessage to retrieve the body content
* Rough refactor to use a lib file (more improvements possible but out of scope here)
* Tested(:tada:) exclusion of embargoed requests

Fixes #3801
Fixes #3803 
Fixes #3852

### DIFF
--- a/spec/lib/data_export_spec.rb
+++ b/spec/lib/data_export_spec.rb
@@ -1,0 +1,153 @@
+# -*- encoding : utf-8 -*-
+require File.expand_path(File.dirname(__FILE__) + '/../spec_helper')
+require File.expand_path(File.dirname(__FILE__) + "./../../lib/data_export.rb")
+
+describe DataExport do
+
+  describe ".exportable_requests" do
+
+    let(:cut_off) { Date.today + 1 }
+
+    it "includes eligible requests " do
+      request = FactoryGirl.create(:info_request)
+      exportable = described_class.exportable_requests(cut_off)
+
+      expect(exportable).to include(request)
+    end
+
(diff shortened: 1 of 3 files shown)


spec/lib/data_export_spec.rb
```diff
@@ -0,0 +1,155 @@
+# -*- encoding : utf-8 -*-
+require File.expand_path(File.dirname(__FILE__) + '/../spec_helper')
+load "lib/data_export.rb"
+
+describe DataExport do
+
+  let(:described_class) { DataExport }
```
lizconlan: (good point, deleting that)

PR #3859, opened 2017-03-24
[Pro] First attempt at guidance text.

crowbot:


stevenday:
Ah whoops, :+1:ed too soon, I was going to ask, should these be translated, or put into the wdtk-theme instead?

app/views/alaveteli_pro/info_requests/_new_request_advice.html.erb
```diff
@@ -0,0 +1,22 @@
+<p>
+  <strong>
+    <%= _("Are you sure the information you’re after isn’t available " \
+       "elsewhere?") %>
+  </strong>
+  <%= _("Check the authority’s responses on {{site_name}} or conduct " \
+        "a quick Internet search.", site_name: site_name) %>
+<p>
```
stevenday: Looks like we're missing a few paragraph closing tags here too

PR #3861, opened 2017-03-24
[#3752] Distinguish domains we're pretty sure are spam from mere suspicious ones

### LINKED ISSUE #3752: Add new spam scorer domains
Since banning over 1000 users from WDTK yesterday, about 20 more from the same domains have appeared so we should consider either banning new signups outright or at least weighting them as part of the spam scorer process.

(We also have a new signup from a webgarden.com address having decided to purge all of the current signups from there, a less pressing problem but probably worth noting.)

<details><summary>Affected domains</summary><p>
allemaling.com  <br />
brmailing.com  <br />
checknowmail.com  <br />
colde-mail.com  <br />
consimail.com  <br />
continumail.com  <br />
contumail.com  <br />
emailber.com  <br />
grow-mail.com  <br />
inemaling.com  <br />
inmailing.com  <br />
itemailing.com  <br />
itmailing.com  <br />
kod-emailing.com  <br />
kod-maling.com  <br />
kodemailing.com  <br />
kodmailing.com  <br />
left-mail.com  <br />
mabermail.com   <br />
mailphar.com  <br />
out-email.com  <br />
semi-mile.com  <br />
sin-mailing.com  <br />
sinemailing.com  <br />
sinmailing.com  <br />
takmailing.com  <br />
themailemail.com  <br />
visitinbox.com  <br />
wowmailing.com  <br />
</p></details>

lizconlan:
...and prevent new user signups from spam domains.

Closes #3752

### DIFF
--- a/app/controllers/user_controller.rb
+++ b/app/controllers/user_controller.rb
@@ -185,6 +185,25 @@ class UserController < ApplicationController
           end
         end
 
+        # Prevent signups from spam domains
+        if UserSpamScorer.new.email_from_spam_domain?(@user_signup)
+          if send_exception_notifications?
+            msg = "Attempted signup from spam domain email: " \
+                  "#{ @user_signup.email }"
+            e = Exception.new(msg)
+            ExceptionNotifier.notify_exception(e, :env => request.env)
+          end
+
+          if AlaveteliConfiguration.enable_anti_spam
+            flash.now[:error] =
+              _("Sorry, we're currently unable to sign up new users, " \
+                "please try again later")
+            error = true
+            render :action => 'sign'
+            return
+          end
+        end
+
         @user_signup.email_confirmed = false
         @user_signup.save!
         send_confirmation_mail @user_signup
--- a/lib/user_spam_scorer.rb
+++ b/lib/user_spam_scorer.rb
@@ -5,7 +5,7 @@ class UserSpamScorer
     :name_is_one_word? => 1,
     :name_includes_non_alpha_characters? => 3,
     :name_is_garbled? => 5,
-    :email_from_spam_domain? => 5,
+    :email_from_suspicious_domain? => 5,
     :email_from_spam_tld? => 3,
     :about_me_includes_currency_symbol? => 2,
     :about_me_is_link_only? => 3,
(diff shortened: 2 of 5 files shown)


lib/user_spam_scorer.rb
```diff
@@ -6,6 +6,7 @@ class UserSpamScorer
     :name_includes_non_alpha_characters? => 3,
     :name_is_garbled? => 5,
     :email_from_suspicious_domain? => 5,
+    :email_from_spam_domain? => 8,
```
lizconlan: 8 seemed like a good number but open to suggestions

app/controllers/user_controller.rb
```diff
@@ -185,6 +185,25 @@ def signup
...
+          if send_exception_notifications?
+            msg = "Attempted signup from spam domain email: " \
+                  " #{ @user_signup.email }"
+            e = Exception.new(msg)
+            ExceptionNotifier.notify_exception(e, :env => request.env)
+          end
+
+          if AlaveteliConfiguration.enable_anti_spam
+            flash.now[:error] =
+              _("Sorry, we're currently unable to sign up new users, " \
```
lizconlan: I'm not convinced this is the correct text - should we be telling people "we don't accept email from **webexample.co.uk**" or just the same generic text as the IP block?

lizconlan: Suggestions welcome

app/controllers/user_controller.rb
```diff
@@ -185,6 +185,25 @@ def signup
           end
         end
 
+        # Prevent signups from spam domains
+        if UserSpamScorer.new.email_from_spam_domain?(@user_signup)
+          if send_exception_notifications?
+            msg = "Attempted signup from spam domain email: " \
+                  " #{ @user_signup.email }"
```
garethrees: Don't need the space at the start of this line given the space at the end of the previous line.

app/controllers/user_controller.rb
garethrees: I think we should keep it as-is for now. We haven't got anti-spam on for most sites, so no one will really see this unless we're dealing with an influx.

lizconlan:
@garethrees good spot on the typo, thanks!

PR #3864, opened 2017-03-27
Extract hashes of translation strings into included modules

lizconlan:
Otherwise the default locale translations are always returned instead of the translations for the requested locale. 

Example:

```rb
I18n.with_locale(:fr) { _("Freedom of Information Act") }
=> "Loi pour la liberté d'information"

I18n.with_locale(:fr) { InfoRequest.first.law_used_human(:act) }
=> "Freedom of Information Act"
```

Connects to mysociety/belgium-theme#45 (will need a hotfix for version 0.28, and any theme that overrides `LAW_USED_READABLE_DATA` will need to be updated - may only affect AskTheEU, currently on 0.25 but worth noting in the Upgrade Notes)

### DIFF
--- a/app/models/info_request.rb
+++ b/app/models/info_request.rb
@@ -38,18 +38,6 @@ class InfoRequest < ActiveRecord::Base
   include AdminColumn
   include Rails.application.routes.url_helpers
 
-  # Two sorts of laws for requests, FOI or EIR
-  LAW_USED_READABLE_DATA =
-    { :foi => { :short => _('FOI'),
-                :full => _('Freedom of Information'),
-                :with_a => _('A Freedom of Information request'),
-                :act => _('Freedom of Information Act') },
-      :eir => { :short => _('EIR'),
-                :full => _('Environmental Information Regulations'),
-                :with_a => _('An Environmental Information request'),
-                :act => _('Environmental Information Regulations') }
-    }
-
   @non_admin_columns = %w(title url_title)
 
   strip_attributes :allow_empty => true
@@ -1646,7 +1634,7 @@ class InfoRequest < ActiveRecord::Base
 
   def applicable_law
     begin
-      LAW_USED_READABLE_DATA.fetch(law_used.to_sym)
+      TranslatedConstants.law_used_readable_data.fetch(law_used.to_sym)
     rescue KeyError
       raise "Unknown law used '#{law_used}'"
     end
--- a/app/helpers/alaveteli_pro/info_requests_helper.rb
+++ b/app/helpers/alaveteli_pro/info_requests_helper.rb
@@ -2,11 +2,13 @@
 module AlaveteliPro::InfoRequestsHelper
   def publish_at_options
     options = { _("Publish immediately") => '' }
-    options.merge(AlaveteliPro::Embargo::DURATION_LABELS.invert)
+    options.
+      merge(AlaveteliPro::Embargo::TranslatedConstants.duration_labels.invert)
   end
 
   def embargo_extension_options(embargo)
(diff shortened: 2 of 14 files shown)


app/models/info_request.rb
```diff
@@ -39,7 +39,7 @@ class InfoRequest < ActiveRecord::Base
   include Rails.application.routes.url_helpers
 
   # Two sorts of laws for requests, FOI or EIR
-  LAW_USED_READABLE_DATA =
+  def law_used_readable_data
```
lizconlan: should this be a private method? I think we can still override it in themes if it is

garethrees: Yeah, +1 for private.

spec/models/info_request_spec.rb
```diff
@@ -1209,6 +1209,15 @@
           with_message( "Unknown key 'random' for '#{info_request.law_used}'")
       end
 
+      context 'using the non default language' do
```
garethrees: I actually don't think we should test for this.

I'd _expect_ this to just work. I wouldn't write a spec each time we added a translated string, so I don't think we should here.

A better thing to do would be to add a custom hound check to assert that we haven't assigned translated values to constants.

garethrees:
> will need a hotfix for versions 0.26 to 0.28

We only really [bugfix against the current version](http://alaveteli.org/docs/running/security/#bug-fixes). It would be better to just upgrade sites that _really_ want this fix.

garethrees:
> I think we should make one commit that fixes this problem across the whole app.

Here's a [quick suggestion](https://gist.github.com/anonymous/1ad5dc5bf28b87d143e5538972bddea7). I'm not set on this, but I can't think of an elegant way to define these before the macro methods that doesn't look like there's a random class method [dumped](https://github.com/bbatsov/rails-style-guide#macro-style-methods) at the top of the file.

spec/models/info_request_spec.rb
lizconlan: good point, although AFAIK this uses the translation file in spec/fixtures/locale so it shouldn't be affected by Transifex changes (shouldn't)

lizconlan:
>> will need a hotfix for versions 0.26 to 0.28
>
> We only really bugfix against the current version. It would be better to just upgrade sites that really want this fix.

Currently causing a problem for Belgium (0.27) so maybe a 0.27 hotfix to avoid retranslating so soon. Will have a proper look next week

spec/models/info_request_spec.rb
lizconlan: confirmed - the spec uses a fixture file for translations

lizconlan:
PR against develop and a hotfix for 0.27 to avoid making Belgium redo their translations?

lizconlan:
@garethrees does fdde543 look ok?

lizconlan:
hotfix for 0.28

lizconlan:
Happy to squash some or all of those post review (although I'd be inclined to keep the Pro one separate)

app/models/mail_server_log/translated_constants.rb
```diff
@@ -0,0 +1,16 @@
+# -*- encoding : utf-8 -*-
+class MailServerLog::DeliveryStatus
```
garethrees: The path to this file should be `app/models/mail_server_log/delivery_status/translated_constants.rb`

lizconlan:
ToDo - cherry pick 23a7e4f onto a new branch off master to create a hotfix comment (holding off in expectation of a translation hotfix which gets priority)

lizconlan:
Closed in favour of a hotfix branch #3929

PR #3865, opened 2017-03-28
Remove committed puts debugging

garethrees:
Looks like these were mistakenly committed.

Removing because they produce unnecessary spec output.


garethrees:
Merged in 4bda3dc

PR #3866, opened 2017-03-28
Unknown delivery status

garethrees:
Fixes https://github.com/mysociety/whatdotheyknow-theme/issues/373

- [x] Add `unknown.png` and `unknown@2x.png` icons


spec/models/mail_server_log/delivery_status_spec.rb
```diff
@@ -0,0 +1,148 @@
...
+
+  end
+
+  describe '#humanize' do
+
+    it 'returns a humanized string for the :delivered status' do
+      status = described_class.new(:delivered)
+      expect(status.humanize).to eq('This message has been delivered.')
+    end
+
```
lizconlan: The `:es` locale and sample translations are available in the test environment, I'd be tempted to a test for a translated string here to see if the `HUMANIZED` constant is providing non-default translations

app/models/mail_server_log/delivery_status.rb
```diff
@@ -0,0 +1,66 @@
+# -*- encoding : utf-8 -*-
+class MailServerLog::DeliveryStatus
+  include Comparable
+
+  # The order of these is important as we use the keys for sorting in #<=>
+  HUMANIZED = {
```
garethrees: I'm going to merge this as is, as it's an existing problem in the MTA-specific classes. We'll catch all of these cases in https://github.com/mysociety/alaveteli/pull/3864

Issue #1719, opened 2014-07-21
Corrupt mail file stopped update-xapian-index running


Flupsy:
The `update-xapian-index` job was failing over the weekend due to a corrupt mail file (`/data/vhost/www.whatdotheyknow.com/raw_emails/217/217302/540916`).  The associated request (https://www.whatdotheyknow.com/request/post_review_correspondence_2) was also generating an error when retrieved.

I've moved the file to `~foi/217302/540916` and replaced it with an empty one; the job and request are no longer generating errors.

garethrees:
Haven't seen this for ages, so closing.

PR #3870, opened 2017-03-29
Hacky fix for stripping null bytes from Tnef

garethrees:
We sometimes receive incoming messages from the Parliamentary and Health
Service Ombudsman that contain only a `winmail.dat` attachment. Inside
this attachment is usually an RTF.

We don’t do a good job at parsing this, as described in
https://github.com/mysociety/alaveteli/issues/1719.

Since upgrading to Rails 4, these types of incoming message don’t have
null bytes stripped. This change just ensures there are no null bytes in
the data that’s about to get saved.

Related links:

- http://stackoverflow.com/questions/28042876/rails-4-2-file-field-string-contains-null-byte-error
- https://github.com/rails/rails/pull/17680
- https://github.com/rails/rails/issues/8957


garethrees:
Merged in 9d91ac7c5ce930409af4d49172566d58f0e09dea

PR #3877, opened 2017-03-30
Send and Publish buttons

### LINKED ISSUE #3842: Change text of "Send" button to "Send and Publish"
Despite the clear notices we still have users not realising their correspondence is being published; leading in some cases to users publishing their own and others' personal information without intending to. 

Perhaps putting the word "publish" on the button they have to click when sending a request / follow-up / reply might reduce the instance of accidental inappropriate use of the service?

lizconlan:
Closes #3842


garethrees:
I think they should be "Send and **p**ublish [THING]".

Should also make this change to the confirmation email I think https://github.com/mysociety/alaveteli/blob/develop/app/controllers/request_controller.rb#L372-L379

Issue #3799, opened 2017-03-02
Upgrade to Rails 4.1


garethrees:
Continuing from https://github.com/mysociety/alaveteli/issues/2968

http://edgeguides.rubyonrails.org/upgrading_ruby_on_rails.html#upgrading-from-rails-4-0-to-rails-4-1

- [x] Update gems
  - [x] Upgrade jquery-ui-rails to 6.0.1
  - [x] Upgrade secure_headers
  - [x] Update strip_attributes (cherry pick commits on to latest compatible master)
  - [ ] ~~Upgrade gettext_i18n_rails~~
  - [ ] ~~Upgrade gettext & locale~~
  - [x] Upgrade coffee-rails to 4.2.1 (seems fine to do, but harder to test)
  - [x] Upgrade uglifier (looks easy)
  - [x] Upgrade capybara (looks easy)
  - [x] Upgrade bullet (looks easy)
  - [x] Upgrade factory_girl_rails (looks easy)
  - [x] Upgrade rspec (bigger changelog)
  - [x] Upgrade rDoc to 4.x (Do we even use it?)
- [x] Update generated templates http://railsdiff.org/4.0.13/4.1.16
- [x] 5.1 CSRF protection from remote `<script>` tags
- [x] 5.2 Add spring preloader? (_if we want Spring, we'll need to [lock it to `< 2.0.0`](https://github.com/rails/spring/blob/master/CHANGELOG.md#200)_)
- [x] 5.3 Migrate to `config/secrets.yml`
- [x] ~~5.4 Changes to test helper~~ N/A
- [x] 5.5 Use new JSON-based cookie serializer?
- [x] 5.6 Compare Flash message keys against strings (no action required - we don't iterate over keys)
- [x] 5.7.1 MultiJSON removal (I don't _think_ we rely on this) [_yeah, just Compass using this now_]
- [x] 5.7.2 JSON gem compatibility – seems like no action needed
- [x] 5.7.3 Check for deprecated JSON encoder dependencies
- [x] 5.7.4 JSON representation of Time objects now includes millisecond precision (does not appear to be harmful, leaving as-is)
- [x] 5.8 Remove usage of return within inline callback blocks
- [x] 5.9 Helper methods that are used in multiple fixtures should be defined on modules included in the newly introduced `ActiveRecord::FixtureSet.context_class`, in `test_helper.rb`.
- [x] 5.10 I18n enforcing available locales
[...]

PR #3883, opened 2017-03-31
[#3799] Update gems

garethrees:
Work on #3799 

This PR:

- Adds some notes about specifying gem versions (27b9b27) – @crowbot & @stevenday weigh in on this if you like.
- Updates version constraints in line with the above
- Adds `<` constraints where we're blocked from updating (e.g. no Ruby 1.9 support or needs a more recent Rails)
- Updates a few gems
- Doesn't touch bootstrap or foundation gems

I've added some todos on #3799 to cover gem upgrades (minor or major version bumps).


Gemfile
```diff
@@ -1,3 +1,78 @@
...
+# This will install the latest version of the 1.2 series. When 1.3 or 2.0 are
+# released, we can assume there may be a breaking change.
+#
+# If we need to be more specific than this, apply one of the following
+# guidelines. Use commit messages to describe why the extra specificity is
+# required. Links to changelogs are very appreciated.
+#
+# If we need greater than a specific PATCH level, specify that:
+#
+#    gem 'foo', '~> 1.2.7'
```
garethrees: I'm wondering whether this should be

```ruby
gem 'foo', '~> 1.2.0', '>= 1.2.7'
```

Its the same, but it does make it more explicit that we really do need `1.2.7+`

stevenday: Yep, I always have to look up what `~>` actually means

Gemfile
```diff
@@ -70,40 +144,40 @@ gem 'mail', :git => 'https://github.com/mikel/mail', :branch => '2-6-stable'
...
-  gem 'rspec-activemodel-mocks', '~> 1.0.1'
+  gem 'rspec-activemodel-mocks', '~> 1.0.0'
   gem 'rspec-rails', '~> 3.4.0'
-  gem 'pry', '~> 0.10.4'
-  gem 'pry-debugger', '~> 0.2.3', :platforms => :ruby_19
-    gem 'public_suffix', '< 1.5.0'
+  gem 'pry', '~> 0.10.0'
+  gem 'pry-debugger', '~> 0.2.0', :platforms => :ruby_19
+    gem 'public_suffix', '~> 1.4.0', '< 1.5.0'
 end
```
lizconlan: I think we can finally add this now:

`gem 'pry-byebug', '~> 3.4.0', :platforms => [:ruby_20, :ruby_21]`

\o/

Gemfile
```diff
@@ -169,7 +169,7 @@ group :test, :development do
   gem 'rspec-rails', '~> 3.4.0'
   gem 'pry', '~> 0.10.0'
   gem 'pry-debugger', '~> 0.2.0', :platforms => :ruby_19
-    gem 'public_suffix', '< 1.5.0'
+    gem 'public_suffix', '~> 1.4.0', '< 1.5.0'
```
lizconlan: bonus feature...  I think we can finally add this now:

`gem 'pry-byebug', '~> 3.4.0', :platforms => [:ruby_20, :ruby_21]`

\o/

garethrees: nope, bundler we rely on is too old so doesn't support the latest ruby platforms. sigh.

PR #3884, opened 2017-04-01
Corrected formatting of Preview Your Request

Keerti-Gautam:
Hi 

Request preview in the current framework of Alaveteli has the text of the class `correspondence_text` out of line and looked something like this in the website of WhatDoTheyKnow: 
screenshot
This was also visible in the current development copy of RightToKnow as currently pointed in one of their [issues](https://github.com/openaustralia/righttoknow/issues/670)

*Correction Made:*
Removed the left padding of the class and now it is in line with the subject and the title and looks like this in the theme of RightToKnow: 
screenshot

This resolves the [issue](https://github.com/openaustralia/righttoknow/issues/670) of RightToKnow of OpenAustralia Foundation and would also improve the formatting of "Preview Your Request" of WhatDoTheyKnow after merging upstream changes.

### DIFF
--- a/app/assets/stylesheets/responsive/_new_request_layout.scss
+++ b/app/assets/stylesheets/responsive/_new_request_layout.scss
@@ -162,6 +162,10 @@ span#to_public_body {
       width: 36.813em;
     }
   }
+
+  .correspondence_text {
+    padding-left: 0;
+  }
 }
 
 


lizconlan:
Hi @Keerti-Gautam, thanks for your PR!

It's been merged into our develop branch (with a credit in the [changelog](https://github.com/mysociety/alaveteli/blob/280c3620c7fc99c2b7003ad6151604a5e1db22a8/doc/CHANGES.md)) and will be included in the next release version of Alaveteli (0.29) 🏆 

I made a small change to your code by moving your new CSS rule so that it only applies to `.correspondence_text` inside the `.preview-pane` block otherwise it would have caused the body text in the request thread to hit the edge of the containing box, like this:

screen shot 2017-04-06 at 12 43 03

I also altered it slightly so that it reads `padding-left: 0` instead of `padding-left: 0em`, this has exactly the same effect but [some CSS checkers see including units with 0 as a fault](https://stylelint.io/user-guide/rules/length-zero-no-unit/) and it helps keep the file size down.

However... as RightToKnow are upgrading to version 0.26, you're going to need to make a PR against their theme project to make your improvement available to them.

Adding something like this to their `responsive/custom.scss` file should work 😉 :

```
.preview-pane {
  .correspondence_text {
    padding-left: 0;
  }
}
```

Keerti-Gautam:
Thanks for the feedback @lizconlan ! I will follow this up on right to know!

PR #3895, opened 2017-04-06
Add a new route for change requests to existing public bodies

### LINKED ISSUE #3816: /change_request/new/ loses the public body url_name during login redirect
Found while manually testing on #2968

lizconlan:
This allows the public body `url_name` to be part of the url rather than
relying on a querystring param as the param is not persisted if the
user clicks on the sign in link before filling in the form

(Saner than trying to persist the `url_name` via sessions through the login - it's a click to login rather than a forced redirect via `authenticated?` so the usual `PostRedirect` trick isn't available.) 

Fixes #3816

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -296,7 +296,11 @@ Alaveteli::Application.routes.draw do
         :via => :get
   ####
 
+  #### PublicBodyChangeRequest controller
   resource :change_request, :only => [:new, :create], :controller => 'public_body_change_requests'
+  match 'change_request/new/:body' => 'public_body_change_requests#new',
+        :as => :new_change_request_body,
+        :via => :get
 
   #### Comment controller
   match '/annotate/request/:url_title' => 'comment#new',
--- a/app/views/public_body/_more_info.html.erb
+++ b/app/views/public_body/_more_info.html.erb
@@ -14,4 +14,4 @@
 
 <%= link_to _('View FOI email address'), view_public_body_email_path(public_body.url_name) %><br>
 
-<%= link_to _("Ask us to update FOI email"), new_change_request_path(:body => public_body.url_name) %><br>
+<%= link_to _("Ask us to update FOI email"), new_change_request_body_path(:body => public_body.url_name) %><br>
--- a/doc/CHANGES.md
+++ b/doc/CHANGES.md
@@ -2,6 +2,8 @@
 
 ## Highlighted Features
 
+* Fixed bug which redirected people trying to request a change to the email of
+  a public body back to the missing body form post sign in (Liz Conlan)
 * Tweak wording of bounce reply to make it easier for admins to locate the
   related request (Gareth Rees)
 * Fix the layout of the request preview page so that the request body text is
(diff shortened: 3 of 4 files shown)


config/routes.rb
```diff
@@ -296,7 +296,11 @@
         :via => :get
   ####
 
+  #### PublicBodyChangeRequest controller
   resource :change_request, :only => [:new, :create], :controller => 'public_body_change_requests'
+  match 'change_request/new/:body' => 'public_body_change_requests#new',
+        :as => :new_change_request_body,
+        :via => :get
```
garethrees: Rather than hacking in a `match` route, I think it would be better to use the Rails conventions and create a nested resource:

```diff
diff --git a/config/routes.rb b/config/routes.rb
index 5591811..a86a2dc 100644
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -297,10 +297,11 @@ Alaveteli::Application.routes.draw do
   ####

   #### PublicBodyChangeRequest controller
-  resource :change_request, :only => [:new, :create], :controller => 'public_body_change_requests'
-  match 'change_request/new/:body' => 'public_body_change_requests#new',
-        :as => :new_change_request_body,
-        :via => :get
+  resources :public_bodies, :only => [] do
+    resources :change_requests,
+              :only => [:new, :create],
+              :controller => 'public_body_change_requests'
+  end

   #### Comment controller
   match '/annotate/request/:url_title' => 'comment#new',
```

```sh
$ be rake routes | g change_request
     public_body_change_requests POST           /public_bodies/:public_body_id/change_requests(.:format)                public_body_change_requests#create
  new_public_body_change_request GET            /public_bodies/:public_body_id/change_requests/new(.:format)            public_body_change_requests#new
```

This has a nice minor benefit of being able to drop the custom url param in the form:

```diff
diff --git a/app/views/public_body_change_requests/new.html.erb b/app/views/public_body_change_requests/new.html.erb
index 12c5879..5a89396 100644
--- a/app/views/public_body_change_requests/new.html.erb
+++ b/app/views/public_body_change_requests/new.html.erb
@@ -2,7 +2,7 @@

 <%= foi_error_messages_for :change_request %>

-<%= form_for(@change_request, :url => change_request_path)  do |f| %>
+<%= form_for(@change_request) do |f| %>
   <% if not @user %>
     <p>
       <label class="form_label" for="user_name">
```

lizconlan: That doesn't look right to me - rather than adding a single new route for `change_request/new/:body` it's done this:

```diff
-     change_request POST               /change_request(.:format)                                               public_body_change_requests#create
-     new_change_request GET            /change_request/new(.:format)                                           public_body_change_requests#new
+     public_body_change_requests POST  /public_bodies/:public_body_id/change_requests(.:format)                public_body_change_requests#create
+  new_public_body_change_request GET   /public_bodies/:public_body_id/change_requests/new(.:format)            public_body_change_requests#new

garethrees: Yes, that's right. Its made `new` and `create` nested actions of the `public_bodies` resource.

lizconlan: I need to double check but I think that would prevent  `/change_request/new` without the public body information (currently the default) for requesting that a new body is added

garethrees:
Spec looks good. +1 after squash and rebuild (looks like a transient spec failure)

PR #3898, opened 2017-04-06
test PR to see if updating bundler lets us use later ruby platform specifiers

lizconlan:


lizconlan:
withdrawn, "proves" Travis is ok needs more work to see if it breaks our machine build scripts

PR #3901, opened 2017-04-07
[#3799] upgrade dev gems

garethrees:
Work on #3799


garethrees:
No use of rdoc that I can find, so removed.

PR #3912, opened 2017-04-18
[#3799] Upgrade secure_headers gem

lizconlan:
3.1.2 -> 3.6.2

https://github.com/twitter/secureheaders/blob/master/CHANGELOG.md#362

This doesn't appear to affect our existing config but we have the opportunity to add configuration for [cookies](https://github.com/twitter/secureheaders/blob/master/CHANGELOG.md#320-cookie-settings-and-csp-hash-sources),  [referrer policy](https://github.com/twitter/secureheaders/blob/master/CHANGELOG.md#330-referrer-policy-support) (not currently implemented) and [clear-site-data]( https://github.com/twitter/secureheaders/blob/master/CHANGELOG.md#360) should we wish to do so.

Work on #3799


garethrees:
> we have the opportunity to add configuration for cookies, referrer policy (not currently implemented) and clear-site-data should we wish to do so.

We should do this as part of the upgrade, or state explicitly that we haven't enabled for whatever reason.

lizconlan:
Have added better notes with highlighted features and links to the changelog.

* It's now clearer that `referrer-policy` is experimental (which suggests it is not for us, at least not yet). 
* I've realised that the `cookie` settings were available previously and they've just been improved rather than being an entirely new feature.
* And there are notes that `clear-site-data` needs some more thought as switching it on could have consequences for users (also tbh, I'm not entirely clear what enabling it in the config would actually do - would it automatically fire on sign out or would we have to implement it in code?)

lizconlan:
Update - secure_headers is the one to worry about

lizconlan:
After re-reading the docs more thoroughly, quietly confident that this is correct

lizconlan:
Having re-read the docs more carefully and examined the current cookie settings, quietly confident that the new settings are right. Although it appears that `SameSite` has limited browser support at the moment http://caniuse.com/#feat=same-site-cookie-attribute

PR #3914, opened 2017-04-18
[#3799] Upgrade internationalisation gems

lizconlan:
#3799


lizconlan:
argh https://gist.github.com/lizconlan/ed353246994a8583d7b22a2af360a3c0

lizconlan:
Ok, this is in a reasonable state now *but* I'm not convinced that the `rake:clean` tasks are working as they should

lizconlan:
Ok, the code is in a reasonable state now **but** I'm not convinced that the rake tasks are behaving as they should (or at least as they were previously).

To fix/Of note:
* `gettext:find` is echoing internal file commands to the screen

e.g.
```
rm -f locale/zh_HK/app.edit.po
cp locale/zh_HK/app.po locale/zh_HK/app.edit.po
touch locale/zh_HK/app.po.time_stamp
```
* `gettext:clean` is
  * stripping out intentional comments (e.g. `-#. "Follow up" in this context means
[...]`)
  * leaving fuzzy translations (for bonus points, these are scattered randomly throughout the file rather than gathered at the end as before, although technically that's probably`find`'s fault)
  * not updating the header version text if it doesn't alter `tx pull`ed translations

Strings containing `\n` are now being split across multiple lines
e.g.
```diff
-msgid ",\\n\\n\\n\\nYours,\\n\\n{{user_name}}"
+msgid ""
+",\n"
+"\n"
+"\n"
+"\n"
+"Yours,\n"
+"\n"
+"{{user_name}}"
```

lizconlan:
ToDo:
- [ ] check the reformatted strings containing `\n` get translated by the app as expected
- [ ] pull out the non rake code from `gettext.rake` into a lib file and write integration tests

lib/alaveteli_gettext/task_methods.rb
```diff
@@ -0,0 +1,137 @@
...
+                          dir,
+                          Dir.glob("#{dir}/*/app.po"),
+                          "alaveteli")
+
+      Dir.glob("#{dir}/*/app.po") do |po_file|
+        GetText::Tools::MsgMerge.run("--output", po_file,
+                                     "--sort-output",
+                                     "--no-location",
+                                     "--no-wrap",
+                                     "--no-fuzzy-matching",
```
lizconlan: the first 3 were, `--no-fuzzy-matching` is new (and might be incorrect) - because it's not being called as a Rake task - it's calling the msgmerge executable - it needs to be passed in directly rather than being able to rely on the task settings

garethrees:
> but extra disk I/O which is not ideal

Yeah, `gettext:find` was pretty slow:

```
$ time be rake gettext:find

real    7m19.073s
user    7m5.687s
sys     0m3.636s
```

garethrees:
`gettext:find` currently (`develop`) prints progress, which is nice:

```sh
$ be rake gettext:find
/home/vagrant/alaveteli/locale/app.pot ........................................................................................................................................................... done.
/home/vagrant/alaveteli/locale/aln/app.po .......................................................................................................................................................... done.
/home/vagrant/alaveteli/locale/ar/app.po ........................................................................................................................................................................ done.
…
```

lib/alaveteli_gettext/task_methods.rb
```diff
@@ -0,0 +1,137 @@
...
+    def theme_locale_path(theme)
+      Rails.root.join "lib", "themes", theme, "locale-theme"
+    end
+
+    def pro_locale_path
+      Rails.root.join "locale_alaveteli_pro"
+    end
+
+    def define_gettext_task(text_domain, locale_path, files_to_translate, version=nil)
+      options = Rails.application.config.gettext_i18n_rails.msgmerge
```
garethrees: For reference, this has changed between updates:

```ruby
# develop
Rails.application.config.gettext_i18n_rails.msgmerge
# => ["--sort-output", "--no-location", "--no-wrap"]
Rails.application.config.gettext_i18n_rails
# => {:msgmerge=>["--sort-output", "--no-location", "--no-wrap"], :use_for_active_record_attributes=>true}

# 3799-upgrade-gettext
Rails.application.config.gettext_i18n_rails.msgmerge
# => nil
Rails.application.config.gettext_i18n_rails
# => {:msgmerge=>nil, :msgcat=>nil, :xgettext=>nil, :use_for_active_record_attributes=>true}
```

Shouldn't matter, given the line below though…

lib/alaveteli_gettext/task_methods.rb
```diff
@@ -0,0 +1,137 @@
...
+      ENV["VERSION_CONTROL"] = "off"
+
+      GetText::Tools::Task.define do |task|
+        task.package_name = text_domain
+        task.package_version = version || "version 0.0.1"
+        task.domain = text_domain
+        task.po_base_directory = locale_path
+        task.files = files_to_translate
+        task.msgmerge_options = options
+        task.msgcat_options = options
```
garethrees: Wonder if this should be `nil`, given:

```ruby
Rails.application.config.gettext_i18n_rails
# => {:msgmerge=>nil, :msgcat=>nil, :xgettext=>nil, :use_for_active_record_attributes=>true}
```

lizconlan: can't literally be `nil` at the point of assignment, but we can try it as an empty array (if you pass it nil, it tries to do Array stuff to it and fails)

garethrees: We'll want to check against https://github.com/grosser/gettext_i18n_rails/blob/v1.7.2/lib/gettext_i18n_rails/tasks.rb#L20

lizconlan:
I _think_ what https://github.com/ruby-gettext/gettext/issues/36 is describing is a process to `tx pull` translations as `app.edit.po` files so that `gettext:find` can compare the `app.po` file it builds with the updated translations in the `.edit.` version of the file. As our process considers updates from Transifex and rebuilding the po files from the templates separately (with the last known good version safely in git), this may not be something we need to worry about.

garethrees:
Links for reference:

- http://projects.theforeman.org/issues/8874
- https://github.com/ruby-gettext/gettext/issues/36
- https://github.com/theforeman/foreman/commit/4406f5af6a8123a93771d4bff5f3dd29769ab0e9
- https://github.com/ruby-gettext/gettext/blob/master/lib/gettext/tools/task.rb#L333

garethrees:
Closing for now

PR #3915, opened 2017-04-18
[Pro] Fixups to core from testing against IPV theme

stevenday:
A few small issues found when testing the upgraded IPV theme branch.


locale/aln/app.po
```diff
@@ -4463,3 +4463,12 @@ msgstr ""
 
 msgid "« Back to search results"
 msgstr ""
+
+#~ msgid "Overdue"
```
garethrees: These shouldn't get committed. See "You may need to manually remove obsolete msgids." on https://github.com/mysociety/alaveteli/wiki/Release-Manager's-checklist

PR #3916, opened 2017-04-18
[Pro] Exclude defunct public bodies from Pro searches.

crowbot:


app/controllers/request_controller.rb
```diff
@@ -44,14 +44,16 @@ def select_authority
...
+      @xapian_requests = perform_search_typeahead(query, :model => PublicBody)
     end
     medium_cache
   end
 
   def select_authorities
     if !params[:public_body_query].nil?
-      @search_bodies = perform_search_typeahead(params[:public_body_query], PublicBody, 1000)
+      @search_bodies = perform_search_typeahead(params[:public_body_query],
+                                                { :model => PublicBody,
```
stevenday: I think you could leave the `{` `}` brackets off the hash param here?

PR #3917, opened 2017-04-18
[Pro] Update selectize from the source at commit 1e2dba

crowbot:
The `dist` versions of files in `selectize` only get updated on a release, so this is a version produced locally from the `master` branch. This incorporates the fixes made in https://github.com/selectize/selectize.js/pull/1193 which appear to fix https://github.com/mysociety/alaveteli-professional/issues/254
@stevenday I haven't restored the edit you made to `selectize.js` in 017180fad400598ddd41cfdb7cb0f0b5e0972e99 as I wasn't sure it's still needed - is it?

### DIFF
--- a/app/assets/javascripts/alaveteli_pro/authority_select.js
+++ b/app/assets/javascripts/alaveteli_pro/authority_select.js
@@ -31,7 +31,7 @@
       // button (which is a link), so this gets only the text node from the
       // DOM, ignoring the remove button. A naive .text() would return the
       // multiplication symbol in that button too.
-      var newAuthorityName = $item.contents().filter(function(){
+      var newAuthorityName = $item.find('.item').contents().filter(function(){
         return this.nodeType === Node.TEXT_NODE;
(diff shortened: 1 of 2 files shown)


stevenday:
@crowbot - it looks like it does unfortunately. The PR we're waiting to be merged is https://github.com/selectize/selectize.js/pull/1151 which seems to be at a bit of a dead end at the moment.

stevenday:
> what interface change should I be seeing as a result of applying this? It seems to me to work the same before and after.

Hmm, the issue this was originally fixing (I think) is that the remove button was added, but didn't actually work. You're right though, it now seems to work regardless! So we can probably drop that fix after all.

There's one more change we need, to make the placeholder filling work again:

In `app/assets/javascripts/alaveteli_pro/authority_select.js` line 34

```diff
- var newAuthorityName = $item.contents().filter(function(){
+ var newAuthorityName = $item.find('.item').contents().filter(function(){
```

PR #3918, opened 2017-04-19
[Pro] Include short_name in search results for selectize.

crowbot:


stevenday:
Looks like there's a spec of the existing behaviour that needs to be changed though: spec/controllers/alaveteli_pro/public_bodies_controller_spec.rb:29

PR #3923, opened 2017-04-24
[Pro] Implement request list styling for batch requests

stevenday:
This adds all of the necessary backend and frontend work to implement the
designs in mysociety/alaveteli-professional#218. This includes:

- A new facade model `RequestSummary` that stores the pertinent details of
  requests, drafts, batches and draft batches in a single place
- Updates to the `RequestFilter` searching model so that it works with
  `RequestSummaries`
- Changes to the request list and dashboard views so that they work with
  `RequestSummaries`
- Some new scopes and model helper methods to make summarising requests and
  displaying them in the request list easier.
- New views for displaying batch requests in the request list, with a progress
  bar and expandable list of child requests.
- New scopes and model helper methods to make that batch request view possible

~~This is still a WIP, as I need to fix up some spec issues caused by these
changes and update the specs for the request list and its views, but I figured
it was worth getting some early review on the setup whilst I did that.~~

Now ready for a full review

I'd also appreciate a look over the css for the progress bar in particular
from @wrightmartin or @zarino to see if I'm doing anything crazy with flexbox.

screen shot 2017-04-24 at 09 01 15

screen shot 2017-04-24 at 09 01 26

screen shot 2017-04-24 at 09 02 08

For #257


app/views/alaveteli_pro/info_request_batches/_info_request_batch.html.erb
```diff
@@ -48,7 +48,9 @@
              locals: { phases: info_request_batch.request_phases_summary } %>
 
   <% unless info_request_batch.info_requests.empty? %>
-    <div class="batch-request__requests">
+    <label class="houdini-label batch-request__show-requests" for="input1"><%= _("Show requests") %></label>
+    <input class="houdini-input" type="checkbox" id="input1">
```
stevenday: Just realised I'm going to need to make sure this id is unique because there could be several on the page at once.

wrightmartin:
I haven't managed to get it running locally, however in the spirit of not holding things up the flexbox css is just fine.

Design-wise it works fine but i'll jump in and polish it up at some point. However, I'm seeing loads of design polish needed across all the new pro screens (just normal tidy-up stuff same as any extended project) so I'll save it for when I (or indeed Mr @zarino) do it all.

app/views/alaveteli_pro/info_request_batches/_progress_bar.html.erb
```diff
@@ -0,0 +1,12 @@
+<div class="batch-summary">
+  <% phases.each do |label, count| %>
+    <% if count > 0 %>
+      <div class="batch-summary__phase
+                  batch-summary__phase--<%= label.parameterize %>"
+           style="flex-grow: <%= count %>">
```
zarino: I wonder whether we need the vendor-prefixed versions here too? Quick google suggests `-webkit-flex-grow` for Safari (desktop and mobile) 2014 and earlier, and `-ms-flex-positive` (!!) for IE10.

stevenday: Thanks - I'll add them in

stevenday:
> OOI, why use flexbox, rather than just floating the progress bar segments left and sizing with percentage widths? (As gender-balance.org does it)

My main worry was keeping the segments big enough to display the label text at all times (or keeping it lined up properly if we let the segments get smaller than the text) - if I did it with percentages, I couldn't see how I'd do that, so the flex-grow thing seemed like a nice way to achieve the compromise I wanted without having to bring any complicated calculations into the code.

stevenday:
@crowbot - I think this is ready for a review now.

Note that there are some migrations that'll need running, and some new seeds for the db.

You can add a batch request by starting at 10.10.10.30:3000/alaveteli_pro/batch_request_authority_searches/new (there's still no link directly to it)

app/models/concerns/alaveteli_pro/request_summaries.rb
```diff
@@ -0,0 +1,29 @@
...
+      has_one :request_summary, :as => :summarisable,
+                                :class_name => "AlaveteliPro::RequestSummary",
+                                :dependent => :destroy
+      after_save :create_or_update_request_summary
+    end
+
+    # Creates a RequestSummary item for this model on first save, or updates
+    # the existing one otherwise.
+    def create_or_update_request_summary
+      if self.should_summarise?
```
crowbot: If a batch request is created, and then the status of one of the requests in the batch is changed (for example, to 'successful'), I don't think that the `RequestSummary` for the batch is getting updated. You can demonstrate this by making a batch, sending it, and then setting one of the requests to successful. The batch should now appear in the request list if you select 'complete' from the LHS nav, but it doesn't.

app/views/alaveteli_pro/draft_info_request_batches/_draft_info_request_batch.html.erb
```diff
@@ -0,0 +1,39 @@
+<div id="draft-info-request-batch-<%= draft_info_request_batch.id %>"
+     class="request">
+  <h3 class="request__title">
+    <% if draft_info_request_batch.embargo_duration %>
+      <span class="embargo-indicator embargo-indicator--small">Private</span>
+    <% end %>
+    <span><%= _("Batch request") %></span>
```
crowbot: If the draft batch request doesn't have a title yet, you can't currently access it - maybe the text "Batch request" should be linked in that case?

spec/controllers/alaveteli_pro/info_requests_controller_spec.rb
```diff
@@ -28,17 +28,17 @@
 
       it "assigns the user's info requests" do
```
stevenday: Fixed in 81fae88ed9d9f42d86cdd2cddb437e95666aa415

spec/views/alaveteli_pro/dashboard/_projects.html.erb_spec.rb
```diff
@@ -0,0 +1,107 @@
...
+
+    FactoryGirl.create(:draft_info_request, user: pro_user)
+
+    public_bodies = FactoryGirl.create_list(:public_body, 10)
+    FactoryGirl.create(:info_request_batch, user: pro_user,
+                                            public_bodies: public_bodies)
+    FactoryGirl.create(:draft_info_request_batch, user: pro_user,
+                                                  public_bodies: public_bodies)
+
+    expect(pro_user.request_summaries.category(:response_received).count).to eq 1
```
stevenday: Whoops - a bit of debugging I forgot to remove. Fixed in b0db9aec123086c40cac311cc1cfcb200e9ad2f6

app/models/info_request/state/action_needed_query.rb
```diff
@@ -0,0 +1,26 @@
...
+    class ActionNeededQuery
+      def initialize(relation = InfoRequest)
+        @relation = relation
+      end
+
+      def call
+        # A combination of response_received, clarification_needed, overdue
+        # and very_overdue
+        query = <<-SQL
+awaiting_description = true
```
stevenday: Fixed in 9eae0ab81c2837a7c24b6cf7c4ad1369c0c8f929

spec/models/info_request/state/calculator_spec.rb
```diff
@@ -253,4 +253,53 @@
     end
   end
 
+  describe '#summarised_phase' do
+    # A summarised version of #phase, grouping the phases down into 3 groups
+    def summarised_phase(cached_value_ok=false)
```
stevenday: An aide-memoire I left in by mistake, removed in b14ec56fb9c603cdd6e7f65c770d1b7c09ec9129

lib/tasks/request_summaries.rake
```diff
@@ -0,0 +1,21 @@
+# -*- encoding : utf-8 -*-
+namespace :request_summaries do
+
+  desc "Generate request summaries for every user"
```
stevenday: Fixed in e25df8589ee9340ba3895870fe9d625ba82bf0eb

lib/tasks/request_summaries.rake
```diff
@@ -0,0 +1,21 @@
...
+  task :generate => :environment do
+    User.all.each do |user|
```
stevenday: Also fixed in e25df8589ee9340ba3895870fe9d625ba82bf0eb

app/models/alaveteli_pro/request_summary.rb
```diff
@@ -26,6 +26,18 @@ class AlaveteliPro::RequestSummary < ActiveRecord::Base
...
                              "AlaveteliPro::DraftInfoRequestBatch"].freeze
 
+  def self.category(category_slug)
+    includes(:request_summary_categories).
+      where("request_summary_categories.slug = ?", category_slug.to_s).
+        references(:request_summary_categories)
+  end
+
+  def self.not_category(category_slug)
+    includes(:request_summary_categories).
```
stevenday: Hmm, yeah this was a bit trickier than I thought. I've fixed it in b049e05f36385b1ae7bf613102b1c5cc66386685 but I've had to resort to two queries to get it done. If you have any thoughts on how to do it more efficiently, let me know :)

spec/models/alaveteli_pro/request_summary_spec.rb
```diff
@@ -28,4 +28,102 @@
...
+        request.save
+        updated_summary = AlaveteliPro::RequestSummary.create_or_update_from(request)
+        expect(updated_summary.id).to eq summary.id
+        expect(updated_summary.title).to eq request.title
+        expect(updated_summary.public_body_names).to eq body.name
+        expect(updated_summary.summarisable).to eq request
+      end
+
+      it "updates the existing summary from a batch request" do
+        bodies = FactoryGirl.create_list(:public_body, 3)
```
stevenday: I've fixed this up (it was in a variety of places) in:
6092d93
87e200c
58af56b
1812be5
3e72a26
bcdfb4a
289a068
1c67ba9
a5744ff

spec/models/alaveteli_pro/request_summary_spec.rb
```diff
@@ -28,4 +28,102 @@
...
+        batch = FactoryGirl.create(:info_request_batch, public_bodies: bodies)
+        summary = FactoryGirl.create(:request_summary, summarisable: batch)
+        body = FactoryGirl.create(:public_body)
```
stevenday: Fixed - see https://github.com/mysociety/alaveteli/pull/3923#discussion_r114945256

app/views/alaveteli_pro/draft_info_request_batches/_draft_info_request_batch.html.erb
```diff
@@ -0,0 +1,39 @@
+<div class="request">
+  <h3 class="request__title">
+    <% if draft_info_request_batch.embargo_duration %>
+      <span class="embargo-indicator embargo-indicator--small">Private</span>
```
stevenday: Fixed in 8795430122175eb7ed0545141d5615d3fa0b1d12

app/views/alaveteli_pro/draft_info_request_batches/_draft_info_request_batch.html.erb
stevenday: Hmm, I thought the link_to helper fell through to using the path as the link - that's what seems to happen for me:

screen shot 2017-05-05 at 09 22 01

Does that not work for you, or do you think I should change it to something generic (like "no title") anyway?

crowbot: So it's strange - it does that for non-embargoed requests, but doesn't seem to for embargoed ones
screen shot 2017-05-05 at 09 30 54

app/views/alaveteli_pro/info_request_batches/_info_request_batch.html.erb
```diff
@@ -0,0 +1,69 @@
...
+
+  </div> <!-- .request__meta -->
+
+  <%= render partial: 'alaveteli_pro/info_request_batches/progress_bar',
+             locals: { phases: info_request_batch.request_phases_summary,
+                       batch: info_request_batch } %>
+
+  <label class="houdini-label batch-request__show-requests"
+         for="houdini-batch-expand-<%= info_request_batch.id %>">
+    <%= _("Show requests") %>
```
stevenday: Added in 2b330ac

app/views/alaveteli_pro/info_request_batches/_info_request_batch.html.erb
```diff
@@ -0,0 +1,37 @@
+<div class="request">
+  <h3 class="request__title">
+    <% if info_request_batch.embargo_duration %>
+      <span class="embargo-indicator embargo-indicator--small">Private</span>
```
stevenday: Fixed in 8795430 too

app/models/alaveteli_pro/request_summary.rb
crowbot: Nope - looks good to me - might be worth reprhrasing `bad_ids` as something that more directly expresses that these are summaries in the category - maybe `summary_ids_to_exclude`?

app/views/alaveteli_pro/info_request_batches/_progress_bar.html.erb
```diff
@@ -0,0 +1,18 @@
+<div class="batch-summary">
+  <% phases.each do |label, count| %>
+    <% if count > 0 %>
+      <div class="batch-summary__phase
+                  batch-summary__phase--<%= label.parameterize %>"
```
stevenday: Fixed in 5edbd53 and f811d18

app/models/concerns/alaveteli_pro/request_summaries.rb
stevenday: Fixed in 8f20007

app/models/alaveteli_pro/request_summary_category.rb
```diff
@@ -0,0 +1,40 @@
+class AlaveteliPro::RequestSummaryCategory < ActiveRecord::Base
+  has_and_belongs_to_many :request_summaries,
+                          :class_name => AlaveteliPro::RequestSummary
+
```
stevenday: I did think about that, but I wasn't sure whether it really gained us anything? We have the categories we rely on in the seeds file already, and I can't think of a good reason to lock down the creation of any more, except perhaps for having another way of making it explicit which ones are available. If that sounds like a good idea, I can add it, what do you reckon?

app/models/info_request_batch.rb
```diff
@@ -117,4 +119,39 @@ def example_request
...
+  # Summarise the phases requests are in into three groups:
+  # in progress, action needed, complete and provide a count of the number of
+  # requests in each group.
+  #
+  # Returns hash of string group names mapped to an integer
+  def request_phases_summary
+    {
+      _('In progress') => self.info_requests.in_progress.count,
+      _('Action needed') => self.info_requests.action_needed.count,
+      _('Complete') => self.info_requests.complete.count
```
stevenday: Ah, whoops. I looked through the list again and I wasn't so sure:
These could definitely be 'in progress'
`'gone_postal'`
`'internal_review'`

These too, although they're not quite so obvious:
`'error_message'`
`'requires_admin'`
`'attention_requested'`

But these two are kind of 'ending' statuses I think, you can't really do any more with a request if you've marked it as either of these?
`'vexatious'`
`'not_foi'`

Maybe a better solution would just be having an 'other' section to the progress bar instead, it could be a grey section at the end perhaps?

app/models/alaveteli_pro/request_summary.rb
```diff
@@ -15,4 +15,58 @@
...
+  def self.attributes_from_request(request)
+    {
+      title: request.title,
+      body: extract_request_body(request),
+      public_body_names: extract_request_public_body_names(request),
+      summarisable: request
+    }
+  end
+
+  def self.extract_request_body(request)
```
stevenday: Done in d6860582fd0c559e92b013fef5e0b85484c01d8f

stevenday:
@crowbot - thanks for the thorough review, I could do with your thoughts on a few things:

- Having a list of allowed categories: https://github.com/mysociety/alaveteli/pull/3923#discussion_r114990169
- Reducing db calls for the authority list sentence: https://github.com/mysociety/alaveteli/pull/3923#discussion_r115004266
- Other statuses not being included in progress bars: https://github.com/mysociety/alaveteli/pull/3923#discussion_r115014709

app/views/alaveteli_pro/draft_info_request_batches/_draft_info_request_batch.html.erb
stevenday: Weird! Ok, I'll look at this on Monday

stevenday: Note to self - there doesn't seem to be anything weird about the urls for embargoed requests, I think what must be happening is that `.title` returns an empty string for those ones, rather than nil.

stevenday: Now fixed in b2e7235f6ee0a68f4c492673390977efdfae6a3a

app/models/alaveteli_pro/request_summary.rb
stevenday: Done in e05747cd59d78e92eae719120b5f8705b90e68df

app/models/alaveteli_pro/request_summary_category.rb
crowbot: Yeah, I was a bit iffy on the value as I was writing the comment - what I was thinking is that it prevents the creation of a new category at runtime. But that would actually have to happen from the console, so I'm happy to leave this.

app/views/alaveteli_pro/info_request_batches/_authority_list.html.erb
```diff
@@ -1,9 +1,9 @@
...
-  <span class="to_public_body_label"><%= _('To') %></span>
-
-  <%= n_('{{count}} recipient, <a href="{{url}}">show</a>',
-         '{{count}} recipients, <a href="{{url}}">show all</a>',
-         batch.public_bodies.length,
-         count: batch.public_bodies.length,
-         url: alaveteli_pro_batch_request_authority_searches_path(draft_id: draft.id)) %>
-</p>
+<% if bodies %>
+  <%= n_('{{count}} recipient, {{body_names}}',
```
crowbot: > I'm also not sure how that would work, because it's a translated string, so we couldn't actually store the finished sentence it in the db anyway. Am I missing something obvious about what you were suggesting?

Nope, I hadn't thought about translation. I think adding the counter cache on the bodies association is probably worth it - I can imagine us wanting to get the number of bodies for a batch quite a lot. Let's not worry about optimizing apart from that for now. 

Looking at the translation now just from a translation point of view, I think it might be better to make the change below to give translators more freedom in word order.

```diff 

- <%= n_('{{count}} recipient, {{public_body_names}}',
-         '{{count}} recipients, including {{public_body_names}}',
-         public_bodies.length,
-         count: public_bodies.length,
-         public_body_names: public_bodies.take(2).map(&:name).join(_(" and "))) %>
+<% example_body_names = public_bodies.take(2).map(&:name) %>
+  <%= n_('{{count}} recipient, {{public_body_names}}',
+         '{{count}} recipients, including {{first_public_body}} and {{second_public_body}}',
+         public_bodies.length,
+         count: public_bodies.length,
+         first_public_body: example_body_names.first, 
+         second_public_body: example_body_names.second) %>
```

app/models/info_request_batch.rb
crowbot: > But these two are kind of 'ending' statuses I think, you can't really do any more with a request if you've marked it as either of these?
> 'vexatious'
> 'not_foi'

Yeah, those are tricky - they're basically statuses only applied by admins (mostly to hide public requests from the site). 

> Maybe a better solution would just be having an 'other' section to the progress bar instead, it could be a grey section at the end perhaps?

Yeah, I think that makes sense in terms of getting this out and into the hands of users. As you say, once we get feedback from them, we may want to split these statuses and put some into 'in progress' but I don't think we should hold up getting the basic implementation out there for some road testing.

stevenday: Done in a1f0d52 and cfc77bc

app/views/alaveteli_pro/info_request_batches/_authority_list.html.erb
stevenday: Fixed in 4afbe1236a5f4af70c1baade3d5215da13695b9c

app/assets/javascripts/alaveteli_pro/batch_request_dropdown.js
```diff
@@ -0,0 +1,16 @@
+(function($){
+  $(function(){
+    var $checkboxes = $('.js-show-batch-requests');
+    $checkboxes.on('change', function() {
+      var $checkbox = $(this);
+      var $label = $('label[for="' + $checkbox.attr('id') + '"]');
```
crowbot: I think we need to store this in a `data` attribute too - otherwise the label just changes on the first click as both labels are the same after that.

stevenday: Done in 7c7cf95 - will squash all the fixups and merge, thanks!

PR #3925, opened 2017-04-25
Install strip_attributes from git ref

garethrees:
Using a branch makes updating the fork from the upstream repo really difficult.

If you rebase globalize3 on upstream/master and force push, you lose the
previous SHA, so old versions become uninstallable.


garethrees:
New git ref from https://github.com/mysociety/strip_attributes/tree/globalize3-1.8

PR #3934, opened 2017-04-27
[#3799] Rails 4.1

lizconlan:
Work on #3799


app/helpers/public_body_helper.rb
```diff
@@ -40,9 +40,9 @@ def public_body_not_requestable_reasons(public_body)
   # Returns a String
   def type_of_authority(public_body)
     categories = PublicBodyCategory.
-      where(:category_tag => public_body.tag_string.split)
+      where(:category_tag => public_body.tag_string.split).to_a
 
-    types = categories.each_with_index.map do |category, index|
+    types = categories.to_a.each_with_index.map do |category, index|
```
garethrees: Why do this twice (L43 and then again on L45)? I'd just do this on L45 – closest to the `map`.

app/controllers/admin_user_controller.rb
```diff
@@ -159,7 +159,7 @@ def changed_role_ids
   def check_role_requirements
     role_names = Role.
                    where(:id => params[:admin_user][:role_ids]).
-                     pluck(:name).map{ |role| role.to_sym }
+                     pluck(:name).to_a.map{ |role| role.to_sym }
```
garethrees: Please add links in commit messages to documentation about changes like this.

spec/mailers/track_mailer_spec.rb
```diff
@@ -28,7 +28,7 @@
                            :url_name => 'test-name',
                            :get_locale => 'en',
                            :should_be_emailed? => true)
-        klass = ActiveRecord::Relation::ActiveRecord_Relation_User
+        klass = ActiveRecord::Relation
```
garethrees: Why?

lib/no_constraint_disabling.rb
```diff
@@ -24,9 +24,9 @@ def disable_referential_integrity(&block)
 module ActiveRecord
   class FixtureSet
 
-    def self.create_fixtures(fixtures_directory, fixture_set_names, class_names = {})
+    def self.create_fixtures(fixtures_directory, fixture_set_names, class_names = {}, config = ActiveRecord::Base)
```
garethrees: Where did you update this from? Description in commit message plz.

app/models/incoming_message.rb
```diff
@@ -55,7 +55,8 @@ class IncomingMessage < ActiveRecord::Base
   # never really has many info_request_events, but could in theory
   has_many :info_request_events, :dependent => :destroy
 
-  belongs_to :raw_email, :dependent => :destroy
```
garethrees: Links / references to what says not to do this?

http://guides.rubyonrails.org/v4.1.16/association_basics.html#options-for-belongs-to-dependent suggests you can?

garethrees: Looks like this got fixed in 4.2. Search for `12380` in https://github.com/rails/rails/blob/4-2-stable/activerecord/CHANGELOG.md. Doesn't exist in https://github.com/rails/rails/blob/4-1-stable/activerecord/CHANGELOG.md

spec/mailers/track_mailer_spec.rb
lizconlan: The spec fails in the original form. Haven't quite worked out why yet

garethrees: What was the failure? How did you come to figure out that removing `::ActiveRecord_Relation_User` fixed it?

app/models/incoming_message.rb
lizconlan: Looks like it's this PR https://github.com/rails/rails/pull/12450/files

garethrees: Yeah; that fixed the issue mentioned in the changelog (https://github.com/rails/rails/issues/12380).

spec/mailers/track_mailer_spec.rb
lizconlan: the error is

```
NameError:
uninitialized constant ActiveRecord::Relation::ActiveRecord_Relation_User
```

lizconlan: (error now noted in commit message)

app/models/incoming_message.rb
lizconlan: possible monkeypatch here if the original fix proves problematic https://gist.github.com/lizconlan/7e577eeff16ad5a83b943601d4241aee

app/controllers/admin_user_controller.rb
garethrees: `map` isn't a mutator; `map!` is. Are you sure this is needed?

lizconlan: uh, also `pluck` returns an Array so... no, not needed!

lizconlan: short answer, yes `map` is a problem where an `ActiveRecord_Relation` is being returned ...

```
reqs = InfoRequest.all
reqs.map
=> NoMethodError: undefined method `map' for InfoRequest::ActiveRecord_Relation:Class
```

*but* I've also called mistakenly called `.to_a` on Hashes and Arrays. Fixing...

spec/mailers/track_mailer_spec.rb
lizconlan: um, I suspect the **correct** version may be that it's a `User::ActiveRecord_Relation` now rather than an `ActiveRecord::Relation::ActiveRecord_Relation_User`

app/controllers/outgoing_messages/delivery_statuses_controller.rb
```diff
@@ -11,7 +11,7 @@ def show
     @show_mail_server_logs = @outgoing_message.is_owning_user?(@user)
 
     if @show_mail_server_logs
-      @mail_server_logs = @outgoing_message.mail_server_logs.to_a.map do |log|
```
lizconlan: child objects are returned ad `ActiveRecord::Associations::CollectionProxy` which has Array methods like `map` and `each` so no need to convert the result

app/helpers/alaveteli_pro/info_requests_helper.rb
```diff
@@ -8,7 +8,7 @@ def publish_at_options
 
   def embargo_extension_options(embargo)
     options = AlaveteliPro::Embargo::TranslatedConstants.
-        duration_labels.to_a.map do |value, label|
```
lizconlan: that's a Hash

app/controllers/admin_user_controller.rb
```diff
@@ -159,7 +159,7 @@ def changed_role_ids
   def check_role_requirements
     role_names = Role.
                    where(:id => params[:admin_user][:role_ids]).
-                     pluck(:name).to_a.map{ |role| role.to_sym }
```
lizconlan: `.pluck` returns an Array anyway

lib/patch_tag_helper.rb
```diff
@@ -11,30 +11,9 @@ def tag_option(key, value, escape)
     if value.is_a?(Array)
       value = escape ? safe_join(value, " ") : value.join(" ")
     else
-      value = escape ? ERB::Util.unwrapped_html_escape(value) : value.to_s
+      value = escape ? ERB::Util.html_escape(value) : value.to_s
```
garethrees: > 4.1's html_escape is identical to unwrapped_html_escape

To me, it doesn't look identical.

- `html_escape` in 4.1 [marks the gsubbed string as `html_safe`](https://github.com/rails/rails/blob/4-1-stable/activesupport/lib/active_support/core_ext/string/output_safety.rb#L25).
- `unwrapped_html_escape` in 4.2 [does not](https://github.com/rails/rails/blob/4-2-stable/activesupport/lib/active_support/core_ext/string/output_safety.rb#L41).

I'm not sure if I've misinterpreted what the commit is doing, but I'm just going to drop this commit to retain the manual patch until we upgrade to Rails 4.2 and get the patch included in the release.

app/helpers/public_body_helper.rb
garethrees: Resolved

app/helpers/public_body_helper.rb
```diff
@@ -42,7 +42,7 @@ def type_of_authority(public_body)
     categories = PublicBodyCategory.
       where(:category_tag => public_body.tag_string.split)
 
-    types = categories.each_with_index.map do |category, index|
+    types = categories.to_a.each_with_index.map do |category, index|
```
garethrees: This seems to work fine without a conversion to `Array`:

```ruby
categories = PublicBodyCategory.where(:category_tag => 'popular_agency')
# => [#<PublicBodyCategory id: 3, category_tag: "popular_agency">]

types = categories.each_with_index.map do |category, index|
  [category.id, index]
end
# => [[3, 0]]
```

garethrees:
```
$ git --no-pager diff
diff --git a/app/helpers/public_body_helper.rb b/app/helpers/public_body_helper.rb
index d76ee9f..62f3a14 100644
--- a/app/helpers/public_body_helper.rb
+++ b/app/helpers/public_body_helper.rb
@@ -42,7 +42,7 @@ module PublicBodyHelper
     categories = PublicBodyCategory.
       where(:category_tag => public_body.tag_string.split)

-    types = categories.to_a.each_with_index.map do |category, index|
+    types = categories.each_with_index.map do |category, index|
       desc = category.description
       if index.zero?
         desc = desc.sub(/\S/) { |m| Unicode.upcase(m) }


$ be rspec spec/helpers/public_body_helper_spec.rb:52
Run options: include {:locations=>{"./spec/helpers/public_body_helper_spec.rb"=>[52]}}

Randomized with seed 13314
......

Finished in 2.91 seconds (files took 55.17 seconds to load)
6 examples, 0 failures

Randomized with seed 13314

[Coveralls] Outside the CI environment, not sending data.
```

spec/mailers/track_mailer_spec.rb
garethrees: Yes, this should be more specific as per the intent of the original spec.

```ruby
User.where(["last_daily_track_email < ?", Time.zone.now - 1.day ]).class
# => User::ActiveRecord_Relation
```

config/application.rb
```diff
@@ -78,46 +78,5 @@ class Application < Rails::Application
       config.action_mailer.default_url_options[:protocol] = "https"
     end
 
-    # Enable the asset pipeline
```
garethrees: Squash in to cb43aaee7511b95bbae97ebc26a83a5a7527750f

Mention http://railsdiff.org/4.0.13/4.1.16 so that we know where this change came from.

PR #3943, opened 2017-05-03
[#3799] update gettext gr

garethrees:
#3799


garethrees:
Simpler attempt at https://github.com/mysociety/alaveteli/pull/3914 – only deals with `gettext:find` and `gettext:clean`. Closing for now.

PR #3945, opened 2017-05-03
[#3799] Update generated templates

lizconlan:
#3799


lizconlan:
> Probably want to move these out in a further commit?

Good catch, thanks!

PR #3947, opened 2017-05-03
[#3799] CSRF protection changes

lizconlan:
Part of #3799 

http://edgeguides.rubyonrails.org/upgrading_ruby_on_rails.html#csrf-protection-from-remote-script-tags

Although it looks as though we may only be requiring forgery protection for logged in users as we have this in our application controller

```rb
  protect_from_forgery :if => :user?
  skip_before_filter :verify_authenticity_token, :unless => :user?
```


lizconlan:
> TL;DR: I don't believe that this change is a) correct or b) required

Yeah, wasn't entirely sure about this one. Closing it seems to be the best thing - thanks for taking the time to review!

PR #3948, opened 2017-05-03
[#3799] Add Spring

lizconlan:
Part of #3799


lizconlan:
Well, I've added it so now I get warnings that read: "Warning: You're using Rubygems 1.8.11 with Spring. Upgrade to at least Rubygems 2.1.0 and run `gem pristine --all` for better startup performance."

I've had to add a workaround for Rubygems < 2.0.0 to make it work at all and there's not a noticeable performance boost. Not convinced at the moment.

Gemfile
```diff
@@ -183,4 +183,5 @@ group :development do
   gem 'mailcatcher', '~> 0.6.0'
   gem 'quiet_assets', '~> 1.1.0'
   gem 'launchy', '~> 2.4.0'
+  gem 'spring', '< 2.0.0'
```
garethrees: Why `< 2.0.0`?

garethrees: Where's the version that should be installed? https://github.com/mysociety/alaveteli/blob/067d4ed948539538339b92566b5320782d38b878/Gemfile#L22-L25

garethrees: I think we also want `spring-commands-rspec` too.

bin/rails
```diff
@@ -1,4 +1,9 @@
 #!/usr/bin/env ruby1.9.1
```
garethrees: I've just noticed that this should probably be `/usr/bin/env ruby`, as per how `bin/string` got generated on your system. I wonder if this is because I'm using `ruby1.9.1` via apt alternatives. I'd be in favour of a commit that updates all of these to `#!/usr/bin/env ruby`.

bin/spring
```diff
@@ -8,6 +8,14 @@ unless defined?(Spring)
   require 'bundler'
 
   if (match = Bundler.default_lockfile.read.match(/^GEM$.*?^    (?:  )*spring \((.*?)\)$.*?^$/m))
+    # the path_separator method was introduced in rubygems 2.0.0
+    # use its default value of File::PATH_SEPARATOR for older versions
+    # https://github.com/rubygems/rubygems/blob/2.0/lib/rubygems/defaults.rb#L65-L67
+    path_separator = if Gem.respond_to?(:path_separator)
```
garethrees: I think this change is fine, given the binstubs are versioned

garethrees: Note that running with Rubygems 1.8.x prints a deprecation warning:

```
Warning: You're using Rubygems 1.8.11 with Spring. Upgrade to at least Rubygems 2.1.0 and run `gem pristine --all` for better startup performance.
```

Gemfile
lizconlan: 2.0.0 drops Rails 4.0 and 4.1 (fixed in squash commit)

lizconlan:
3f942e0 is a temp commit to get a Travis run against develop

lizconlan:
test passing https://travis-ci.org/mysociety/alaveteli/builds/229144411

garethrees:
Looks good to me. Lets try it and see how it goes!

PR #3950, opened 2017-05-04
[#3799] rails 4 1 gr

garethrees:
#3799


garethrees:
Marking for review as discussed in standup. Main thing is to have a check that bd96c54 isn't going to break anything. Some links in the commit comment that are of interest.

PR #3953, opened 2017-05-04
[#3799] rails 4 1 lc

lizconlan:
Part of #3799 

The new session cookie format looks useful - [supposed to be more secure](https://www.happybearsoftware.com/almost-protect-yourself-from-cookie-session-store) and I wonder if using JSON instead of Marshal might be slightly faster but I haven't been able to test that. Did need to tweak the code that checks for expiry to work under Ruby 1.9 tho.

(The long term suggestion for the cookie serializer is that it should be switched to `Rails.application.config.action_dispatch.cookies_serializer = :json` once we are confident that all the old format sessions have been migrated across.)


config/secrets.yml
```diff
@@ -0,0 +1,8 @@
+development:
+  secret_key_base: AlaveteliConfiguration.secret_key_base
```
garethrees: Err, wouldn't this literally make the value  `"AlaveteliConfiguration.secret_key_base"`.

config/initializers/secret_token.rb
```diff
@@ -1,21 +0,0 @@
...
-# if you're sharing your code publicly.
-
-# Example Rails 4 secret_key_base option
-Rails.application.config.secret_key_base =
-  AlaveteliConfiguration.secret_key_base
-
-# Just plopping an extra character on the secret_token so that any sessions on upgrading from
-# Rails 2 to Rails 3 version of Alaveteli are invalidated.
-# See http://blog.carbonfive.com/2011/03/19/rails-3-upgrade-tip-invalidate-session-cookies/
-Rails.application.config.secret_token = "3" + AlaveteliConfiguration::cookie_store_session_secret
```
garethrees: Is this not required in Rails 4.1? Where does it get set now?

lizconlan: Looks like it isn't - according to http://guides.rubyonrails.org/upgrading_ruby_on_rails.html#action-pack it's superseded by `secret_key_base` in Rails 4. Have added a note tom the fixup commit

config/secrets.yml
lizconlan: oops, fixed (and no longer using "real" config for dev and test) - thanks!

app/controllers/application_controller.rb
```diff
@@ -132,7 +132,12 @@ def session_remember_me
   # is not replayable forever
   SESSION_TTL = 3.hours
   def validate_session_timestamp
-    if session[:user_id] && session[:ttl] && session[:ttl] < SESSION_TTL.ago
+    session_ttl = if session[:ttl].is_a?(String)
+      Time.zone.parse(session[:ttl]) # for Ruby 1.9
```
garethrees: No controller specs seem to fail before this change – should there be a spec to demonstrate the problem?

config/initializers/secret_token.rb
garethrees: 👍

app/controllers/application_controller.rb
lizconlan: That seemed to fails consistently on 1.9 specs with the new`config.action_dispatch.cookies_serializer` setting but without that code change

garethrees: 👍

PR #3958, opened 2017-05-05
[Pro] Issues/alaveteli pro 126/marketing page

crowbot:
To do
- [x] Check what we want in the 'contact us' section
- [x] Confirm text - @MyfanwyNixon looking at it. Should somewhere mention other anticipated users e.g. campaigners and researchers
- [x] Add screenshots to the Pro section
- [x] Add screenshots to the batch section
- [x] Needs to send an email to us
- [x] Remove `pro_invite` route
- [x] Somewhere (either here or in our response) we need to make clear that people are getting a year's free access


app/assets/stylesheets/responsive/alaveteli_pro/_pro_marketing.scss
```diff
@@ -116,6 +116,9 @@ label.input-beta-button {
   input[type="radio"] {
     margin-right: 0.35em;
   }
+  .fieldWithErrors{
```
stevenday: Ah, this is a rails thing, ignore

Issue #1240, opened 2013-11-27
Use a queue for offline jobs


crowbot:
Possible job types to include: 
- Updates to the xapian index (see also https://github.com/mysociety/alaveteli/issues/1168#issuecomment-27673460)
- Processing of incoming messages (https://github.com/mysociety/alaveteli/issues/974)
- Sending email alerts
- Sending batch requests

Possible advantages: 
- Separation of concerns - web app interface versus more resource intensive processes
- Closer to realtime updates of the index
- Resilience to memory leaks, bloat
- One coherent system rather than multiple cron-jobs and daemons

garethrees:
Rails 4.2 (https://github.com/mysociety/alaveteli/issues/2968) has a built-in framework for integrating [background jobs](http://weblog.rubyonrails.org/2014/12/19/Rails-4-2-final/).

garethrees:
Rails 6.1 (#6011) has the ability to [destroy associated records asynchronously](https://guides.rubyonrails.org/v6.1/6_1_release_notes.html#destroy-associations-async) which would help prevent timeouts when managing WDTK.

gbp:
To help 3rd party installs, @garethrees wanted to look at using a PG backed queue over Sidekiq/Redis - this seems like a good option: https://github.com/bensheldon/good_job

garethrees:
https://github.com/que-rb/que was also something I'd seen that seems to be pretty well maintained.

garethrees:
Our initial stab at this should be quite basic; mainly with the intention of getting the job infrastructure in place and running, with an initial concrete implementation of ActiveJob so that we actually gain some benefit from doing the setup work.

I think we have 4 main tasks.

1. Decide on a QueueAdapter
1. Add ActiveJob configuration (ensure we handle errors, etc)
1. Add infrastructure to run the backend processor (unix daemon?)
1. Add a concrete implementation of ActiveJob

2 and 3 don't need too much discussion at this point, so I'll focus on 1 and 4.

To reiterate, the *main focus is getting the ActiveJob infrastructure in place and running* so that over time we can more easily add jobs in future; we're not going to get everything migrated to a job in one shot.

## 1. Decide on a QueueAdapter

As previously mentioned, I'm keen to avoid additional infrastructure costs for re-users who don't have in-house sysadmins. Sidekiq requires Redis, which adds another thing to configure, secure, run and maintain.

[...]

PR #3968, opened 2017-05-10
Expire associated requests when adding an authority censor rule

### LINKED ISSUE #3959: New authority censor rules are not applied to existing responses
If you add a censor rule to an authority after a response has been published, the intended redaction only applies to the correspondence header until the message cache (e.g. `cached_main_body_text_unfolded`) is cleared.

The censor rule intention was to replace the last name of the sender.

screen shot 2017-05-05 at 17 56 58

(May be looking in the wrong place but I can't see how an admin can fix this)

lizconlan:
Fixes #3959

### DIFF
--- a/app/models/censor_rule.rb
+++ b/app/models/censor_rule.rb
@@ -56,6 +56,20 @@ class CensorRule < ActiveRecord::Base
     info_request_id.nil? && user_id.nil? && public_body_id.nil?
   end
 
+  def expire_requests
+    if info_request
+      info_request.expire
+    elsif user
+      user.expire_requests
+    elsif public_body
+      public_body.expire_requests
+    else # global rule
+      InfoRequest.find_in_batches do |group|
+        group.each { |request| request.expire }
+      end
+    end
+  end
+
   private
 
   def single_char_regexp
--- a/spec/controllers/admin_censor_rule_controller_spec.rb
+++ b/spec/controllers/admin_censor_rule_controller_spec.rb
@@ -218,6 +218,16 @@ describe AdminCensorRuleController do
 
       context 'successfully saving the censor rule' do
 
+        it 'calls expire_requests on the new censor_rule' do
+          censor_rule = FactoryGirl.build(:global_censor_rule)
+          allow(CensorRule).to receive(:new) { censor_rule }
+          allow(censor_rule).to receive(:expire_requests)
+
+          create_censor_rule
+
+          expect(censor_rule).to have_received(:expire_requests)
+        end
+
         it 'redirects to the censor rules index' do
           create_censor_rule
           expect(response).to redirect_to(
@@ -295,19 +305,17 @@ describe AdminCensorRuleController do
           expect(flash[:notice]).to eq(msg)
         end
 
-        it 'purges the cache for the info request' do
-          info_request = FactoryGirl.create(:info_request)
-          censor_rules = double
-          allow(info_request).to receive(:censor_rules) { censor_rules }
+        it 'calls expire_requests on the new censor_rule' do
           allow(InfoRequest).to receive(:find) { info_request }
           censor_rule = FactoryGirl.build(:info_request_censor_rule,
                                           :info_request => info_request)
-          allow(censor_rules).to receive(:build) { censor_rule }
-
-          expect(info_request).to receive(:expire)
+          allow(info_request.censor_rules).to receive(:build) { censor_rule }
+          allow(censor_rule).to receive(:expire_requests)
 
           post :create, :censor_rule => censor_rule_params,
                         :request_id => info_request.id
+
+          expect(censor_rule).to have_received(:expire_requests)
         end
 
         it 'redirects to the associated info request' do
@@ -378,15 +386,16 @@ describe AdminCensorRuleController do
 
       context 'successfully saving the censor rule' do
 
-        it 'purges the cache for the info request' do
-          expect(User).to receive(:find) { user }
-          censor_rules = double
-          allow(user).to receive(:censor_rules) { censor_rules }
-          censor_rule = FactoryGirl.build(:user_censor_rule, :user => user)
-          allow(censor_rules).to receive(:build) { censor_rule }
+        it 'calls expire_requests on the new censor_rule' do
+          allow(User).to receive(:find) { user }
+          censor_rule = FactoryGirl.build(:user_censor_rule,
+                                          :user => user)
+          allow(user.censor_rules).to receive(:build) { censor_rule }
+          allow(censor_rule).to receive(:expire_requests)
 
-          expect(censor_rule.user).to receive(:expire_requests)
           create_censor_rule
+
+          expect(censor_rule).to have_received(:expire_requests)
         end
 
         it 'redirects to the associated info request' do
@@ -466,6 +475,19 @@ describe AdminCensorRuleController do
                         :body_id => public_body.id
           msg = 'Censor rule was successfully created.'
           expect(flash[:notice]).to eq(msg)
+        end
+
+        it 'calls expire_requests on the new censor_rule' do
+          allow(PublicBody).to receive(:find) { public_body }
+          censor_rule = FactoryGirl.build(:public_body_censor_rule,
+                                          :public_body => public_body)
+          allow(public_body.censor_rules).to receive(:build) { censor_rule }
+          allow(censor_rule).to receive(:expire_requests)
+
+          post :create, :censor_rule => censor_rule_params,
+                        :body_id => public_body.id
+
+          expect(censor_rule).to have_received(:expire_requests)
         end
 
         it 'redirects to the associated public body' do
@@ -624,6 +646,15 @@ describe AdminCensorRuleController do
           expect(flash[:notice]).to eq(msg)
         end
 
+        it 'calls expire_requests on the censor_rule' do
+          allow(CensorRule).to receive(:find) { censor_rule }
+          allow(censor_rule).to receive(:expire_requests)
+          put :update, :id => censor_rule.id,
+            :censor_rule => { :text => 'different text' }
+
+          expect(censor_rule).to have_received(:expire_requests)
+        end
+
         it 'redirects to the censor rule index' do
           put :update, :id => censor_rule.id,
             :censor_rule => { :text => 'different text' }
@@ -691,14 +722,13 @@ describe AdminCensorRuleController do
           expect(flash[:notice]).to eq(msg)
         end
 
-        it 'purges the cache for the info request' do
-          info_request = FactoryGirl.create(:info_request)
-          allow(CensorRule).to receive(:find).and_return(censor_rule)
-          allow(censor_rule).to receive(:info_request).and_return(info_request)
-          expect(info_request).to receive(:expire)
-
+        it 'calls expire_requests on the censor_rule' do
+          allow(CensorRule).to receive(:find) { censor_rule }
+          allow(censor_rule).to receive(:expire_requests)
           put :update, :id => censor_rule.id,
-              :censor_rule => { :text => 'different text' }
+            :censor_rule => { :text => 'different text' }
+
+          expect(censor_rule).to have_received(:expire_requests)
         end
 
         it 'redirects to the associated info request' do
@@ -769,12 +799,13 @@ describe AdminCensorRuleController do
           expect(flash[:notice]).to eq(msg)
         end
 
-        it 'purges the cache for the info request' do
-          expect(CensorRule).to receive(:find) { censor_rule }
-          expect(censor_rule.user).to receive(:expire_requests)
-
+        it 'calls expire_requests on the censor_rule' do
+          allow(CensorRule).to receive(:find) { censor_rule }
+          allow(censor_rule).to receive(:expire_requests)
           put :update, :id => censor_rule.id,
-              :censor_rule => { :text => 'different text' }
+            :censor_rule => { :text => 'different text' }
+
+          expect(censor_rule).to have_received(:expire_requests)
         end
 
         it 'redirects to the associated info request' do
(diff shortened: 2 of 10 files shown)


spec/integration/admin_censor_rule_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+    using_session(@admin) do
+      visit new_admin_body_censor_rule_path(@authority)
+      fill_in 'censor_rule_text', :with => 'rubbish question'
+      fill_in 'censor_rule_replacement', :with => '[REDACTED]'
+      fill_in 'censor_rule_last_edit_comment', :with => 'test'
+      click_button 'Create'
+    end
+
+    incoming_message.reload
+    expect(incoming_message.cached_main_body_text_unfolded).to be_nil
```
garethrees: I think I'd feel better with an even more explicit spec here.

* create request with specific text e.g. "I have a rubbish question for you"
* add new censor rule as you've done
* visit request page
* check that request page has the contents "I have a [REDACTED] for you"

spec/controllers/admin_censor_rule_controller_spec.rb
```diff
@@ -835,6 +835,16 @@ def create_censor_rule
...
         end
 
+        it 'purges the cache for the info request' do
+          public_body = double("public_body")
+          allow(CensorRule).to receive(:find) { @censor_rule }
+          allow(@censor_rule).to receive(:public_body) { public_body }
+          expect(public_body).to receive(:expire_requests)
+
+          put :update, :id => @censor_rule.id,
+            :censor_rule => { :text => 'different text' }
```
garethrees: I appreciate this is an existing pattern, but the level of stubbing here makes me uncomfortable.

I don't think the controller should have the logic of what to expire.  I think the spec should be this:

```ruby
it 'purges the cache for the info request' do
  expect(@censor_rule).to receive(:expire_requests)
  put :update, :id => @censor_rule.id,
               :censor_rule => { :text => 'different text' }
end
```
I know that `CensorRule#expire_requests` doesn't yet exist; I think it should.

Here's what I think we want:

```ruby
class CensorRule #…
  def expire_requests
    if info_request
      info_request.expire
    elsif user
      user.expire_requests
    elsif public_body
      public_body.expire_requests
    end
  end
 end

class AdminCensorRuleController #…
   def expire_requests_and_redirect
     @censor_rule.expire_requests

     if @censor_rule.info_request
       redirect_to admin_request_url(@censor_rule.info_request)
     elsif @censor_rule.user
       redirect_to admin_user_url(@censor_rule.user)
     elsif @censor_rule.public_body
       redirect_to admin_body_url(@censor_rule.public_body)
     else
       redirect_to admin_censor_rules_path
     end
   end
end
```

- `InfoRequest#expire` should be robustly tested
- `PublicBody#expire_requests` and `User#expire_requests` can just expect that `#expire` is called on the correct association(/s)
- `CensorRule#expire_requests` can check that `#expire_requests` is called on the correct associations
- The controller spec here can just check that `#expire_requests` is called on the `CensorRule`
- Integration tests can do a full-stack test to ensure that the end-to-end process results in the correct redactions.

spec/integration/admin_censor_rule_spec.rb
garethrees: This PR should also include similar end to end tests for each of the censor rule types

lizconlan:
> I haven't mentioned global censor rules. The problem of not expiring old requests exists here too. I think we should handle it in the same way, and add a warning to the inline help that creating/editing a global censor rule will iterate over all requests in the request cycle (so will probably time out).

Hmm. Is that bad? Will it complete but "just" break in the browser?

garethrees:
> Hmm. Is that bad? Will it complete but "just" break in the browser?

IDK exactly; I think its better than expecting requests to be redacted and then not being redacted.

Really all these should be background jobs, which will get enabled by #1240, which is on our immediate roadmap, so I'm not too worried about this as a long term solution. That said, might be worth bringing up in standup and getting some opinions.

garethrees:
The alternative is to temporarily disable global censor rule creation / edits?

lizconlan:
> The alternative is to temporarily disable global censor rule creation / edits?

Yeah, I did wonder if we were effectively deprecating global rules with this

lizconlan:
Decision - allow the timeout to occur, fix it at a later date by implementing DelayedJob (or similar)

spec/models/public_body_spec.rb
```diff
@@ -140,6 +140,19 @@
 
   end
 
+  describe '#expire_requests' do
+
+    it 'calls expire on all associated requests' do
+      authority = FactoryGirl.create(:public_body)
+      request = FactoryGirl.create(:info_request, :public_body => authority)
+      expect(authority).to receive(:info_requests).and_return([request])
```
stevenday: I don't think it's necessary to mock this, you might just need to reload the public body? (You could just create a request and then get the body out afterwards?)

app/models/censor_rule.rb
```diff
@@ -56,6 +56,18 @@ def is_global?
...
   end
 
+  def expire_requests
+    if info_request
+      info_request.expire
+    elsif user
+      user.expire_requests
+    elsif public_body
+      public_body.expire_requests
+    else # global rule
```
stevenday: Is this the same set of requests that global rules were applied to before this change? It seems like it will miss hidden requests and embargoed requests - is that intended?

spec/models/censor_rule_spec.rb
```diff
@@ -143,6 +143,40 @@
 
   end
 
+  describe '#expire_requests' do
+
+    it 'calls expire on the request if it is a request rule' do
```
stevenday: Super minor nitpick, betterspecs.org (which I think we've been following elsewhere) suggests to use contexts for this kind of thing: http://www.betterspecs.org/#contexts

spec/integration/admin_censor_rule_spec.rb
```diff
@@ -0,0 +1,53 @@
+# -*- encoding : utf-8 -*-
+require File.expand_path(File.dirname(__FILE__) + '/../spec_helper')
+require File.expand_path(File.dirname(__FILE__) + '/alaveteli_dsl')
+
+describe 'Adding an authority censor rule' do
+  before do
+    allow(AlaveteliConfiguration).to receive(:skip_admin_auth).and_return(false)
+
+    confirm(:admin_user)
+    @admin = login(:admin_user)
```
stevenday: More spec pendantisim: we've been using `let` (and `let!`) elsewhere in new specs, rather than setting instance variables like this.

spec/integration/admin_censor_rule_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+    @authority = FactoryGirl.create(:public_body)
+  end
+
+  it 'clears the cache for existing requests' do
+    raw_email_data = <<-EOF.strip_heredoc
```
stevenday: You could move all of this set up into `let` blocks or the `before`, to make the spec clearer in what it's actually testing vs what is just data for it to run?

stevenday: Maybe this should also be inside a context to explain what kind of censor rule it's creating, and then the data set up can be scoped appropriately in case we add other specs for user/global censor rules?

stevenday: Aha - great minds! sorry, I was taking this commit by commit so I didn't see this.

It might be clearer if you moved the integration specs out of this and pushed them as a fixup to the previous commit?  

I'd also still opt for more use of `let` :)

spec/models/info_request_spec.rb
```diff
@@ -1058,6 +1058,29 @@
...
+      info_request.expire
+    end
+
+    it 'removes itself from the varnish cache' do
+      expect(info_request).to receive(:purge_in_cache)
+      info_request.expire
+    end
+
+    it 'removes any zip files' do
+      allow(info_request).to receive(:foi_fragment_cache_directories) {[]}
```
stevenday: Maybe worth a comment here to say that if we don't do this, we'll get spurious calls to remove other directories?

spec/models/info_request_spec.rb
```diff
@@ -1058,6 +1058,29 @@
...
+      expect(FileUtils).to receive(:rm_rf).with(info_request.download_zip_dir)
+      info_request.expire
+    end
+
+    it 'removes any file caches' do
+      # called for each cache directory + the zip file directory
+      expected_calls = info_request.foi_fragment_cache_directories.count + 1
```
stevenday: We should probably expect the number of `foi_fragment_cache_directories` to be > 0 here, just in case?

spec/controllers/admin_censor_rule_controller_spec.rb
```diff
@@ -280,18 +290,18 @@ def create_censor_rule
...
 
-        it 'purges the cache for the info request' do
-          info_request = FactoryGirl.create(:info_request)
-          censor_rules = double
-          allow(info_request).to receive(:censor_rules) { censor_rules }
-          allow(InfoRequest).to receive(:find) { info_request }
-          censor_rule = FactoryGirl.build(:info_request_censor_rule, :info_request => info_request)
-          allow(censor_rules).to receive(:build) { censor_rule }
+        it 'calls expire_requests on the new censor_rule' do
+          allow(InfoRequest).to receive(:find) { @info_request }
```
stevenday: My personal preference would be to actually create all these test objects and mock less things, is that possible? I'm thinking that at least `@info_request` could be real and not need a mocking of the find method to return it?

stevenday: (It would also be an opportunity to replace @info_request with a let-ed equivalent)

spec/controllers/admin_censor_rule_controller_spec.rb
```diff
@@ -359,15 +369,17 @@ def create_censor_rule
...
-          expect(User).to receive(:find) { @user }
-          censor_rules = double
-          allow(@user).to receive(:censor_rules) { censor_rules }
-          censor_rule = FactoryGirl.build(:user_censor_rule, :user => @user)
-          allow(censor_rules).to receive(:build) { censor_rule }
-
-          expect(censor_rule.user).to receive(:expire_requests)
+
+        it 'calls expire_requests on the new censor_rule' do
+          allow(User).to receive(:find) { @user }
```
stevenday: Same here for @user and User.find ?

spec/controllers/admin_censor_rule_controller_spec.rb
```diff
@@ -443,6 +455,19 @@ def create_censor_rule
           expect(flash[:notice]).to eq(msg)
         end
 
+        it 'calls expire_requests on the new censor_rule' do
+          allow(PublicBody).to receive(:find) { @public_body }
```
stevenday: And again with the public body?

spec/controllers/admin_censor_rule_controller_spec.rb
```diff
@@ -609,6 +634,15 @@ def create_censor_rule
           expect(flash[:notice]).to eq(msg)
         end
 
+        it 'calls expire_requests on the censor_rule' do
+          allow(CensorRule).to receive(:find) { @censor_rule }
```
stevenday: And for all the CensorRule bits throughout this section

spec/models/public_body_spec.rb
lizconlan: have fixed so the authority is now `request.public_body` but if I just `authority.reload` rather than mocking `info_requests`, I get the request, expire is called but my test doesn't see it happen

spec/models/censor_rule_spec.rb
lizconlan: not for a method call, surely?

lizconlan: or do I need a nap? I probably need a nap

spec/integration/admin_censor_rule_spec.rb
lizconlan: \*sheepish  hadn't-worked-out-what-let!-does-yet face\* (thank you)

spec/controllers/admin_censor_rule_controller_spec.rb
lizconlan: \*factors to replace all the `before do` assigns to `let`s\* 👍

lizconlan: > My personal preference would be to actually create all these test objects and mock less things, is that possible? 

***This has taken me a while to work out and check to see what's going on - please correct me if I have any of this wrong***

Unfortunately because this a bit arm's length, the mocking is pretty awkward...

We need to stub a method on an object (the censor rule) which is created during the test (by the call to post) so ideally we need to send in a spy. A further complication is that the censor rule is built via a call on the info_request but it's not using *our* info_request object, it's calling `InfoRequest.find` with the id which returns the same row of the database but it turns out each call to `find` is treated as a different object which brings us back round to stubbing `InfoRequest.find` \*sigh\*.

A different way of doing this could be to use the [deprecated `allow_any_instance_of` syntax](https://github.com/rspec/rspec-mocks#settings-mocks-or-stubs-on-any-instance-of-a-class) to avoid stubbing the call to `InfoRequest.find` but that still leaves us with something fairly messy:

```rb
censor_rule_spy = FactoryGirl.build(:info_request_censor_rule,
                                    :info_request => info_request)
mock_rules = double()
allow(mock_rules).to receive(:build).and_return(censor_rule_spy)
allow_any_instance_of(InfoRequest).to receive(:censor_rules).
  and_return(mock_rules)

allow(censor_rule_spy).to receive(:expire_requests)

post :create, :censor_rule => censor_rule_params,
              :request_id => info_request.id

expect(censor_rule_spy).to have_received(:expire_requests)
```

(**note:** I've renamed `censor_rule` as `censor_rule_spy` in the code in an attempt to make the intention a bit clearer.)

lizconlan:
## Outstanding issues/questions for reviewers

### Can the spec be any cleaner?
https://github.com/mysociety/alaveteli/pull/3968#discussion_r119888789 (hidden above as it registers as "outdated")
This is the best that I could do - have I missed something?

### Would this be better as a context?
https://github.com/mysociety/alaveteli/pull/3968#discussion_r118053426

### Is this doing the right thing?
https://github.com/mysociety/alaveteli/pull/3968#discussion_r118052606
I was trying to minimise the amount of stuff that got expired but this may be a mistake

Oh and I've rebased so that it makes sense and faked the author date on the Refactor commit to trick GitHub into show commits in a good reading order.

spec/models/public_body_spec.rb
```diff
@@ -140,6 +140,19 @@
 
   end
 
+  describe '#expire_requests' do
+
+    it 'calls expire on all associated requests' do
```
lizconlan: Great, I'll adopt that! That last points seems particularly important

PR #3974, opened 2017-05-11
[#3971] Always render HTML in flash messages

### LINKED ISSUE #3971: Classifying own requests results in unrendered HTML
Partially successful:

ss 2017-05-10 at 09 02 10

Happens in Firefox and in Chrome. 

As noted elsewhere, classifying as Successful results in "502 Bad Gateway". "Refused" seems to work OK. "not held" results in:

ss 2017-05-10 at 09 06 25

"Waiting for an internal review" results in raw HTML, and I'm stopping checking the rest now.

cheers

garethrees:
Since Rails 4.1 flash messages have been preventing HTML rendering. This
always renders the flash without escaping tags. We should ensure that
all user input is escaped before adding it to the flash.

Fixes #3971

### DIFF
--- a/app/views/general/_responsive_header.html.erb
+++ b/app/views/general/_responsive_header.html.erb
@@ -38,10 +38,10 @@
   <div class="wrapper">
     <div class="homepage-flash">
       <% if flash[:notice] %>
-        <div id="notice"><%= flash[:notice] %></div>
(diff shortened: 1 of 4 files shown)


garethrees:
I _think_ this is safe to do, but a couple more eyes would be appreciated! :)

Issue #216, opened 2011-09-29
Improve request readability by collapsing uninteresting responses


sebbacon:
Fold up request pages more by default - don't show known acknowledgements until you click and some (javascript) expands them.

At the same time, make the fold/unfold of quoted sections to be javascripty as well.

PR #3975, opened 2017-05-11
[Pro] Add a use_notifications flag and disable emails for requests with it

stevenday:
Step one of #216, this adds a new `use_notifications` field to InfoRequest, a
new role `notifications_tester` and then sets `use_notifications` to true on
requests in notifications testers' batches.

Each type of basic request notification/reminder email is then disabled for
requests where use_notifications is true.

Requires a migration, a run of `rake db:seed` and a run of
`rake temp:set_use_notifications` to initialise it to `false` on all existing
requests.

For mysociety/alaveteli-professional#272

### DIFF
--- a/db/migrate/20170509210708_add_use_notifications_to_info_request.rb
+++ b/db/migrate/20170509210708_add_use_notifications_to_info_request.rb
@@ -1,0 +1,5 @@
+class AddUseNotificationsToInfoRequest < ActiveRecord::Migration
+  def change
+    add_column :info_requests, :use_notifications, :boolean
+  end
+end
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -8,8 +8,10 @@
 #   Mayor.create(:name => 'Daley', :city => cities.first)
 include AlaveteliFeatures::Helpers
 
-if Role.where(:name => 'admin').empty?
-  Role.create(:name => 'admin')
+['admin', 'notifications_tester'].each do |role_name|
+  if Role.where(:name => role_name).empty?
+    Role.create(:name => role_name)
+  end
 end
 
 if feature_enabled?(:alaveteli_pro)
--- a/spec/models/info_request_spec.rb
+++ b/spec/models/info_request_spec.rb
@@ -3779,4 +3779,49 @@ describe InfoRequest do
     end
   end
 
+  describe "setting use_notifications" do
+    context "when user is a notifications tester and it's in a batch" do
+      it "sets use_notifications to true" do
+        user = FactoryGirl.create(:notifications_tester_user)
+        public_bodies = FactoryGirl.create_list(:public_body, 3)
+        batch = FactoryGirl.create(:info_request_batch, user: user, public_bodies: public_bodies)
+        batch.create_batch!
+        info_request = batch.info_requests.first
+        expect(info_request.use_notifications).to be true
+      end
+    end
+
+    context "when user is a notifications tester and it's not in a batch" do
+      it "sets use_notifications to false" do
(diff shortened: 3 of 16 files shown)


stevenday:
@crowbot - I think this is ready for a review now

spec/models/info_request_spec.rb
```diff
@@ -3642,4 +3642,49 @@ def email_and_raw_email(opts = {})
...
+    context "when user is not a notification tester and it's not in a batch" do
+      it "sets use_notifications to false" do
+        info_request = FactoryGirl.create(:info_request)
+        expect(info_request.use_notifications).to be false
+      end
+    end
+
+    context "when the request has a value set manually" do
+      it "doesn't override it" do
+        info_request = FactoryGirl.create(:use_notifications_request)
```
stevenday: Whoops, will re-arrange

db/migrate/20170509210708_add_use_notifications_to_info_request.rb
```diff
@@ -0,0 +1,5 @@
+class AddUseNotificationsToInfoRequest < ActiveRecord::Migration
+  def change
+    add_column :info_requests, :use_notifications, :boolean
```
stevenday: I had that initially, but it means that we can't detect the difference between someone *forcing* it to false and rails initialising it with the database default, which in turn means we can't tell [whether to apply our own initialisation or not](https://github.com/mysociety/alaveteli/pull/3975/files#diff-5f7dfe1185cac97cbc9f9dc356abce7cR1636).

crowbot: I don't _think_ a database default would affect that as it would only take effect once the record is created. But I don't think setting a default really adds much here. I was just trying to avoid nulls because of the confusing thing where a boolean comparison with null at the database level will always return false.

stevenday: > I don't think a database default would affect that as it would only take effect once the record is created.

Hmm, I did spend a bit of time trying to make it work, but it did seem that rails was also defaulting it. I'll try applying a migration again over this and see what happens. I agree it would make things clearer all round.

crowbot: Hmm...looks like [you're right](http://blog.endpoint.com/2014/02/rails-activerecord-with-database-column.html) - the defaults affect initialisation.

stevenday:
@crowbot - thanks for the review - I've fixed up all these issues and solved the merge conflicts with develop, so I think we're good to go now. One thought I've just had regarding the default value would be to set a db default, reset the initialised value to nil in `set_defaults` and *then* check again in the `before_create` hook based on the user/request data. Do you think that's worth the hassle?

db/seeds.rb
```diff
@@ -13,7 +13,7 @@
 end
 
 if feature_enabled?(:alaveteli_pro)
-  ['pro', 'pro_admin'].each do |role_name|
+  ['pro', 'pro_admin', 'notifications_tester'].each do |role_name|
```
crowbot: Doesn't this need to go above with `:admin` ? It gets checked in code whether or not pro is enabled. Or is the `try` in that check to account for the fact that pro might not be enabled?

stevenday: fixed in 0e0e027

stevenday:
> Maybe we should just add a scarier changelog note that says that people will silently stop getting notifications unless they run bundle exec rake temp:set_use_notifications

I've done that now in ee1bb18, I'll just wait for travis to double check things, then I think I can merge this?

PR #3976, opened 2017-05-11
Use hw.physicalcpu to get cpu count for Mac OS

lizconlan:
Vagrant kept setting my number of cpus as 4 instead of 2 (which
VirtualBox then complains about) as `hw.ncpu` counts a hyperthreaded
core as 2 processors. `hw.physicalcpu` should return the number of
physical cores instead.

Syntax borrowed from https://coolaj86.com/articles/get-a-count-of-cpu-cores-on-linux-and-os-x/

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -116,7 +116,7 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
     host = RbConfig::CONFIG['host_os']
     # Give VM access to all cpu cores on the host
     if host =~ /darwin/
-      cpus = `sysctl -n hw.ncpu`.to_i
+      cpus = `sysctl -n hw.physicalcpu`.to_i
     elsif host =~ /linux/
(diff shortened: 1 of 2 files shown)


garethrees:
Ah interesting! I noticed the same warning just yesterday. I had thought that the total count including hyperthreading was the correct thing to do, but perhaps not! I did switch it to `2` manually but it didn't seem to make much difference to `time bundle exec rake environment`, but I think this is probably the correct thing to do. Will leave it for others to have a bit of a deeper look in to though.

lizconlan:
I'd seen my box try to do this recently and didn't like it (my attention was drawn by the fan noise)

screen shot 2017-05-08 at 23 36 11

It peaked at over 300% before I managed to shut it down but didn't get the screengrab. Possibly brought on by Spring doing clever stuff with `Process.fork` to maximise the use of CPUs I've told it to go ahead and use but it doesn't actually have.

PR #3980, opened 2017-05-12
[Pro] Fix new response links for pro users

stevenday:
This should hopefully fix the issue with pro users not getting a login prompt
for new responses on their embargoed requests. I didn't limit it to embargoed
requests only because I wanted to ensure pros always get the same experience
from the emails.

This is just copied from what other links which require a logged in user do,
so I might have missed some nuance, but it seems to work ok for me.

For mysociety/alaveteli-professional#222

### DIFF
--- a/app/mailers/request_mailer.rb
+++ b/app/mailers/request_mailer.rb
@@ -104,7 +104,6 @@ class RequestMailer < ApplicationMailer
       :uri => respond_to_last_url(info_request, :anchor => "followup"),
       :user_id => user.id)
     post_redirect.save!
-    url = confirm_url(:email_token => post_redirect.email_token)
 
     @url = confirm_url(:email_token => post_redirect.email_token)
     @info_request = info_request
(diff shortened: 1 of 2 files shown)


stevenday:
> This is just copied from what other links which require a logged in user do,
so I might have missed some nuance

So, not quite a nuance as much as quite a major issue!

@crowbot - I've done as you suggested in c9ec05a7facbf71975bef2e462989b944fd01986 (obviously I'll need to reword the commit too when I squash it). I've also been through the other emails and noted which ones currently send a direct login link:

- overdue alerts (links to the response form)
- very overdue alerts (links to the response form too)
- new response reminders (links to the describe state form)
- not clarified reminders (links to the response form)
- batch sent notifications (links to the batch page)
- the user's tracks digest (links to the user's profile)
- Obviously mails like the confirmation of a user's account/password changes

I'm minded to think that it's not worth working through all of these now and adding exceptions for pro users, but rather to wait until we redo pro notifications in the next few planned tickets and bear this in mind there instead?

PR #3982, opened 2017-05-12
Fix the load-sample-data script

### LINKED ISSUE #3977: The load-sample-data script breaks in Rails 4.1
The `/dev/stdin <<END` syntax we were no longer works and produces this instead 

```
$ bundle exec ./script/load-sample-data
/home/vagrant/bundle/ruby/1.9.1/gems/railties-4.1.16/lib/rails/commands/runner.rb:60:in `load': No such file or directory - /tmp/sh-thd-849333849 (deleted) (Errno::ENOENT)
	from /home/vagrant/bundle/ruby/1.9.1/gems/railties-4.1.16/lib/rails/commands/runner.rb:60:in `<top (required)>'
	from /home/vagrant/bundle/ruby/1.9.1/gems/railties-4.1.16/lib/rails/commands/commands_tasks.rb:128:in `require'
	from /home/vagrant/bundle/ruby/1.9.1/gems/railties-4.1.16/lib/rails/commands/commands_tasks.rb:128:in `require_command!'
	from /home/vagrant/bundle/ruby/1.9.1/gems/railties-4.1.16/lib/rails/commands/commands_tasks.rb:95:in `runner'
	from /home/vagrant/bundle/ruby/1.9.1/gems/railties-4.1.16/lib/rails/commands/commands_tasks.rb:40:in `run_command!'
	from /home/vagrant/bundle/ruby/1.9.1/gems/railties-4.1.16/lib/rails/commands.rb:17:in `<top (required)>'
	from bin/rails:9:in `require'
	from bin/rails:9:in `<main>'
Loaded fixtures. You may now wish to run ./script/update-xapian-index
```

(AFAICT it looks to see whether `/dev/stdin` is a file, decides that it is, tries to open it and falls over)

Someone else has fallen over the same thing and suggested a fix https://github.com/rails/rails/pull/26343 but it doesn't seem to have made it into the Rails codebase.

lizconlan:
* Sets a load order for the fixture files to avoid foreign key constraints
* Uses a temporary file for the loading script rather than using the /dev/stdin hack as that doesn't appear to work anymore (rails/rails#26343)
* Updates no_constraint_disabling and uses it in the data loader script

Fixes #3977

### DIFF
--- a/doc/CHANGES.md
+++ b/doc/CHANGES.md
@@ -56,6 +56,9 @@
   disappear without warning!
 * Prevent Vagrant assigning more CPU cores than VirtualBox recognises (Liz
   Conlan)
+* Improvements to the `load-sample-data` script to make it possible to run
+  without the superuser db permissions and for Rails 4.1 compatibility
+  (Liz Conlan)
 
 ## Upgrade Notes
 
--- a/lib/no_constraint_disabling.rb
+++ b/lib/no_constraint_disabling.rb
@@ -12,10 +12,12 @@ require 'active_record/connection_adapters/postgresql_adapter'
 module ActiveRecord
   module ConnectionAdapters
     class PostgreSQLAdapter < AbstractAdapter
-      def disable_referential_integrity(&block)
-        transaction {
-          yield
-        }
+      module ReferentialIntegrity
+        def disable_referential_integrity
+          transaction {
+            yield
+          }
+        end
       end
     end
   end
--- a/script/load-sample-data
+++ b/script/load-sample-data
@@ -5,7 +5,8 @@
 # have a filesystem representation of their contents
 
 export LOC=`dirname "$0"`
-bundle exec rails runner /dev/stdin <<END
+
+cat > "tmp/data-loader.rb" <<EOF
 require 'no_constraint_disabling'
 require 'rspec/rails'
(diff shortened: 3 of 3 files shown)


lizconlan:
Spent some time trying to work around the `RI_ConstraintTrigger_a_16410` type problems when loading data (this [loosely translates](http://blog.endpoint.com/2012/10/postgres-system-triggers-error.html) to `ERROR:  Cannot disable triggers on a table containing foreign keys unless superuser`) - aka the db and its constraints have different owners - but only found a few `SET CONSTRAINTS DEFERRED` workarounds, which don't fit our use case.

It's a lovely idea [explained (with code) here](http://kopongo.com/2008/7/25/postgres-ri_constrainttrigger-error) and [there's even a plugin for it](https://github.com/greyblake/rails3_pg_deferred_constraints/blob/master/lib/rails3_pg_deferred_constraints/postgresql_adapter.rb) but it turns out our constraints aren't [`DEFERRABLE`](http://stackoverflow.com/a/28681238) and that foreign key constraints are `NOT DEFERRABLE` by design ([gory details here](http://stackoverflow.com/a/10035119)), presumably to prevent non superusers from circumventing data integrity mechanisms.

PR #3988, opened 2017-05-15
[Pro] Check that a body is requestable before previewing/creating requests

stevenday:
This adds the same `is_requestable?` checking that the existing request
controller does. I note from testing that we still seem to allow people to
select defunct bodies in the search, but I thought we'd fixed this already?

For mysociety/alaveteli-professional#228

### DIFF
--- a/app/controllers/alaveteli_pro/public_bodies_controller.rb
+++ b/app/controllers/alaveteli_pro/public_bodies_controller.rb
@@ -3,8 +3,12 @@ class AlaveteliPro::PublicBodiesController < AlaveteliPro::BaseController
   def search
     query = params[:query] || ""
     xapian_results = typeahead_search(query, :model => PublicBody,
-                                             :exclude_tags => [ 'defunct' ])
+                                             :exclude_tags => [ 'defunct',
+                                                                'not_apply' ])
     results = xapian_results.present? ? xapian_results.results : []
+    # Exclude any bodies we can't make a request to (in addition to the ones
+    # we've already filtered out by the excluded tags above)
+    results.select! { |result| result[:model].is_requestable? }
     # Xapian's results include things we don't want to publish, like the
     # request email and api_key, so we map these results into a simpler object
     # with only some whitelisted attributes.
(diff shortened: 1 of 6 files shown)


stevenday:
> What do you think about the approach of excluding these results in public_bodies_controller? 

So I guess the initial UX question is whether it's better to see that we know who you're looking for, (but then be told you can't send a request to them), or to never be able to find them in the first place? I think the ideal situation is that they're there, but are disabled, or maybe that as soon as you select them, you see a message that tells you why you can't continue. 

Maybe the message should be different for pro users, but I just re-read it and I don't think it's that bad - at the end of the day it's an acceptance of the fact that the "public body" landscape is complicated and ever-changing :) 

I have no idea how big a problem it is though, so maybe as a compromise/starter, rendering that page should trigger some kind of alert for pro admins to see, so we can measure it?

> We could add not_apply to the :exlude_tags param to typeahead_search, and remove non requestable bodies from the results there to cover those that don't have a request email address. What do you think?

Aside from the question above, I guess there's always the chance that things will change between selecting a body and submitting the form (especially with drafts maybe taking a while to write). So I think whatever happens we'll still need some kind of check like this to avoid an error when sending?

stevenday:
Because there's an unfinished debate here, I'm going to park this until @crowbot is back from holidays.

stevenday:
@crowbot - I've made those changes as suggested and added a couple of specs for the public body searching, so I think this is now good to go?

app/controllers/alaveteli_pro/public_bodies_controller.rb
```diff
@@ -3,7 +3,8 @@ class AlaveteliPro::PublicBodiesController < AlaveteliPro::BaseController
   def search
     query = params[:query] || ""
     xapian_results = typeahead_search(query, :model => PublicBody,
-                                             :exclude_tags => [ 'defunct' ])
+                                             :exclude_tags => [ 'defunct',
+                                                                'not_apply' ])
```
crowbot: I think in order to exclude authorities with no contact address from the results presented in Pro, we'd need to remove them from the xapian results returned in the `results.map!` call below. We could do this by removing anything that responds `false` to `is_requestable?`.

stevenday: Done in fa8eaf2 which also fixes an oversight in the specs that meant we weren't really testing the search results properly.

crowbot: Yep, looks good to me - ready for deploy! :shipit:

PR #3991, opened 2017-05-15
Make Conditionallyprependlocale use its own copy of the class variables

### LINKED ISSUE #3957: Locales with underscores no longer accepted in the URL
e.g. `/nl_BE` doesn't work but `/nl-BE` does

(behaviour apparently introduced in Rails 4.0.0)

Given a config with underscore locales (e.g. the Belgian theme) the language switcher puts the language name in the dropdown but the URL is for the default locale (e.g. `/`)

lizconlan:
Not entirely sure what changed in Rails 4.0 to cause this (by the looks of things, something in `ActiveSupport`) but our custom `RoutingFilter` which inherits from `RoutingFilter::Locale` ([supplied by the routing-filter gem](https://github.com/svenfuchs/routing-filter/blob/0-4-stable/lib/routing_filter/filters/locale.rb)) wasn't updating `locales` or `locales_pattern` properly, letting the default I18n values (e.g. fr-BE) remain set rather than our FastGettext/POSIX friendly versions (e.g. fr_BE).

Suggestion - should this be a hotfix for 0.28?

Fixes #3957

### DIFF
--- a/lib/routing_filters.rb
+++ b/lib/routing_filters.rb
@@ -1,6 +1,8 @@
 # -*- encoding : utf-8 -*-
 module RoutingFilter
   class Conditionallyprependlocale < RoutingFilter::Locale
+    cattr_accessor :locales
+
     # Override core Locale filter not to prepend locale path segment
     # when there's only one locale
     def prepend_locale?(locale)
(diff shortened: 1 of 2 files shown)


stevenday:
I'm a bit confused by this, I don't think my Ruby foo is strong enough to see the link between the `cattr_accessor` and the other overrides we've added to our own filter. Maybe we can go over this together tomorrow?

lizconlan:
Closed in favour of [hotfix/0.28.0.7](https://github.com/mysociety/alaveteli/tree/hotfix/0.28.0.7)

PR #3992, opened 2017-05-15
Deprecate Ubuntu Precise  in favour of Trusty

### LINKED ISSUE #3928: Update Travis CI config for Precise EOL
https://blog.travis-ci.com/2017-04-17-precise-EOL

lizconlan:
Fixes #3928 

Of interest - there is a ruby2.0 package (and ruby2.0-dev) available for Trusty (included in trusty-updates), not installed here by default but we could include it when we drop 1.9 support


lizconlan:
Note to self - add CHANGELOG note

lizconlan:
> I wonder if this aspect of it is worth waiting for @garethrees or @crowbot to discuss?

Happy to wait, it's not a critical change. It looks like we're pretty close to being able to move to Ruby 2 so we could potentially roll that in at the same time with a packages file rather than making people rebuild boxes twice. (I've been running on Trusty for the past few weeks and it seems fine in terms of dev features.)

garethrees:
I left the default as Ubuntu because that was what the Vagrantfile used at the time I introduced switchable versions.

I think that given we're now changing the default, we should change it to our production default (`wheezy64`) and have Ubuntu available as an option.

Vagrantfile
```diff
@@ -60,7 +60,7 @@
 ALAVETELI_FQDN = ENV['ALAVETELI_VAGRANT_FQDN'] || "alaveteli.10.10.10.30.xip.io"
 ALAVETELI_MEMORY = ENV['ALAVETELI_VAGRANT_MEMORY'] || 1536
 ALAVETELI_THEMES_DIR = ENV['ALAVETELI_THEMES_DIR'] || '../alaveteli-themes'
-ALAVETELI_OS = ENV['ALAVETELI_VAGRANT_OS'] || 'trusty64'
+ALAVETELI_OS = ENV['ALAVETELI_VAGRANT_OS'] || 'wheezy64'
```
garethrees: This doesn't update the changelog, which states that "trusty" is being made default.

PR #3993, opened 2017-05-16
Fix transient spec failure

lizconlan:


stevenday:
@lizconlan - Sorry, this is a flakey spec, but it's not one that match_array can solve, as it's comparing two strings (which are built out of query results that vary in order). I think @crowbot should have fixed this in https://github.com/mysociety/alaveteli/commit/e6545ed4726467e9aa9e61ae17c73128511e5579

lizconlan:
@stevenday - cool! Was just letting it run through to Travis as I've broken my test box \*sigh\*

PR #3994, opened 2017-05-16
[Pro] Add a Notification model and use it for new response notifications

stevenday:
This does the following:

- Adds a new `Notification` model. Notifications can be instant or daily and set a sending timestamp on themselves as appropriate, so that we can easily query for the ones we need to send at any given time.
- Adds a new `User#notify` method for "notifying" users of events. The user can control when they want to receive the notification, although for now there's no logic for it, I just use the notifications_tester role to set everything to daily. 
- Adds new hour/minute fields to `Users` to store when they should be sent daily summaries. My thinking here is that we might want to spread this out over the day eventually, because otherwise we'll be sending lots of emails at one specific time, so this is one way to control that.
- Moves the mailer stuff for new responses from `RequestMailer` to a new `NotificationMailer`. I didn't want to do this, I initially thought I could just proxy to it, but Rails' mailers don't work like that because they do weird things to process mail (allowing you to call instance methods as class methods) which results in the NotificationMailer returning a `NullMail` object instead of the mail object (because `mail` wasn't called in the instance of the NotificationMailer class you're interacting with). I added some more specs of this because I was a bit nervous about changing it, so please scrutinise that bit especially :)
- Adds an `NotificationMailer#instant_notification` mailer method which proxies to the right internal method depending on the type of notification, so that we can send notifications about any kind of event using a single method. Right now only new response events are implemented, but eventually this will allow us to port all the other mails over to the notification system in a similar way
- Adds a `NotificationMailer#daily_summary` method and associated views, which bring all of the notifications for a user into a single mail, grouped by the request they relate to.
- Adds class methods to `NotificationMailer` to actually send all the notifications that need sending at any given point.
- Adds the necessary scripts and crontab entries to call these class methods regularly.

### Questions
- [ ] I've plucked some cron timings out of fairly thin air. Does anyone have suggestions for good ones for WDTK in particular, or better defaults for the example file?
- [ ] Have I missed any knock-on effects on other parts of the codebase? I've thought about the `UserInfoRequestSentAlerts`, but this doesn't affect them right now. Eventually though, it seems like a seen notification could replace a record in that table and perhaps reduce a lot of the logic around sending those alerts? I don't think we actually record the sending of a message about a new response anywhere at the moment, so I don't think there's anything to update regarding that.
- [x] One day the Notifications table will get very big, is that likely to be a problem? Have I missed an index that could help?
- [x] I've gone for `seen_at` not `sent_at` because I'm thinking that one day we might offer a "no emails" notification option where you can see them on a page on the site instead ala github. Premature optimisation?
- [x] Is tying this to an `InfoRequestEvent` going to bite us in the future? Are there notifications we send out which aren't related to events?

For mysociety/alaveteli-professional#273


stevenday:
I've requested a review from @crowbot - but @garethrees might want to weigh in on my questions too.

lib/tasks/temp.rake
```diff
@@ -323,4 +323,9 @@ namespace :temp do
   task :set_use_notifications => :environment do
     InfoRequest.update_all use_notifications: false
   end
+
+  desc 'Set a default time for users daily summary notifications'
+  task :set_daily_summary_times => :environment do
+    User.update_all daily_summary_hour: 8, daily_summary_minute: 30
```
crowbot: How do daily summary times get set for users created after this temp cron job is run?

app/views/notification_mailer/info_request_batches/_info_request_batch.text.erb
```diff
@@ -0,0 +1,14 @@
+<%= info_request_batch.title %>
```
crowbot: I think some underscores or equals signs under the names of requests or batches would make this email easier to read.

app/models/notification.rb
```diff
@@ -4,4 +4,22 @@ class Notification < ActiveRecord::Base
   belongs_to :user
 
   validates_presence_of :info_request_event, :user, :frequency, :send_after
+
+  before_validation :calculate_send_after
+
+  INSTANTLY = 'instantly' # Notify me about this as soon as possible
+  DAILY = 'daily' # Notify me about this at most once a day
```
stevenday: I chose a string because the db stores it the `frequency` as a string and I didn't want to have to remember to `to_sym` or `to_s` it everywhere we need to compare the two. Is there a trick I'm missing to make that painless though (without using a gem like symbolize?).

Ah - I guess I should use http://api.rubyonrails.org/v4.1/classes/ActiveRecord/Enum.html now that we're on 4.1

stevenday: I've done this in 28dc63d and I'll just remove 284501d when I squash everything down.

Note to test this, you'll need to run `rake db:rollback STEP=2` in both your test and dev dbs *before* you pull my changes, and then pull and run `rake db:migrate`.

app/models/notification.rb
```diff
@@ -15,6 +15,14 @@ def self.unseen
     where(seen_at: nil)
   end
 
+  def self.instantly
+    where(frequency: INSTANTLY)
+  end
+
+  def self.daily
+    where(frequency: DAILY)
+  end
```
stevenday: These are no longer required because of me switching to use an Enum, because that adds them automatically. I'll remove 284501d when I squash this to get rid of them.

stevenday: I've temporarily added a revert commit in 162947e, otherwise the tests fail.

app/models/notification.rb
```diff
@@ -11,6 +11,10 @@ class Notification < ActiveRecord::Base
   DAILY = 'daily' # Notify me about this at most once a day
   FREQUENCIES = [INSTANTLY, DAILY].freeze
 
+  def self.unseen
+    where(seen_at: nil)
+  end
```
stevenday: Fixed in b281694

spec/mailers/request_mailer_spec.rb
```diff
@@ -16,15 +16,10 @@
       receive_incoming_mail('incoming-request-plain.email', ir.incoming_email)
       expect(ir.incoming_messages.size).to eq(2) # one more arrives
       expect(ir.info_request_events[-1].incoming_message_id).not_to be_nil
-
-      deliveries = ActionMailer::Base.deliveries
-      expect(deliveries.size).to eq(1)
-      mail = deliveries[0]
-      expect(mail.to).to eq([ 'bob@localhost' ]) # to the user who sent fancy_dog_request
-      deliveries.clear
```
stevenday: In the cases where I've removed these, I think the only mail being cleared was the new response notification I've now removed. I'll double check though.

spec/models/info_request_spec.rb
```diff
@@ -307,32 +307,30 @@
...
-      it 'emails the user that a response has been received' do
+      it 'notifies the user that a response has been received' do
         info_request = FactoryGirl.create(:info_request)
         email, raw_email = email_and_raw_email
+        user = info_request.user
+
+        # Without this, the user retrieved in the model is not the same one
+        # in memory as this one, so we can't put expectations on it
+        allow(info_request).to receive(:user).and_return(user)
+        expect(user).
```
stevenday: With the mocking still in place this can work, i.e.:

```diff
-        user = info_request.user

         # Without this, the user retrieved in the model is not the same one
         # in memory as this one, so we can't put expectations on it
+        allow(info_request).to receive(:user).and_return(info_request.user)
+        expect(info_request.user).
```

Is that what you meant? I'm happy to make that change if it's clearer, I've pushed it as 7cd9b3995f1b1cb1072752a10c9150b414eb9187

spec/mailers/request_mailer_spec.rb
stevenday: I went through them all and in every case the tests were either expecting the mailbox to be empty or contain one request - I can reinstate the former expectation if you think it's going to be a problem, as at least that way we'd see when something bad happens, but I don't *think* these changes should cause a problem.

spec/models/notification_spec.rb
```diff
@@ -79,4 +79,22 @@
...
+    let(:daily_notification) { FactoryGirl.create(:daily_notification) }
+
+    it "only returns instant notifications" do
+      expect(Notification.instantly).to match_array([instant_notification])
+    end
+  end
+
+  describe ".daily" do
+    let(:instant_notification) { FactoryGirl.create(:instant_notification) }
+    let(:daily_notification) { FactoryGirl.create(:daily_notification) }
```
stevenday: This is now obsoleted by 162947e

app/views/alaveteli_pro/info_request_batches/_authority_list.text.erb
```diff
@@ -0,0 +1,11 @@
+<% if public_bodies.size %>
+<% example_body_names = public_bodies.take(2).map(&:name) %>
+<%= n_('{{count}} recipient, {{first_public_body}}',
```
stevenday: Ah - so I just read through https://github.com/mysociety/alaveteli/pull/3443 and realised what's going on here. I initially assumed that rails wouldn't escape this because it was in a text file (due to https://github.com/rails/rails/pull/8235), but then I saw that it was when I tested it in the console. So I *think* this is actually ok, thanks to the magic in our translation helpers: https://github.com/mysociety/alaveteli/blob/develop/lib/i18n_fixes.rb#L39

*However* when we use this in a text email, we'll actually want to mark the body names as `html_safe` so that they *don't* get escaped (along with a variety of other things we pass into translation strings like the site name, law name, etc). Given that, I think I should make the text and html versions of this partial separate, and explicitly mark these body names as html_safe in the text one.

stevenday: I've fixed this issue up in c2824bb, 9188ab3, e57c020 and 071e57a. In the end, I duplicated the code from this view back into the html version, with appropriate calls to `raw` or `html_safe` in the text version. I also went through all the text email templates and made sure (with lots of new specs) that they were escaping everything they should.

app/views/notification_mailer/info_request_batches/_info_request_batch.text.erb
stevenday: Done in 9c5b641 and 840cb72, with changes to the specs in 5c0745a38a23335a5e086763500d4956d2b887bf

app/views/notification_mailer/daily_summary.text.erb
```diff
@@ -0,0 +1,10 @@
+<%= _("Here's your daily request summary from {{pro_site_name}}",
+      pro_site_name: AlaveteliConfiguration.pro_site_name) %>
+
+<% @grouped_notifications.each do |request, notifications| %>
+<%= render :partial => "notification_mailer/#{request.to_partial_path}",
+           :locals => { request.class.model_name.param_key.to_sym => request,
+                        :notifications => notifications } %>
+<% end %>
+
+-- <%= _('the {{site_name}} team', :site_name => site_name.html_safe) %>
```
stevenday: Yep, fair enough - I'll squash that down later.

app/views/notification_mailer/info_request_batches/_info_request_batch.text.erb
```diff
@@ -0,0 +1,14 @@
...
+<%= render partial: 'alaveteli_pro/info_request_batches/authority_list',
+           locals: { public_bodies: info_request_batch.public_bodies } %>
+<%= render partial: 'alaveteli_pro/info_request_batches/progress_bar',
+          locals: { phases: info_request_batch.request_phases_summary,
+                    batch: info_request_batch } %>
+<%= _("What's happened:") %>
+<% notifications_by_event_type = notifications.group_by { |n| n.info_request_event.event_type } %>
```
stevenday: Done in 8940262 as a separate commit for introducing the helper and 0a088e7 to fix this up to use it. I'll shuffle the commits around when I squash this down.

script/send-daily-notifications
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
```
stevenday: Done in f6e2651

script/send-instant-notifications
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
```
stevenday: Done in 107dba3

app/models/user.rb
```diff
@@ -593,6 +595,39 @@ def for_admin_column(complete = false)
...
+      info_request_event: info_request_event,
+      frequency: self.notification_frequency,
+      user: self
+    )
+  end
+
+  # Return a timestamp for the next time a user should be sent a daily summary
+  def next_daily_summary_time
+    summary_time = Time.zone.now.change(self.daily_summary_time)
+    summary_time + 1.day if summary_time < Time.zone.now
```
stevenday: Fixed in 51220dc

spec/models/user_spec.rb
```diff
@@ -1152,4 +1152,88 @@
...
+      end
+    end
+
+    context "when the time is in the past" do
+      let(:expected_time) { Time.zone.now.change(hour: 7, min: 56) + 1.day }
+
+      it "returns tomorrow's date with the daily summary time set" do
+        time_travel_to(expected_time + 1.minute) do
+          expect(user.next_daily_summary_time).
+            to be_within(1.second).of(expected_time)
```
stevenday: Whoops, thanks! Fixed now in 51220dc

lib/tasks/temp.rake
stevenday: Hmm, good point! I've added some initialisation in dde11e3 copying what we already do for alert timings (setting a random time). I've also added 87673b5 to make the rake task work like this too, so that it sets a random hour and minute for each user. However, to make this fast I did it in SQL which I've now realised is Postgres-specific. Let me know if this is an issue.

spec/mailers/request_mailer_spec.rb
crowbot: What you've done sounds good to me, thanks!

stevenday:
Closing in favour of #4043 and a separate pull request for new response notifications for other users

PR #3998, opened 2017-05-17
Prevent autoreplies causing withdrawn requests to require classification

### LINKED ISSUE #3939: Users cannot set "withdrawn" status if authority has autoresponder
Setting a request as "I would like to withdraw this request" results in the user being presented with a follow-up form so they can inform the authority that they wish to withdraw the request. The request is only set as "User withdrawn" when this form has been completed.
If the authority replies to the message, e.g. with an autoresponder, the user is again asked to classify the response. They are not offered the option to set the request as "user withdrawn" - they can do so through the Actions menu but then they have to send another message to the authority...

lizconlan:
As withdrawing a request emails the authority, an automated response
would automatically set the request back to awaiting_description

Fixes #3939

### DIFF
--- a/app/models/info_request.rb
+++ b/app/models/info_request.rb
@@ -1669,7 +1669,9 @@ class InfoRequest < ActiveRecord::Base
       raw_email.data = raw_email_data
       raw_email.save!
 
-      self.awaiting_description = true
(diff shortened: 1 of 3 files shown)


lizconlan:
Note to self - add CHANGELOG note

PR #4004, opened 2017-05-19
Ensure the admin link for misdelivered messages has text

### LINKED ISSUE #2689: Messages without body text are unclickable in the 'misdelivered responses' section of the admin summary page
We use `excerpt(message.get_body_for_quoting, "", :radius => 60)` to display the text of the link generated here. If there's no body text, this is empty. We should add a fallback.

lizconlan:
As there is already a guard for empty messages, it could be
that unicode spaces are causing blank text links to be created.

Possibly related to a known bug in String.strip that affects Ruby 1.9.2
and 1.9.3 - https://bugs.ruby-lang.org/issues/7845

Fixes #2689

### DIFF
--- a/app/views/admin_general/index.html.erb
+++ b/app/views/admin_general/index.html.erb
@@ -35,7 +35,7 @@
             <% for message in @holding_pen_messages %>
               <tr>
                 <td>
-                  <% if message.get_body_for_quoting.strip.size == 0 %>
+                  <% if message.get_body_for_quoting.gsub(/[[:space:]]/, "").size == 0 %>
                     <%= link_to "(no body)", admin_raw_email_path(message.raw_email_id) %>
                   <% else %>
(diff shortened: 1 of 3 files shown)


app/views/admin_general/index.html.erb
```diff
@@ -35,7 +35,7 @@
             <% for message in @holding_pen_messages %>
               <tr>
                 <td>
-                  <% if message.get_body_for_quoting.strip.size == 0 %>
+                  <% if message.get_body_for_quoting.strip.gsub(/[[:space:]]/, "").size == 0 %>
```
stevenday: is the `.strip` necessary here if we're going a `gsub` too?

lizconlan: \*fiddles around on the command line\* apparently not - good spot!

PR #4006, opened 2017-05-22
Pro/may design polish

wrightmartin:
Adds some surface level tweaks and polish to pro screens


stevenday:
Whoops, I fixed this up and merged before pushing the fixed up version here, it's done.

PR #4009, opened 2017-05-24
[#3978] Uses partials to render complex flash messages

### LINKED ISSUE #3978: Don't render HTML into flash messages
These are now (like all cookie content) serialized via json and so don't retain `html_safe` markings. Where we need to render HTML in flashes, it should be in a partial template, and the flash should just indicate which template (and params if needed) should be rendered.

lizconlan:
Fixes #3978


spec/controllers/request_game_controller_spec.rb
```diff
@@ -27,6 +27,45 @@
...
+        })
+      end
+
+      it 'displays the flash message' do
+        get :play
+        expect(response.body).
+          to include("<p>All done! Thank you very much for your help.</p>")
+
+        expect(response.body).
+          to include("There are <a href=\"#{test_url}\">more things " \
```
stevenday: Here you could use `have_link` and `have_content` if you decide to go down the capybara route.

spec/controllers/user_controller_spec.rb
```diff
@@ -1178,6 +1178,27 @@ def do_signin(email, password)
...
+      profile_photo = ProfilePhoto.
+                        create(:data => load_file_fixture("parrot.png"),
+                               :user => user)
+
+      post :set_profile_photo, { :id => user.id,
+                                 :file => @uploadedfile,
+                                 :submitted_crop_profile_photo => 1,
+                                 :draft_profile_photo_id => profile_photo.id }
+
+      expect(flash[:notice]).
```
stevenday: elsewhere you tested the partial key explicitly, rather than the whole hash, it would be good to be consistent either way (I think just checking the partial is probably good where we can and there's no logic in that partial, since testing the text makes the spec more fragile).

spec/integration/admin_spec.rb
```diff
@@ -147,4 +147,30 @@
...
+      authority_name = request.public_body.name
+      authority_email = request.public_body.request_email
+
+      using_session(@admin) do
+        visit admin_request_path :id => request.id
+        find_button('Generate URL').click
+
+        url = confirm_url(PostRedirect.last.email_token)
+
+        message = "Send \"#{authority_name}\" &lt;" \
```
stevenday: Since this is an integration spec, I'd definitely use the capybara matchers here.

app/controllers/followups_controller.rb
```diff
@@ -76,7 +76,9 @@ def check_request_matches_incoming_message
 
   def check_responses_allowed
     if @info_request.allow_new_responses_from == "nobody"
-      flash[:error] = _('Your follow up has not been sent because this request has been stopped to prevent spam. Please <a href="{{url}}">contact us</a> if you really want to send a follow up message.', :url => help_contact_path.html_safe)
+      flash[:error] = { :partial => "followup_not_sent.html.erb",
```
stevenday: I think this should be `flash.now[:error]` ? There's no redirect, so the flash will persist until the next request.

app/controllers/request_game_controller.rb
```diff
@@ -25,9 +25,11 @@ def play
               order('random()')
 
     if @missing == 0
-      flash[:notice] = _('<p>All done! Thank you very much for your help.</p><p>There are <a href="{{helpus_url}}">more things you can do</a> to help {{site_name}}.</p>',
-                         :helpus_url => help_credits_path(:anchor => "helpus"),
-                         :site_name => site_name)
+      flash[:notice] = {:partial => "game_over.html.erb",
```
stevenday: This should be `flash.now[:notice]` too I think , because there's not redirect afterwards, I added a commit but I think the specs will need a tweak too, because IIRC it's harder to test flash.now in controller specs?

spec/controllers/request_controller_spec.rb
```diff
@@ -1042,6 +1042,69 @@ def expect_hidden(hidden_template)
...
+                    "the email address"
+
+        post :new, :info_request => {
+                     :public_body_id => @body.id,
+                     :title => "Test Request" },
+                   :outgoing_message => { :body => "me@here.com" },
+                   :submitted_new_request => 1,
+                   :preview => 1
+
+        expect(response.body).to include(expected1)
```
stevenday: snap about `have_content`

spec/controllers/request_controller_spec.rb
```diff
@@ -1042,6 +1042,69 @@ def expect_hidden(hidden_template)
...
+    it 'displays an error message warning about the postcode' do
+      expected = "Your request contains a <strong>postcode</strong>"
+      post :new, :info_request => {
+                   :public_body_id => @body.id,
+                   :title => "Test Request" },
+                 :outgoing_message => { :body => "SW1A 1AA" },
+                 :submitted_new_request => 1,
+                 :preview => 1
+
+      expect(response.body).to include(expected)
```
stevenday: > snap about `have_content`

(Just marking all these places so it's easy to fix up if you want to)

spec/controllers/user_profile/about_me_controller_spec.rb
```diff
@@ -127,10 +127,8 @@
 
         it 'sets a message suggesting they add one' do
           put :update, :user => { :about_me => 'My bio' }
-          msg = "<p>Thanks for changing the text about you on your " \
-                "profile.</p><p><strong>Next...</strong> You can " \
-                "upload a profile photograph too.</p>"
-          expect(flash[:notice]).to eq(msg)
+          expect(flash[:notice]).
+            to eq({ :partial => "update_profile_text.html.erb" })
```
stevenday: Same about testing the whole hash here

spec/integration/request_controller_spec.rb
```diff
@@ -37,4 +37,50 @@
...
+
+    it 'shows a flash alert to users' do
+      expected_message = '<p>Alaveteli is currently in maintenance. You ' \
+                         'can only view existing requests. You cannot make ' \
+                         'new ones, add followups or annotations, or ' \
+                         'otherwise change the database.</p> '\
+                         '<p>Down for maintenance</p>'
+
+      visit new_request_path
+      expect(page.body).to include(expected_message)
```
stevenday: > snap about `have_content`

(Just marking all these places so it's easy to fix up if you want to)

spec/integration/request_controller_spec.rb
```diff
@@ -37,4 +37,50 @@
...
+      end
+
+      it 'shows a flash alert to users' do
+        expected_message = '<p>Alaveteli is currently in maintenance. You ' \
+                           'can only view existing requests. You cannot make ' \
+                           'new ones, add followups or otherwise change the ' \
+                           'database.</p> <p>Down for maintenance</p>'
+
+        visit new_request_path
+        expect(page.body).to include(expected_message)
```
stevenday: > snap about `have_content`

(Just marking all these places so it's easy to fix up if you want to)

spec/integration/request_game_controller_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+        choose("rejected1")
+        within "#describe_state_form_1" do
+          find_button("Submit status").click
+        end
+
+        escaped = "Awkward &gt; Title"
+        msg = "Thank you for updating the status of the request " \
+              "'<a href=\"#{request_path(request)}\">#{escaped}</a>'. " \
+              "There are some more requests below for you to classify."
+        expect(page.body).to include(msg)
```
stevenday: > snap about `have_content`

(Just marking all these places so it's easy to fix up if you want to)

spec/integration/user_profile_updates_spec.rb
```diff
@@ -0,0 +1,102 @@
...
+      it "page displays thank you message with nudge to upload photo" do
+        using_session(login(user)) do
+          msg = "<p>Thanks for changing the text about you on your " \
+                  "profile.</p><p><strong>Next...</strong> You can " \
+                  "upload a profile photograph too.</p>"
+          visit edit_profile_about_me_path
+          fill_in :user_about_me, :with => "I am a researcher"
+          click_button "Save"
+
+          expect(page.body).to include(msg)
```
stevenday: > snap about `have_content`

(Just marking all these places so it's easy to fix up if you want to)

spec/integration/user_profile_updates_spec.rb
```diff
@@ -0,0 +1,102 @@
...
+      end
+
+      it "displays a thank you message without upload photo nudge" do
+        using_session(login(user)) do
+          msg = "You have now changed the text about you on your profile."
+          visit edit_profile_about_me_path
+          fill_in :user_about_me, :with => "I am a researcher"
+          click_button "Save"
+
+          expect(page).to have_content msg
```
stevenday: > snap about `have_content`

(Just marking all these places so it's easy to fix up if you want to)

spec/integration/user_profile_updates_spec.rb
```diff
@@ -0,0 +1,102 @@
...
+                                   :user => user)
+
+          page.driver.post set_profile_photo_path,
+               :id => user.id,
+               :file => photo_file,
+               :submitted_crop_profile_photo => 1,
+               :draft_profile_photo_id => profile_photo.id
+
+          visit page.driver.response.location
+          expect(page.body).to include(msg)
```
stevenday: > snap about `have_content`

(Just marking all these places so it's easy to fix up if you want to)

app/controllers/request_controller.rb
```diff
@@ -781,7 +781,7 @@ def upload_response
       flash[:notice] = _("Thank you for responding to this FOI request! " \
                            "Your response has been published below, and a " \
                            "link to your response has been emailed to {{user_name}}.",
-                         :user_name => CGI.escapeHTML(@info_request.user.name))
+                         :user_name => @info_request.user.name.html_safe)
```
stevenday: Why does this need a `.html_safe`, but the one in the admin view doesn't?

spec/integration/request_controller_spec.rb
```diff
@@ -83,4 +83,31 @@
...
   end
 
+  describe 'FOI officer uploading a reponse' do
+
+    let(:user) { FactoryGirl.create(:user, :name => "Awkward > Name") }
+    let(:request) { FactoryGirl.create(:info_request, :user => user) }
+    let(:officer) { FactoryGirl.create(:user) }
+
+    before do
+      allow(request.public_body).
```
stevenday: I think this would be true anyway (aren't they both localhost?). Either way, since this is an integration test, I think it'd be better to set the request email and the user's email to the same domain, rather than mock things.

spec/integration/request_controller_spec.rb
```diff
@@ -83,4 +83,31 @@
...
+
+    it 'should render a message confirming the response has been published' do
+      message = "Thank you for responding to this FOI request! " \
+                "Your response has been published below, and a " \
+                "link to your response has been emailed to Awkward &gt; Name."
+      using_session(login(officer)) do
+        visit upload_response_path :url_title => request.url_title
+        fill_in(:body, :with => 'Additional information')
+        click_button("Upload FOI response")
+        expect(page.body).to include(message)
```
stevenday: `have_content` option again

spec/integration/send_user_message_spec.rb
```diff
@@ -0,0 +1,23 @@
+# -*- encoding : utf-8 -*-
+require File.expand_path(File.dirname(__FILE__) + '/../spec_helper')
```
stevenday: This can just be require `spec_helper` now I think (see pro specs for examples)

spec/integration/send_user_message_spec.rb
```diff
@@ -0,0 +1,23 @@
...
+require File.expand_path(File.dirname(__FILE__) + '/alaveteli_dsl')
+
+describe 'Sending a message to another user' do
+
+  let(:sender) { FactoryGirl.create(:user) }
+  let(:recipient) { FactoryGirl.create(:user, :name => "Awkward > Name") }
+
```
stevenday: Extra line break

app/controllers/request_game_controller.rb
```diff
@@ -25,11 +25,13 @@ def play
               order('random()')
 
     if @missing == 0
-      flash[:notice] = {:partial => "game_over.html.erb",
-                        :locals => {
-                          :helpus_url => help_credits_path(:anchor => "helpus"),
-                          :site_name => site_name }
-                        }
+      flash.now[:notice] = {
+        :partial => "request_game/game_over.html.erb",
```
stevenday: I made this specific, but it probably doesn't need to be

spec/integration/request_controller_spec.rb
lizconlan: IIRC, I switched because `have_content` didn't work with HTML in the string - I'll have another go (either escaping the HTML or using a different matcher should help - cheers for the feedback! Don't let me take the lazy option when there's a better one available :slightly_smiling_face: )

spec/controllers/request_controller_spec.rb
```diff
@@ -1042,6 +1042,69 @@ def expect_hidden(hidden_template)
...
+                    "the email address"
+
+        post :new, :info_request => {
+                     :public_body_id => @body.id,
+                     :title => "Test Request" },
+                   :outgoing_message => { :body => "me@here.com" },
+                   :submitted_new_request => 1,
+                   :preview => 1
+
+        expect(response.body).to include(expected1)
```
lizconlan: I really wanted to check that the HTML wasn't being escaped as that was the error we had before from just passing the HTML into flash with the new cookie settings

lizconlan: ...and there's probably a better way of doing that 🤔

stevenday: Ah ok, that's probably a good idea - I think the only way to do it with Capybara would be to use an xpath selector which is probably going to be harder work to understand in the future.

lizconlan:
Oh, I missed something essential - this is going to go wrong for all sorts of reasons:

https://github.com/mysociety/alaveteli/blob/0.28.0.7/app/controllers/request_controller.rb#L540-L548

(it doesn't - yet - set up the partial (easy) and it passes in an object which doesn't work with JSON cookies)

app/helpers/application_helper.rb
```diff
@@ -93,6 +93,21 @@ def cache_if_caching_fragments(*args, &block)
...
   end
 
+  def render_flash_message(flash_msg)
+    return flash_msg if flash_msg.is_a?(String)
+    msg = {}
+    # JSON deserializes the cookie with stringified keys so we need to switch
+    # them - and any contained hashes - back to symbols to work with render
+    flash_msg.keys.each do |key|
+      msg[key.to_sym] = if flash_msg[key].is_a?(Hash)
+        flash_msg[key].symbolize_keys
```
lizconlan: whut?

app/views/request/_preview_errors.html.erb
```diff
@@ -0,0 +1,24 @@
+<% if contains_email
+  if user.nil? %>
+    <%= _("<p>You do not need to include your email in the " \
+          "request in order to get a reply, as we will ask " \
+          "for it on the next screen (<a href=\"{{url}}\">" \
+          "details</a>).</p>",
+          :url => (help_privacy_path(:anchor => "email_address")).html_safe) %>
```
lizconlan: looks like I can get away with passing that in, that will be simpler

spec/controllers/request_controller_spec.rb
lizconlan: this is nicer > `expect(…).to_not have_content('&lt;p&gt;')`

app/controllers/request_controller.rb
lizconlan: I suspect because I got the spec for the admin view wrong /o\

lizconlan: yeah, I inadvertently check that it **had** escaped the value rather than that it hadn't

app/controllers/request_game_controller.rb
lizconlan: won't hurt if there's any chance it will get confused - did break the tests tho 🤣

app/views/request/_preview_errors.html.erb
```diff
@@ -0,0 +1,24 @@
+<% if contains_email
+  if user.nil? %>
```
garethrees: This would be cleaner and less confusing:

```erb
<% if contains_email %>
  <% if user.nil? %>
     …
```

app/controllers/user_controller.rb
```diff
@@ -338,7 +338,8 @@ def contact
           params[:contact][:subject],
           params[:contact][:message]
         ).deliver
-        flash[:notice] = _("Your message to {{recipient_user_name}} has been sent!",:recipient_user_name=>CGI.escapeHTML(@recipient_user.name))
+        flash[:notice] = _("Your message to {{recipient_user_name}} has been sent!",
+                           :recipient_user_name => @recipient_user.name.html_safe)
```
garethrees: I assume this ends up getting double-escaped as per the commit message?

Seems strange to need to mark it as html safe?

spec/integration/send_user_message_spec.rb
```diff
@@ -1,13 +1,12 @@
 # -*- encoding : utf-8 -*-
-require File.expand_path(File.dirname(__FILE__) + '/../spec_helper')
+require 'spec_helper'
```
garethrees: How come this changes here?

stevenday:
The fixups to my comments all look good, thanks!

app/controllers/user_controller.rb
lizconlan: if the name contains markup that's not marked as html safe, it gets escaped during translation then escaped by JSON when the cookie is read

spec/integration/send_user_message_spec.rb
lizconlan: Oops, tagged the wrong commit :(

doc/CHANGES.md
```diff
@@ -110,8 +113,17 @@
...
+    app/views/layouts/admin.html.erb
+    app/views/layouts/default.html.erb
+    app/views/layouts/no_chrome.html.erb
     app/views/reports/new.html.erb
     app/views/request/describe_notices/_error_message.html.erb
+    app/views/request/describe_notices/_internal_review.html.erb
+    app/views/request/describe_notices/_not_held.html.erb
+    app/views/request/describe_notices/_successful.html.erb
+    app/views/request/describe_notices/_waiting_response.html.erb
+    app/views/request/describe_notices/_waiting_response_overdue.html.erb
```
lizconlan: I've only added the _changed_ templates - should this included the newly created ones as well?

garethrees: No – newly created ones wouldn't really be overridden so reusers wouldn't need to reconcile changes.

doc/CHANGES.md
```diff
@@ -102,6 +102,9 @@
   opt all existing requests out of the new notifications feature. This is VERY
   IMPORTANT to run, as without it, existing requests won't trigger any alert
   emails at all.
+* The way that flash messages are rendered has changed so if you have overridden
+  a template which renders flash (e.g. `<%= flash[:notice] %>`), you will need
+  to use the new `<%= render_flash(flash[:notice] %>` style instead.
```
garethrees: Missed the closing bracket of `render_flash`

lizconlan: argh! ta :)

doc/CHANGES.md
lizconlan: 👍

PR #4011, opened 2017-05-26
Add Shopify's Bootsnap gem

stevenday:
I saw this today (https://github.com/Shopify/bootsnap) and figured it was
worth a try on Alaveteli to see what difference it makes. They claim to be
able to reduce their app's boot time by about 75%, with even quick loading
apps being reduced by 50% or so.

In case anyone else wants to try it out, I've made this branch, but to be
honest, I can't reall discern a difference!

Note this is using the latest 0.3.0.pre branch because that's got linux
support, otherwise it seems to be mac only.

I've been using:

```
time bundle exec rake environment
```

As a quick way to test how long the app takes to boot.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -171,7 +171,8 @@ group :test, :development do
   gem 'rspec-rails', '~> 3.5.0'
   gem 'pry', '~> 0.10.0'
   gem 'pry-debugger', '~> 0.2.0', :platforms => :ruby_19
-    gem 'public_suffix', '~> 1.4.0', '< 1.5.0'
+  gem 'public_suffix', '~> 1.4.0', '< 1.5.0'
+  gem 'bootsnap', '0.3.0.pre'
 end
 
 group :development do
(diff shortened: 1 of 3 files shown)


Gemfile
```diff
@@ -171,7 +171,8 @@ group :test, :development do
   gem 'rspec-rails', '~> 3.5.0'
   gem 'pry', '~> 0.10.0'
   gem 'pry-debugger', '~> 0.2.0', :platforms => :ruby_19
-    gem 'public_suffix', '~> 1.4.0', '< 1.5.0'
+  gem 'public_suffix', '~> 1.4.0', '< 1.5.0'
```
garethrees: This was an intentional indent as we don't use `public_suffix` directly. Pry depends on it, but we need to pin the specific version for compatibility reasons.

garethrees: Although looking at what the gem does, it doesn't look related to Pry & Co. I might have been mistaken in where the dependency was.

PR #4022, opened 2017-06-05
Use Rails 4 syntax for grouped counts

### LINKED ISSUE #3984: Stats broken in admin interface
https://www.whatdotheyknow.com/admin/stats

ss 2017-05-13 at 12 35 10

lizconlan:
Fixes #3984

### DIFF
--- a/spec/controllers/admin_general_controller_spec.rb
+++ b/spec/controllers/admin_general_controller_spec.rb
@@ -158,4 +158,66 @@ describe AdminGeneralController do
     end
 
   end
+
+  describe 'GET #stats' do
+
+    it 'assigns the number of public bodies to the view' do
+      get :stats
+      expect(assigns[:public_body_count]).to eq PublicBody.count
+    end
+
+    it 'assigns the number of requests to the view' do
+      get :stats
+      expect(assigns[:info_request_count]).to eq InfoRequest.count
+    end
+
+    it 'assigns the number of users to the view' do
+      get :stats
+      expect(assigns[:user_count]).to eq User.count
+    end
+
+    it 'assigns the number of tracks to the view' do
+      get :stats
+      expect(assigns[:track_thing_count]).to eq TrackThing.count
+    end
+
+    it 'assigns the number of comments to the view' do
+      get :stats
+      expect(assigns[:comment_count]).to eq Comment.count
+    end
+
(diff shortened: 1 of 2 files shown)


lizconlan:
Should this be a hotfix if we're about to upgrade Belgium to 0.28?

lizconlan:
ToDo: change commit comment to indicate Rails 4.1 rather than Rails 4 syntax

doc/CHANGES.md
```diff
@@ -65,6 +65,8 @@
   are no requests (Liz Conlan)
 * Switch to Trusty as the preferred OS for Travis CI and use Debian Wheezy as
   the new Vagrant default (Liz Conlan)
+* Fix bug which prevented stats from being displayed in the admin interface
```
garethrees: I'd say we don't need a changelog entry as there wasn't a release introducing the bug

spec/controllers/admin_general_controller_spec.rb
```diff
@@ -150,4 +150,66 @@
...
+
+      8.times { FactoryGirl.create(:successful_request) }
+      2.times { FactoryGirl.create(:info_request) }
+      FactoryGirl.create(:attention_requested_request)
+
+      get :stats
+      expect(assigns[:request_by_state]).to be_a(Hash)
+      expect(assigns[:request_by_state]).to include("successful" => 8)
+      expect(assigns[:request_by_state]).to include("waiting_response" => 2)
+      expect(assigns[:request_by_state]).to include("attention_requested" => 1)
```
garethrees: +1 to the extra tests!

Any reason this isn't a [single expect](http://www.betterspecs.org/#single), e.g:

```ruby
expect(assigns[:request_by_state]).
  to eq({ 'successful' => 8,
          'waiting_response' => 2,
          'attention_requested' => 1 })
```

Ditto for the spec below?

lizconlan: only that I wasn't entirely sure what the order would be

garethrees: Ah! Don't think it matters though:

```ruby
{ :a => 'b', :c => 3 } == { :c => 3, :a => 'b' }
# => true
```

lizconlan: \o/

PR #4026, opened 2017-06-06
[Pro] Issues/alaveteli pro 298/embargoed to do section

crowbot:


app/controllers/admin_general_controller.rb
```diff
@@ -42,6 +46,50 @@ def index
...
+                                        find_in_state('attention_requested').
+                                          embargoed
+
+      @embargoed_request_tasks = [ @embargoed_requires_admin_requests,
+                                   @embargoed_error_message_requests,
+                                   @embargoed_attention_requests,
+                                 ].any?{ |to_do_list| ! to_do_list.empty? }
+
+      @embargoed_attention_comments = Comment.
+                                  where(:attention_requested => true).
```
stevenday: indentation seems to have gone a bit awry here

app/models/info_request.rb
```diff
@@ -761,6 +761,9 @@ def log_event(type, params, options = {})
     if options[:created_at]
       event.update_column(:created_at, options[:created_at])
     end
+    if !self.last_event_time || (event.created_at > self.last_event_time)
```
stevenday: It would be nice to have a couple of model specs for this date setting

app/views/admin_general/index.html.erb
```diff
@@ -17,282 +17,204 @@
...
+  </div>
+<% end %>
 
-            <%= link_to "Add authority",
-              new_admin_body_path(:change_request_id => @change_request.id),
-              :class => 'btn btn-primary' %>
-          <% end %>
-        <% end %>
-      </div>
+<% if can?(:admin, AlaveteliPro::Embargo) && @embargoed_request_tasks %>
```
stevenday: Is this extra `can?` check necessary? We only set `@embargoed_request_tasks` if the user can admin the right things in the controller.

lib/tasks/temp.rake
```diff
@@ -1,6 +1,14 @@
 # -*- coding: utf-8 -*-
 namespace :temp do
 
+  desc 'Populate last_event_time column of InfoRequest'
+  task :populate_last_event_time => :environment do
+    InfoRequest.where('last_event_time is NULL').find_each do |info_request|
```
stevenday: Super minor pedantry: `is` could be capitalised too if we're capitalising `NULL` in the SQL.

stevenday: Also - would it speed this up at all if we eager loaded the info_request_events?

Issue #3800, opened 2017-03-02
Upgrade to Rails 4.2


garethrees:
Following on from https://github.com/mysociety/alaveteli/issues/3799.

http://edgeguides.rubyonrails.org/upgrading_ruby_on_rails.html#upgrading-from-rails-4-1-to-rails-4-2

- [x] Add web-console gem
- [x] Remove class-level `respond_to` & `respond_with` in controllers
- [x] Add error handling in `after_rollback` & `after_commit` transaction callbacks
- [x] ~~Set `test_order`~~ N/A because of rspec
- [x] Check `nil` behaviour of custom serializers
- [x] Set production log level if we don't want to default to `:debug`
- [x] ~~after_bundle in Rails templates~~ N/A
- [x] `sanitize`, `sanitize_css`, `strip_tags` and `strip_links` are backed by a new implementation – check/test
- [x] ~Pull in rails-dom-testing gem?~ Pulled in as a dependency of the Rails gems
- [x] Masked Authenticity Tokens – any strategies for verifying requests from non-rails forms that relied on a static session CSRF token have to take this into account. _seems to be fine_
- [x] Changelog notes
- [x] Upgrade notes

garethrees:
Review all Rails 4.0 / 4.1 commits for temporary fixes and monkeypatches to see if we can remove them.

garethrees:
http://railsdiff.org/4.1.16/4.2.8

lizconlan:
Looks like we can't upgrade  the `routing-filter` gem for Rails 4.2 until we [adopt Ruby 2.0](https://github.com/svenfuchs/routing-filter/blob/v0.6.0/routing-filter.gemspec) (The version we are currently using only [supports '< Rails 4.2'](https://github.com/svenfuchs/routing-filter/blob/0-4-stable/routing-filter.gemspec#L18))

lizconlan:
- [ ] Resolve conflict between Globalize and Rails 4.2

We've got validation and NOT NULL checks on top of translated columns. Globalize expects translated data to be migrated to the translation tables and dropped from the originating table. Under Rails 4.2, creating an instance of a model where this is the case fails as no values are supplied for required fields in the INSERT statement.

Known to be affected - PublicBody and anything that depends on it (e.g. InfoRequest)

lizconlan:
- [ ] Check whether `acts_as_versioned` (not being updated and still targeted at Rails 3.x) still works as required

[...]

PR #4028, opened 2017-06-07
WIP - [#3800] Rails 4.2

lizconlan:
For #3800


spec/models/public_body_spec.rb
```diff
@@ -532,7 +532,6 @@ def set_default_attributes(public_body)
   before(:each) do
     # Make sure that there's no internal_admin_body before each of
     # these tests:
-    PublicBody.connection.delete("DELETE FROM public_bodies WHERE url_name = 'internal_admin_body'")
```
garethrees: Just noting that this probably needs a replacement – the line below deletes the translation, but the `PublicBody` record will still exist.

app/models/public_body.rb
```diff
@@ -424,7 +424,10 @@ def self.import_csv_from_file(csv_filename, tag, tag_behaviour, dry_run, editor,
         # Give an error listing ones that are to be deleted
         deleted_ones = set_of_existing - set_of_importing
         if deleted_ones.size > 0
-          notes.push "Notes: Some " + tag + " bodies are in database, but not in CSV file:\n    " + Array(deleted_ones).sort.join("\n    ") + "\nYou may want to delete them manually.\n"
+          notes.push "Notes: Some #{tag} bodies are in database, but " \
+                     "not in CSV file:\n    " +
+                     Array(deleted_ones).reject(&:nil?).sort.join("\n    ") +
```
garethrees: s/`reject(&:nil?)`/`compact`/?

app/models/public_body.rb
```diff
@@ -75,6 +75,9 @@ class ImportCSVDryRun < StandardError ; end
   after_update :reindex_requested_from
 
 
+  # Force eager loading of translations everywhere
+  default_scope { includes(:translations) }
```
garethrees: Maybe should be a proc? https://stackoverflow.com/questions/4731125/always-eager-load-association-with-entity#comment49457616_4731370

```diff
-  default_scope { includes(:translations) }
+  default_scope -> { includes(:translations) }
```

lizconlan: Not sure, went off the Rails docs which suggest the syntax I used: http://api.rubyonrails.org/v4.2/classes/ActiveRecord/Scoping/Default/ClassMethods.htmll
http://api.rubyonrails.org/v5.1/classes/ActiveRecord/Scoping/Default/ClassMethods.html

(Followup comment - on that SO thread reveals the user to have been running MongoDB via the Mongoid gem rather than relying on ActiveRecord)

app/models/public_body.rb
lizconlan: yes, much nicer/clearer! thanks

Gemfile
```diff
@@ -183,4 +183,5 @@ group :development do
   gem 'launchy', '~> 2.4.0'
   gem 'spring', '~> 1.7.0', '< 2.0.0'
   gem 'spring-commands-rspec', '~> 1.0.0'
+  gem 'web-console', '~> 2.3.0', '< 3.0.0'
```
garethrees: I think this can get applied before updating rails version

garethrees:
Superseded by https://github.com/mysociety/alaveteli/pull/4114

Issue #4032, opened 2017-06-09
AlaveteliPro::RequestSummary should not set default database value for timestamps


garethrees:
https://github.com/mysociety/alaveteli/commit/0bce1d74b5542dc99bebf313aa4d1997ea25e787 adds database default values for `request_created_at` and `request_updated_at`. These values get fixed at the point of migration – not evaluated on record initialisation.

In practice, these default values _should_ get updated by the parent summarisable, but invalid records _can_ be created:

```ruby
Time.zone.now
# => Thu, 08 Jun 2017 12:11:42 BST +01:00
AlaveteliPro::RequestSummary.new.request_created_at
# => Mon, 05 Jun 2017 10:19:11 BST +01:00
```

I think that:

- The default value should be `nil`
- The not null constraint should be kept

```ruby
class RemoveDefaultValueFromRequestCreatedAtAndRequestUpdatedAtToRequestSummary < ActiveRecord::Migration
  def up
    change_column_default :request_summaries, :request_created_at, nil
    change_column_default :request_summaries, :request_updated_at, nil
  end

  def down
    change_column_default :request_summaries, :request_created_at, Time.now
    change_column_default :request_summaries, :request_updated_at, Time.now
  end
end
```

This causes spec failures because now the `AlaveteliPro::RequestSummary` doesn't have values for those attributes.

I had a bit of a stab at this, but it got more complicated than I was expecting. I think the way the two classes (`AlaveteliPro::RequestSummary` and the `Summarisable`) collaborate is a bit strange, but that might just be because I don't have any background knowledge on this.

IMO, this should work:

```ruby
summary = AlaveteliPro::RequestSummary.new
# => #<AlaveteliPro::RequestSummary id: nil, title: nil, body: nil, public_body_names: nil, summarisable_id: nil, summarisable_type: nil, created_at: nil, updated_at: nil, user_id: nil, request_created_at: nil, request_updated_at: nil>
summary.summarisable = InfoRequest.first
# => #<InfoRequest id: 101, …>
summary.save
# ActiveRecord::StatementInvalid: PG::NotNullViolation: ERROR:  null value in column "request_created_at" violates not-null constraint
# : INSERT INTO "request_summaries" ("created_at", "summarisable_id", "summarisable_type", "updated_at") VALUES ($1, $2, $3, $4) RETURNING "id"
[...]

PR #4035, opened 2017-06-09
Fix issues with request summary callbacks

stevenday:
This changes when the request summary callbacks are fired, fixing issues with them not working when a request is deleted, and with needing to disable callbacks during specs that involve request summaries.

I think with this change @garethrees - #4032 should be much easier to solve.

For mysociety/alaveteli-professional#290


stevenday:
This is now included in #4048 (it had to be, to make the specs pass) and improved there based on the comments here, so I'm closing this.

PR #4039, opened 2017-06-13
[#3800] Rails 4.2 Globalize Compatibility

garethrees:
Split out from https://github.com/mysociety/alaveteli/pull/4028 for easier reviewing.

### DIFF
--- a/app/models/concerns/public_body_derived_fields.rb
+++ b/app/models/concerns/public_body_derived_fields.rb
@@ -8,12 +8,12 @@ module PublicBodyDerivedFields
 
     # When name or short name is changed, also change the url name
     def short_name=(short_name)
-      super
+      write_attribute("short_name", short_name)
       update_url_name
     end
 
     def name=(name)
-      super
+      write_attribute("name", name)
       update_url_name
     end
 
--- a/app/models/public_body.rb
+++ b/app/models/public_body.rb
@@ -586,8 +586,8 @@ class PublicBody < ActiveRecord::Base
       # information
       # :version, :last_edit_editor, :last_edit_comment
       :home_page => calculated_home_page,
-      :notes => notes,
-      :publication_scheme => publication_scheme,
+      :notes => notes.to_s,
+      :publication_scheme => publication_scheme.to_s,
       :tags => tag_array,
       :info => {
         :requests_count => info_requests_count,
--- a/app/models/public_body.rb
+++ b/app/models/public_body.rb
@@ -75,6 +75,9 @@ class PublicBody < ActiveRecord::Base
   after_update :reindex_requested_from
 
 
+  # Force eager loading of translations everywhere
+  default_scope { includes(:translations) }
+
   # Every public body except for the internal admin one is visible
   scope :visible, -> { where("public_bodies.id <> #{ PublicBody.internal_admin_body.id }") }
 
--- a/app/models/public_body.rb
+++ b/app/models/public_body.rb
@@ -309,10 +309,12 @@ class PublicBody < ActiveRecord::Base
 
   # The "internal admin" is a special body for internal use.
   def self.internal_admin_body
-    # Use find_by_sql to avoid the search being specific to a
-    # locale, since url_name is a translated field:
-    sql = "SELECT * FROM public_bodies WHERE url_name = 'internal_admin_authority'"
-    matching_pbs = PublicBody.find_by_sql sql
+    matching_pbs = PublicBody.
+                     joins(:translations).
+                       where('public_body_translations.locale = ?',
+                             I18n.default_locale).
+                         where(:url_name => 'internal_admin_authority')
+
     case
     when matching_pbs.empty? then
       I18n.with_locale(I18n.default_locale) do
@@ -702,9 +704,9 @@ class PublicBody < ActiveRecord::Base
                         joins(:translations)
       else
         bodies = where("public_body_translations.locale = ?
-                        AND public_bodies.url_name in (?)",
-                        underscore_locale, body_short_names).
-                  joins(:translations)
+                        AND public_body_translations.url_name in (?)",
+                        underscore_locale, body_short_names
+                      ).joins(:translations)
       end
     end
     return bodies
--- a/app/models/public_body.rb
+++ b/app/models/public_body.rb
@@ -193,6 +193,18 @@ class PublicBody < ActiveRecord::Base
     self.api_key = SecureRandom.base64(33)
   end
 
+  def self.find_by_name(name)
+    PublicBody.joins(:translations).
+      where("public_body_translations.name = ?", name).
+       first
+  end
+
+  def self.find_by_url_name(url_name)
+    PublicBody.joins(:translations).
+      where("public_body_translations.url_name = ?", url_name).
+       first
+  end
+
   # like find_by_url_name but also search historic url_name if none found
   def self.find_by_url_name_with_historic(name)
     # If many bodies are found (usually because the url_name is the same
--- a/app/models/public_body.rb
+++ b/app/models/public_body.rb
@@ -424,7 +424,10 @@ class PublicBody < ActiveRecord::Base
         # Give an error listing ones that are to be deleted
         deleted_ones = set_of_existing - set_of_importing
         if deleted_ones.size > 0
-          notes.push "Notes: Some " + tag + " bodies are in database, but not in CSV file:\n    " + Array(deleted_ones).sort.join("\n    ") + "\nYou may want to delete them manually.\n"
+          notes.push "Notes: Some #{tag} bodies are in database, but " \
+                     "not in CSV file:\n    " +
+                     Array(deleted_ones).reject(&:nil?).sort.join("\n    ") +
+                     "\nYou may want to delete them manually.\n"
         end
 
         # Rollback if a dry run, or we had errors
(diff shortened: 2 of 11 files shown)


db/migrate/20170609100310_migrate_public_body_data.rb
```diff
@@ -0,0 +1,19 @@
...
+    PublicBody.find_each do |record|
+      translation = record.translation_for(I18n.default_locale) ||
+        record.translations.build(:locale => I18n.default_locale)
+      fields = record.translated_attributes
+      fields.each do |attribute_name, attribute_type|
+        translation[attribute_name] = record.
+                                        read_attribute(attribute_name,
+                                                       {:translated => false})
+      end
+      translation.save!
```
garethrees: Should consider failure here?

db/migrate/20170609101302_drop_public_body_translated_columns.rb
```diff
@@ -0,0 +1,12 @@
...
+class DropPublicBodyTranslatedColumns < ActiveRecord::Migration
+  def change
+    remove_column :public_bodies, :name, :text
+    remove_column :public_bodies, :short_name, :text
+    remove_column :public_bodies, :request_email, :text
+    remove_column :public_bodies, :url_name, :text
+    remove_column :public_bodies, :notes, :text
+    remove_column :public_bodies, :first_letter, :text
+    remove_column :public_bodies, :publication_scheme, :text
+  end
```
garethrees: Do we want to consider re-adding from translation table for revert?

db/migrate/20170609100310_migrate_public_body_data.rb
```diff
@@ -0,0 +1,19 @@
+# -*- encoding : utf-8 -*-
+class MigratePublicBodyData < ActiveRecord::Migration
+  def up
+    PublicBody.find_each do |record|
+      translation = record.translation_for(I18n.default_locale) ||
+        record.translations.build(:locale => I18n.default_locale)
+      fields = record.translated_attributes
+      fields.each do |attribute_name, attribute_type|
```
garethrees: To make it clearer that we're not using `attribute_type`

```diff
-      fields.each do |attribute_name, attribute_type|
+      fields.each do |attribute_name, _|
```

db/migrate/20170609100310_migrate_public_body_data.rb
```diff
@@ -0,0 +1,19 @@
...
+        translation[attribute_name] = record.
+                                        read_attribute(attribute_name,
+                                                       {:translated => false})
+      end
+      translation.save!
+    end
+  end
+
+  def down
```
garethrees: Delete default locale translations here?

```ruby
PublicBody::Translation.where(:locale => I18n.default_locale).destroy_all
```

db/migrate/20170609100310_migrate_public_body_data.rb
```diff
@@ -0,0 +1,19 @@
+# -*- encoding : utf-8 -*-
+class MigratePublicBodyData < ActiveRecord::Migration
+  def up
+    PublicBody.find_each do |record|
```
garethrees: Check `find_each` vs `find_in_batches` performance here

db/migrate/20170609100310_migrate_public_body_data.rb
```diff
@@ -0,0 +1,19 @@
+# -*- encoding : utf-8 -*-
+class MigratePublicBodyData < ActiveRecord::Migration
```
garethrees: s/Data/Translations/

db/migrate/20170609100310_migrate_public_body_data.rb
garethrees: I'd like to be able to run this and verify it has worked as expected before immediately dropping the tables (in the next migration). I think we should make this a class in `lib` so that we can add it to a temp rake task that we try before merging & deploying the destructive migrations.

Maybe something along the lines of:

```ruby
# lib/tasks/temp.task
task :migrate_public_body_translations_to_translation_table do
  PublicBodyTranslationMigration.new.migrate_to_translation_table
end

# db/migrate/20170609100310_migrate_public_body_translations.rb
def up
  PublicBodyTranslationMigration.new.migrate_to_translation_table
end

def down
  PublicBodyTranslationMigration.new.revert_to_public_bodies_table
end
```

app/models/concerns/public_body_derived_fields.rb
```diff
@@ -8,12 +8,12 @@ module PublicBodyDerivedFields
 
     # When name or short name is changed, also change the url name
     def short_name=(short_name)
-      super
+      write_attribute("short_name", short_name)
```
garethrees: Don't understand this change without more context.

Should there be tests changed/added here?

app/models/public_body.rb
```diff
@@ -193,6 +193,18 @@ def set_api_key!
     self.api_key = SecureRandom.base64(33)
   end
 
+  def self.find_by_name(name)
```
garethrees: Tests needed

garethrees: This probably needs to get squashed in to the commit where we drop the columns, as that's the point at which these methods would disappear?

garethrees: I think we can use hash keys for the `where` arguments:

```ruby
where(public_body_translations: { name: name })
```

spec/models/public_body_spec.rb
```diff
@@ -532,7 +532,6 @@ def set_default_attributes(public_body)
   before(:each) do
     # Make sure that there's no internal_admin_body before each of
     # these tests:
-    PublicBody.connection.delete("DELETE FROM public_bodies WHERE url_name = 'internal_admin_body'")
```
garethrees: Just noting that this probably needs a replacement – the line below deletes the translation, but the `PublicBody` record will still exist.

app/models/public_body.rb
```diff
@@ -424,7 +424,10 @@ def self.import_csv_from_file(csv_filename, tag, tag_behaviour, dry_run, editor,
         # Give an error listing ones that are to be deleted
         deleted_ones = set_of_existing - set_of_importing
         if deleted_ones.size > 0
-          notes.push "Notes: Some " + tag + " bodies are in database, but not in CSV file:\n    " + Array(deleted_ones).sort.join("\n    ") + "\nYou may want to delete them manually.\n"
+          notes.push "Notes: Some #{tag} bodies are in database, but " \
+                     "not in CSV file:\n    " +
+                     Array(deleted_ones).reject(&:nil?).sort.join("\n    ") +
```
garethrees: s/`reject(&:nil?)`/`compact`/?

app/models/public_body.rb
```diff
@@ -586,8 +586,8 @@ def json_for_api
       # information
       # :version, :last_edit_editor, :last_edit_comment
       :home_page => calculated_home_page,
-      :notes => notes,
-      :publication_scheme => publication_scheme,
+      :notes => notes.to_s,
+      :publication_scheme => publication_scheme.to_s,
```
garethrees: Why? Presumably because `nil` gets set as `null`?

```ruby
{ :a => nil }.to_json
# => "{\"a\":null}"
{ :a => '' }.to_json
# => "{\"a\":\"\"}"
```

Should have specs to document this.

app/models/public_body.rb
garethrees: Is this a new problem that's been introduced somehow?

app/models/public_body.rb
```diff
@@ -129,6 +129,17 @@ class Translation
   #
   # [1] http://git.io/vIetK
   class Version
+    before_save :copy_translated_attributes
```
garethrees: - Commit message should explain _why_ this change is being made; not just _what_ it is doing. I assume now that `PublicBody` doesn't have e.g. a `name` column, it doesn't get copied across. It should still have a `name` method that returns the translation though, so not sure why this behaviour has changed?
- Should have specs.

app/models/public_body.rb
```diff
@@ -129,6 +129,17 @@ class Translation
...
+
+    def copy_translated_attributes
+      public_body.attributes.each do |name, value|
+        if self.public_body.translated?(name) &&
+            !public_body.non_versioned_columns.include?(name)
+          self.send("#{name}=", value)
```
garethrees: Unnecessary `self` calls

garethrees: Probably also want to check that the sent method name exists:

```ruby
send("#{ name }=", value) if respond_to?("#{ name }=")
```

app/models/public_body.rb
```diff
@@ -75,6 +75,9 @@ class ImportCSVDryRun < StandardError ; end
   after_update :reindex_requested_from
 
 
+  # Force eager loading of translations everywhere
+  default_scope { includes(:translations) }
```
garethrees: Specs & explanation of _why_ we're doing this now.

db/migrate/20170609100310_migrate_public_body_data.rb
garethrees: As far as I can tell, the data is _already_ mirrored to the translation tables. I'm not saying we don't want to do this to be sure, but I don't know if you found cases where it was _necessary_ to do this.

garethrees:
Closing this as we're targeting this at develop in https://github.com/mysociety/alaveteli/pull/4050

app/models/public_body.rb
lizconlan: This has been removed again as it didn't feel like the right thing to do

PR #4041, opened 2017-06-15
Public body validations

garethrees:
Improve `PublicBody`validations for the upcoming Globalize update.

Some strange things happening with the attributes that set a `NOT NULL` constraint and an empty string `""` default value. In some cases this gets set on object initialization; in some it remains as `nil`:

```ruby
PublicBody.new(:disclosure_log => nil).disclosure_log
# => nil
PublicBody.new(:notes => nil).notes
# => ""
PublicBody.new(:home_page => nil).home_page
# => nil
PublicBody.new(:publication_scheme => nil).publication_scheme
# => ""
```

EDIT: Ahhh. `notes` and `publication_scheme` are translated attributes. `home_page` and `disclosure_log` are not.

- [x] Add changelog note


app/models/public_body.rb
```diff
@@ -64,6 +64,8 @@ class ImportCSVDryRun < StandardError ; end
   validates_presence_of :url_name, :message => N_("URL name can't be blank")
   validates_presence_of :last_edit_editor,
                         :message => N_("Last edit editor can't be blank")
+  validates_presence_of :last_edit_comment,
```
garethrees: This should be a `not_nil` validation, as the database has records where the comment is empty string.

app/models/public_body.rb
```diff
@@ -67,6 +67,8 @@ class ImportCSVDryRun < StandardError ; end
 
   validates :request_email,
             not_nil: { message: N_("Request email can't be nil") }
+  validates :last_edit_comment,
+            not_nil: { message: N_("Last edit comment can't be blank") }
```
garethrees: Message should be "can't be nil"

PR #4043, opened 2017-06-15
Add a Notification model and use it to send daily summaries to pros

stevenday:
This splits out the first half of #3994, not interfering with email notifications for normal users
at all, using new Notification system just for opted-in pro users' batch requests.

Specifically:

- Adds a new `Notification` model. Notifications can be instant or daily and set a sending timestamp on themselves as appropriate, so that we can easily query for the ones we need to send at any given time.
- Adds a new `User#notify` method for "notifying" users of events. The user can control when they want to receive the notification, although for now there's no logic for it, I just use the notifications_tester role to set everything to daily. 
- Modifies the method in `InfoRequest` that receives incoming mail so that it calls `User#notify` instead if the request is set to `use_notifications`. Currently this will only be true for batch requests on users with the notifications_tester role.
- Adds new hour/minute fields to `Users` to store when they should be sent daily summaries. These are set to a random time daily like existing tracked things alerts and a rake task is included to set these up initially.
- Adds a `NotificationMailer#daily_summary` method and associated views, which bring all of the notifications for a user into a single mail, grouped by the request they relate to.
- Adds class methods to `NotificationMailer` to actually send all the notifications that need sending at any given point.
- Adds the necessary scripts and daemon files to either send notifications manually, or to run a daemon that tries to send them continuously.


spec/mailers/notification_mailer_spec.rb
```diff
@@ -139,4 +139,177 @@
       expect(mail.header["X-Auto-Response-Suppress"].value).to eq "OOF"
     end
   end
+
+  describe '.send_instant_notifications' do
```
crowbot: I think we want to drop these instant notification specs from this commit.

app/models/user.rb
```diff
@@ -593,6 +595,39 @@ def for_admin_column(complete = false)
     end
   end
 
+  # Notify a user about an info_request_event, allowing the user's preferences
+  # to determine how that notification is delivered.
+  def notify(info_request_event)
+    Notification.create(
+      info_request_event: info_request_event,
+      frequency: self.notification_frequency,
```
crowbot: Looks like this needs to be `Notification.frequencies[self.notification_frequency]

doc/CHANGES.md
```diff
@@ -114,6 +114,11 @@
   [notes](https://git.io/vLNg0) on migrating away from 1.8.7; migrating to
   Ruby 2+ should be a similar process. Debian Jessie and Ubuntu 14.04+ include
   packaged versions of Ruby 2+.
+* Run `bundle exec rake temp:set_daily_summary_times` to set some default
+  times for users to receive daily summaries of notifications. This won't have
+  any effect on emails for most users until the new notifications system is
+  rolled out.
+>>>>>>> Add a rake task to set default daily summary times for each user
```
stevenday: Whoops, fixed in f611836

stevenday:
@crowbot - sorry, a few things slipped through the next as I built this branch up there. It should be good for a re-review now, but since some issues were with merges/cherry-picks and I wanted a clean history to base #4046 on, I've rebased everything again.

spec/models/notification_spec.rb
```diff
@@ -0,0 +1,82 @@
...
+      end
+
+      it "doesn't overwrite it" do
+        expect(notification.send_after).to be_within(1.second).of(delayed_time)
+      end
+    end
+  end
+
+  describe ".unseen" do
+    let(:unseen_notification) do
```
crowbot: Looks like these might need to be `let!` calls to make sure the objects are created?

stevenday: Sorry, I missed that - done in 969a286ddc18364889297d1f167bac793b02550f

spec/mailers/notification_mailer_spec.rb
stevenday: Good spot, I've rebased from this point so that it's fixed throughout.

app/views/notification_mailer/info_requests/_info_request.text.erb
```diff
@@ -0,0 +1,12 @@
+<%= info_request.title.html_safe %>
+<%= "-" * info_request.title.html_safe.length %>
```
stevenday: Ah, yes - that's because they aren't being translated, so because they're in a text template, they're not being escaped. I'll remove the specs and the html_safe stuff to make it clearer

app/views/alaveteli_pro/info_request_batches/_authority_list.text.erb
```diff
@@ -0,0 +1,15 @@
...
+    that the public body names are marked as html_safe, so that they don't get
+    escaped by our translation helpers (because this a text view so they don't
+    need to be and we don't want html entities in it. %>
+<% if public_bodies.size > 0 %>
+<% example_body_names = public_bodies.take(2).map(&:name) %>
+<%= n_('{{count}} recipient, {{first_public_body}}',
+       '{{count}} recipients, including {{first_public_body}} and {{second_public_body}}',
+       public_bodies.size,
+       count: public_bodies.size,
+       first_public_body: raw(example_body_names.first),
```
stevenday: So I initially used `html_safe` but it doesn't work when a body name is nil (which it might be in the singular case here), so used raw() instead. My understanding was that they're essentially interchangeable, and indeed it does seem to be doing something to the string that's passed into the translation helper, because without it the spec fails. I've updated the comment in 700dd45 to make all that clearer.

PR #4046, opened 2017-06-16
Extend notification usage to cover new responses for opted in users

stevenday:
This PR is part two of the split up #3994, following on from #4043. It covers:

- Copying the new response notification methods and emails into the
  notification mailer. This is a change from #3994 where we moved them
  wholesale because we want to keep that old code path alive for the time
  being.
- Adding a new set of methods for sending instant notifications
- Updating the daemon and associated scripts in #4043 to send both types of
  notifications
- Adding an additional feature flag condition when we choose what to do with a
  new incoming mail in `InfoRequest#receive`. This caused me some confusion,
  hence the slightly elaborate comment there, but I think it's necessary for
  now to keep these two slightly competing sets of controls on when we send
  emails. Eventually, we should be able to reduce it down to just the feature
  flag as we slowly roll out notifications across all types of emails.

### DIFF
--- a/app/views/notification_mailer/new_response.text.erb
+++ b/app/views/notification_mailer/new_response.text.erb
@@ -1,7 +1,7 @@
 <%= _('You have a new response to the {{law_used_full}} request',
       :law_used_full => @info_request.law_used_human(:full).html_safe) %>
-'<%= raw @info_request.title %>' <%=_('that you made to') %>
-<%= raw @info_request.public_body.name %>.
+'<%= @info_request.title %>' <%=_('that you made to') %>
+<%= @info_request.public_body.name %>.
 
 <%= _('To view the response, click on the link below.') %>
 
--- a/app/views/notification_mailer/new_response.text.erb
+++ b/app/views/notification_mailer/new_response.text.erb
@@ -1,0 +1,18 @@
+<%= _('You have a new response to the {{law_used_full}} request',
+      :law_used_full => @info_request.law_used_human(:full).html_safe) %>
+'<%= raw @info_request.title %>' <%=_('that you made to') %>
+<%= raw @info_request.public_body.name %>.
+
+<%= _('To view the response, click on the link below.') %>
+
+<%= new_response_url(@info_request, @incoming_message) %>
+
+<%= _('When you get there, please update the status to say if the response ' \
+      'contains any useful information.' ) %>
+
+<%= _('Although all responses are automatically published, we depend on ' \
+      'you, the original requester, to evaluate them.') %>
+
+-- <%= _('the {{site_name}} team', :site_name => site_name.html_safe) %>
+
+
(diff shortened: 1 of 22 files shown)


app/views/notification_mailer/new_response.text.erb
```diff
@@ -0,0 +1,18 @@
+<%= _('You have a new response to the {{law_used_full}} request',
+      :law_used_full => @info_request.law_used_human(:full).html_safe) %>
+'<%= raw @info_request.title %>' <%=_('that you made to') %>
```
stevenday: Fixed in 88b6dd5 - I kept this separate rather than as a fixup because it's present in the old view I copied this from and this is supposed to be a 'temporary' duplication, so this fix changes both.

stevenday:
> This looks good to me - the only thing I would consider doing is using another role instead of an alaveteli feature, as a role can be added or remove from users in the admin interface (how do we switch on the notifications alaveteli feature for a user?) and then it's all being controlled from one mechanism. We could make the role we have already 'batch_notifications_tester' and add 'notifications_tester' as a new role, and that might be clearer?

I do understand about the one mechanism, but if anything, I think I'd prefer to use the features more and the roles less. Since the roles are a documented part of alaveteli and (apart from this) have clear connections to user access control, it seems like we should only add a new role if it's going to stick around for a while and is also for that kind of use. What we're doing here seems like something different because of its temporary nature?

The feature flag would be enabled for a specific user in the console with a line like:
```ruby
AlaveteliFeatures.backend[:notifications].enable_actor user
```

What I thought was nice about that is that we could later do something like:
```ruby
AlaveteliFeatures.backend.register(:pros) do |actor|
  actor.respond_to?(:pro?) && actor.pro?
end
AlaveteliFeatures.backend[:notifications].enable_group :pros
```

To turn it on for all pro users, and then eventually do something like:
```ruby
AlaveteliFeatures.backend[:notifications].enable_percentage_of_actors 10
```
to also turn it on for 10% of normal users. We could then ramp that up to gradually load test the notifications code.

I'd be happy to add removing the existing role and basing the whole disabling of existing emails on the feature to this PR too (It would just be a small tweak to the InfoRequest model) if that made things clearer?

(Also - the features gem does have a [web ui](https://github.com/jnunemaker/flipper/blob/master/docs/ui/README.md) for enabling these things, I just didn't want to bring that in immediately until we were sure we'd use it, but it might be helpful rather than using the console?)

stevenday:
I've just pushed a few commits that remove the notifications_tester role and swap it for the feature flag instead - I think that makes things clearer all round.

One caveat here is that if we enable/disable the feature, it's effectively being cached on the request because of the use_notifications feature. This means that any change will only take effect for new requests made after the change. For now this is only for batch requests so I think it's ok, (i.e. it would be a relatively small number to manually change if we had to panic abort) but eventually I think we should drop that field and just check the feature at runtime - this will get easier to do when we have other emails in the notificaiton system

PR #4050, opened 2017-06-19
[#3800] globalize migration

garethrees:
- [x] Address feedback from https://github.com/mysociety/alaveteli/issues/4039
- [x] Fix internal admin authority find by sql https://github.com/mysociety/alaveteli/blob/develop/app/models/public_body.rb#L307-L330

### DIFF
--- a/app/controllers/admin_general_controller.rb
+++ b/app/controllers/admin_general_controller.rb
@@ -114,8 +114,7 @@ class AdminGeneralController < AdminController
       end
       # get all the models in the slice, eagerly loading the associations we use in the view
       public_body_versions = PublicBody.versioned_class.
-        includes(:public_body => :translations).
-          find(public_body_version_ids.keys)
+        find(public_body_version_ids.keys)
       info_request_events = InfoRequestEvent.
         includes(:info_request).
           find(info_request_event_ids.keys)
--- a/app/controllers/admin_raw_email_controller.rb
+++ b/app/controllers/admin_raw_email_controller.rb
@@ -25,8 +25,9 @@ class AdminRawEmailController < AdminController
             []
           else
             PublicBody.
-              where("lower(request_email) like lower('%'||?||'%')", domain).
-                order('name')
+              joins(:translations).
+                where("lower(request_email) like lower('%'||?||'%')", domain).
+                  order('name')
           end
 
           # 2. Match the email address in the message without matching the hash
--- a/app/models/alaveteli_pro/draft_info_request_batch.rb
+++ b/app/models/alaveteli_pro/draft_info_request_batch.rb
@@ -17,8 +17,12 @@ class AlaveteliPro::DraftInfoRequestBatch < ActiveRecord::Base
   include AlaveteliPro::RequestSummaries
 
   belongs_to :user
-  has_and_belongs_to_many :public_bodies,
-     -> { reorder('public_bodies.name asc') }
+  has_and_belongs_to_many :public_bodies, -> {
+    I18n.with_locale(I18n.locale) do
+      includes(:translations).
+        reorder('public_body_translations.name asc')
+    end
+  }
 
   validates_presence_of :user
 
@@ -43,7 +47,9 @@ class AlaveteliPro::DraftInfoRequestBatch < ActiveRecord::Base
 
   # @see RequestSummaries#request_summary_public_body_names
   def request_summary_public_body_names
-    self.public_bodies.pluck(:name).join(" ")
+    self.public_bodies.
+      joins(:translations).
+        pluck(:name).join(" ")
   end
 
   # @see RequestSummaries#request_summary_categories
--- a/app/models/concerns/public_body_derived_fields.rb
+++ b/app/models/concerns/public_body_derived_fields.rb
@@ -7,13 +7,13 @@ module PublicBodyDerivedFields
     before_save :set_first_letter
 
     # When name or short name is changed, also change the url name
-    def short_name=(short_name)
-      super
+    def short_name=(value)
+      write_attribute("short_name", value)
       update_url_name
     end
 
-    def name=(name)
-      super
+    def name=(value)
+      write_attribute("name", value)
       update_url_name
     end
 
@@ -41,7 +41,9 @@ module PublicBodyDerivedFields
 
   def update_url_name
     if changed.include?('name') || changed.include?('short_name')
-      self.url_name = MySociety::Format.simplify_url_part(self.short_or_long_name, 'body')
+      write_attribute("url_name",
+                      MySociety::Format.
+                        simplify_url_part(self.short_or_long_name, 'body'))
     end
   end
 
--- a/app/models/public_body.rb
+++ b/app/models/public_body.rb
@@ -202,6 +202,18 @@ class PublicBody < ActiveRecord::Base
     self.api_key = SecureRandom.base64(33)
   end
 
+  def self.find_by_name(name)
+    PublicBody.joins(:translations).
+      where("public_body_translations.name = ?", name).
+       first
+  end
+
+  def self.find_by_url_name(url_name)
+    PublicBody.joins(:translations).
+      where("public_body_translations.url_name = ?", url_name).
+       first
+  end
+
   # like find_by_url_name but also search historic url_name if none found
   def self.find_by_url_name_with_historic(name)
     # If many bodies are found (usually because the url_name is the same
--- a/app/models/public_body.rb
+++ b/app/models/public_body.rb
@@ -203,25 +203,30 @@ class PublicBody < ActiveRecord::Base
   end
 
   def self.find_by_name(name)
-    PublicBody.joins(:translations).
-      where("public_body_translations.name = ?", name).
-       first
+    PublicBody.
+      unscoped.
+        where(:name => name).
+          with_translations.
+            first
   end
 
   def self.find_by_url_name(url_name)
-    PublicBody.joins(:translations).
-      where("public_body_translations.url_name = ?", url_name).
-       first
+    PublicBody.
+      unscoped.
+        where(:url_name => url_name).
+          with_translations.
+            first
   end
 
   # like find_by_url_name but also search historic url_name if none found
   def self.find_by_url_name_with_historic(name)
     # If many bodies are found (usually because the url_name is the same
     # across locales) return any of them.
-    found = joins(:translations).
-      where("public_body_translations.url_name = ?", name).
-      readonly(false).
-      first
+    found = PublicBody.
+              unscoped.
+                where(:url_name => name).
+                  with_translations.
+                    first
 
     return found if found
 
@@ -318,10 +323,12 @@ class PublicBody < ActiveRecord::Base
 
   # The "internal admin" is a special body for internal use.
   def self.internal_admin_body
-    # Use find_by_sql to avoid the search being specific to a
-    # locale, since url_name is a translated field:
-    sql = "SELECT * FROM public_bodies WHERE url_name = 'internal_admin_authority'"
-    matching_pbs = PublicBody.find_by_sql sql
+    matching_pbs = PublicBody.
+                     joins(:translations).
+                       where('public_body_translations.locale = ?',
+                             I18n.default_locale).
+                         where(:url_name => 'internal_admin_authority')
+
     case
     when matching_pbs.empty? then
       I18n.with_locale(I18n.default_locale) do
@@ -702,9 +709,9 @@ class PublicBody < ActiveRecord::Base
                         joins(:translations)
       else
         bodies = where("public_body_translations.locale = ?
-                        AND public_bodies.url_name in (?)",
-                        underscore_locale, body_short_names).
-                  joins(:translations)
+                        AND public_body_translations.url_name in (?)",
+                        underscore_locale, body_short_names
+                      ).joins(:translations)
       end
     end
     return bodies
--- a/app/models/public_body.rb
+++ b/app/models/public_body.rb
@@ -438,7 +438,10 @@ class PublicBody < ActiveRecord::Base
         # Give an error listing ones that are to be deleted
         deleted_ones = set_of_existing - set_of_importing
         if deleted_ones.size > 0
-          notes.push "Notes: Some " + tag + " bodies are in database, but not in CSV file:\n    " + Array(deleted_ones).sort.join("\n    ") + "\nYou may want to delete them manually.\n"
+          notes.push "Notes: Some #{tag} bodies are in database, but " \
+                     "not in CSV file:\n    " +
+                     Array(deleted_ones).compact.sort.join("\n    ") +
+                     "\nYou may want to delete them manually.\n"
         end
 
         # Rollback if a dry run, or we had errors
--- a/app/models/public_body.rb
+++ b/app/models/public_body.rb
@@ -591,8 +591,8 @@ class PublicBody < ActiveRecord::Base
       # information
       # :version, :last_edit_editor, :last_edit_comment
       :home_page => calculated_home_page,
-      :notes => notes,
-      :publication_scheme => publication_scheme,
+      :notes => notes.to_s,
+      :publication_scheme => publication_scheme.to_s,
       :tags => tag_array,
       :info => {
         :requests_count => info_requests_count,
--- a/app/models/public_body.rb
+++ b/app/models/public_body.rb
@@ -129,6 +129,17 @@ class PublicBody < ActiveRecord::Base
   #
   # [1] http://git.io/vIetK
   class Version
+    before_save :copy_translated_attributes
+
+    def copy_translated_attributes
+      public_body.attributes.each do |name, value|
+        if self.public_body.translated?(name) &&
+            !public_body.non_versioned_columns.include?(name)
+          self.send("#{name}=", value)
+        end
+      end
+    end
+
     def last_edit_comment_for_html_display
       text = self.last_edit_comment.strip
       text = CGI.escapeHTML(text)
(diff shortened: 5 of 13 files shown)


app/models/public_body.rb
```diff
@@ -217,15 +217,19 @@ def set_api_key!
   end
 
   def self.find_by_name(name)
-    PublicBody.joins(:translations).
+    found = PublicBody.joins(:translations).
       where("public_body_translations.name = ?", name).
        first
+
+    PublicBody.find(found.id)
```
garethrees: Could we just call `reload`?


```diff
diff --git a/spec/controllers/admin_public_body_controller_spec.rb b/spec/controllers/admin_public_body_controller_spec.rb
index f451d81..044b43b 100644
--- a/spec/controllers/admin_public_body_controller_spec.rb
+++ b/spec/controllers/admin_public_body_controller_spec.rb
@@ -200,7 +200,7 @@ describe AdminPublicBodyController do
         post :create, @params

         body = PublicBody.find_by_name('New Quango')
-
+        body.reload
         I18n.with_locale(:es) do
           expect(body.name).to eq('Los Quango')
           expect(body.url_name).to eq('lq')
```

```
$ be rspec ./spec/controllers/admin_public_body_controller_spec.rb:199
Run options: include {:locations=>{"./spec/controllers/admin_public_body_controller_spec.rb"=>[199]}}
.

Finished in 2.91 seconds (files took 41.49 seconds to load)
1 example, 0 failures
```

garethrees: So, what's happening here? Why are the translations not getting loaded through `where`?

lizconlan: It's possible the `default_scope` is doing harm here as adding `unscoped` to the query seems to help, and adding `with_translations` appears to ensure the translations are loaded (the localisation integration spec appears to be the best test for this)

Looking into removing default_scope on that basis, but may keep the `unscoped` syntax for safety.

db/migrate/096_create_translation_tables.rb
```diff
@@ -24,6 +24,6 @@ def self.up
 
 
   def self.down
-    PublicBody.drop_translation_table!
+    PublicBody.drop_translation_table! :migrate_data => true
```
garethrees: Drop this

db/migrate/20170609100310_migrate_public_body_data.rb
```diff
@@ -0,0 +1,19 @@
+# -*- encoding : utf-8 -*-
+class MigratePublicBodyData < ActiveRecord::Migration
+  def up
+    PublicBody.find_each do |record|
```
garethrees: This isn't really needed.

We do need to figure out how to handle the internal admin authority, but it would be better to have a migration that just affects that, rather than every authority.

db/migrate/20170609101302_drop_public_body_translated_columns.rb
```diff
@@ -0,0 +1,35 @@
...
+    remove_column :public_bodies, :publication_scheme
+  end
+
+  def down
+    add_column :public_bodies, :name, :text
+    add_column :public_bodies, :short_name, :text
+    add_column :public_bodies, :request_email, :text
+    add_column :public_bodies, :url_name, :text
+    add_column :public_bodies, :notes, :text
+    add_column :public_bodies, :first_letter, :text
```
garethrees: Some of these have default values and constraints that need to get reapplied. Also check types – some might be `:string` rather than `:text`

db/migrate/20170609101302_drop_public_body_translated_columns.rb
```diff
@@ -0,0 +1,35 @@
...
+    add_column :public_bodies, :publication_scheme, :text
+
+    # Migrate the data back
+    PublicBody.find_each do |pb|
+      translated = pb.translation_for(I18n.default_locale)
+
+      # Create a hash containing the translated column names and their values
+      pb.translated_attribute_names.inject(fields_to_update={}) do |f, name|
+        f.update({name.to_sym => translated[name.to_s]})
+      end
```
garethrees: I'm not sure of the scope of `fields_to_update` here.

- Does it exist outside the block?
- Does it get mutated each iteration of the loop?

To me it looks like the `inject` block should return some value, which is then passed to `update_all` below.

db/migrate/20170609101302_drop_public_body_translated_columns.rb
```diff
@@ -0,0 +1,35 @@
...
+
+      # Now, update the actual model's record with the hash
+      PublicBody.where(:id => translated['id']).update_all(fields_to_update)
```
garethrees: Why re-find this? Why not `pb.update_all(fields_to_update)`?

Also, this is finding the `PublicBody` record by the `id` attribute of the `PublicBody::Translation`, so this will result in the wrong data being applied to the `PublicBody`.

db/migrate/20170609101302_drop_public_body_translated_columns.rb
garethrees: Why does this even have to happen in the migration? Might be able to move this to a post-migrate rake task.

Only problem might be if a `nil` value in e.g. `PublicBody#name` prevents the translation record being found.

garethrees: Or does globalize provide a data migration method?

app/models/concerns/public_body_derived_fields.rb
```diff
@@ -7,13 +7,13 @@ module PublicBodyDerivedFields
     before_save :set_first_letter
 
     # When name or short name is changed, also change the url name
-    def short_name=(short_name)
-      super
+    def short_name=(value)
+      write_attribute("short_name", value)
```
garethrees: Lets link to the globalize source in the commit message to show where this was taken from.

Use symbol keys for the attr names if possible.

app/models/concerns/public_body_derived_fields.rb
```diff
@@ -41,7 +41,9 @@ def set_first_letter
 
   def update_url_name
     if changed.include?('name') || changed.include?('short_name')
-      self.url_name = MySociety::Format.simplify_url_part(self.short_or_long_name, 'body')
+      write_attribute("url_name",
+                      MySociety::Format.
+                        simplify_url_part(self.short_or_long_name, 'body'))
```
garethrees: can probably remove the `self` here

app/models/concerns/public_body_derived_fields.rb
```diff
@@ -41,7 +41,9 @@ def set_first_letter
 
   def update_url_name
     if changed.include?('name') || changed.include?('short_name')
```
garethrees: Check that these `changed` attributes still work correctly now that they don't exist on the `PublicBody`

app/models/public_body.rb
```diff
@@ -202,6 +202,18 @@ def set_api_key!
     self.api_key = SecureRandom.base64(33)
   end
 
+  def self.find_by_name(name)
+    PublicBody.joins(:translations).
+      where("public_body_translations.name = ?", name).
+       first
+  end
+
+  def self.find_by_url_name(url_name)
```
garethrees: We need to do this because these no longer exist now that we've dropped the columns on this table.

- Commit message needs to indicate why these are required.
- It surprises me that globalize doesn't reimplement these. Maybe they've moved to a gem or something?
- Doesn't look like these take locale in to account
- These need testing – specifically around locale fallbacks etc
- Probably don't need the class prefix. just start with `joins`
- Can probably use hash keys for a bit nicer syntax while we're at it `where(:public_body_translations => { :url_name => url_name } )`

garethrees: This comment is for 79a82c06439df7c01a0aca4c02609383d546036f

app/controllers/admin_general_controller.rb
```diff
@@ -114,8 +114,7 @@ def timeline
       end
       # get all the models in the slice, eagerly loading the associations we use in the view
       public_body_versions = PublicBody.versioned_class.
-        includes(:public_body => :translations).
-          find(public_body_version_ids.keys)
+        find(public_body_version_ids.keys)
```
garethrees: Don't know what this is doing.

If we need it, then:
- tests
- commit message explaining why we need the change.

app/controllers/admin_raw_email_controller.rb
```diff
@@ -25,8 +25,9 @@ def show
             []
           else
             PublicBody.
-              where("lower(request_email) like lower('%'||?||'%')", domain).
-                order('name')
+              joins(:translations).
+                where("lower(request_email) like lower('%'||?||'%')", domain).
+                  order('name')
```
garethrees: Don't know what this is doing. Is this taking locale in to account?

If we need it, then:
- tests
- commit message explaining why we need the change.

app/models/alaveteli_pro/draft_info_request_batch.rb
```diff
@@ -17,8 +17,12 @@ class AlaveteliPro::DraftInfoRequestBatch < ActiveRecord::Base
   include AlaveteliPro::RequestSummaries
 
   belongs_to :user
-  has_and_belongs_to_many :public_bodies,
-     -> { reorder('public_bodies.name asc') }
+  has_and_belongs_to_many :public_bodies, -> {
+    I18n.with_locale(I18n.locale) do
```
garethrees: Explain why this is needed

- tests?
- commit message explaining why we need the change.

app/models/alaveteli_pro/draft_info_request_batch.rb
```diff
@@ -17,8 +17,12 @@ class AlaveteliPro::DraftInfoRequestBatch < ActiveRecord::Base
...
+      includes(:translations).
+        reorder('public_body_translations.name asc')
```
garethrees: order vs reorder?

app/models/alaveteli_pro/draft_info_request_batch.rb
```diff
@@ -43,7 +47,9 @@ def request_summary_body
 
   # @see RequestSummaries#request_summary_public_body_names
   def request_summary_public_body_names
-    self.public_bodies.pluck(:name).join(" ")
+    self.public_bodies.
+      joins(:translations).
```
garethrees: Shouldn't globalize make this join? If not then possibly want a spec & definitely an explanation of why we need this and what's breaking in the commit message.

app/models/public_body.rb
garethrees: Comments for 7813a5d3abd7ae6383c3376926b8f15663759a3c

Explain why we need the `with_translations`. Add specs.

app/models/public_body.rb
```diff
@@ -318,10 +323,12 @@ def calculated_home_page
...
   def self.internal_admin_body
-    # Use find_by_sql to avoid the search being specific to a
-    # locale, since url_name is a translated field:
-    sql = "SELECT * FROM public_bodies WHERE url_name = 'internal_admin_authority'"
-    matching_pbs = PublicBody.find_by_sql sql
+    matching_pbs = PublicBody.
+                     joins(:translations).
+                       where('public_body_translations.locale = ?',
+                             I18n.default_locale).
+                         where(:url_name => 'internal_admin_authority')
```
garethrees: Lets add some tests to make sure this is working as expected. See belgium where they have empty `fr_BE`, `nl_BE` translations and a populated `en` translation (but `en` is not an available locale).

app/models/public_body.rb
```diff
@@ -702,9 +709,9 @@ def self.popular_bodies(locale)
                         joins(:translations)
       else
         bodies = where("public_body_translations.locale = ?
-                        AND public_bodies.url_name in (?)",
-                        underscore_locale, body_short_names).
-                  joins(:translations)
+                        AND public_body_translations.url_name in (?)",
+                        underscore_locale, body_short_names
+                      ).joins(:translations)
```
garethrees: specs? or commit message saying what specs this was breaking and how the change fixes it

spec/models/public_body_spec.rb
```diff
@@ -532,8 +532,11 @@ def set_default_attributes(public_body)
   before(:each) do
     # Make sure that there's no internal_admin_body before each of
     # these tests:
-    PublicBody.connection.delete("DELETE FROM public_bodies WHERE url_name = 'internal_admin_body'")
-    PublicBody.connection.delete("DELETE FROM public_body_translations WHERE url_name = 'internal_admin_body'")
+    if internal_admin_body = PublicBody::Translation.
```
garethrees: Just destroy all the bodies.

```ruby
PublicBody.destroy_all
```

Check that all translation records are also destroyed.

app/models/public_body.rb
```diff
@@ -438,7 +438,10 @@ def self.import_csv_from_file(csv_filename, tag, tag_behaviour, dry_run, editor,
         # Give an error listing ones that are to be deleted
         deleted_ones = set_of_existing - set_of_importing
         if deleted_ones.size > 0
-          notes.push "Notes: Some " + tag + " bodies are in database, but not in CSV file:\n    " + Array(deleted_ones).sort.join("\n    ") + "\nYou may want to delete them manually.\n"
+          notes.push "Notes: Some #{tag} bodies are in database, but " \
+                     "not in CSV file:\n    " +
+                     Array(deleted_ones).compact.sort.join("\n    ") +
+                     "\nYou may want to delete them manually.\n"
```
garethrees: Explain what was breaking and how this fixes it.

Might need specs?

app/models/public_body.rb
```diff
@@ -591,8 +591,8 @@ def json_for_api
       # information
       # :version, :last_edit_editor, :last_edit_comment
       :home_page => calculated_home_page,
-      :notes => notes,
-      :publication_scheme => publication_scheme,
+      :notes => notes.to_s,
+      :publication_scheme => publication_scheme.to_s,
```
garethrees: Why? Presumably because `nil` gets set as `null`?

```ruby
{ :a => nil }.to_json
# => "{\"a\":null}"
{ :a => '' }.to_json
# => "{\"a\":\"\"}"
```

Should have specs to document this.

app/models/public_body.rb
```diff
@@ -129,6 +129,17 @@ class Translation
...
+    before_save :copy_translated_attributes
+
+    def copy_translated_attributes
+      public_body.attributes.each do |name, value|
+        if self.public_body.translated?(name) &&
+            !public_body.non_versioned_columns.include?(name)
+          self.send("#{name}=", value)
+        end
+      end
+    end
```
garethrees: failing because of broken specs? if so, explain what specs were broken and why this fixes.

If no failing specs, then we probably need some extra coverage here.

app/models/concerns/public_body_derived_fields.rb
lizconlan: Better commit message:

ActiveRecord attributes that are not bound to database columns are handled through `method_missing` so there's no named method for `super` to find. The handler is a wrapper for `write_attribute` so we can safely call that directly ourselves.

https://apidock.com/rails/v4.1.8/ActiveRecord/AttributeMethods/Write/attribute=

app/models/public_body.rb
lizconlan: This might have been a missing migration - the public_body_translations table doesn't have the default values from public_bodies so translated fields could be returning nil instead of ""

db/migrate/20170609101302_drop_public_body_translated_columns.rb
lizconlan: the syntax is a bit eyewatering but it appears that `(inject(fields_to_update={})` is creating a new Hash before the loop starts and passing it in one go so it's available afterwards

pinched from https://github.com/globalize/globalize/blob/4-0-stable/lib/globalize/active_record/migration.rb#L127-L133

lizconlan: > Why does this even have to happen in the migration?

in the revised version, we'd need to (re)create the columns, copy the data back and then apply the original constraints otherwise creating the columns immediately triggers an error for e.g. the name field being NULL (name is the first one it hits so it stops there)

lizconlan: (still thinking of a way to make that more testable so we can check whether it's doing the right thing - may involve 2 separate migrations: one of adding the columns, one for calling the a data "migration" from a lib file and applying the constraints)

db/migrate/20170609101302_drop_public_body_translated_columns.rb
lizconlan: originally pinched blindly from Globalize
https://github.com/globalize/globalize/blob/4-0-stable/lib/globalize/active_record/migration.rb#L127-L133

...that might turn out to be a necessary evil - `update_all` is a method that can only be called on a Relation, not the actual model instance.
https://apidock.com/rails/v4.0.2/ActiveRecord/Relation/update_all

The equivalent for model, `update_attributes`, is going to call attribute setters which I imagine will write to the translation table
https://apidock.com/rails/ActiveRecord/Base/update_attributes

lizconlan: (Yep - if you use `pb.update_attributes` it appears to have succeeded but has actually updated the translations table and the migration fails when adding the constraints as none of the values have been set and it all gets rolled back)

app/models/public_body.rb
lizconlan: Fixed by extra migration to add the missing constraints and defaults

app/models/public_body.rb
lizconlan: Fixed by combo of extra migration to set defaults and using a persisted factory instance rather than using `.new` which doesn't really test real-world use and produces null values

lizconlan:
Closed in favour of #4066

PR #4051, opened 2017-06-20
[#3978] Track flash

garethrees:
WIP on #3978

Covers the subscribe and already subscribed flash messages.

Does not cover the unsubscribe alert.


garethrees:
Lets get these commits out as soon as we can. Still a few questions over the [unsubscribe alert](https://github.com/mysociety/alaveteli/issues/3978#issuecomment-309741555).

garethrees:
Merged in https://github.com/mysociety/alaveteli/commit/5dce746cb21e2a6b6a05c651b1997bcdf134026b

PR #4052, opened 2017-06-20
WIP - Make it clearer that search totals are an estimate

lizconlan:


app/views/general/search.html.erb
```diff
@@ -202,6 +202,8 @@
     <div class="results_section">
       <% if @xapian_requests_hits == 1 && @page == 1 %>
         <h2 class="foi_results"><%= _("One FOI request found") %></h2>
+      <% elsif @xapian_requests.matches_estimated > @requests_per_page %>
+        <h2 class="foi_results"><%= _("FOI requests {{start_count}} to {{end_count}} of about {{total_count}}", :start_count => ((@page-1)*@requests_per_page+1).to_s, :end_count => [@page*@requests_per_page, @xapian_requests.matches_estimated].min.to_s, :total_count =>  @xapian_requests.matches_estimated.to_s) %></h2>
```
garethrees: Why the extra conditionals?

Can't we make this much simpler and just s/`{{total_count}}`/`about {{total_count}}`/?

lizconlan: Logically it would be nicer to see e.g. "1 to 4 of 4" if there's only a single page, but probably not worth the overhead. Experimental commits pushed pre-cleanup so the code wasn't only on my laptop in case it overheats some more. Reverting

spec/controllers/request_controller_spec.rb
```diff
@@ -17,6 +20,11 @@
     expect(response).to render_template('list')
   end
 
+  it "renders the approximate number of requests" do
+    get :list, :view => 'all'
+    expect(response.body).to match(/About \d+ FOI requests found/)
```
garethrees: I'm not sure a controller spec is really appropriate here.

We're now calling `render_views` which will have a performance impact, and I don't think this is really testing any logic as such. In a controller spec we might want to check that `@results[:matches_estimated]` is set/accurate, but I don't really think that how we then interpolate that in to HTML is massively relevant here.

This would be better in a view spec if anything, but tbh, I'd just drop this.

PR #4055, opened 2017-06-21
Check for existing events when backloading.

crowbot:


lib/tasks/temp.rake
```diff
@@ -123,9 +123,21 @@ namespace :temp do
...
-      if verbose
-        puts "Logging #{event_type} for #{overdue_alert.info_request.id}"
+
+      existing_event = InfoRequestEvent.where("info_request_id = ?
+                                              AND event_type = ?
+                                              AND created_at > ?",
+                                              overdue_alert.info_request,
+                                              event_type,
+                                              event_forming_request.created_at)
+      unless existing_event.empty?
```
stevenday: I think this should be an `if`, not an `unless`? (or `empty?` switched to `any?`)

stevenday:
I've pushed a fixup to speed things up, will let @crowbot double-check I've not mistaken the intent here.

Issue #2945, opened 2015-12-08
Add Google Analytics to the admin


stevenday:
So that we can identify stuff like:
- The most popular/used admin pages
- The least popular/used admin pages
- How much time is spent on the site by admins

And perhaps:
- Mark goals and track conversions over time
- Find the best places to make improvements

crowbot:
Is the idea here to add one GA account for all Alaveteli installs?

stevenday:
@crowbot - I guess that would be best - the idea was that we'd be looking at them, rather than the partners running the sites, but I hadn't really thought about it either way to be honest. 

It depends whether we think there'd ever be a need to share it with them too? (Even if so, probably easier to make an exception for that later rather than make it the default now)

garethrees:
Don't think we're likely to do this, so closing.

PR #4066, opened 2017-06-26
[#3800] Drop Globalize columns ahead of 4.2 upgrade

lizconlan:
Previous PR: https://github.com/mysociety/alaveteli/issues/4050
Prerequisite of #3800

### DIFF
--- a/app/controllers/admin_public_body_controller.rb
+++ b/app/controllers/admin_public_body_controller.rb
@@ -142,10 +142,12 @@ class AdminPublicBodyController < AdminController
   def missing_scheme
     # There might be a way to do this in ActiveRecord, but I can't find it
     @public_bodies = PublicBody.find_by_sql("
-            SELECT a.id, a.name, a.url_name, COUNT(*) AS howmany
-              FROM public_bodies a JOIN info_requests r ON a.id = r.public_body_id
-             WHERE a.publication_scheme = ''
-             GROUP BY a.id, a.name, a.url_name
+            SELECT a.id, t.name, t.url_name, COUNT(*) AS howmany
+              FROM public_bodies a
+                JOIN public_body_translations t on a.id = t.public_body_id
+                JOIN info_requests r ON a.id = r.public_body_id
+             WHERE t.publication_scheme = '' AND t.locale = '#{I18n.locale}'
+             GROUP BY a.id, t.name, t.url_name
              ORDER BY howmany DESC
              LIMIT 20
         ")
--- a/app/models/public_body.rb
+++ b/app/models/public_body.rb
@@ -708,7 +708,7 @@ class PublicBody < ActiveRecord::Base
                         joins(:translations)
       else
         bodies = where("public_body_translations.locale = ?
-                        AND public_bodies.url_name in (?)",
+                        AND public_body_translations.url_name in (?)",
                         underscore_locale, body_short_names).
                   joins(:translations)
       end
--- a/app/models/public_body.rb
+++ b/app/models/public_body.rb
@@ -202,6 +202,14 @@ class PublicBody < ActiveRecord::Base
     self.api_key = SecureRandom.base64(33)
   end
 
+  def self.find_by_name(name)
+    find_by(name: name)
+  end
+
+  def self.find_by_url_name(url_name)
+    find_by(url_name: url_name)
+  end
+
   # like find_by_url_name but also search historic url_name if none found
   def self.find_by_url_name_with_historic(name)
     # If many bodies are found (usually because the url_name is the same
--- a/app/models/public_body.rb
+++ b/app/models/public_body.rb
@@ -129,6 +129,17 @@ class PublicBody < ActiveRecord::Base
   #
   # [1] http://git.io/vIetK
   class Version
+    before_save :copy_translated_attributes
+
+    def copy_translated_attributes
+      public_body.attributes.each do |name, value|
+        if public_body.translated?(name) &&
+            !public_body.non_versioned_columns.include?(name)
+          send("#{name}=", value)
+        end
+      end
+    end
+
     def last_edit_comment_for_html_display
       text = self.last_edit_comment.strip
       text = CGI.escapeHTML(text)
--- a/app/models/public_body.rb
+++ b/app/models/public_body.rb
@@ -314,21 +314,31 @@ class PublicBody < ActiveRecord::Base
 
   # The "internal admin" is a special body for internal use.
   def self.internal_admin_body
-    # Use find_by_sql to avoid the search being specific to a
-    # locale, since url_name is a translated field:
-    sql = "SELECT * FROM public_bodies WHERE url_name = 'internal_admin_authority'"
-    matching_pbs = PublicBody.find_by_sql sql
+    matching_pbs = PublicBody.where(:url_name => 'internal_admin_authority')
     case
     when matching_pbs.empty? then
-      I18n.with_locale(I18n.default_locale) do
-        PublicBody.create!(:name => 'Internal admin authority',
-                           :short_name => "",
-                           :request_email => AlaveteliConfiguration::contact_email,
-                           :home_page => "",
-                           :notes => "",
-                           :publication_scheme => "",
-                           :last_edit_editor => "internal_admin",
-                           :last_edit_comment => "Made by PublicBody.internal_admin_body")
+      # "internal admin" exists but has the wrong default locale - fix & return
+      if invalid_locale = PublicBody::Translation.
+                            find_by_url_name('internal_admin_authority')
+        found_pb = PublicBody.find(invalid_locale.public_body_id)
+        I18n.with_locale(I18n.default_locale) do
+          found_pb.name = "Internal admin authority"
+          found_pb.save
+        end
+        found_pb
+      else
+        I18n.with_locale(I18n.default_locale) do
+          PublicBody.
+            create!(:name => 'Internal admin authority',
+                    :short_name => "",
+                    :request_email => AlaveteliConfiguration.contact_email,
+                    :home_page => "",
+                    :notes => "",
+                    :publication_scheme => "",
+                    :last_edit_editor => "internal_admin",
+                    :last_edit_comment =>
+                      "Made by PublicBody.internal_admin_body")
+        end
       end
     when matching_pbs.length == 1 then
       matching_pbs[0]
--- a/db/migrate/20170626122608_set_public_body_translation_defaults.rb
+++ b/db/migrate/20170626122608_set_public_body_translation_defaults.rb
@@ -1,0 +1,20 @@
+# -*- encoding : utf-8 -*-
+class SetPublicBodyTranslationDefaults < ActiveRecord::Migration
+  def up
+    change_column :public_body_translations,
+                  :short_name, :text, :null => false, :default => ""
+    change_column :public_body_translations,
+                  :notes, :text, :null => false, :default => ""
+    change_column :public_body_translations,
+                  :publication_scheme, :text, :null => false, :default => ""
+  end
+
+  def down
+    change_column :public_body_translations,
+                  :short_name, :text, :null => false
+    change_column :public_body_translations,
+                  :notes, :text, :null => false
+    change_column :public_body_translations,
+                  :publication_scheme, :text, :null => false
+  end
+end
--- a/db/migrate/20170626141302_drop_public_body_translated_columns.rb
+++ b/db/migrate/20170626141302_drop_public_body_translated_columns.rb
@@ -1,0 +1,54 @@
+# -*- encoding : utf-8 -*-
+class DropPublicBodyTranslatedColumns < ActiveRecord::Migration
+  def up
+    remove_column :public_bodies, :name
+    remove_column :public_bodies, :short_name
+    remove_column :public_bodies, :request_email
+    remove_column :public_bodies, :url_name
+    remove_column :public_bodies, :notes
+    remove_column :public_bodies, :first_letter
+    remove_column :public_bodies, :publication_scheme
+  end
+
+  def down
+    add_column :public_bodies, :name, :text
+    add_column :public_bodies, :short_name, :text
+    add_column :public_bodies, :request_email, :text
+    add_column :public_bodies, :url_name, :text
+    add_column :public_bodies, :notes, :text
+    add_column :public_bodies, :first_letter, :string
+    add_column :public_bodies, :publication_scheme, :text
+
+    # Migrate the data back - wrapped in its own transaction so the data will
+    # be updated before the constraints are reapplied
+    PublicBody.transaction do
+      PublicBody.find_each do |pb|
+        translated = pb.translation_for(I18n.default_locale)
+
+        # Create a hash containing the translated column names and their values
+        pb.translated_attribute_names.inject(fields_to_update={}) do |f, name|
+          f.update({name.to_sym => translated[name.to_s]})
+        end
+
+        # Now, update the actual model's record with the hash (using the
+        # ActiveRecord::Relation method update_all to force the use of an
+        # UPDATE statement rather than pb.update_atrributes which will
+        # use the overridden attribute setters and update translations instead)
+        PublicBody.where(:id => pb.id).update_all(fields_to_update)
+      end
+    end
+
+    # Re-add the constraints
+    change_column :public_bodies,
+                  :name, :text, :null => false
+    change_column :public_bodies,
+                  :short_name, :text, :null => false, :default => ""
+    change_column :public_bodies, :request_email, :text, :null => false
+    change_column :public_bodies, :url_name, :text, :null => false
+    change_column :public_bodies,
+                  :notes, :text, :null => false, :default => ""
+    change_column :public_bodies, :first_letter, :string, :null => false
+    change_column :public_bodies,
+                  :publication_scheme, :text, :null => false, :default => ""
+  end
+end
--- a/db/migrate/20170626141302_drop_public_body_translated_columns.rb
+++ b/db/migrate/20170626141302_drop_public_body_translated_columns.rb
@@ -1,6 +1,26 @@
 # -*- encoding : utf-8 -*-
 class DropPublicBodyTranslatedColumns < ActiveRecord::Migration
   def up
+    PublicBody.transaction do
+      PublicBody.find_each do |record|
+        translation = record.translation_for(I18n.default_locale) ||
+          record.translations.build(:locale => I18n.default_locale)
+
+        if translation.new_record?
+          id = record.id
+          puts "Creating default locale translation for public body #{ id }"
+
+          fields = record.translated_attribute_names
+          fields.each do |attribute_name|
+            translation[attribute_name] =
+              record.read_attribute(attribute_name, :translated => false)
+          end
+
+          translation.save!
+        end
+      end
+    end
+
     remove_column :public_bodies, :name
     remove_column :public_bodies, :short_name
     remove_column :public_bodies, :request_email
@@ -22,19 +42,26 @@ class DropPublicBodyTranslatedColumns < ActiveRecord::Migration
     # Migrate the data back - wrapped in its own transaction so the data will
     # be updated before the constraints are reapplied
     PublicBody.transaction do
-      PublicBody.find_each do |pb|
-        translated = pb.translation_for(I18n.default_locale)
+      PublicBody.find_each do |record|
+        translated = record.translation_for(I18n.default_locale)
+        translated = record.translations.first unless translated.persisted?
+
+        if translated.new_record? || translated.nil?
+          puts "No translations for public body #{record.id}"
+        end
 
         # Create a hash containing the translated column names and their values
-        pb.translated_attribute_names.inject(fields_to_update={}) do |f, name|
+        attr_names = record.translated_attribute_names
+        attr_names.inject(fields_to_update={}) do |f, name|
           f.update({name.to_sym => translated[name.to_s]})
         end
 
         # Now, update the actual model's record with the hash (using the
         # ActiveRecord::Relation method update_all to force the use of an
-        # UPDATE statement rather than pb.update_atrributes which will
+        # UPDATE statement rather than record.update_attributes which will
         # use the overridden attribute setters and update translations instead)
-        PublicBody.where(:id => pb.id).update_all(fields_to_update)
+        puts "Migrating default locale translation to public body #{record.id}"
+        PublicBody.where(:id => record.id).update_all(fields_to_update)
       end
     end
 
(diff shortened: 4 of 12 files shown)


db/migrate/20170626122608_set_public_body_translation_defaults.rb
```diff
@@ -0,0 +1,19 @@
+class SetPublicBodyTranslationDefaults < ActiveRecord::Migration
```
garethrees: Its there, so deleting

app/models/public_body.rb
```diff
@@ -129,6 +129,17 @@ class Translation
   #
   # [1] http://git.io/vIetK
   class Version
+    before_save :copy_translated_attributes
```
garethrees: squash in to the previous commit

db/migrate/20170626122608_set_public_body_translation_defaults.rb
```diff
@@ -0,0 +1,20 @@
+# -*- encoding : utf-8 -*-
+class SetPublicBodyTranslationDefaults < ActiveRecord::Migration
+  def up
+    change_column :public_body_translations,
+                  :short_name, :text, :null => false, :default => ""
+    change_column :public_body_translations,
+                  :notes, :text, :null => false, :default => ""
+    change_column :public_body_translations,
+                  :publication_scheme, :text, :null => false, :default => ""
+  end
```
garethrees: `request_email` [should be](https://github.com/mysociety/alaveteli/blob/develop/app/models/public_body.rb#L9)  `:null => false`?

garethrees: `name` [should be](https://github.com/mysociety/alaveteli/blob/develop/app/models/public_body.rb#L7)  `:null => false`?

garethrees: `url_name` [should be](https://github.com/mysociety/alaveteli/blob/develop/app/models/public_body.rb#L15)  `:null => false`?

garethrees: `first_letter` [should be](https://github.com/mysociety/alaveteli/blob/develop/app/models/public_body.rb#L18)  `:null => false`?

garethrees: `disclosure_log` [should be](https://github.com/mysociety/alaveteli/blob/develop/app/models/public_body.rb#L22)  `:null => false, :default => ''`?

app/models/public_body.rb
```diff
@@ -306,21 +325,31 @@ def calculated_home_page
...
-                           :home_page => "",
-                           :notes => "",
-                           :publication_scheme => "",
-                           :last_edit_editor => "internal_admin",
-                           :last_edit_comment => "Made by PublicBody.internal_admin_body")
+      # "internal admin" exists but has the wrong default locale - fix & return
+      if invalid_locale = PublicBody::Translation.
+                            find_by_url_name('internal_admin_authority')
+        found_pb = PublicBody.find(invalid_locale.public_body_id)
+        I18n.with_locale(I18n.default_locale) do
```
lizconlan: might need to make this `I18n.with_locale(FastGettext.locale) do` as using Belgium's settings when working with translations, this could accidentally use the wrong locale format...

```rb
translation = PublicBodyHeading::Translation.new(:locale => I18n.default_locale)
=> #<PublicBodyHeading::Translation id: nil, public_body_heading_id: nil, locale: "fr-BE", created_at: nil, updated_at: nil, name: nil>
```

garethrees: ugh maybe we need to make a wrapper around all these so we always do `Alaveteli.default_locale` etc.

db/migrate/20170626141302_drop_public_body_translated_columns.rb
```diff
@@ -0,0 +1,54 @@
...
+    remove_column :public_bodies, :name
+    remove_column :public_bodies, :short_name
+    remove_column :public_bodies, :request_email
+    remove_column :public_bodies, :url_name
+    remove_column :public_bodies, :notes
+    remove_column :public_bodies, :first_letter
+    remove_column :public_bodies, :publication_scheme
+  end
+
+  def down
```
garethrees: I've just run this (`db:rollback`) on a fresh dev db.

```
git checkout develop
be rake db:drop db:create db:migrate
script/load-sample-data

git checkout 3800-globalize
be rake db:migrate
be rake db:rollback
```

I get errors:

```
$ be rake db:rollback
== 20170626141302 DropPublicBodyTranslatedColumns: reverting ==================
-- add_column(:public_bodies, :name, :text)
   -> 0.0024s
-- add_column(:public_bodies, :short_name, :text)
   -> 0.0014s
-- add_column(:public_bodies, :request_email, :text)
   -> 0.0011s
-- add_column(:public_bodies, :url_name, :text)
   -> 0.0011s
-- add_column(:public_bodies, :notes, :text)
   -> 0.0010s
-- add_column(:public_bodies, :first_letter, :string)
   -> 0.0010s
-- add_column(:public_bodies, :publication_scheme, :text)
   -> 0.0011s
-- change_column(:public_bodies, :name, :text, {:null=>false})
rake aborted!
StandardError: An error has occurred, this and all later migrations canceled:

PG::NotNullViolation: ERROR:  column "name" contains null values
: ALTER TABLE "public_bodies" ALTER "name" SET NOT NULL
/home/vagrant/bundle/ruby/1.9.1/gems/activerecord-4.1.16/lib/active_record/connection_adapters/postgresql/database_statements.rb:128:in `async_exec'
/home/vagrant/bundle/ruby/1.9.1/gems/activerecord-4.1.16/lib/active_record/connection_adapters/postgresql/database_statements.rb:128:in `block in execute'
/home/vagrant/bundle/ruby/1.9.1/gems/activerecord-4.1.16/lib/active_record/connection_adapters/abstract_adapter.rb:378:in `block in log'
/home/vagrant/bundle/ruby/1.9.1/gems/activesupport-4.1.16/lib/active_support/notifications/instrumenter.rb:20:in `instrument'
/home/vagrant/bundle/ruby/1.9.1/gems/activerecord-4.1.16/lib/active_record/connection_adapters/abstract_adapter.rb:372:in `log'
/home/vagrant/bundle/ruby/1.9.1/gems/activerecord-4.1.16/lib/active_record/connection_adapters/postgresql/database_statements.rb:127:in `execute'
/home/vagrant/bundle/ruby/1.9.1/gems/activerecord-4.1.16/lib/active_record/connection_adapters/postgresql/schema_statements.rb:427:in `change_column_null'
/home/vagrant/bundle/ruby/1.9.1/gems/activerecord-4.1.16/lib/active_record/connection_adapters/postgresql/schema_statements.rb:410:in `change_column'
/home/vagrant/bundle/ruby/1.9.1/gems/activerecord-4.1.16/lib/active_record/migration.rb:656:in `block in method_missing'
/home/vagrant/bundle/ruby/1.9.1/gems/activerecord-4.1.16/lib/active_record/migration.rb:628:in `block in say_with_time'
/home/vagrant/bundle/ruby/1.9.1/gems/activerecord-4.1.16/lib/active_record/migration.rb:628:in `say_with_time'
/home/vagrant/bundle/ruby/1.9.1/gems/activerecord-4.1.16/lib/active_record/migration.rb:648:in `method_missing'
/home/vagrant/alaveteli/db/migrate/20170626141302_drop_public_body_translated_columns.rb:42:in `down'
/home/vagrant/bundle/ruby/1.9.1/gems/activerecord-4.1.16/lib/active_record/migration.rb:605:in `exec_migration'
/home/vagrant/bundle/ruby/1.9.1/gems/activerecord-4.1.16/lib/active_record/migration.rb:586:in `block (2 levels) in migrate'
/home/vagrant/bundle/ruby/1.9.1/gems/activerecord-4.1.16/lib/active_record/migration.rb:585:in `block in migrate'
/home/vagrant/bundle/ruby/1.9.1/gems/activerecord-4.1.16/lib/active_record/connection_adapters/abstract/connection_pool.rb:294:in `with_connection'
/home/vagrant/bundle/ruby/1.9.1/gems/activerecord-4.1.16/lib/active_record/migration.rb:584:in `migrate'
/home/vagrant/bundle/ruby/1.9.1/gems/activerecord-4.1.16/lib/active_record/migration.rb:759:in `migrate'
/home/vagrant/bundle/ruby/1.9.1/gems/activerecord-4.1.16/lib/active_record/migration.rb:998:in `block in execute_migration_in_transaction'
/home/vagrant/bundle/ruby/1.9.1/gems/activerecord-4.1.16/lib/active_record/migration.rb:1044:in `block in ddl_transaction'
/home/vagrant/bundle/ruby/1.9.1/gems/activerecord-4.1.16/lib/active_record/connection_adapters/abstract/database_statements.rb:201:in `block in transaction'
/home/vagrant/bundle/ruby/1.9.1/gems/activerecord-4.1.16/lib/active_record/connection_adapters/abstract/database_statements.rb:209:in `within_new_transaction'
/home/vagrant/bundle/ruby/1.9.1/gems/activerecord-4.1.16/lib/active_record/connection_adapters/abstract/database_statements.rb:201:in `transaction'
/home/vagrant/bundle/ruby/1.9.1/gems/activerecord-4.1.16/lib/active_record/transactions.rb:208:in `transaction'
/home/vagrant/bundle/ruby/1.9.1/gems/activerecord-4.1.16/lib/active_record/migration.rb:1044:in `ddl_transaction'
/home/vagrant/bundle/ruby/1.9.1/gems/activerecord-4.1.16/lib/active_record/migration.rb:997:in `execute_migration_in_transaction'
/home/vagrant/bundle/ruby/1.9.1/gems/activerecord-4.1.16/lib/active_record/migration.rb:959:in `block in migrate'
/home/vagrant/bundle/ruby/1.9.1/gems/activerecord-4.1.16/lib/active_record/migration.rb:955:in `each'
/home/vagrant/bundle/ruby/1.9.1/gems/activerecord-4.1.16/lib/active_record/migration.rb:955:in `migrate'
/home/vagrant/bundle/ruby/1.9.1/gems/activerecord-4.1.16/lib/active_record/migration.rb:821:in `down'
/home/vagrant/bundle/ruby/1.9.1/gems/activerecord-4.1.16/lib/active_record/migration.rb:910:in `move'
/home/vagrant/bundle/ruby/1.9.1/gems/activerecord-4.1.16/lib/active_record/migration.rb:803:in `rollback'
/home/vagrant/bundle/ruby/1.9.1/gems/activerecord-4.1.16/lib/active_record/railties/databases.rake:119:in `block (2 levels) in <top (required)>'
/home/vagrant/bundle/ruby/1.9.1/gems/rake-12.0.0/exe/rake:27:in `<top (required)>'
Tasks: TOP => db:rollback
(See full trace by running task with --trace)
```

garethrees: Okay here's what I think we should do:

- Add a data migration in the `up` migration, creating a default locale translation _only if_ one doesn't already exist. This should be a no-op but it covers any edge cases that might have happened.
- If we can't find a translation for the default locale (shouldn't be a problem, given the above) fall back to the first translation we can find so that the public body constraints are met.

```diff
diff --git a/db/migrate/20170626141302_drop_public_body_translated_columns.rb b/db/migrate/20170626141302_drop_public_body_translated_columns.rb
index 060cc54..006973b 100644
--- a/db/migrate/20170626141302_drop_public_body_translated_columns.rb
+++ b/db/migrate/20170626141302_drop_public_body_translated_columns.rb
@@ -1,6 +1,26 @@
 # -*- encoding : utf-8 -*-
 class DropPublicBodyTranslatedColumns < ActiveRecord::Migration
   def up
+    PublicBody.transaction do
+      PublicBody.find_each do |record|
+        translation = record.translation_for(I18n.default_locale) ||
+          record.translations.build(:locale => I18n.default_locale)
+
+        if translation.new_record?
+          id = record.id
+          puts "Creating default locale translation for public body #{ id }"
+
+          fields = record.translated_attributes
+          fields.each do |attribute_name, attribute_type|
+            translation[attribute_name] =
+              record.read_attribute(attribute_name, :translated => false)
+          end
+
+          translation.save!
+        end
+      end
+    end
+
     remove_column :public_bodies, :name
     remove_column :public_bodies, :short_name
     remove_column :public_bodies, :request_email
@@ -22,19 +42,25 @@ class DropPublicBodyTranslatedColumns < ActiveRecord::Migration
     # Migrate the data back - wrapped in its own transaction so the data will
     # be updated before the constraints are reapplied
     PublicBody.transaction do
-      PublicBody.find_each do |pb|
-        translated = pb.translation_for(I18n.default_locale)
+      PublicBody.find_each do |record|
+        translated = record.translation_for(I18n.default_locale)
+        translated = record.translations.first unless translated.persisted?
+
+        if translated.new_record? || translated.nil?
+          puts "No translations for public body #{record.id}"
+        end
 
         # Create a hash containing the translated column names and their values
-        pb.translated_attribute_names.inject(fields_to_update={}) do |f, name|
+        record.translated_attribute_names.inject(fields_to_update={}) do |f, name|
           f.update({name.to_sym => translated[name.to_s]})
         end
 
         # Now, update the actual model's record with the hash (using the
         # ActiveRecord::Relation method update_all to force the use of an
-        # UPDATE statement rather than pb.update_atrributes which will
+        # UPDATE statement rather than record.update_atrributes which will
         # use the overridden attribute setters and update translations instead)
-        PublicBody.where(:id => pb.id).update_all(fields_to_update)
+        puts "Migrating default locale translation to public body #{record.id}"
+        PublicBody.where(:id => record.id).update_all(fields_to_update)
       end
     end
```

lizconlan: Looks good, although I think we can tidy the `up` translation a bit by doing something like:

```diff
-          fields = record.translated_attributes
-          fields.each do |attribute_name, attribute_type|
+          attribute_names = record.translated_attributes_names
+          attribute_names.each do |attribute_name|
```

...as the `attribute_type` doesn't appear to have an effect

lizconlan: also I typoed "attributes" in the comment line that starts `# UPDATE` \*sigh\*

spec/models/public_body_spec.rb
```diff
@@ -192,6 +192,42 @@
...
+    end
+
+    it "repairs the internal_admin_body if the default locale has changed" do
+      iab = PublicBody.internal_admin_body
+      expect(iab).not_to be_nil
+
+      with_default_locale(:es) do
+        I18n.with_locale(:es) do
+          found_iab = PublicBody.internal_admin_body
+          expect(found_iab).to eq(iab)
```
lizconlan: you're right - it should demonstrate more clearly that:

a) initially `iab.translations.map(&:locale).include(:es)` is false
b) `iab.translations.map(&:locale).include(:es)` is true
c) (as current) it's the same internal_admin_body we first thought of (or at least that there's only one of them, we haven't created a fresh one in error)

db/migrate/20170626122608_set_public_body_translation_defaults.rb
```diff
@@ -0,0 +1,20 @@
...
+
+  def down
+    change_column :public_body_translations,
+                  :short_name, :text, :null => false
```
lizconlan: yes! good spot

app/controllers/admin_public_body_controller.rb
```diff
@@ -142,10 +142,12 @@ def mass_tag_add
   def missing_scheme
     # There might be a way to do this in ActiveRecord, but I can't find it
     @public_bodies = PublicBody.find_by_sql("
-            SELECT a.id, a.name, a.url_name, COUNT(*) AS howmany
```
lizconlan: hard to say for sure without running the entire test suite but the controller spec doesn't catch it so, at it's simplest, something like...

```rb
  describe 'GET #missing_scheme' do

    it 'returns successfully' do
      get :missing_scheme
      expect(response).to be_success
    end

  end
```

added to `spec/controllers/admin_public_body_controller_spec.rb` will trip it up

lizconlan: (but not until the columns are dropped)

Also, the controller code needs:

```diff
-    @stats = {
-      "total" => PublicBody.count,
-      "entered" => PublicBody.where("publication_scheme != ''").count
-    }
+    @stats = {
+      "total" => PublicBody.count,
+      "entered" => PublicBody.
+                    with_translations(FastGettext.locale).
+                     where("public_body_translations.publication_scheme != ''").
+                     count
+    }

File: db/migrate/20170626122608_set_public_body_translation_defaults.rb (hunk 3 above)
Comment (garethrees) [inline db/migrate/20170626122608_set_public_body_translation_defaults.rb]: Annoyingly we have a load of data that would make this migration fail:

```sql
SELECT COUNT(*) FROM public_body_translations WHERE name IS NULL;
 count
-------
   268

SELECT COUNT(*) FROM public_body_translations WHERE short_name IS NULL;
 count
-------
   308
```

I think we need a rake task to clean up existing records, _then_ apply these migrations.

app/controllers/admin_public_body_controller.rb
garethrees: So:

- no specs fail when the suite is run before this change is made
- adding a simple `expect(response).to be_success` spec illustrates a failure exists
- there's a problem with a different query in this action
- you believe there to be a problem with the `find_by_sql` query, but there aren't any specific specs to demonstrate its behaviour?

Is my understanding of this correct?

If so, I think plan of action is:

- Push the `find_by_sql` in to `PublicBody.missing_scheme` and add some tests
- Add specs for `AdminPublicBodyController#missing_scheme`

…or we could just leave it broken / delete it because it's not linked from anywhere so presumably (we could answer this if we had https://github.com/mysociety/alaveteli/issues/2945) isn't used.

db/migrate/20170626122608_set_public_body_translation_defaults.rb
lizconlan: \*sigh\*

```rb
PublicBody::Translation.where(name: nil).map(&:locale).uniq
=> [:cy]
PublicBody::Translation.where(short_name: nil).map(&:locale).uniq
=> [:cy]
```

Maybe we also need to check what creating a new PublicBody via the admin interface is doing as these don't look like old data.

```rb
PublicBody::Translation.where(name: nil).last.created_at
=> Wed, 28 Jun 2017 13:21:48 BST +01:00
```

garethrees: Presumably creates and saves a translation for each locale when you submit the form. Because we don't have the same defaults and constraints on the `public_body_translations` table we get the mismatch.

lizconlan: Although the most recently created PublicBody was from 6 Feb 2017

```rb
PublicBody::Translation.
  where(name: nil).
    joins("JOIN public_bodies ON public_body_translations.public_body_id = public_bodies.id").
      order('public_bodies.created_at desc').
        first.
          public_body_id
```

lizconlan: **UNLESS** putting NULL constraints on the translation table is a mistake...

app/controllers/admin_public_body_controller.rb
lizconlan: > Is my understanding of this correct?

I think so...
* 2 of the 3 queries in the [`missing_scheme` action](https://github.com/mysociety/alaveteli/blob/0.29.0.0/app/controllers/admin_public_body_controller.rb#L141-L155) break when we drop the translated columns from the `public_bodies` table (easily verifiable by copy-pasting the code into the console, and I've only just picked up on the second failure)
* there are no tests for this action
* it's an admin action - if we don't link to it, maybe it's not being used

app/models/public_body.rb
```diff
@@ -202,6 +202,14 @@ def set_api_key!
     self.api_key = SecureRandom.base64(33)
   end
 
+  def self.find_by_name(name)
+    find_by(name: name)
```
garethrees: Check that this isn't problematic as per https://github.com/spree-contrib/spree_globalize/issues/21

spec/models/public_body_spec.rb
garethrees: Fixed

spec/models/public_body_spec.rb
```diff
@@ -170,6 +170,28 @@
...
+      expect(public_body.first_letter).to be_nil
+    end
+
+    it 'handles mutlibyte characters correctly' do
+      public_body = PublicBody.new(:name => 'Åccented')
+      public_body.set_first_letter
+      expect(public_body.first_letter).to eq('Å')
+    end
+
+  end
```
garethrees: Fixed

app/models/public_body.rb
lizconlan: Possibly related, in which case we should be ok https://github.com/globalize/globalize/issues/458

garethrees: Not sure if we need this method – doesn't appear to be called anywhere in `app/`

lizconlan:
It looks as though our guess at how Globalize behaves with regard to nil translated values was correct, if the translation has a nil (but not a blank) value, it returns the equivalent value from the default locale instead:

```rb

I18n.default_locale
# => :fr
pb = PublicBody.find(9)
pb.translations.where(locale: :en)
# => #<ActiveRecord::AssociationRelation
# [#<PublicBody::Translation id: 14, public_body_id: 9, locale: "en", 
#                            created_at: "2017-07-04 15:29:49", updated_at: "2017-07-04 15:40:08",
#                            name: nil, short_name: "testfr", request_email: "testfr@localhost",
#                            url_name: "testfr", notes: "", first_letter: "T", publication_scheme: "",
#                            disclosure_log: nil>]>

I18n.with_locale(I18n.default_locale) { pb.name }
# => "Test fr"

I18n.with_locale(:en) { pb.name }
# => "Test fr"

I18n.with_locale(:en) { pb.name  = ''; pb.save! }
I18n.with_locale(:en) { pb.name }
# => ""
```

(Note, this cheats heavily for testing purposes - the translated columns have been dropped and the validations for not nil and uniqueness have been disabled)

app/models/public_body.rb
```diff
@@ -314,21 +314,31 @@ def calculated_home_page
...
-                           :publication_scheme => "",
-                           :last_edit_editor => "internal_admin",
-                           :last_edit_comment => "Made by PublicBody.internal_admin_body")
+      # "internal admin" exists but has the wrong default locale - fix & return
+      if invalid_locale = PublicBody::Translation.
+                            find_by_url_name('internal_admin_authority')
+        found_pb = PublicBody.find(invalid_locale.public_body_id)
+        I18n.with_locale(I18n.default_locale) do
+          found_pb.name = "Internal admin authority"
+          found_pb.save
```
garethrees: This save doesn't handle failure

garethrees:
Superseded by https://github.com/mysociety/alaveteli/pull/4107[](url)

PR #4070, opened 2017-06-27
Log embargo expiring events against info requests

stevenday:
This adds a new type of event 'expiring_embargo' which gets logged when an embargo is about to expire, paving the way for us to notify people about those events.


stevenday:
I've pushed this on top of https://github.com/mysociety/alaveteli/pull/4046 because it's a logical follow on, but I've set the merge base to develop so that this gets tested.

The real action is in 7ea5af2 and onwards.

stevenday:
@crowbot - I've rebased this, [pushed a fixup](https://github.com/mysociety/alaveteli/pull/4070/commits/3804a70bc3fda97b682b1591f056c14a9234079a) to the task name and [made the migration a fixup to the earlier commit](https://github.com/mysociety/alaveteli/pull/4070/commits/4ffc3df8b21e8f1c2a47f3c972f854800ceb379a) as [you suggested](https://github.com/mysociety/alaveteli/pull/4070#discussion_r125078005) so I think it's ready for another look.

lib/tasks/stats.rake
```diff
@@ -262,6 +262,12 @@ namespace :stats do
     InfoRequest.log_very_overdue_events
   end
 
+  desc 'Add InfoRequestEvents for the points when requests embargoes began to
+        expire'
+  task :update_expiring_embargo_info_request_events => :environment do
+    Embargo.log_expiring_events
```
crowbot: Sorry, late comment - looks like this needs the `AlaveteliPro` namespace.

stevenday: Oops, thanks, I've fixed that and rebased.

PR #4073, opened 2017-06-28
Fix handling of underscores in default_locale

lizconlan:
This fixes some bugs with setting an underscore locale as the default_locale for a site. This also now uses `AlaveteliLocalization.locale` wherever we were previously using `I18n.locale` although the tests suggest this wasn't causing any problems with the current code so this is mostly for consistency.

The 2 (independent) problems are: 
1. the custom routing filter didn't have an override for a new method that checked against `I18n.default_locale` instead of `FastGettext.default_locale`, causing the underscore locale to be added to the URL
1. we're mostly using `I18n` instead of `FastGettext` versions of locales which means we're using hyphenated rather than underscore locales (or defensively pasting `.to_s.gsub` where there are key uses and it's not clear what's being passed in)

This PR provides wrapper methods in `AlaveteliLocalization` and replaces current uses of `default_locale`, `locale` and `available_locales` with the new wrapper methods.

Fixes mysociety/belgium-theme#54
Connects to mysociety/belgium-theme#57

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -69,7 +69,7 @@ class ApplicationController < ActionController::Base
   def set_gettext_locale
     if AlaveteliConfiguration::include_default_locale_in_urls == false
       params_locale = params.fetch(:locale) do
-        AlaveteliLocalization.default_locale.to_s
+        AlaveteliLocalization.default_locale
       end
     else
       params_locale = params[:locale]
@@ -78,11 +78,11 @@ class ApplicationController < ActionController::Base
       requested_locale = params_locale || session[:locale] ||
                            cookies[:locale] ||
                            request.env['HTTP_ACCEPT_LANGUAGE'] ||
-                           AlaveteliLocalization.default_locale.to_s
+                           AlaveteliLocalization.default_locale
     else
       requested_locale = params_locale || session[:locale] ||
                            cookies[:locale] ||
-                           AlaveteliLocalization.default_locale.to_s
+                           AlaveteliLocalization.default_locale
     end
     requested_locale = FastGettext.best_locale_in(requested_locale)
     # set the currrent locale to the requested_locale
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -63,34 +63,32 @@ class ApplicationController < ActionController::Base
     anonymous_cache(24.hours)
   end
 
-  # This is an override of the method provided by gettext_i18n_rails - note the explicit
-  # setting of I18n.locale, required due to the I18nProxy used in Rails 3 to trigger the
+  # This is an override of the method provided by gettext_i18n_rails.
+  # AlaveteliLocalization.set_session_locale explicitly sets I18n.locale,
+  # required due to the I18nProxy used in Rails to trigger the
   # lookup_context and expire the template cache
   def set_gettext_locale
-    if AlaveteliConfiguration::include_default_locale_in_urls == false
+    if AlaveteliConfiguration.include_default_locale_in_urls == false
       params_locale = params.fetch(:locale) do
         AlaveteliLocalization.default_locale
       end
     else
       params_locale = params[:locale]
     end
-    if AlaveteliConfiguration::use_default_browser_language
-      requested_locale = params_locale || session[:locale] ||
-                           cookies[:locale] ||
-                           request.env['HTTP_ACCEPT_LANGUAGE'] ||
-                           AlaveteliLocalization.default_locale
+    browser_locale = if AlaveteliConfiguration.use_default_browser_language
+      request.env['HTTP_ACCEPT_LANGUAGE']
     else
-      requested_locale = params_locale || session[:locale] ||
-                           cookies[:locale] ||
-                           AlaveteliLocalization.default_locale
+      nil
     end
-    requested_locale = FastGettext.best_locale_in(requested_locale)
+    AlaveteliLocalization.set_session_locale(params_locale,
+                                             session[:locale],
+                                             cookies[:locale],
+                                             browser_locale)
     # set the currrent locale to the requested_locale
-    session[:locale] = FastGettext.set_locale(requested_locale)
-    I18n.locale = FastGettext.locale.to_s.gsub("_", "-")
+    session[:locale] = AlaveteliLocalization.locale
     if !@user.nil?
-      if @user.locale != requested_locale
-        @user.locale = session[:locale]
+      if @user.locale != AlaveteliLocalization.locale
+        @user.locale = AlaveteliLocalization.locale
         @user.save!
       end
     end
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -68,17 +68,26 @@ class ApplicationController < ActionController::Base
   # lookup_context and expire the template cache
   def set_gettext_locale
     if AlaveteliConfiguration::include_default_locale_in_urls == false
-      params_locale = params[:locale] ? params[:locale] : I18n.default_locale
+      params_locale = params.fetch(:locale) do
+        AlaveteliLocalization.default_locale.to_s
+      end
     else
       params_locale = params[:locale]
     end
     if AlaveteliConfiguration::use_default_browser_language
-      requested_locale = params_locale || session[:locale] || cookies[:locale] || request.env['HTTP_ACCEPT_LANGUAGE'] || I18n.default_locale
+      requested_locale = params_locale || session[:locale] ||
+                           cookies[:locale] ||
+                           request.env['HTTP_ACCEPT_LANGUAGE'] ||
+                           AlaveteliLocalization.default_locale.to_s
     else
-      requested_locale = params_locale || session[:locale] || cookies[:locale] || I18n.default_locale
+      requested_locale = params_locale || session[:locale] ||
+                           cookies[:locale] ||
+                           AlaveteliLocalization.default_locale.to_s
     end
     requested_locale = FastGettext.best_locale_in(requested_locale)
-    session[:locale] = I18n.locale = FastGettext.set_locale(requested_locale)
+    # set the currrent locale to the requested_locale
+    session[:locale] = FastGettext.set_locale(requested_locale)
+    I18n.locale = FastGettext.locale.to_s.gsub("_", "-")
     if !@user.nil?
       if @user.locale != requested_locale
         @user.locale = session[:locale]
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -517,9 +517,9 @@ class ApplicationController < ActionController::Base
   #
   # Returns a Hash
   def collect_locales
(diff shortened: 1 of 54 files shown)


garethrees:
Could we monkeypatch `I18n.default_locale` with a warning, so that if we use the former the developer gets a notice that they shouldn't?

```ruby
I18n.instance_eval do
  def self.default_locale
    warn "Use FastGettext.default_locale…"
    super
  end
end

**Comment by lizconlan** [comment]
> Could we monkeypatch I18n.default_locale with a warning

I like this idea but could there be a case when using `I18n.default_locale` is correct?

**Comment by lizconlan** [comment]
From https://github.com/mysociety/alaveteli/pull/4066/files#r124752221

> **garethrees**
ugh maybe we need to make a wrapper around all these so we always do `Alaveteli.default_locale` etc.

**Comment by garethrees** [comment]
FYI I've just noticed wdtk-theme has a routing filter override – worth a check for compatibility https://github.com/mysociety/whatdotheyknow-theme/blob/master/lib/routingfilter.rb

**Comment by lizconlan** [comment]
> FYI I've just noticed wdtk-theme has a routing filter override

That looks ok - it's overriding a method on the underlying RoutingFilter::Filter which is not altered by LocaleFilter and doesn't appear to have changed:
https://github.com/svenfuchs/routing-filter/blob/0-4-stable/lib/routing_filter/filter.rb#L27-L30

File: spec/spec_helper.rb [hunk 1]
```diff
@@ -225,14 +225,18 @@ def with_hidden_and_successful_requests
...
 # is also cleared from the fallbacks
 def with_default_locale(locale)
-  original_default_locale = I18n.default_locale
-  original_fallbacks = I18n.fallbacks
-  I18n.fallbacks = nil
-  I18n.default_locale = locale
+  original_default_locale = AlaveteliLocalization.default_locale
+  original_fallbacks = AlaveteliLocalization.available_locales.join(" ")
+  AlaveteliLocalization.set_locales(
+      locale.to_s,
```
garethrees: Oh, [it does](https://github.com/mysociety/alaveteli/blob/54c139ba6e258d1c07a36a24f28fb703454364f1/lib/alaveteli_localization.rb#L7-L11) for `available_locales`.

Looks like this could be:

```ruby
AlaveteliLocalization.set_locales(locale, locale)
```

lib/alaveteli_localization.rb
```diff
@@ -4,10 +4,12 @@ class << self
...
-      I18n.locale = default_locale.gsub("_", '-').to_sym
-      I18n.available_locales = available_locales.map { |locale_name| locale_name.gsub("_", '-').to_sym }
-      I18n.default_locale = default_locale.gsub("_", '-').to_sym
+      FastGettext.default_available_locales = available_locales.
+                                                map { |x| x.to_sym }
+      I18n.available_locales = available_locales.map do |locale_name|
+        locale_name.gsub("_", '-').to_sym
+      end
+      I18n.locale = I18n.default_locale = default_locale.to_s.gsub("_", '-')
       FastGettext.default_locale = default_locale
```
garethrees: Make sure `default_locale` is the correct type (call `to_sym`?)

app/controllers/public_body_controller.rb
```diff
@@ -99,7 +99,7 @@ def view_email
     @public_body = PublicBody.find_by_url_name_with_historic(params[:url_name])
     raise ActiveRecord::RecordNotFound.new("None found") if @public_body.nil?
 
-    I18n.with_locale(I18n.locale.to_s) do
+    I18n.with_locale(AlaveteliLocalization.locale.to_s) do
```
garethrees: I think we should wrap `with_locale` too, as its not super obvious where to use `AlaveteliLocalization` vs `I18n` vs `FastGettext`.

spec/lib/alaveteli_localization_spec.rb
```diff
@@ -0,0 +1,78 @@
...
+    it 'sets FastGettext.locale' do
+      expect(FastGettext.locale).to eq("en_GB")
+    end
+
+    it 'sets FastGettext.default_locale' do
+      expect(FastGettext.default_locale).to eq("en_GB")
+    end
+
+    it 'sets FastGettext.default_available_locales' do
+      expect(FastGettext.default_available_locales).to eq([:en_GB, :es])
```
lizconlan: The mistake there may be checking `available_default_locales` instead of `default_locales` - one returns an array of strings, the other an array of symbols.

(*wait*  as these are already handily formatted as symbols they're **returned as-is** as `AlaveteliLocalization.available_locales` so maybe this test is fine but would benefit from a comment. And I would benefit from some more coffee. So not a stupid question, no.)

To answer a broader question, it seems to be that `FastGettext` returns strings whereas `I18n` returns symbols, which is why historically we were casting everything to string (or to symbol) before taking any action to avoid type errors as it was hard to know where `locale` had come from at any given point.

```
2.3.0 :001 > FastGettext.default_available_locales
 => [:en, :es]
2.3.0 :002 > FastGettext.available_locales
 => ["en", "es"]
2.3.0 :003 > FastGettext.locale
 => "en"
2.3.0 :004 > I18n.locale
 => :en
```

From https://github.com/grosser/gettext_i18n_rails/blob/v1.1.0/spec/gettext_i18n_rails_spec.rb#L61-L64

```
it "links FastGettext with I18n locale" do
  FastGettext.locale = 'xx'
  I18n.locale.should == :xx
end
```

spec/lib/alaveteli_localization_spec.rb
```diff
@@ -0,0 +1,78 @@
+# -*- encoding : utf-8 -*-
+require 'spec_helper'
+
+describe AlaveteliLocalization do
+
+  context '.set_locales' do
+
+    before do
+      AlaveteliLocalization.set_locales('en_GB es', 'en_GB')
```
lizconlan: I don't think it does. Replying as a note to self to check

spec/lib/alaveteli_localization_spec.rb
```diff
@@ -0,0 +1,78 @@
...
+  end
+
+  context '.locale' do
+
+    it 'returns the current locale' do
+      expect(AlaveteliLocalization.locale).to eq(:en)
+    end
+
+    it 'returns the locale in the undescore format' do
+      AlaveteliLocalization.set_locales('en_GB', 'en_GB')
```
lizconlan: that ... does not feel right

lizconlan:
ToDo list (from review feedback):

- [ ] add a `default_locale?` method and use throughout
- [ ] always return locales as strings (symbols were a nice idea but when we want to use them in URLs etc they need to be strings so we're wasting the tiny efficiency saving of symbols in constantly converting them)
- [ ] `set_locale` should automatically sort out hyphens and underscores (TBD - what, if anything, this should return)
- [ ] replace `FastGettext` calls in the app code with `AlaveteliLocalization` (to be fair to past me, some - but not all - of these have arrived via rebasing onto current develop)
- [ ] think about `#with_locale` wrapper

spec/lib/alaveteli_localization_spec.rb
lizconlan: can confirm, not persisted

db/migrate/20140716131107_create_category_translation_tables.rb
```diff
@@ -112,16 +112,16 @@ def down
     new_categories = []
     PublicBodyCategory.all.each do |category|
       category.locale = category.translation.locale.to_s
-      FastGettext.default_available_locales.each do |locale|
-        if locale.to_s != category.locale
+      AlaveteliLocalization.available_locales.each do |locale|
+        if locale != category.locale
```
lizconlan: shouldn't be, it gets set to `category.translation.locale.to_s` on line 114

garethrees: But won't it then get [converted to a Symbol](https://github.com/globalize/globalize/blob/v4.0.3/lib/globalize/active_record/translation.rb#L29) on the way back out in the call to `#locale`?

app/models/public_body_category.rb
```diff
@@ -56,7 +56,8 @@ def self.without_headings
   end
 
   def default_locale?
-    locale == I18n.default_locale
+    return false if locale.nil?
+    AlaveteliLocalization.default_locale?(locale)
```
lizconlan: handling it explicitly as `false` in case locale is ever purposely set to `""` as that will eval to `true` which is probably not what we want

app/views/admin_general/index.html.erb
```diff
@@ -54,12 +54,17 @@
...
+                 id: 'unclassified',
+                 parent: 'public-request-things-to-do',
+                 items: @old_unclassified,
+                 label:  "Classify responses that are still unclassified " \
+                         "#{ AlaveteliLocalization.with_locale(:en) do
+                               distance_of_time_in_words(
+                                 InfoRequest.OLD_AGE_IN_DAYS
+                               )
+                            }} after response"
+               } %>
```
lizconlan: I think it's ok - weirdly - as it needs to close the `#{` as well as the `do` (and then `locals: {`) but closing `do` with `}` is pretty hard to look at so adopting suggested fix 👍

garethrees: You can't close `do` with `}` AFAIK:

```ruby
irb(main):001:0> 10.times do |i|
irb(main):002:1* puts i
irb(main):003:1> }
SyntaxError: (irb):3: syntax error, unexpected '}', expecting keyword_end
        from /usr/bin/irb:12:in `<main>'
```

db/migrate/20140716131107_create_category_translation_tables.rb
lizconlan: need to take a closer look at that after rebasing onto `develop`, at present it's denying that either categories or headings have `.locale`

lizconlan: ok, at that point we've just reinstated the `locale` database column so `category.locale` is a `String` drawn from the database, not a translated attribute (until we do that, there is no `.locale` method)

db/migrate/20140716131107_create_category_translation_tables.rb
```diff
@@ -136,16 +136,16 @@ def down
     new_headings = []
     PublicBodyHeading.all.each do |heading|
       heading.locale = heading.translation.locale.to_s
-      FastGettext.default_available_locales.each do |locale|
-        if locale.to_s != heading.locale
+      AlaveteliLocalization.available_locales.each do |locale|
+        if locale != heading.locale
```
lizconlan: It's a `String`, it's not translated, it's being read from the database column we've just [re]created

app/controllers/application_controller.rb
```diff
@@ -63,25 +63,32 @@ def long_cache
     anonymous_cache(24.hours)
   end
 
-  # This is an override of the method provided by gettext_i18n_rails - note the explicit
-  # setting of I18n.locale, required due to the I18nProxy used in Rails 3 to trigger the
+  # This is an override of the method provided by note that
```
lizconlan: thanks, fixed!

PR #4075, opened 2017-06-30
Globalize minor tweaks

garethrees:
Uncontroversial commits extracted from #4066


garethrees:
Just going to merge this because I know it passes and this PR passes against most rubies, but has had a few memory-related fails on 1.9.3.

PR #4076, opened 2017-06-30
Public body validations specs only

garethrees:
Extracted from #4041 so that this PR only adds specs.

Will pull in the specs that add additional validations later.


garethrees:
Already reviewed in https://github.com/mysociety/alaveteli/pull/4041, so just waiting for a successful build.

PR #4081, opened 2017-07-03
Add a temporary sleep to batch request sending

stevenday:


app/models/info_request_batch.rb
```diff
@@ -60,6 +60,12 @@ def create_batch!
         else
           unrequestable << public_body
         end
+        # Sleep between requests in production, in case we're sending a huge
+        # batch which may result in a torrent of auto-replies coming back to
+        # us and overloading the server.
+        if Rails.env.production?
+          sleep 60
+        end
```
crowbot: I think we want to put the delay in the loop below that actually sends the requests, not this on which creates the objects? In fact though, if we're delaying sending the request after it's been created, there's a risk of it being indexed whilst it hasn't been sent, which will cause `last_event_forming_initial_request` to raise an error. How about something like 
```
    public_bodies.each do |public_body|
      if public_body.is_requestable?
        ActiveRecord::Base.transaction do
          created << create_request!(public_body)
        end
        outgoing_message = created.outgoing_messages.first
        outgoing_message.sendable?
        mail_message = OutgoingMailer.initial_request(outgoing_message.info_request, outgoing_message).deliver
        outgoing_message.record_email_delivery(mail_message.to_addrs.join(', '), mail_message.message_id)
        sleep 60
      else
        unrequestable << public_body
      end
      self.sent_at = Time.zone.now
      self.save!
    end
```

The commit comment when I initially separated the creation and sending into two loops was "wrap model creation in a transaction and do the message sending separately - we may ultimately want to do this outside the request cycle.", so I don't think re-merging them should cause any particular problems.

stevenday: D'oh, sorry, I missed that (I always think that just creating a request sends it, I'm not sure why).

In your example though, I think we've lost a feature of the previous code where if we have an error creating one request, the transaction will get rolled back and none of the requests will be created or sent. 

Without that, I think we're going to have issues if we ever have to resend a batch, because it won't be marked as sent, but some of the requests might have been?

Is it possible to create the request but stop xapian indexing it until we send the request, so we could move the sleep to the second loop?

crowbot: I don't think we can stop xapian indexing it easily as that happens in another process and as a side effect of request creation. We could move the `sent_at` setting into the transaction block. Then if a particular request causes an error, we should get an error message, and the batch should have been marked as sent if any previous request was successfully created. This is, in effect, the same situation we end up in when some requests are unsendable i.e. requests unsent = public bodies attached to the batch that don't have a request. What do you think?

stevenday: That works for me - I've just pushed an updated commit along those lines.

Issue #4068, opened 2017-06-27
[ERROR] [www.whatdotheyknow.com] user#signin (NoMethodError) "undefined method `[]' for nil:NilClass"


stevenday:
```
A NoMethodError occurred in user#signin:

  undefined method `[]' for nil:NilClass
  lib/alaveteli_pro/post_redirect_handler.rb:15:in `override_post_redirect_for_pro'

-------------------------------
Request:
-------------------------------

  * URL        : https://www.whatdotheyknow.com/profile/sign_in
  * HTTP Method: POST
  * IP address : FILTERED
  * Parameters : {"utf8"=>"✓", "authenticity_token"=>"", "user_signin"=>{"email"=>"peter.g.yeung@[GMAIL]", "password"=>"[FILTERED]"}, "token"=>"FILTERED", "modal"=>"", "commit"=>"Sign in", "controller"=>"user", "action"=>"signin"}
  * Timestamp  : 2017-06-27 13:11:46 +0100
  * Server : FILTERED
  * Rails root : /FILTERED/www.whatdotheyknow.com/alaveteli-2017-06-21T12-52-52
  * Process: FILTERED

-------------------------------
Backtrace:
-------------------------------

  lib/alaveteli_pro/post_redirect_handler.rb:15:in `override_post_redirect_for_pro'
  app/controllers/application_controller.rb:338:in `do_post_redirect'
  app/controllers/user_controller.rb:130:in `signin'
  app/controllers/application_controller.rb:108:in `record_memory'
  lib/strip_empty_sessions.rb:13:in `call'
```

stevenday:
I think what's happened here is that the user has created a new account, I've promoted it to pro and then they've written a request before confirming their account, and so the post redirect created for them is to confirm their account (and then go to /new) rather than to create the request.

I'm not exactly sure where in the controllers this is happening though, because it *seems* like the code should just update the existing post redirect, which would keep the details. Perhaps it is, but we lose the post_params along the way somewhere?

garethrees:
`authenticity_token` is blank, so that might have caused a problem

stevenday:
Hmm, I've pushed a spec to test my theory in https://github.com/mysociety/alaveteli/commit/e68edc4f869435bdc337ee5f63b0e20edcfc2541 but that doesn't seem to be an issue because the spec works fine.

Looking back through the logs some more, there's some weird stuff going on with this user's access - almost as if they're not allowing cookies to be set perhaps? 

[...]

PR #4082, opened 2017-07-04
Add an extra spec for pro signup

stevenday:
#4068 relates to an error (I thought) we saw when a pro user began a request before confirming their account, so I added a spec to test what was going on.

In the end, I think this was related to a cookies setting on their behalf, but I thought we might as well keep this spec anyway.


garethrees:
Closed this as we haven't done anything with it.

PR #4087, opened 2017-07-06
Persistent Vagrantfile config

garethrees:


Vagrantfile
```diff
@@ -57,11 +63,42 @@
...
-ALAVETELI_THEMES_DIR = ENV['ALAVETELI_THEMES_DIR'] || '../alaveteli-themes'
-ALAVETELI_OS = ENV['ALAVETELI_VAGRANT_OS'] || 'wheezy64'
-ALAVETELI_USE_NFS = ENV['ALAVETELI_VAGRANT_USE_NFS'] || false
+#
+# Defaults can be overridden either in `.vagrant.yml` with the same key name, or
+# via the environment by prefixing the key with `ALAVETELI_VAGRANT_` and
+# upcasing. Boolean values can be set to `false` in the environment with "0",
+# "false" or "no". The environment takes precedence over `.vagrant.yml`.
+DEFAULTS = {
+  'fqdn' => 'alaveteli.10.10.10.30.xip.io',
```
garethrees: Note that I used `String` keys here because that's what standard YAML keys (e.g. `fqdn:` get parsed as). To get `Symbol` keys you need to use `:fqdn:` which is not obvious at all, so we'll just stick with strings.

garethrees: I've also used "settings" instead of "config" because `config` is used in `Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|`, so it makes it pretty confusing.

davea:
Drive-by comment: Would a `.vagrant.yml-example` be useful to demonstrate the intended use?

Vagrantfile
```diff
@@ -33,18 +35,22 @@
 # This Vagrantfile allows customisation of some aspects of the virtaual machine
 # See the customization options below for details.
 #
-# The options can be set either by prefixing the vagrant command, or by
-# exporting to the environment.
+# The options can be set either by prefixing the vagrant command, using
+# `.vagrant.yml`, or by exporting to the environment.
```
lizconlan: worth documenting that the environment setting will win if both are set

Vagrantfile
```diff
@@ -33,18 +35,22 @@
...
 #
 #   # Prefixing the command
 #   $ ALAVETELI_VAGRANT_MEMORY=2048 vagrant up
 #
+#   # .vagrant.yml
+#   $ echo "memory: 2048" >> .vagrant.yml
```
lizconlan: having a yml dotfile seems pretty non-standard and makes it hard to edit or have an example file for

Vagrantfile
```diff
@@ -57,11 +63,43 @@
...
+end
+
+SETTINGS = DEFAULTS.merge(settings_file).merge(env).freeze
+
+if SETTINGS['show_settings']
+  puts 'Current machine settings:'
+  puts "\n"
+  pp SETTINGS
+  puts "\n"
+end
```
lizconlan: this is pretty cool - I like being able to see what settings I'm supposed to have :cake:

Vagrantfile
lizconlan: although maybe we could put an example as a block in the README file instead...

Vagrantfile
```diff
@@ -87,55 +125,58 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
     box
   end
   config.vm.box_url = box_url
-  config.vm.network :private_network, :ip => "10.10.10.30"
+  config.vm.network :private_network, ip: SETTINGS['ip']
 
-  if ALAVETELI_USE_NFS
-    config.vm.synced_folder ".", "/home/vagrant/alaveteli", :nfs => true
+  if SETTINGS['use_nfs']
```
lizconlan: Can't override this on the command line if it's on in the .vagrant.yml file (pretty sure this is how it was working before)

lizconlan: (whatever you set gets read as a string and evaluates to true)

lizconlan:
My `.vagrant.yml` file for testing purposes ended up as:

```
memory: 4096
show_settings: true
use_nfs: false
os: precise64
```

(although I probably wouldn't keep the `show_settings: true` all the time)

Vagrantfile
garethrees: I think its fairly standard, even in Alaveteli:

```
$ find . -name *.yml -d 1
./.codeclimate.yml
./.hound.yml
./.ruby-style.yml
./.travis.yml
```

I did wonder about an example file, but I had thought that this example was enough.

I can pop an example file in if you think it would help.

Vagrantfile
lizconlan: oh hang on, it's not that...

Vagrantfile
```diff
@@ -57,11 +63,43 @@
...
+  'themes_dir' => '../alaveteli-themes',
+  'os' => 'wheezy64',
+  'use_nfs' => false,
+  'show_settings' => false
+}.freeze
+
+env = DEFAULTS.keys.reduce({}) do |memo, key|
+  value = ENV["ALAVETELI_VAGRANT_#{ key.upcase }"]
+  value = false if %w(0 false no).include?(value)
+  memo[key] = value if value
```
lizconlan: needs to be the more awkward

```
memo[key] = value unless value.nil?
```

to allow false to be set!

Vagrantfile
lizconlan: Fair enough, comment withdrawn. Not sure about the example file, I got it wrong on first try but that was because I'd written `ALAVETELI_VAGRANT_MEMORY` out of habit while I was still reading the documentation

Vagrantfile
garethrees: Ah, I did on L70 but agree that it should say that here

Vagrantfile
garethrees: 👍

PR #4089, opened 2017-07-07
Add expiring embargoes to daily summary emails

stevenday:
This adds:
- A call to notify users when logging expiring embargo events against
  requests. This uses the same `uses_notifications` logic that new response
  notifications use, so that it's only batch requests belonging to
  users with the `:notifications` feature enabled that trigger notifications.
- A new 'instant' notification mailer for embargoes expiring. This isn't
  strictly necessary right now (right now having the notifications feature
  turned on means getting daily summaries of everything) but it seemed
  wise to build it anyway.
- A few extra specs and bugfixes around html escaping because I realised
  it was wrong in the current embargo emails
- New email views for the daily summary notification mailer to include
  these new types of notification.
- A slight refactor of `mail_user` from ApplicationMailer so that I can
  use it for these notification mails

For mysociety/alaveteli-professional#309 and mysociety/alaveteli-professional#274


app/mailers/notification_mailer.rb
```diff
@@ -90,13 +90,10 @@ def response_notification(notification)
...
-      :subject => _("New response to your FOI request - {{request_title}}",
-                    :request_title => @info_request.title.html_safe),
-      :charset => "UTF-8",
-      :template_name => 'new_response'
-    )
+    subject = _("New response to your FOI request - {{request_title}}",
+                :request_title => @info_request.title.html_safe)
+    mail_user(@info_request.user,
+              subject,
+              template_name: 'response_notification')
```
stevenday: Because we're using send in `instant_notification`, it seems like we have to specify the template name here regardless because otherwise rails will look for a template called `instant_notification.text.erb`

app/views/notification_mailer/info_requests/messages/_embargo_expiring.text.erb
```diff
@@ -0,0 +1,6 @@
+<%= _('This request will be made public on {{site_name}} in the next week.' \
```
crowbot: Needs a space after the full stop

stevenday: fixed in ab1ddf4 and f8c6000

spec/mailers/alaveteli_pro/embargo_mailer_spec.rb
```diff
@@ -157,5 +157,18 @@
...
+      allow(AlaveteliConfiguration).
+        to receive(:site_name).and_return("Something & something")
+      AlaveteliPro::EmbargoMailer.expiring_alert(pro_user, [expiring_1])
+      mail = ActionMailer::Base.deliveries[0]
+      expected_subject = "1 request will be made public on Something & " \
+                         "something this week"
+      escaped_subject = "1 request will be made public on Something &amp; " \
+                        "something this week"
+      expect(mail.subject).to eq expected_subject
+      expect(mail.subject).not_to eq escaped_subject
```
stevenday: Done in 7295865

stevenday:
I've fixed up those inline issues now, re: 

> Looks good - but I think people with the notifications feature enabled will get embargo expiring emails for individual requests via alert-embargoes-expiring and via daily notifications - so they'll be reminded twice. Is that intentional?

`EmbargoMailer.alert_expiring` [only alerts about requests where `use_notifications` is `false`](https://github.com/mysociety/alaveteli/blob/ab1ddf4274e74132bdb8ac7c9d2381730be6a060/app/mailers/alaveteli_pro/embargo_mailer.rb#L10-L14), so at the moment they're not getting any alerts, and this should bring that back via the notifications I think, or am I missing something else? (Highly possible :) )

stevenday:
Ah sorry, I see you mean just for individual requests, but those won't get notified about at the moment because they won't have `use_notifications: true`, only batch requests get that.

stevenday:
I think I've fixed the spec failures here via 581de57 and I've also brought it up to speed with develop, so we should be good to merge if Travis agrees..

stevenday:
Whilst reviewing the over-arching ticket for this work, I've just remembered that we need to also check that the notification is still valid when we try to send it, so there's a little bit more to add here.

stevenday:
@crowbot - I've added some mechanisms to check if a notification is still valid to be sent in dabfee1, 9bd04da and cbf2adb. I went with the term "expired" because "valid" has another connotation in ActiveRecord obviously, but better suggestions for naming are welcome :)

Basically the idea is to check at runtime that whatever caused the notification is still true/the current state of things and then to mark that in the db so that we don't lose track of notifications that expired, but ignore them whenever we send things out.

app/mailers/notification_mailer.rb
```diff
@@ -10,15 +10,24 @@ def self.send_daily_notifications
...
+                false)
     users.find_each do |user|
-      notifications = user.notifications.daily.unseen.order(created_at: :desc)
+      notifications = user.
+        notifications.
+          daily.
+            unseen.
+              where(expired: false).
+                order(created_at: :desc)
+      notifications = Notification.reject_and_mark_expired(notifications)
```
crowbot: Sorry, another bit of late review here - notifications is now an array as `reject_and_mark_expired` causes it to be lazily loaded. So we can't call `update_all` on it in a couple of lines time.

stevenday: @crowbot good spot! I've just pushed a fix and a new spec to catch that, thanks

PR #4097, opened 2017-07-12
Don't say we'll publish embargoed request on the submit button

crowbot:


app/views/followups/preview.html.erb
```diff
@@ -49,7 +49,11 @@
         <%= hidden_field_tag(:internal_review, 1 ) %>
       <% end %>
       <%= submit_tag _("Re-edit this message"), :name => 'reedit' %>
-      <%= submit_tag _("Send and publish message"), :name => 'submit' %>
+      <% if @info_request.embargo %>
+       <%= render partial: 'alaveteli_pro/followups/submit' %>
```
lizconlan: minor quibble - could use one more space on that indent!

PR #4099, opened 2017-07-13
Menu side set in css by page context, not in markup

crowbot:
The action menus at the top and bottom of the page were
either both projecting to the left (on embargoed requests)
or to the right (on public requests). This commit alters
the css so that the left projection is applied only to the
action menu in the header (which is positioned to the
right hand side of the request).

Given that this is possible, I don't think there's a need
to specify in the element classes whether they are right
or left positioned.

### DIFF
--- a/app/assets/stylesheets/responsive/_global_layout.scss
+++ b/app/assets/stylesheets/responsive/_global_layout.scss
@@ -207,10 +207,12 @@ textarea{
  *  Drop-down style action button
  */
 
-.action-menu__menu--right {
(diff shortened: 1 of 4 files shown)


wrightmartin:
I'm happy with the change, but just wanted to flag one side-effect of this change

This assumes we only ever use the ability to change the side of the action menu in this context. 

If we wanted to re-use the action menu component else where in alaveteli, it loses the ```--right``` ```--left``` functionality.

PR #4101, opened 2017-07-14
Display phase and status in request list

crowbot:
Fixes https://github.com/mysociety/alaveteli-professional/issues/312

This is intended to clarify the relationship between the two,
and make it less confusing that phase is displayed on the left
hand side nav for the request list, and status is used in the
rows (and on the request page).


wrightmartin:
It certainly makes it less confusing, but I'm not sure the design holds up now, as the status labels are so much longer (see screenshot)

screen shot 2017-07-19 at 12 12 16

There are a couple of other things about the request view I've become less happy with (how it behaves on small screens) so maybe we should spin this out into a new ticket for a minor-redesign (or realign)

garethrees:
Closing as this is nearly 3 years old. Always here for reference if we take another stab at this.

PR #4107, opened 2017-07-17
3800 globalize after drop constraints

garethrees:
1. Checkout develop
1. Drop and recreate database
1. Load sample data
1. Checkout this branch
1. Run migrations
1. Test as follows...

As normal user on site:
- [x] View public body as a user
- [x] Search for public body as a user

Through admin interface:
- [x] Create public body as an admin
- [x] Edit public body as an admin
- [x] Edit public body translations as an admin
- [x] View public body as an admin
- [x] Search for public body as an admin

### DIFF
--- a/app/models/public_body.rb
+++ b/app/models/public_body.rb
@@ -328,21 +328,32 @@ class PublicBody < ActiveRecord::Base
 
   # The "internal admin" is a special body for internal use.
   def self.internal_admin_body
-    # Use find_by_sql to avoid the search being specific to a
-    # locale, since url_name is a translated field:
-    sql = "SELECT * FROM public_bodies WHERE url_name = 'internal_admin_authority'"
-    matching_pbs = PublicBody.find_by_sql sql
+    matching_pbs = PublicBody.where(:url_name => 'internal_admin_authority')
     case
     when matching_pbs.empty? then
-      I18n.with_locale(I18n.default_locale) do
-        PublicBody.create!(:name => 'Internal admin authority',
-                           :short_name => "",
-                           :request_email => AlaveteliConfiguration::contact_email,
-                           :home_page => "",
-                           :notes => "",
-                           :publication_scheme => "",
-                           :last_edit_editor => "internal_admin",
-                           :last_edit_comment => "Made by PublicBody.internal_admin_body")
+      # "internal admin" exists but has the wrong default locale - fix & return
+      if invalid_locale = PublicBody::Translation.
+                            find_by_url_name('internal_admin_authority')
+        found_pb = PublicBody.find(invalid_locale.public_body_id)
+        I18n.with_locale(I18n.default_locale) do
+          found_pb.name = "Internal admin authority"
+          found_pb.request_email = AlaveteliConfiguration.contact_email
+          found_pb.save!
+        end
+        found_pb
+      else
+        I18n.with_locale(I18n.default_locale) do
+          PublicBody.
+            create!(:name => 'Internal admin authority',
+                    :short_name => "",
+                    :request_email => AlaveteliConfiguration.contact_email,
+                    :home_page => nil,
+                    :notes => nil,
+                    :publication_scheme => nil,
+                    :last_edit_editor => "internal_admin",
+                    :last_edit_comment =>
+                      "Made by PublicBody.internal_admin_body")
+        end
       end
     when matching_pbs.length == 1 then
       matching_pbs[0]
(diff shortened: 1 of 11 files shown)


app/models/public_body.rb
```diff
@@ -216,6 +216,14 @@ def set_api_key!
     self.api_key = SecureRandom.base64(33)
   end
 
+  def self.find_by_name(name)
```
lizconlan: it looks as though `PublicBody.find_by_name` is only used in specs so we could get rid of this relatively easily

app/models/public_body.rb
```diff
@@ -216,6 +216,14 @@ def set_api_key!
...
+    find_by(name: name)
+  end
+
+  def self.find_by_url_name(url_name)
```
lizconlan: `PublicBody.find_by_url_name` only appears in a couple of controllers (and some specs) so this can also be removed relatively easily

garethrees: I think actually we can keep these – dynamic finders are [still supported](http://guides.rubyonrails.org/active_record_querying.html#dynamic-finders) so lets keep this commit as is?

app/models/public_body.rb
```diff
@@ -328,21 +328,31 @@ def calculated_home_page
...
+      if invalid_locale = PublicBody::Translation.
+                            find_by_url_name('internal_admin_authority')
+        found_pb = PublicBody.find(invalid_locale.public_body_id)
+        I18n.with_locale(I18n.default_locale) do
+          found_pb.name = "Internal admin authority"
+          found_pb.save
+        end
+        found_pb
+      else
+        I18n.with_locale(I18n.default_locale) do
```
garethrees: Need to make sure we have a test for this `else` case.

db/migrate/20170626141302_drop_public_body_translated_columns.rb
```diff
@@ -1,6 +1,26 @@
...
+          id = record.id
+          puts "Creating default locale translation for public body #{ id }"
+
+          fields = record.translated_attribute_names
+          fields.each do |attribute_name|
+            translation[attribute_name] =
+              record.read_attribute(attribute_name, :translated => false)
+          end
+
+          translation.save!
```
garethrees: I think we should handle failure here:

```ruby
unless translation.save
  puts "WARNING: Could not create default locale translation for public body #{ id }" 
end

File: db/migrate/20170626141302_drop_public_body_translated_columns.rb [hunk 5]
```diff
@@ -22,19 +42,26 @@ def down
...
     # be updated before the constraints are reapplied
     PublicBody.transaction do
-      PublicBody.find_each do |pb|
-        translated = pb.translation_for(I18n.default_locale)
+      PublicBody.find_each do |record|
+        translated = record.translation_for(I18n.default_locale)
+        translated = record.translations.first unless translated.persisted?
+
+        if translated.new_record? || translated.nil?
+          puts "No translations for public body #{record.id}"
```
garethrees: Do we want to call `next` at this point?

db/migrate/20170626141302_drop_public_body_translated_columns.rb
garethrees: …and move the success feedback here too:

```ruby
    PublicBody.transaction do
      PublicBody.find_each do |record|
        translation = record.translation_for(I18n.default_locale) ||
          record.translations.build(:locale => I18n.default_locale)

        if translation.new_record?
          fields = record.translated_attribute_names
          fields.each do |attribute_name|
            translation[attribute_name] =
              record.read_attribute(attribute_name, :translated => false)
          end

          if translation.save
            puts "Created default locale translation for public body #{ record.id }"
          else
            puts "WARNING: Could not create default locale translation for public body #{ record.id }" 
          end
        end
      end
    end
```

app/models/public_body.rb
```diff
@@ -143,6 +143,17 @@ class Translation
   #
   # [1] http://git.io/vIetK
   class Version
+    before_save :copy_translated_attributes
```
garethrees: Squash in to f973eb946084e87ba00dbf5abb7adb363a75dad1

app/models/public_body.rb
lizconlan: Sure, simplest option :)

app/models/public_body.rb
lizconlan: is covered by `spec/models/public_body_spec.rb:628`

PR #4111, opened 2017-07-19
Issues/alaveteli pro 324/embargoed image cookies

crowbot:
Adds a `cookie_passthrough` param to attachment urls that we can use in Varnish to avoid stripping cookies from them.

### DIFF
--- a/app/helpers/info_request_helper.rb
+++ b/app/helpers/info_request_helper.rb
@@ -236,4 +236,48 @@ module InfoRequestHelper
        user: user_link_for_request(info_request))
   end
 
+  def attachment_link(incoming_message, attachment)
+    img_filename = "icon_#{attachment.content_type.sub('/', '_')}_large.png"
+    full_filename = File.expand_path(Rails.root.join('app',
+                                                     'assets',
+                                                     'images',
+                                                     img_filename))
+    image = if File.exist?(full_filename)
+      img_filename
+    else
+      "icon_unknown.png"
+    end
+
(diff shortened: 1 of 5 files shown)


app/helpers/info_request_helper.rb
```diff
@@ -236,4 +236,48 @@ def status_text_awaiting_description_other(info_request, new_responses_count)
...
+                                                     img_filename))
+    image = if File.exist?(full_filename)
+      img_filename
+    else
+      "icon_unknown.png"
+    end
+
+    link_to image_tag(image, :class => "attachment__image",
+                             :alt => "Attachment"),
+                      attachment_path(incoming_message, attachment)
```
stevenday: I'm not sure if this is actually the line Hound is complaining about above, but I think `attachment_path` should be aligned to `image_tag`? (It's a param to `link_to` right?)

spec/helpers/info_request_helper_spec.rb
```diff
@@ -503,4 +503,72 @@
...
+    end
+
+    context 'when given an html format option' do
+
+      it 'returns the path to the HTML version of the attachment' do
+        expect(attachment_path(incoming_message,
+                               jpeg_attachment,
+                               :html => true)).
+          to eq("/request/#{incoming_message.info_request_id}" \
+                "/response/#{incoming_message.id}/" \
```
stevenday: Just to make this clearer, can we put the trailing `/` at the start of the next line, like all the others?

PR #4114, opened 2017-07-19
[WIP][#3800] Rails 4.2 upgrade

lizconlan:
#3800


lizconlan:
Still to do:
- [ ] look at http://railsdiff.org/4.1.16/4.2.9
- [ ] have a proper look at 'Masked Authenticity Tokens'

garethrees:
I've made a new PR with the prerequisites in https://github.com/mysociety/alaveteli/pull/4124. We can get those deployed first, and then rebase this.

app/helpers/sanitizer_helper.rb
```diff
@@ -0,0 +1,14 @@
...
+  # Strip out script tags entirely and remove unwanted HTML markup
+  def sanitize_html(text)
+    # remove script tags and their content nodes
+    scrubber = Loofah::Scrubber.new do |node|
+      node.remove if node.name == 'script'
+    end
+    output = Loofah.fragment(text).scrub!(scrubber).to_s
+
+    # sanitize HTML (strip unwanted tags leaving content intact)
+    ActionController::Base.helpers.sanitize(output)
```
garethrees: I don't fully understand what's happening here.

Here's how I'm reading it:

* `#sanitize` doesn't strip out the contents between `<script>` tags any more…
* …so we need a custom scrubber to do that.
* We manually pre-scrub the `text` because passing the `scrubber` to `sanitize` (e.g. `sanitize(text, scrubber: scrubber)`) prevents the default scrubbers removing some other stuff that we want removed.

---

* Is that a correct assessment?
* If so, why can't we construct a scrubber that does all that we want, and pass it to `sanitize`?

I'm also in two minds about adding `#sanitize_html`.

My instinct says that its quite obscure, so `#sanitize` is going to creep in because of familiarity. Because of that, maybe we should change how `sanitize` works?

Its annoying that we can't just create our own sanitizer and [set it as default](https://github.com/rails/rails/issues/28587).

What we can do is create a custom sanitzer:

```ruby
class AlaveteliWhiteListSanitizer < Rails::Html::WhiteListSanitizer
  def sanitize(html, options = {})
    scrubber = Loofah::Scrubber.new do |node|
      node.remove if node.name == 'script'
    end

    html = Loofah.fragment(html).scrub!(scrubber).to_s

    super(html, options)
  end
end
```

…and override the [method](https://github.com/rails/rails-html-sanitizer/blob/master/lib/rails-html-sanitizer.rb#L18-L20) that returns the default sanitzer class

```ruby
Rails::Html::Sanitizer.instance_eval do
  def white_list_sanitizer
    ::AlaveteliWhiteListSanitizer
  end
end
```

This means we can use `sanitize` with our desired behaviour.

```ruby
Rails::Html::Sanitizer.white_list_sanitizer
# => AlaveteliWhiteListSanitizer

helper.sanitize('<script>alert("hi!");</script>')
# => ""
```

On the other hand, it means we're monkey patching internals.

Any thoughts?

config/application.rb
```diff
@@ -25,7 +25,12 @@ class Application < Rails::Application
     I18n.config.enforce_available_locales = true
 
     # Allow some extra tags to be whitelisted in the 'sanitize' helper method
-    config.action_view.sanitized_allowed_tags = 'html', 'head', 'body', 'table', 'tr', 'td', 'style'
+    config.action_view.sanitized_allowed_tags = Set.new(
+      %w(strong em b i p code pre tt samp kbd var sub sup dfn cite big small
+         address hr br div span h1 h2 h3 h4 h5 h6 ul ol li dl dt dd abbr
+         acronym a img blockquote del ins table tr td style
+        )
+    )
```
garethrees: And I think this should move to `config/initializers/action_view.rb`

app/views/general/_opengraph_tags.html.erb
```diff
@@ -9,7 +9,7 @@
 <%= content_for :open_graph_tags %>
 
 <meta property="og:site_name" content="<%= site_name %>" />
-<meta property="og:url" content="<%= request.original_url %>" />
+<meta property="og:url" content="<%= request.original_url.encode("utf-8", invalid: :replace, undef: :replace, replace: "_") %>" />
```
garethrees: Some more info on what caused this would be good. Spec failures I assume?

Would be worth adding any reference material you found that helped diagnose this.

Could we use `force_encoding('utf-8')`?

app/models/outgoing_message.rb
```diff
@@ -305,7 +306,11 @@ def get_body_for_html_display
     text = CGI.escapeHTML(text)
     text = MySociety::Format.make_clickable(text, { :contract => 1, :nofollow => true })
     text.gsub!(/\[(email address|mobile number)\]/, '[<a href="/help/officers#mobiles">\1</a>]')
-    text = ActionController::Base.helpers.simple_format(text)
+
+    # sanitize the text separately to avoid stripping out the link element
```
lizconlan: yep

lizconlan: (unless you pass `sanitize: false` to prevent it: https://apidock.com/rails/ActionView/Helpers/TextHelper/simple_format)

spec/controllers/request_controller_spec.rb
```diff
@@ -604,9 +604,9 @@ def get_attachment(params = {})
                            :part => 2,
                            :file_name => 'interesting.html',
                            :skip_cache => 1
-      expect(response.body).not_to match("script")
-      expect(response.body).not_to match("interesting")
-      expect(response.body).to match('dull')
+      expected =
+        "<html>\n  <head>\n  </head>\n  <body>dull\n    \n  </body>\n</html>\n"
+      expect(response.body).to eq(expected)
```
garethrees: First, some examples:

```ruby
content = <<-EOF
<html>
  <head>
  </head>
  <body>dull
    <script>alert('interesting')</script>
  </body>
</html>
EOF

# Rails 4.1
helper.sanitize(content)
# => "<html>\n  <head>\n  </head>\n  <body>dull\n    \n  </body>\n</html>\n"

# Rails 4.2
#
# Default sanitize method with custom set of
# config.action_view.sanitized_allowed_tags
helper.sanitize(content)
# => "\n  \n  \n  dull\n    alert('interesting')\n  \n\n"

# Liz's new helper:
helper.sanitize_html(content)
# => "\n  \n  \n  dull\n    \n  \n\n"

# Default with a supplied list of allowed tags.
# For some reason these don't work as I'd expect
helper.sanitize(content, tags: %w(html head body))
# => "\n  \n  \n  dull\n    alert('interesting')\n  \n\n"
helper.sanitize(content, tags: %w(body))
# => "\n  \n  \n  dull\n    alert('interesting')\n  \n\n"

# Try _allowing_ the script tag. Works, but still strips out the other tags.
helper.sanitize(content, tags: %w(script))
# => "\n  \n  \n  dull\n    <script>alert('interesting')</script>\n  \n\n"
```

What's going on?

We're [parsing](https://github.com/rails/rails-html-sanitizer/blob/master/lib/rails/html/sanitizer.rb#L33) `content` as a Loofah **fragment**. Since `content` is a full HTML document, we [should](https://github.com/flavorjones/loofah#side-note-fragments-vs-documents) be parsing it as a Loofah **document**.

I think we _do_ need a custom sanitzer helper for sanitizing attachments (as per ed87595fc52d0aa020b32e124fdbfd4bbe672e84), but just adding an additional scrubbing step is incorrect.

I think we should define a new sanitzer in the same way as [`Rails::Html::FullSanitizer`](https://github.com/rails/rails-html-sanitizer/blob/master/lib/rails/html/sanitizer.rb#L22-L40), replacing the Loofer line with `Loofer.document`.

garethrees: Actually, we might just want to call Loofah directly:

```ruby
Loofah.scrub_document(content, :prune).to_html(encoding: 'UTF-8')
# => "<!DOCTYPE html PUBLIC \"-//W3C//DTD HTML 4.0 Transitional//EN\" \"http://www.w3.org/TR/REC-html40/loose.dtd\">\n<html>\n  <head>\n<meta http-equiv=\"Content-Type\" content=\"text/html; charset=UTF-8\">\n  </head>\n  <body>dull\n    \n  </body>\n</html>\n"

Loofah.scrub_document(content, :escape).to_html(encoding: 'UTF-8')
# => "<!DOCTYPE html PUBLIC \"-//W3C//DTD HTML 4.0 Transitional//EN\" \"http://www.w3.org/TR/REC-html40/loose.dtd\">\n<html>\n  <head>\n<meta http-equiv=\"Content-Type\" content=\"text/html; charset=UTF-8\">\n  </head>\n  <body>dull\n    &lt;script&gt;alert('interesting')&lt;/script&gt;\n  </body>\n</html>\n"
```

app/views/general/_opengraph_tags.html.erb
lizconlan: `force_encoding('utf-8')` should work, I'll give it a try (and add some more info)

Gemfile
```diff
@@ -182,7 +182,7 @@ group :development do
   gem 'mailcatcher', '~> 0.6.0'
   gem 'quiet_assets', '~> 1.1.0'
   gem 'launchy', '~> 2.4.0'
-  gem 'spring', '~> 1.7.0', '< 2.0.0'
+  gem 'spring', '~> 2.0.0'
```
garethrees: Added in d0a8274

Gemfile
```diff
@@ -135,7 +135,7 @@ gem 'gender_detector', '~> 1.0.0'
 gem 'gettext_i18n_rails', '~> 0.9.0', '< 1.0.0'
   gem 'fast_gettext', '< 1.2.0'
 gem 'gettext', '~> 2.3.0'
-gem 'globalize', '~> 4.0.0', '< 5.0.0'
+gem 'globalize', '~> 5.0.0', '< 5.1.0'
```
garethrees: Fixed in bbb5758

app/models/user.rb
```diff
@@ -347,7 +347,7 @@ def name
   # When name is changed, also change the url name
   def name=(name)
     write_attribute(:name, name.try(:strip))
-    update_url_name
+    update_url_name unless banned?
```
garethrees: This gets fixed by https://github.com/mysociety/alaveteli/pull/4137

lib/no_constraint_disabling.rb
```diff
@@ -51,7 +51,7 @@ def self.create_fixtures(fixtures_directory, fixture_set_names, class_names = {}
               ::File.join(fixtures_directory, fs_name))
           end
 
-          all_loaded_fixtures.update(fixtures_map)
+          update_all_loaded_fixtures fixtures_map
```
garethrees: Fixed by b202adf

app/views/general/_opengraph_tags.html.erb
lizconlan: fixed in #4139

lib/i18n_fixes.rb
```diff
@@ -67,28 +67,3 @@ def locale
     end
   end
 end
-
```
lizconlan: I don't think it's a globalize change, I think it's that Rails 4.2 doesn't accidentally hold on to the cached value the way that 4.1 did. I suspect we ended up patching Globalize to work around a Rails 4.1 bug.

(Also as `@attributes_cache` has been removed in 4.2, the patch crashes the app post upgrade)

The Globalize code looks pretty similar between the versions:
- [4.0.3](https://github.com/globalize/globalize/blob/v4.0.3/lib/globalize/active_record/instance_methods.rb#L22-L59)
- [5.0.1](https://github.com/globalize/globalize/blob/v5.0.1/lib/globalize/active_record/instance_methods.rb#L22-L57)

garethrees: This information should be in the commit message.

What prompted the change? Failing specs?

lizconlan: > This information should be in the commit message

Oh absolutely, was making notes here as was too tired yesterday to make enough sense for a commit message :)

> What prompted the change?

Mostly remembering that it was Rails 4.1 specific...

> Failing specs?

...and also that - it tries to access cache variables which were removed as one of the big things in Rails 4.2 was an overhaul of object caching so our patch causes breakages. I'll think of a sensible way to phrase this in a minute. One more coffee should do it

lizconlan: \*hunts for a useful link in the 4.2 release notes\*

lizconlan: Fixed in #4140

lizconlan:
(Hopefully) last thing - `User.name` with " (Account suspended)" is still being persisted to the database when `update_attributes` is called. This is the currently failing spec.

PR #4128, opened 2017-07-31
Use nip.io by default

crowbot:
Doesn't seem to have the same transient failure issues as
xip.io


stevenday:
Hmm, my fqdn is currently set to `alaveteli.10.10.10.30.xip.io` but I actually just always hit the ip address directly too

PR #4132, opened 2017-08-01
Remove commenting (presumably added by mistake)

crowbot:
This line is the one that stops the popup from displaying if
you've already closed it. @wrightmartin it does result in FOUC though - any ideas?

### DIFF
--- a/app/assets/javascripts/general.js
+++ b/app/assets/javascripts/general.js
@@ -60,10 +60,6 @@ $('#standard-popup .js-popup__close').click(function() {
      })
    })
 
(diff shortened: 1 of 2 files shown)


stevenday:
The easiest way to reduce/remove the FOUC is probably to add a css rule like:

```css
.js-loaded #standard-popup {
  display: none;
}
```

So that it works off the inline script for setting that.

garethrees:
Added a combination of your fix and steve's suggestion in https://github.com/mysociety/alaveteli/pull/4176 for review.

PR #4134, opened 2017-08-02
Add Rails 4.2 generated template changes

lizconlan:
Extracted from http://railsdiff.org/4.1.16/4.2.9

Mostly harmless / boilerplate changes except for the `squash!` commits which introduce config or other changes that would benefit from closer attention:

* adding `assets.digest = true` sounds like it might have unintended consequences but I'm not certain
* the `bin/setup` file feels to me like we should replace most of the content with a TODO comment and consider it separately
* `ActiveSupport.to_time_preserves_timezone` looks like it might save us some major headaches down the line
 
The commit message for moving the cookies_serializer config is a bit misleading - pulling it out into its own file was introduced in 4.1 but it appears I opted to configure it inside application.rb instead.

### DIFF
--- a/config/environments/development.rb
+++ b/config/environments/development.rb
@@ -28,6 +28,10 @@ Rails.application.configure do
   # number of complex assets.
   config.assets.debug = ENV.key?('ASSETS_DEBUG') || false
 
+  # Asset digests allow you to set far-future HTTP expiration dates on all assets,
+  # yet still be able to expire them through the digest params.
+  config.assets.digest = true
+
   # Adds additional error checking when serving assets at runtime.
   # Checks for improperly declared sprockets dependencies.
   # Raises helpful error messages.
--- a/config/environments/development.rb
+++ b/config/environments/development.rb
@@ -48,6 +48,9 @@ Rails.application.configure do
     config.action_mailer.delivery_method = :sendmail
   end
 
+  # Allow any IP address in the range 10.10.10.x to access the web console
+  config.web_console.whitelisted_ips = '10.10.10.0/16'
+
   # Writes useful log files to debug memory leaks, of the sort where have
   # unintentionally kept references to objects, especially strings.
   # require 'memory_profiler'
--- a/config/initializers/cookies_serializer.rb
+++ b/config/initializers/cookies_serializer.rb
@@ -1,0 +1,4 @@
+# -*- encoding : utf-8 -*-
+# Set the cookie serializer to :hybrid to migrate the old format Marshalled
+# cookies to the new, more secure, JSON format
+Rails.application.config.action_dispatch.cookies_serializer = :hybrid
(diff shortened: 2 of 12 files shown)


config/initializers/to_time_preserves_timezone.rb
```diff
@@ -0,0 +1,10 @@
+# Be sure to restart your server when you modify this file.
+
+# Preserve the timezone of the receiver when calling to `to_time`.
+# Ruby 2.4 will change the behavior of `to_time` to preserve the timezone
+# when converting to an instance of `Time` instead of the previous behavior
+# of converting to the local system timezone.
+#
+# Rails 5.0 introduced this config option so that apps made with earlier
+# versions of Rails are not affected when upgrading.
+ActiveSupport.to_time_preserves_timezone = true
```
lizconlan: Rails told me to do it!

config/environments/development.rb
```diff
@@ -48,6 +48,8 @@
     config.action_mailer.delivery_method = :sendmail
   end
 
+  config.web_console.whitelisted_ips = '10.10.10.0/16'
```
lizconlan: this isn't from Railsdiff but it felt like a related place to put it

config/initializers/cookies_serializer.rb
```diff
@@ -0,0 +1,3 @@
+# Set the cookie serializer to :hybrid to migrate the old format Marshalled
+# cookies to the new, more secure, JSON format
+Rails.application.config.action_dispatch.cookies_serializer = :hybrid
```
garethrees: No `\n` at end of file.

config/initializers/to_time_preserves_timezone.rb
garethrees: > Rails told me to do it!

?

See https://robots.thoughtbot.com/no-newline-at-end-of-file

garethrees:
I had this notification in email but doesn't look like we're setting the web_console value now? Guessing thats intentional?

> > @@ -48,6 +48,8 @@
>    config.action_mailer.delivery_method = :sendmail
>  end
> 
> +  config.web_console.whitelisted_ips = '10.10.10.0/16'
> this isn't from Railsdiff but it felt like a related place to put it

config/initializers/to_time_preserves_timezone.rb
lizconlan: Bizarrely http://railsdiff.org/4.1.16/4.2.9 (scroll down to 'config/initializers/cookies_serializer.rb'), taking it out though

lizconlan:
> I had this notification in email but doesn't look like we're setting the web_console value now? Guessing thats intentional?

I think I missed cherry-picking it when I made the new branch. Will add it back rather than have yet another pull request just for that

lizconlan:
[ waiting for spec run ]

lizconlan:
no new fails \o/

config/initializers/to_time_preserves_timezone.rb
garethrees: To me it looks like 4.2.x is [_adding_](https://github.com/rails/rails/commit/bb63e9c5902f4f6374e9c3fa5bd8041d7be9fa91) the final newline.

A final newline is not the same as a final **blank** line.

PR #4136, opened 2017-08-03
3800 post globalize rails 4 2 sanitizers

garethrees:


config/application.rb
```diff
@@ -24,8 +24,16 @@ class Application < Rails::Application
...
+                               sup dfn cite big small address hr br div span h1
+                               h2 h3 h4 h5 h6 ul ol li dl dt dd abbr acronym a
+                               img blockquote del ins)
+    alaveteli_extra_tags = %w(html head body table tr td style)
+
+    allowed_tags = rails_41_default_tags + alaveteli_extra_tags
+
     # Allow some extra tags to be whitelisted in the 'sanitize' helper method
-    config.action_view.sanitized_allowed_tags = 'html', 'head', 'body', 'table', 'tr', 'td', 'style'
+    config.action_view.sanitized_allowed_tags = allowed_tags
```
garethrees: should move this to `config/initializers/action_view.rb`

garethrees:
> Assuming I've understood this correctly and we're accepting the new `sanitize` functionality except for within `RequestController#get_attachment` 

Yeah, that's correct. All the current specs pass and I've manually tested some incoming message / outgoing message HTML and it all seems to work as expected.

app/controllers/request_controller.rb
```diff
@@ -654,7 +654,10 @@ def get_attachment
             apply_masks(@attachment.default_body, @attachment.content_type)
 
     if content_type == 'text/html'
-      body = ActionController::Base.helpers.sanitize(body)
+      body =
```
garethrees: Add note to say why we don't use `#sanitize`

garethrees:
Merged

PR #4137, opened 2017-08-03
Do not create a variable with the same name as the AR model attribute inside getter methods

lizconlan:
- [x] Fix `User#url_name` spec fail
- [x] Check the rest of the code for similar issues
- [x] Fix for other models
  - [x] `User#name` (for completeness)
  - [x] `InfoRequest#title`
  - [x] `MailServerLog#line`

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -368,7 +368,7 @@ class User < ActiveRecord::Base
   # When name is changed, also change the url name
   def name=(name)
     write_attribute(:name, name.try(:strip))
-    update_url_name
+    update_url_name unless banned?
   end
 
   def update_url_name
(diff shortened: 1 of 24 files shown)


app/models/user.rb
```diff
@@ -358,17 +358,17 @@ def get_locale
...
 
   def name
-    name = read_attribute(:name)
+    tmp_name = read_attribute(:name)
     if banned?
-      name = _("{{user_name}} (Account suspended)", :user_name => name)
+      tmp_name = _("{{user_name}} (Account suspended)", :user_name => tmp_name)
     end
-    name
+    tmp_name
```
lizconlan: 👍

lizconlan:
Merged

PR #4138, opened 2017-08-03
Add form to manage embargoes across a whole batch request

### LINKED ISSUE #3985: This request is hidden until XY - does not change
The text "this request is hidden until XY" (loosely translated, having Czech interface) does not change when changing to (e.g.) "publish immediately) at /alaveteli_pro/info_requests/new
and still shows the default date (after 3 months), which is a bit misleading.
(e.g., http://www.infoprovsechny.cz/alaveteli_pro/info_requests/new )

stevenday:
This PR:
  - Makes it so that you can't edit an embargo on a batch request singularly
  - Adds new controller actions for extending and publishing a whole batch
  - Adds a form to do those things to the batch request page
  - Updates existing request forms to act on the whole batch when showing a
    request from a batch, with a suitable confirmation step
  - Fixes a bug which I think resolves #3985 / mysociety/alaveteli-professional#278
  - Fixes a couple of other bugs in specs or missing specs entirely for existing embargo controllers

For mysociety/alaveteli-professional#258

### DIFF
--- a/app/controllers/alaveteli_pro/embargo_extensions_controller.rb
+++ b/app/controllers/alaveteli_pro/embargo_extensions_controller.rb
@@ -7,20 +7,23 @@
 
 class AlaveteliPro::EmbargoExtensionsController < AlaveteliPro::BaseController
   def create
-    @embargo = AlaveteliPro::Embargo.find(embargo_extension_params[:embargo_id])
+    @embargo = AlaveteliPro::Embargo.find(
+      embargo_extension_params[:embargo_id])
     authorize! :update, @embargo
     @info_request = @embargo.info_request
     # Embargoes cannot be updated individually on batch requests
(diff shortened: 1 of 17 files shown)


app/controllers/alaveteli_pro/embargo_extensions_controller.rb
```diff
@@ -25,9 +32,44 @@ def create
         url_title: @info_request.url_title)
   end
 
+  def create_batch
+    @info_request_batch = InfoRequestBatch.find(
+      params[:info_request_batch_id])
+    authorize! :update, @info_request_batch
+    begin
+      ActiveRecord::Base.transaction do
```
crowbot: For a large batch, this will be quite a lot of object creation - in the longer term, we might want to move it out of the request cycle.

PR #4139, opened 2017-08-03
squash! Force request.original_url to UTF-8

lizconlan:
Slightly bafflingly, reverting the template code back to its original state and running the specs locally at the moment (with Ruby 2.0.0) gives the much less helpful `Capybara::ElementNotFound` error which could mean anything (although I'm pretty sure in this case it means - "I did not receive valid HTML but I've eaten the error message")

Example Travis log from when the error first appeared: https://travis-ci.org/mysociety/alaveteli/jobs/256872459

Extract:
```
2) creating requests in alaveteli_pro when writing a new request from scratch allows us to preview the request
     Failure/Error: <%= render :partial => 'general/locale_switcher' %>
     Encoding::CompatibilityError:
       incompatible character encodings: ASCII-8BIT and UTF-8
     # ./.bundle/ruby/2.0.0/gems/activesupport-4.2.8/lib/active_support/core_ext/string/output_safety.rb:185:in `concat'
     # ./.bundle/ruby/2.0.0/gems/activesupport-4.2.8/lib/active_support/core_ext/string/output_safety.rb:185:in `concat'
     # ./.bundle/ruby/2.0.0/gems/actionview-4.2.8/lib/action_view/buffers.rb:12:in `<<'
     # ./app/views/general/_responsive_header.html.erb:16:
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 /tmp
-/log
+/log/*
+!/log/.keep
 /coverage
 /sqlitedbs
 /cache
--- a/app/controllers/request_controller.rb
+++ b/app/controllers/request_controller.rb
@@ -654,7 +654,10 @@ class RequestController < ApplicationController
             apply_masks(@attachment.default_body, @attachment.content_type)
 
     if content_type == 'text/html'
-      body = ActionController::Base.helpers.sanitize(body)
+      body =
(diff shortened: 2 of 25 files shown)


lizconlan:
Merged

PR #4141, opened 2017-08-04
Override the user name param with the attribute value before update

lizconlan:
params from form gets "decorated" version of name and passes it to update_attributes which will then update the database value to "Bad User (Account suspended)" and next time will report it as "Bad User (Account suspended) (Account suspended)"

Not convinced that I've fixed this the right way but I can't immediately think of a better one - a specialised `strip_attributes` would require a minor refactor to hold `" (Account suspended)"` (which gets translated) as a new (private?) attribute (probably a method)


lizconlan:
withdrawn, this was a horrible idea

PR #4149, opened 2017-08-08
Add a new helper to return the contents of an asset file

### LINKED ISSUE #4148: [ERROR] request#download_entire_request (NoMethodError) "undefined method `[]' for nil:NilClass...
Introduced by Rails 4.2, seems to be a Sprockets (or SprocketRails) issue

```
A NoMethodError occurred in request#download_entire_request:

  undefined method `[]' for nil:NilClass
  app/views/general/_stylesheet_includes.html.erb:4:in `_app_views_general__stylesheet_includes_html_erb___3349839874082148774_66274660'


Request:

  * URL        : https://www.whatdotheyknow.com/request/organisational_structure_chart_f_3474/download
  * HTTP Method: GET
  * IP address : FILTERED
  * Parameters : {"controller"=>"request", "action"=>"download_entire_request", "url_title"=>"organisational_structure_chart_f_3474"}
  * Timestamp  : 2017-08-08 12:11:58 +0100
  * Rails root : FILTERED/www.whatdotheyknow.com/alaveteli-2017-08-08T09-11-37
  * Process: 21279

Backtrace:

  app/views/general/_stylesheet_includes.html.erb:4:in `_app_views_general__stylesheet_includes_html_erb___3349839874082148774_66274660'
  app/views/layouts/default.html.erb:17:in `_app_views_layouts_default_html_erb__1695932967868905343_62073620'
  app/controllers/request_controller.rb:974:in `make_request_summary_file'
  app/controllers/request_controller.rb:951:in `block in make_request_zip'
  app/controllers/request_controller.rb:950:in `make_request_zip'
  app/controllers/request_controller.rb:842:in `block in download_entire_request'
  app/controllers/request_controller.rb:821:in `download_entire_request'
  app/controllers/application_controller.rb:108:in `record_memory'
```

lizconlan:
Connects to #4148 

Fix inspired by https://github.com/rails/sprockets-rails/issues/237#issuecomment-231940436
(and the immediate followups).

Placed in a helper wrapper to allow us to force UTF-8 and handle the minor faff of checking whether the Enumerator actually contains anything without a bunch of extra stuff to the template.

After deploying, we may need to clear zip file caches for attempted downloads as I suspect it's cached an HTML file containing the File Not Found message

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -78,6 +78,16 @@ module ApplicationHelper
     return "#{exact_date} (#{ago_text})"
   end
 
+  def read_asset_file(asset_name)
+    assets = Rails.application.assets_manifest.find_sources(asset_name)
+    if assets.first
+      assets.first.force_encoding('utf-8')
+    else
+      raise Sprockets::FileNotFound,
+            "Asset file '#{asset_name}' was not found in the assets directory"
+    end
+  end
+
   # Note that if the admin interface is proxied via another server, we can't
   # rely on a sesssion being shared between the front end and admin interface,
   # so need to check the status of the user.
--- a/app/views/general/_stylesheet_includes.html.erb
+++ b/app/views/general/_stylesheet_includes.html.erb
@@ -1,8 +1,8 @@
 <% if AlaveteliConfiguration::responsive_styling || params[:responsive] %>
   <%- if @render_to_file %>
     <style>
-      <%= Rails.application.assets["responsive/application.css"].to_s %>
-      <%= Rails.application.assets["responsive/print.css"].to_s %>
+        <%= read_asset_file('responsive/application.css') %>
+        <%= read_asset_file('responsive/print.css') %>
     </style>
   <%- else %>
     <%= render :partial => 'general/responsive_stylesheets' %>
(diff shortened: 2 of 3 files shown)


app/helpers/application_helper.rb
```diff
@@ -78,6 +78,11 @@ def admin_date(date)
     return "#{exact_date} (#{ago_text})"
   end
 
+  def read_asset_file(asset_name)
+    assets = Rails.application.assets_manifest.find_sources(asset_name)
+    assets.first.force_encoding('utf-8') if assets.first
```
garethrees: I think it should raise an exception if it can't be found, otherwise this would silently fail.

app/helpers/application_helper.rb
```diff
@@ -78,6 +78,16 @@ def admin_date(date)
     return "#{exact_date} (#{ago_text})"
   end
 
+  def read_asset_file(asset_name)
+    assets = Rails.application.assets_manifest.find_sources(asset_name)
+    if assets.first
+      assets.first.force_encoding('utf-8')
+    else
+      raise(Sprockets::FileNotFound,
```
garethrees: Always omit parentheses for Methods that have "keyword" status in Ruby:

https://github.com/bbatsov/ruby-style-guide#method-invocation-parens

PR #4172, opened 2017-08-15
Fix documentation pages

### LINKED ISSUE #4109: Documentation pages not loading assets correctly
When I redeployed the docs site just now (after merging #4080 which only touches `/docs/glossary.md`), pages within folders (including the /es/ pages) were being built with the wrong assets path so were being rendered without CSS.

Instead of the expected path prefix from `{{site.baseurl}}` of '/', it was trying to load from the local folder path e.g. `/docs/assets/css/...` and 404ing. To avoid the docs pages appearing broken, I've made the simplest possible fix of changing the asset paths in the layout file to start with a literal '/'.

My working assumption is that something has changed in how paths are evaluated and that merging the changes to the glossary page triggered a complete rebuild of the gh-pages site rather than the changes themselves having caused the error directly. (Full disclosure - I did not visit the page immediately before deploying but as @davewhiteland included screenshots I am reasonably confident in assuming the pages were working as expected pre deploy.)

It would be good to allocate some time to investigate this fully and possibly come up with a better fix. 

***Note:*** images are currently broken on the Spanish pages as these also rely on the `{{site.baseurl}}/assets/` syntax

lizconlan:
Closes #4109


Gemfile
```diff
@@ -1,2 +1,2 @@
 source 'https://rubygems.org'
-gem 'github-pages'
+gem 'github-pages', '148'
```
garethrees: I think we always want the latest of this, so don't need to pin.

_config.yml
```diff
@@ -1,6 +1,6 @@
 safe: true
 highlighter: rouge
-baseurl: /
+baseurl: ""
```
lizconlan: we also use "/es/" so it's probably safest to set the default

Gemfile
lizconlan: the latest is beyond what GitHub have documented as current

_config.yml
lizconlan: (I'll try it without, see what happens)

lizconlan: it does seem to work locally without this being set here, removing

PR #4175, opened 2017-08-16
Add fixtures for pro users

garethrees:
These get loaded as sample data to make development on pro easier.

Adds:

* A pro user
* A pro admin
* A user who is a pro, an admin _and_ a pro admin


garethrees:
Needs a `ProAccount` for users who have the `pro` role.

PR #4176, opened 2017-08-16
squash! Remove commenting (presumably added by mistake)

garethrees:
Improvements to https://github.com/mysociety/alaveteli/pull/4132.

Prevent FOUC (Flash Of Unstyled Content) when showing popup.

We hide popups with CSS by default, and then show the popup unless the
user has already closed it.

### DIFF
--- a/app/assets/javascripts/general.js
+++ b/app/assets/javascripts/general.js
@@ -61,7 +61,7 @@ $('#standard-popup .js-popup__close').click(function() {
    })
 
    if($.cookie('seen_foi2') == 1) {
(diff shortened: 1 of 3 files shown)


garethrees:
> If we can make it such that it doesn't show at all when it shouldn't and is immediately present, rather than appearing, when it should, I think that would be the ideal.

Ah gotcha. Amended in d8c37fd by checking the cookies in the template rendering, and then just relying on JS to close the popup and set the cookie.

garethrees:
Squashed in to 8909fb0

PR #4180, opened 2017-08-18
Allow filtering of users by role on admin user list page

crowbot:


app/controllers/admin_user_controller.rb
```diff
@@ -22,18 +22,26 @@ class AdminUserController < AdminController
   def index
     @query = params[:query]
 
+    @roles = params[:roles] || []
     @sort_options = index_sort_options
 
     @sort_order =
       @sort_options.key?(params[:sort_order]) ? params[:sort_order] : 'name_asc'
 
     users = if @query
```
lizconlan: Now that we are using the search button to apply a filter, this would work better as `users = if @query.present?` otherwise it will always add the extra WHERE clauses, even when the submitted query is blank

spec/controllers/admin_user_controller_spec.rb
```diff
@@ -106,6 +106,26 @@
...
+    it 'filters the records by multiple role' do
+      User.destroy_all
+      admin_user = FactoryGirl.create(:admin_user)
+      pro_user = FactoryGirl.create(:pro_user)
+      user = FactoryGirl.create(:user)
+      user.add_role :pro
+      user.add_role :admin
+      get :index, :roles => [ 'admin', 'pro' ]
+      expect(assigns[:admin_users]).to eq([user])
+
```
lizconlan: extra blank line

crowbot: Fixed in be0a7b90cc6c3fffe2c7a626d014e1b40e21f051

app/controllers/admin_user_controller.rb
crowbot: Fixed in fbebe1dd0291c9e165b5236a7315845f95db10da

PR #4182, opened 2017-08-18
[#Pro329] Payment stub pages

garethrees:
Fixes https://github.com/mysociety/alaveteli-professional/issues/329.

I think this also closes https://github.com/mysociety/alaveteli-professional/issues/328 – it doesn't look like I need to do any more than declare a feature flag for it to operate correctly?

Adds some stub pages for the new Alaveteli Pro payment flow.

I haven't added subs for the following pages, because I need a better understanding of https://github.com/mysociety/alaveteli-professional/issues/331.

* Contract screen
* Enter card details
* Upgrade confirmation screen

In any case, they'll be hinged off the `Users::AccountsController` actions

## /pro (existing)

screen shot 2017-08-18 at 15 13 09

## /pro/pricing

screen shot 2017-08-18 at 15 12 53

## /pro/sign_in

screen shot 2017-08-18 at 15 13 52

## /alaveteli_pro/users (rendered after pro account creation)

screen shot 2017-08-18 at 15 23 11

## /profile/account

screen shot 2017-08-18 at 15 14 43

## /profile/account/edit

screen shot 2017-08-18 at 15 15 25

### DIFF
--- a/app/controllers/alaveteli_pro/pricing_controller.rb
+++ b/app/controllers/alaveteli_pro/pricing_controller.rb
@@ -1,0 +1,14 @@
+# -*- encoding : utf-8 -*-
+# Does not inherit from AlaveteliPro::BaseController as is pre-login
+class AlaveteliPro::PricingController < ApplicationController
+  before_filter :check_pro_pricing_enabled
+
+  def index
+  end
+
+  private
+
+  def check_pro_pricing_enabled
+    raise ActiveRecord::RecordNotFound unless feature_enabled?(:pro_pricing)
+  end
+end
--- a/app/controllers/users/accounts_controller.rb
+++ b/app/controllers/users/accounts_controller.rb
@@ -1,0 +1,20 @@
+# -*- encoding : utf-8 -*-
+class Users::AccountsController < ApplicationController
+  before_filter :check_pro_pricing_enabled
+
+  def show
+  end
+
+  def edit
+  end
+
+  def update
+    redirect_to users_account_path
+  end
+
+  private
+
+  def check_pro_pricing_enabled
+    raise ActiveRecord::RecordNotFound unless feature_enabled?(:pro_pricing)
+  end
+end
--- a/app/controllers/alaveteli_pro/users/sessions_controller.rb
+++ b/app/controllers/alaveteli_pro/users/sessions_controller.rb
@@ -1,0 +1,14 @@
+# -*- encoding : utf-8 -*-
+# Does not inherit from AlaveteliPro::BaseController as is pre-login
+class AlaveteliPro::Users::SessionsController < ApplicationController
+  before_filter :check_pro_pricing_enabled
+
+  def new
+  end
+
+  private
+
+  def check_pro_pricing_enabled
+    raise ActiveRecord::RecordNotFound unless feature_enabled?(:pro_pricing)
+  end
+end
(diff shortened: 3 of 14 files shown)


app/controllers/alaveteli_pro/pricing_controller.rb
```diff
@@ -0,0 +1,14 @@
+# -*- encoding : utf-8 -*-
+# Does not inherit from AlaveteliPro::BaseController as is pre-login
+class AlaveteliPro::PricingController < ApplicationController
```
garethrees: Hmm, maybe this should be called `AlaveteliPro::PlansController` after looking at https://github.com/mysociety/alaveteli-professional/issues/331#issuecomment-324043946. I was going to add `AlaveteliPro::PlansController#show` for the "contract" page (`GET /plans/pro`), so it would make sense for `AlaveteliPro::PlansController#index` to be the view of all plans. We could add a custom route if we prefer `/pro/pricing` instead of `/pro/plans`.

lizconlan:
> I think this also closes mysociety/alaveteli-professional#328 – it doesn't look like I need to do any more than declare a feature flag for it to operate correctly?

I think you might need to add something here to configure (enable/disable) it:
https://github.com/mysociety/alaveteli/blob/develop/config/initializers/alaveteli_features.rb

(Usage notes here: https://github.com/mysociety/alaveteli/tree/develop/gems/alaveteli_features#usage)

app/controllers/users/accounts_controller.rb
```diff
@@ -0,0 +1,20 @@
...
+  def update
+    redirect_to users_account_path
+  end
+
+  private
+
+  def check_pro_pricing_enabled
+    raise ActiveRecord::RecordNotFound unless feature_enabled?(:pro_pricing)
+  end
+end
```
lizconlan: probably worth having a stub filter method or `# TODO` in here as a visual reminder that these pages will be restricted to the owning user (makes the purpose of the controller more obvious at a glance)

lizconlan:
Oh and +1 on refactoring the user controller methods at some point

garethrees:
> a bit confused between the multiple naming conventions of `/alaveteli_pro`, `/pro`, `/profile` and `/account`

Here's how I've been distinguishing them:

* `/alaveteli_pro` – unfortunately not just `/pro` but we can fix that later – in any case, I'm sort of taking it to mean "logged in to a pro area"
* `/pro` – marketing stuff for Pro before you're signed-in in a pro area
* `/profile` – nothing to do with Pro. The root of your user settings
* `/account` – doesn't exist
* `/profile/account` – a subsection of your "profile" specific to your payment level. It would be better to rename this to `/profile/subscriptions` now that we know the stripe terminology better.

app/controllers/alaveteli_pro/pricing_controller.rb
```diff
@@ -0,0 +1,14 @@
...
+  before_filter :check_pro_pricing_enabled
+
+  def index
+  end
+
+  private
+
+  def check_pro_pricing_enabled
+    raise ActiveRecord::RecordNotFound unless feature_enabled?(:pro_pricing)
```
garethrees: It would be better to do this with a [feature constraint in `config/routes.rb`](https://github.com/mysociety/alaveteli/blob/develop/config/routes.rb#L608) – wasn't aware of this until today.

app/controllers/alaveteli_pro/pricing_controller.rb
garethrees: Superseded by https://github.com/mysociety/alaveteli/pull/4203

garethrees:
Removing labels and assignments as this is now just a reference for specific PRs

garethrees:
These are all implemented now.

PR #4184, opened 2017-08-23
[WIP][#4078] Require id in CSV import files

### LINKED ISSUE #4078: Spreadsheet (CSV) import broken
I can't work out why this won't import into WhatDoTheyKnow: (had to zip it as Github doesn't allow attaching .csv)
[schoolsimport.zip](https://github.com/mysociety/alaveteli/files/1117907/schoolsimport.zip)
On "dry runs" it says "error: line 1: invalid email 'tag_string' for authority 'request_email'
error: line 2: invalid email 'fei school' for authority 'office@tomlinscoteschool.com'"
etc.
Odd. Wrong character encoding? Wrong CR,LF? something else? I don't know but would appreciate your assistance.
thanks

garethrees:
Fixes #4078.

See CSV in https://github.com/mysociety/alaveteli/issues/4078 – doesn't
work.

Adding an `id` field with blank values allows the CSV to be imported
correctly.

I _think_ the changes made here are fine, but the spec failures need to
be looked at in more detail. I think what's happening is that the output
is just showing the newly added `id` attribute with the persisted value.

### DIFF
--- a/app/models/public_body.rb
+++ b/app/models/public_body.rb
@@ -38,6 +38,7 @@ class PublicBody < ActiveRecord::Base
   # [field_name, 'short description of field (basic html allowed)']
   cattr_accessor :csv_import_fields do
     [
+      ['id', ''],
       ['name', '(i18n)<strong>Existing records cannot be renamed</strong>'],
       ['short_name', '(i18n)'],
       ['request_email', '(i18n)'],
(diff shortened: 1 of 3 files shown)


garethrees:
Marking as new so that we assign some time in next sprint planning to take a bit of time on this.

lizconlan:
> I think the changes made here are fine, but the spec failures need to
be looked at in more detail. I think what's happening is that the output
is just showing the newly added id attribute with the persisted value.

Yeah, stale tests is all. 2 of the failures were comparing the list of expected fields and complaining about it including the new requirement we'd just added. The other was looking at values as well so I've included the actual ID number rather than trying to do something clever as it's being pulled from a fixture file.

garethrees:
I actually don't think this is valid given the newer comments in https://github.com/mysociety/alaveteli/issues/4078.

PR #4185, opened 2017-08-23
[pro#227] Add embargo date to padlock

lizconlan:
Show the embargo expiry on mouseover using the title element.

screen shot 2017-08-23 at 15 26 52

Extended to the sidebar of the request and draft request page views as it looks as though it would be useful there as well:

screen shot 2017-08-23 at 15 09 05

**Edit:** Now with more text, per review comments:

screen shot 2017-08-29 at 12 53 33

Closes mysociety/alaveteli-professional#227

### DIFF
--- a/app/views/alaveteli_pro/info_request_batches/_embargo_form.html.erb
+++ b/app/views/alaveteli_pro/info_request_batches/_embargo_form.html.erb
@@ -1,7 +1,13 @@
 <% if info_request_batch.embargo_duration.present? %>
 <div class="sidebar__section update-embargo">
   <h2 class="embargo-sidebar-heading">
-    <i class="embargo-indicator embargo-indicator--small"></i>
+    <i class="embargo-indicator embargo-indicator--small"
+       title="<%= _('Private until {{date}}',
+                    :date => simple_date(
+                               info_request_batch.
+                                 info_requests.first.embargo.publish_at,
+                               :format => :text)
+                   ) %>">
     <%= _("Privacy") %>
   </h2>
   <label class="houdini-label" for="input1"><%= _("Change privacy") %></label>
--- a/app/views/alaveteli_pro/info_requests/_info_request.html.erb
+++ b/app/views/alaveteli_pro/info_requests/_info_request.html.erb
@@ -2,7 +2,11 @@
      class="request request--<%= info_request.state.summarised_phase %>">
   <h3 class="request__title">
     <% if info_request.embargo %>
-      <span class="embargo-indicator embargo-indicator--small">
+      <span class="embargo-indicator embargo-indicator--small"
+            title="<%= _('Private until {{date}}',
+                         :date => simple_date(info_request.embargo.publish_at,
+                                              :format => :text)
+                        ) %>">
         <%= _("Private") %>
       </span>
     <% end %>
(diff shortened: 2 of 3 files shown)


app/views/alaveteli_pro/info_requests/_info_request.html.erb
```diff
@@ -2,7 +2,9 @@
      class="request request--<%= info_request.state.summarised_phase %>">
   <h3 class="request__title">
     <% if info_request.embargo %>
-      <span class="embargo-indicator embargo-indicator--small">
+      <span class="embargo-indicator embargo-indicator--small"
+                   title="<%= I18n.l(info_request.embargo.publish_at,
+                                    format: '%d %B %Y') %>">
```
garethrees: I think we should give more context here:

```erb
title="<%= _('Embargoed until {{date}}', :date => I18n.l(info_request.embargo.publish_at, format: '%d %B %Y')) %>
```

Also align `title` with `class`.

garethrees: There's also [a helper](https://github.com/mysociety/alaveteli/blob/develop/app/helpers/date_time_helper.rb#L16) that looks like it would be good to use to convert the date.

app/views/alaveteli_pro/info_request_batches/_embargo_form.html.erb
```diff
@@ -1,7 +1,9 @@
 <% if info_request_batch.embargo_duration.present? %>
 <div class="sidebar__section update-embargo">
   <h2 class="embargo-sidebar-heading">
-    <i class="embargo-indicator embargo-indicator--small"></i>
+    <i class="embargo-indicator embargo-indicator--small"
+       title="<%= I18n.l(info_request_batch.info_requests[0].embargo.publish_at,
```
garethrees: Use `first` instead of `[0]`.

`first` adds a `LIMIT` to the SQL, but the array accessor does not.

app/views/alaveteli_pro/info_requests/_info_request.html.erb
lizconlan: Cool, thanks - just having a look round to see if 'Embargoed until' is the right phrasing as I half remember a conversation that being a potentially confusing term

lizconlan: Looks like 'Private until' will fit in better with what we have in the Pro new request form. Will happily correct it later if I got that wrong

app/views/alaveteli_pro/info_request_batches/_embargo_form.html.erb
lizconlan: 👍

app/views/alaveteli_pro/info_requests/_info_request.html.erb
garethrees: Nope sounds good to me – have a quick check to see if we already do this anywhere else (I guess search for `publish_at`) but +1 on UI consistency.

lizconlan: The wording around making a new request (or viewing an existing one) from the dashboard 'Privacy' / 'private' so that feels like the right thing

lizconlan: Hmm, actually we have "This request will be private on {{site_name}} until {{date}}" but that seems too long for rollover text. Easy enough to tweak the working once we have the code so going to leave it as-is

PR #4186, opened 2017-08-23
[pro#162] extra menu tweaks

lizconlan:
screen shot 2017-08-23 at 16 57 31

This is intended as an addition to mysociety/whatdotheyknow-theme#405 rather than a replacement but the 2 fixes are independent.

The WDTK theme work fixes the theme-specific styling issues, whereas this adds a new class (rather hackily, improvements welcome) to identify a logged in Pro user (there's currently no way to work this out in CSS) and uses some additional margin space to force the `logged_in_bar` to be wider than the available space so it will appear below the login link as per @stevenday's [original suggestion](https://github.com/mysociety/whatdotheyknow-theme/pull/405#pullrequestreview-44873829)

Connects to mysociety/alaveteli-professional#162

### DIFF
--- a/app/assets/stylesheets/responsive/_header_layout.scss
+++ b/app/assets/stylesheets/responsive/_header_layout.scss
@@ -120,6 +120,14 @@
 
 }
 
+.no-js #banner {
+  @include respond-min( $main_menu-mobile_menu_cutoff ){
+    #logged_in_bar {
+      margin-left: 50%;
+    }
+  }
+}
+
 .site-title__logo {
   @extend .image-replacement;
   width: $logo-width;
--- a/doc/CHANGES.md
+++ b/doc/CHANGES.md
@@ -2,6 +2,10 @@
 
 ## Highlighted Features
 
+* Added some extra margin space to the `#logged_in_bar` when javascript is
+  disabled to avoid the user's name from overlapping the 'Sign out' link -
+  otherwise if there is enough space to do so, the secondary menu will try to
+  float alongside the nav bar content (Liz Conlan)
 * Make it clearer to users that they must complete an action when receiving the
   email to remind them to update the status of a request (Gareth Rees)
 * Removed non-responsive assets (Gareth Rees)


app/views/general/_log_in_bar.html.erb
```diff
@@ -1,4 +1,4 @@
-<li id="logged_in_bar" class="logged_in_bar account-link-menu-item">
+<li id="logged_in_bar" class="logged_in_bar account-link-menu-item <%= 'pro-account' if @user && @user.is_pro? %>">
```
garethrees: This class is unused.

Also, no idea _why_ this was added from the commit message.

app/assets/stylesheets/responsive/_header_layout.scss
```diff
@@ -120,6 +120,14 @@
 
 }
 
+.no-js #banner {
+  @include respond-min( $main_menu-mobile_menu_cutoff ){
+    #logged_in_bar {
+      margin-left: 3em;
```
garethrees: This isn't a great fix because some sites may have much shorter navs:

screen shot 2017-09-01 at 13 04 24

I think we can fix this for all eventualities with `margin-left: 70%;`:

screen shot 2017-09-01 at 13 04 18

garethrees:
Also add a changelog note :)

app/assets/stylesheets/responsive/_header_layout.scss
lizconlan: `margin-left: 40%` seems safer as 70% with a non-pro user logged in does this:

screen shot 2017-09-01 at 13 21 03

the first value that works is 20%, the last is 60% so choosing the one in the middle!

lizconlan: more scientifically, `50%` seems right, if the nav bar is shorter than 12 characters, that's different problem:

screen shot 2017-09-01 at 13 41 42

PR #4190, opened 2017-08-29
[#Pro337] Payment page

garethrees:
pro-payments

Work on https://github.com/mysociety/alaveteli-professional/issues/337.

TODO:

- [x] Finish calculating prices as Plan Price + VAT
- [x] Handle Stripe exceptions
- [x] Constrain payment routes by feature flag
- [x] Prevent user signing up to the same plan twice
- [x] Handle invalid plan (e.g. `/plans/iqwecb`)
- [x] Check without JS


garethrees:
Marked for initial review – I think a pairing session to look over what we have would be good and flesh out the remaining todo list?

lizconlan:
Dev/review setup docs, notes to self:
- create own Stripe account and paste the test key and secret into general.yml
- set up a Plan with an id of 'pro' (under the Subscriptions menu in the Stripe dashboard), can set the name to whatever you would like the "customer" to see

app/views/alaveteli_pro/plans/show.html.erb
```diff
@@ -12,7 +12,7 @@
           data-key="<%= AlaveteliConfiguration.stripe_publishable_key %>"
           data-name="<%= AlaveteliConfiguration.pro_site_name %>"
           data-description="A monthly subscription"
-          data-amount="<%= @plan.amount %>"
+          data-amount="<%= @display_price %>"
```
lizconlan: it tried to charge me 10p instead of £10 until I added ` * 100`

Gemfile.lock
```diff
@@ -504,3 +516,6 @@ DEPENDENCIES
   xapian-full-alaveteli (~> 1.2.21.1)
   xml-simple (~> 1.1.0)
   zip (~> 2.0.0)
+
+BUNDLED WITH
```
lizconlan: would prefer not to have the bundler version checked in

app/views/alaveteli_pro/plans/show.html.erb
```diff
@@ -3,7 +3,7 @@
 
 <%= form_tag(subscriptions_path) do %>
   <label class="amount">
-    <span>Amount: £<%= @plan.amount.to_f / 100 %></span>
+    <span>Amount: <%= number_to_currency(@display_price) %></span>
```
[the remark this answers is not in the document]
garethrees: Sorry should have added better notes – Yeah, need to make this locale specific

app/views/alaveteli_pro/plans/show.html.erb
garethrees: soz that's from the WIP VAT commit

Gemfile.lock
garethrees: ugh yeah

app/views/alaveteli_pro/plans/show.html.erb
lizconlan: just noting so it doesn't get lost!

garethrees:
Haven't handled most of the Stripe errors in `AlaveteliPro::PlansController#show` because they're unrecoverable – at the moment we _need_ to hit Stripe to get the plan as we don't have a local cache. The exception notification will alert us to any ongoing problems here.

app/controllers/alaveteli_pro/plans_controller.rb
```diff
@@ -4,7 +4,8 @@ class AlaveteliPro::PlansController < ApplicationController
   before_filter :authenticate
 
   def show
-    @plan = Stripe::Plan.retrieve(params[:id])
+    stripe_plan = Stripe::Plan.retrieve(params[:id])
```
lizconlan: this is going to be a bit clunky for reuse in the subscription code as we don't fetch the plan directly, it's access through `customer.subscription.plan`, having think

app/controllers/alaveteli_pro/plans_controller.rb
```diff
@@ -4,7 +4,8 @@ class AlaveteliPro::PlansController < ApplicationController
...
+    @plan = AlaveteliPro::PlanWithTax.new(stripe_plan)
```
lizconlan: this is going to be a bit clunky for reuse in the subscription code as we don't fetch the plan directly, it's accessed through `customer.subscription.plan`, having think

lizconlan: it's not too bad inside a helper method (or similar) https://github.com/mysociety/alaveteli/blob/d3f7c14b9cff179c40bd2c80e39f04b3dce00f62/app/helpers/alaveteli_pro/account_helper.rb#L29-L31

garethrees: `@plan = AlaveteliPro::PlanWithTax.new(customer.subscription.plan)` seems fine to me

garethrees: Its not much different to e.g. needing a `plus_vat` helper…

```ruby
@plan = Stripe::Plan.retrieve(params[:id])
@price = add_vat(@plan.amount)

@plan = customer.subscription.plan
@price = add_vat(@plan.amount)
```

…still a 2-step process?

lizconlan: sure, working with it but if I expand the pattern to make something similar to calculate discounts, I have 2 object instantiations, one to multiply by 1.2, the other to work out whether to take some money off

garethrees: You mean like?

```ruby
stripe_plan = Stripe::Plan.retrieve(params[:id])
@plan = PlanWithTax.new(PlanWithDiscount.new(plan, 20.0))
```
I quite like [this pattern](http://nithinbekal.com/posts/ruby-decorators/#applying-the-decorator-pattern).

lizconlan: I don't think that's going to work here as I'm working with a `Stripe::Subscription` not a `Stripe::Plan` as discounts belong to `Subscriptions`

It looks like discount is applied net, then the tax is calculated, let me try to work out Stripe's discount structure a bit better...

lizconlan: (that is nice, I want one - trying to work out how to get one)

lizconlan: bizarrely...

```
stripe_sub = Stripe::Subscription.retrieve(params[:id])
@plan = PlanWithTax.new(PlanWithDiscount.new(subscription))
```

spec/controllers/alaveteli_pro/plans_controller_spec.rb
```diff
@@ -0,0 +1,119 @@
...
+
+        it 'tells the user they already have a plan' do
+          expect(flash[:error]).to eq('You are already subscribed to this plan')
+        end
+
+        it 'redirects to the pro dashboard' do
+          expect(response).to redirect_to(alaveteli_pro_dashboard_path)
+        end
+
+        pending 'redirects to the account page' do
```
lizconlan: this syntax appears to cause 2.0.0 to throw actual RSpec errors - it doesn't like the undefined path variable - instead of treating the block as pending :(

removing the do block and supplying the code as a comment instead seems to work

spec/controllers/alaveteli_pro/subscriptions_controller_spec.rb
```diff
@@ -0,0 +1,284 @@
...
+      end
+
+      context 'when invalid params are submitted' do
+
+        it 'redirects to the plan page if there is a plan' do
+          post :create, :plan_id => 'pro'
+          expect(response).to redirect_to(plan_path('pro'))
+        end
+
+        pending 'redirects to the pricing page if there is no plan' do
```
lizconlan: this syntax appears to cause 2.0.0 to throw actual RSpec errors - it doesn't like the undefined path variable - instead of treating the block as pending :(

removing the do block and supplying the code as a comment instead seems to work

app/models/alaveteli_pro/plan_with_tax.rb
```diff
@@ -0,0 +1,22 @@
...
+#   @plan = PlanWithTax.new(plan)
+#   @plan.amount
+#   # => 833
+#   @plan.amount_with_tax
+#   # => 1000
+class AlaveteliPro::PlanWithTax < SimpleDelegator
+  TAX_RATE = BigDecimal('0.20').freeze
+
+  def amount_with_tax
+    net = BigDecimal(amount * 0.01, 0).round(2)
```
lizconlan: it looks (from the spec failure) like this will fail on Ruby 2.0.0 unless called as

`BigDecimal.new(amount * 0.01, 0).round(2)`

but the `TAX_RATE` assignment appears to be fine as-is o_0

Gemfile
```diff
@@ -163,6 +163,7 @@ group :test do
     gem 'term-ansicolor', '~> 1.3.0', '< 1.4'
   gem 'capybara', '~> 2.15.0'
   gem 'delorean', '~> 2.1.0'
+  gem 'stripe-ruby-mock', '~> 2.5.0'
```
lizconlan: we should add the platform spec from the `stripe` gem here as well

lizconlan: (not that that in itself will fix `rake spec` as it tries to load everything, finds `require 'stripe_mock'` and falls over but individual spec files will be runnable)

app/controllers/alaveteli_pro/plans_controller.rb
```diff
@@ -22,9 +22,19 @@ def authenticate
...
-      flash[:error] = _('You are already subscribed to this plan')
-      redirect_to alaveteli_pro_dashboard_path
+    customer_id = @user.pro_account.try(:stripe_customer_id)
+
+    if customer_id
+      customer = Stripe::Customer.retrieve(customer_id)
+      # TODO: This doesn't take the plan in to account
+      active_subscriptions =
+        customer.subscriptions.map(&:status).
+          any? { |status| status != 'canceled' }
```
garethrees: Actually a 'canceled' subscription is still "active" until the end of the billing period, so we want to block if `customer.subscriptions.any?`

config/initializers/stripe.rb
```diff
@@ -0,0 +1,2 @@
+# -*- encoding : utf-8 -*-
+Stripe.api_key = AlaveteliConfiguration.stripe_secret_key
```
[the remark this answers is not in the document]
garethrees: Going to drop 1.9

Gemfile
```diff
@@ -120,6 +120,7 @@ gem 'sass', '3.4.21'
 gem 'secure_headers', '~> 3.6.0'
 gem 'statistics2', '~> 0.54'
 gem 'strip_attributes', :git => 'https://github.com/mysociety/strip_attributes.git', :ref => 'c1c14da'
+gem 'stripe', '~> 3.3.1', platforms: [:ruby_20, :ruby_21, :ruby_22, :ruby_23]
```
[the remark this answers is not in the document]
garethrees: Going to drop 1.9

Gemfile
garethrees: Going to drop 1.9

app/controllers/alaveteli_pro/plans_controller.rb
garethrees: This _should_ change to `customer.subscriptions.any?`, but a subscription with a `'canceled'` state is just so that you have the object available after you delete the subscription. When you reload a customer after deleting their subscription, `customer.subscriptions.any?` is `false`

app/models/alaveteli_pro/plan_with_tax.rb
```diff
@@ -14,7 +14,9 @@ class AlaveteliPro::PlanWithTax < SimpleDelegator
   TAX_RATE = BigDecimal('0.20').freeze
 
   def amount_with_tax
-    net = BigDecimal((amount * 0.01), 0).round(2)
+    # Need to use BigDecimal.new here because SimpleDelegator is forwarding
```
lizconlan: Aha!

Issue #4166, opened 2017-08-14
Improve performance of AlaveteliPro::DashboardController#index


garethrees:
This has a 7-8 second average response time, mostly due to looking up `InfoRequestEvent`s:

![screen shot 2017-08-14 at 13 15 13](https://user-images.githubusercontent.com/282788/29271300-a68c68c6-80f2-11e7-8912-7e0c3c0999e4.png)

lizconlan:
The `Postgres AlaveteliPro::RequestSummary select` queries are ` SELECT COUNT` statements  inside the `_projects` template to calculate the total number or requests and the requests per 'phase'. If this is slowing the page load, we could add some counter caches to the user model instead.

lizconlan:
I've worked out how to reduce the number of `Postgres AlaveteliPro::RequestSummary select` ops by using a SQL group statement to retrieve the totals in a single lookup in the controller rather than running a separate COUNT for each one. This makes the `_projects` partial render a lot faster.

I need to work out how to do something similar for the `_to_do_list` bits to cut down the number of lookups on `InfoRequestEvent` but that's proving trickier at the moment.

There's probably some scope for optimisation in the `_activity_list` section as well.

Ideally these should share a single data load from the database but it's currently passing the user object around and fetching events and requests on demand which slows down when there's a lot of data.

HelenWDTK:
This issue has been automatically closed due to a lack of discussion or resolution for over 12 months.
Should we decide to revisit this issue in the future, it can be reopened.

PR #4195, opened 2017-08-30
[#4166] Fewer summary selects in AP Dashboard

lizconlan:
Connects to #4166

This addresses the second item in the Breakdown table:

screen shot 2017-08-30 at 15 31 54

The code in the original `projects` partial runs a SELECT COUNT statement for every possible category as it iterates over the request_summaries data which, in the dashboard template, is never used for anything else. This should reduced the number of `RequestSummary select`s down from 9+ to 1 on the Dashboard page and 2 in `InfoRequestsController#index` as it shares the `projects` partial and loads the request_summaries data for display.

For a user with 5 requests and 1 draft, reduces the number of ActiveRecord calls on the page from 54 down to 41. Replaces this N+1 SQL call

screen shot 2017-09-26 at 12 22 20

with this which is only called once per page view and runs in under a second:

screen shot 2017-09-26 at 12 18 02


app/controllers/alaveteli_pro/dashboard_controller.rb
```diff
@@ -13,5 +13,14 @@ def index
     @page = 1 if @page < 1
     @per_page = 10
     @activity_list = AlaveteliPro::ActivityList::List.new(@user, @page, @per_page)
+    @phase_counts = @user.request_summaries.
```
crowbot: This and the identical code in `info_requests_controller.rb` should be refactored to a single method. I'm thinking probably the `User` model would be a good place for it. That method should have some specs and should also handle the conversion of the counts to integers currently being done in templates.

app/controllers/alaveteli_pro/dashboard_controller.rb
```diff
@@ -13,5 +13,14 @@ def index
...
+                      joins(:request_summary_categories).
+                      references(:request_summary_categories).
+                      group("request_summary_categories.slug").
+                      count("request_summary_categories.id")
+    @phase_counts = @phase_counts.with_indifferent_access
+    @phase_counts[:total] = @phase_counts.values.reduce(0, :+)
```
crowbot: Unfortunately, there is a `RequestSummaryCategory` `embargo_expiring` that isn't mutually exclusive with the other slugs, so I think the total will need a separate query that just gets a total count of request summaries. This code will overcount the total in the case where some requests have an expiring embargo.

app/controllers/alaveteli_pro/dashboard_controller.rb
lizconlan: note to self: the `to_i` is a lazy way of handling nil as 0 - this could be better

lizconlan: > probably the `User` model

trying to do that indirectly in the first instance as referencing pro-only code directly in `User` feels a bit weird - may well abandon this approach

lizconlan:
**Note to reviewer** the first commit here gets reverted but I've left it in as I think it's helpful for understanding the initial approach and the first round review comments. Please let me know if you'd prefer that I remove it instead

spec/models/alaveteli_pro/user_phase_counts_spec.rb
```diff
@@ -71,6 +71,25 @@
 
     end
 
+    it 'caches the results' do
+      before = user.phase_count('awaiting_response')
+      AlaveteliPro::RequestSummary.
+        create_or_update_from(FactoryGirl.create(:info_request, user: user))
+      expect(user.phase_count('awaiting_response')).to eq(before)
```
garethrees: This isn't a great test, because `not_drafts` is currently not memoized and would therefore fail.

There's a design problem with this this module which I've addressed in the main review section, but this test should effectively be:

```ruby
expect(user.phase_counts).to eq({ awaiting_response: 2, not_drafts: 3 })
```

garethrees:
Merged via https://github.com/mysociety/alaveteli/pull/4325

PR #4204, opened 2017-09-07
[#4208] Run problematic query through #find_by_sql

garethrees:
See https://docs.google.com/document/d/1_bDeethGcXyvHFhVBnucpr1EXW5h7YC_WvVCk3Fgzu4/edit#

This query has started to run indefinitely on WDTK. It locks a passenger
process, so eventually results in a queue full error.

The query itself runs fine (in an SQL console), but for whatever reason
running it through arel just hangs.

Running it through `#find_by_sql` seems to work okay in the console.

Fixes https://github.com/mysociety/alaveteli/issues/4208.

### DIFF
--- a/app/controllers/admin_general_controller.rb
+++ b/app/controllers/admin_general_controller.rb
@@ -52,9 +52,9 @@ class AdminGeneralController < AdminController
                          @body_update_requests ].
       any?{ |to_do_list| ! to_do_list.empty? }
 
+    # HACK: Running this query through ActiveRecord freezes…
     @attention_comments = Comment.
-      where(:attention_requested => true).
-        not_embargoed
+      find_by_sql(Comment.where(attention_requested: true).not_embargoed.to_sql)
 
     @comment_tasks = [ @attention_comments ].
       any?{ |to_do_list| ! to_do_list.empty? }


app/controllers/admin_general_controller.rb
```diff
@@ -52,9 +52,14 @@ def index
                          @body_update_requests ].
       any?{ |to_do_list| ! to_do_list.empty? }
 
-    @attention_comments = Comment.
-      where(:attention_requested => true).
-        not_embargoed
+    # HACK: Running this query through ActiveRecord freezes…
+    attention_comments_sql = Comment.connection.unprepared_statement do
```
garethrees: Doesn't look like we need the `unprepared_statement` block actually. I'm convinced I needed this yesterday because [`to_sql` displays `$1` instead of real value after upgrade to Rails 4.1.1](https://github.com/rails/rails/issues/15331).

garethrees:
I generated a [load of data](https://gist.github.com/garethrees/bba5244fa42ff629bccdf93014435e87) (requires https://github.com/mysociety/alaveteli/pull/4207) to try to simulate this, but locally the summary page still works (112 public & 54 embargoed requiring attention):

screen shot 2017-09-08 at 09 34 05

It doesn't _seem_ like a problem with the amount of embargoes:

```ruby
# WDTK
AlaveteliPro::Embargo.count
# => 2084

# Local
AlaveteliPro::Embargo.count
# => 9661
```

WDTK has lots more InfoRequests and Comments, but I'm not convinced this is a load problem given the fix.

Object allocation maybe? But we _should_ only be loading Comment – the rest is just filtering.

PR #4209, opened 2017-09-08
Drop support for Ruby 1.9.x

garethrees:
*tentative new proposal - drop 1.9 without updating the gems to sneak it into release 0.31 to tell a better story about dropping support for Wheezy and updating the Vagrant configs, split gem updates into a new ticket*

- [x] Remove from CI build
- [x] Remove references to `ruby1.9.1` package ~#4471~
- [x] Drop Wheezy support (EOL May 2018) #4462
- [x] Drop Precise support (EOL April 2017) https://github.com/mysociety/alaveteli/pull/4330 & https://github.com/mysociety/alaveteli/pull/4331
- [x] Note suggested workarounds (rbenv, BrightBox packages) #4460
- [x] Check `Gemfile` for gems we can now upgrade – ticket any complicated ones
  - https://github.com/mysociety/alaveteli/issues/4246

### DIFF
--- a/lib/quiet_opener.rb
+++ b/lib/quiet_opener.rb
@@ -1,8 +1,5 @@
 # -*- encoding : utf-8 -*-
 require 'open-uri'
-if RUBY_VERSION.to_f < 2.0
-  require 'net/http/local'
-end
 
 def quietly_try_to_open(url, timeout=60)
   begin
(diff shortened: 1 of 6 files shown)


lib/quiet_opener.rb
```diff
@@ -30,23 +27,17 @@ def quietly_try_to_open(url, timeout=60)
...
 
-# On Ruby versions before 2.0, we need to use the net-http-local gem
+# On Ruby versions before 2.0, we needed to use the net-http-local gem
 # to force the use of 127.0.0.1 as the local interface for the
 # connection.  However, at the time of writing this gem doesn't work
 # on Ruby 2.0 and it's not necessary with that Ruby version - one can
 # supply a :local_host option to Net::HTTP:start.  So, this helper
 # function is to abstract away that difference, and can be used as you
 # would Net::HTTP.start(host) when passed a block.
 def http_from_localhost(host)
```
garethrees: I can't see this method even being called – could we remove it entirely?

lizconlan: I couldn't find where it was being called from either ... I'll have a quick look online while waiting for specs to run (if neither of those turn up anything interesting, I'll remove it)

lizconlan: (Got a clean spec run, no clues online so rebasing to remove that method before merge)

lizconlan:
Merged into the release branch

Issue #4210, opened 2017-09-11
Search results show embargoed request count


garethrees:
> Another query about a "missing request" ..  I've replied even though I
> can't actually see the request in question exists.
> 
> I suspect the number of requests appearing in the user search results is a bug?
> 
> https://www.whatdotheyknow.com/search/Jenna%20Corderoy/all
> 
> "1151 requests made"
> 
> vs
> 
> https://www.whatdotheyknow.com/user/jenna_corderoy
> 
> "This person has made no Freedom of Information requests using this site."

crowbot:
Fixed in https://github.com/mysociety/alaveteli/pull/4212.

PR #4212, opened 2017-09-11
Show visible, not total, request count in user search results

crowbot:
Related to #4210

### DIFF
--- a/spec/views/user/_user_listing_single.html.erb_spec.rb
+++ b/spec/views/user/_user_listing_single.html.erb_spec.rb
@@ -1,0 +1,34 @@
+# -*- encoding : utf-8 -*-
+require 'spec_helper'
+
+describe 'when displaying user listings' do
+  let(:highlighted_words) { [] }
+  let(:user) { FactoryGirl.create(:user) }
+
+  before do
+    assign :highlight_words, highlighted_words
+  end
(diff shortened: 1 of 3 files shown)


spec/views/user/_user_listing_single.html.erb_spec.rb
```diff
@@ -0,0 +1,34 @@
+# -*- encoding : utf-8 -*-
+require 'spec_helper'
+
+describe 'when displaying user listings' do
+  let(:highlighted_words) { [] }
+  let(:user){ FactoryGirl.create(:user) }
```
garethrees: missed a space

spec/views/user/_user_listing_single.html.erb_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+    assign :highlight_words, highlighted_words
+  end
+
+  def render_view
+    render :partial => 'user/user_listing_single',
+           :locals => { :display_user => user }
+  end
+
+  it 'displays a normal request' do
+    info_request = FactoryGirl.create(:info_request, :user => user)
```
garethrees: Don't really need to assign these as they're unused – initially added a little confusion of "where are these being used?"

PR #4219, opened 2017-09-12
Click a correspondence header to collapse it

### LINKED ISSUE #4085: Allow collapse and expansion of request correspondence
This is related to #748 but doesn't require any intelligence about responses - a Gmail-like collapse/expand for each bit of correspondence, and one for the whole thread.

zarino:
Fixes #4085.

To do:
* Integration tests for the javascript behaviour?


garethrees:
Individual:

collapse-expand-correspondence-individual

All:

collapse-expand-correspondence-all

wrightmartin:
This is all great.

Having the 'collapse all' control buried in the action menu wasn't where I'd have looked for it though. Enough to block the release though? I hope not.

garethrees:
> Having the 'collapse all' control buried in the action menu wasn't where I'd have looked for it though.

Same…

> Enough to block the release though? I hope not.

…but yeah, its a huge improvement as is. We can always move it later.

There are a couple of issues with this PR that I've found:

* When you click "collapse all", the menu item still says "Collapse all". Clicking it expands all. Really, it should say "Expand all". This sounds like a relatively easy fix.
* I found that in some cases in Chrome, clicking collapse all and then refreshing / coming back to the page resulted in the text getting tiny. I'm not 100% sure how to reproduce this; it seems hit or miss:

screen shot 2018-02-06 at 11 57 22

wrightmartin:
> I found that in some cases in Chrome, clicking collapse all and then refreshing / coming back to the page resulted in the text getting tiny. I'm not 100% sure how to reproduce this; it seems hit or miss:

I haven't looked into this deeply, but that kind of bug is usually due to an unclosed tag somewhere

zarino:
@garethrees The menu item label now alternates between "Collapse all correspondence" and "Expand all correspondence", as you’d expect.

But I think maybe I need to change the way the "Collapse all" menu item actually collapses the messages, because right now it just _toggles_ the state of all the messages in the thread, rather than _explicitly_ showing or hiding them all. So if you’ve already collapsed one (or more!) of the messages, and then you use the "Collapse all" menu item, the two messages you’d previously collapsed will expand themselves, while all the others will collapse as you intended. Grrrr. I’ll look at this tomorrow morning.

In the meantime, I can’t seem to replicate the "tiny text" bug in Chrome. I’ve tried various combinations of fiddling with the collapsable state and reloading the page, and I’ve tried a few different request pages.

zarino:
Ok, the "Collapse all" / "Expand all" now explicitly collapses or expands all the correspondence on the page, rather than just toggling the current state. 🎉

garethrees:
> I can’t seem to replicate the "tiny text" bug in Chrome

I can't now either. Wonder if I just had some weird cache issue.

Anyway, fixups look good so will get this rolled out

Issue #3233, opened 2016-04-28
Excel .xlsx file corrupted


kingqueen3065:
The spreadsheet attached to this response https://www.whatdotheyknow.com/request/complete_non_residential_busines_90#incoming-804454 won't open when downloaded from the site, but will when extracted from the original email; it appears that Alaveteli is in some way corrupting it.

garethrees:
I've just re-parsed the attachment and it seems okay now (won't work through view as html, but I can successfully open in Apple Preview / Numbers).

Possibly related to https://github.com/mysociety/alaveteli/issues/3106 or similar. Worth keeping an eye out for other cases like this.

RichardTaylor:
There is another case of an Excel file which works when the whole mail message is imported into a mail application but doesn't work when downloaded via WhatDoTheyKnow in the thread at 

https://www.whatdotheyknow.com/request/list_of_industrial_trade_effluen_5

garethrees:
I can open both of the `xlsx` files in preview on Mac, but I don't have excel installed so can't check there. Might be worth uploading to the file store and adding a comment?

kingqueen3065:
I did a direct comparison: I downloaded the XLS spreadsheet from the WhatDoTheyKnow request page, and I then downloaded the original raw email, imported it into my mail client and them saved the spreadsheet. This resulted in two files with the same name and same file size:

[161019 Whatdotheyknow com DOWNLOADED.xlsx](https://github.com/mysociety/alaveteli/files/548371/161019.Whatdotheyknow.com.DOWNLOADED.xlsx)
[161019_Whatdotheyknow_com EXPORTED.xlsx](https://github.com/mysociety/alaveteli/files/548372/161019_Whatdotheyknow_com.EXPORTED.xlsx)

But a binary comparison of the two reveals that there is a difference between the two. To be precise, there is a difference in 9 specific Hex values at 9 consecutive Hex addresses. The original numbers were a variety, but they were all replaced with the same Hex number 78.

```fc /b "161019 Whatdotheyknow com DOWNLOADED.xlsx" "161019_Whatdothey
know_com EXPORTED.xlsx"
Comparing files 161019 Whatdotheyknow com DOWNLOADED.xlsx and 161019_WHATDOTHEYK
NOW_COM EXPORTED.XLSX
000AE38E: 78 37
000AE38F: 78 5A
000AE391: 78 4E
000AE392: 78 50
000AE394: 78 72
000AE395: 78 72
[...]

PR #4224, opened 2017-09-13
[#4161] Replace zip gem with rubyzip

### LINKED ISSUE #4161: All "Download a zip file of all correspondence" zip files corrupted
Example: https://www.whatdotheyknow.com/request/prow_data/download - both the correspondence PDF and the attachment PDF fail CRC check, resulting in this in 7-Zip
ss 2017-08-13 at 12 50 21
and this in Windows Explorer.
ss 2017-08-13 at 12 51 32
I've tried with multiple requests, browsers and user accounts, all with the same result.

lizconlan:
Replaces the `zip` gem with the newer  (and actively maintained) `rubyzip` gem

Although it passes on Travis, the test introduced in a430046 passes locally under 1.9.3 but fails when testing the file size under 2.0.0 and 2.3.0 (using rvm) with

```
Failures:

  1) zipping pdf files correctly records the size of the zipped file
     Failure/Error: expect(zipped_size).to eq expected_size

       expected: 12228
            got: 12132
```

which mirrors the problem we're seeing on the live site

Fixes #4161


lizconlan:
adopting rubyzip may help us fix #3233

lizconlan:
> The also pushed a small change which could fix some of the XLSX file issues we've had - basically an extra carriage return at the end of zipped files - the only other differences I have seen are valid redactions of email addresses.

Ace, thanks! Unless you get to it first, I'm going to add a note to the commit to that effect so we can see why it's there (I need to rebase it anyway to prep it for merge - and to get rid of the rubocop warnings)

Note to self: write CHANGELOG note!

PR #4238, opened 2017-09-19
Reduce free period for Pro beta accounts

garethrees:
Discussed in "Memberships" meeting on 2017-09-18.


garethrees:
Reviewed via email thread – going to deploy on 2017-09-22.

PR #4243, opened 2017-09-21
[#Pro333] Cancel Account

garethrees:
Fixes https://github.com/mysociety/alaveteli-professional/issues/333

Adds basic cancellation of the `Stripe::Subscription`.

- [x] basic functionality / proof of concept
- [x] think about how/where to record account status changes as Stripe won't keep the history for us
- [x] how to record a former pro user - new flag? set `stripe_customer_id` to "cancelled"? (we might need to allow them read-only access to some of their pro data)

cancel-subscription

### DIFF
--- a/app/models/pro_account.rb
+++ b/app/models/pro_account.rb
@@ -14,6 +14,10 @@ class ProAccount < ActiveRecord::Base
   belongs_to :user,
              :inverse_of => :pro_account
 
+  def active?
+    stripe_customer.subscriptions.any?
+  end
+
   def stripe_customer
     Stripe::Customer.retrieve(stripe_customer_id) if stripe_customer_id
   end
--- a/app/models/pro_account.rb
+++ b/app/models/pro_account.rb
@@ -13,4 +13,8 @@
 class ProAccount < ActiveRecord::Base
   belongs_to :user,
              :inverse_of => :pro_account
+
+  def stripe_customer
+    Stripe::Customer.retrieve(stripe_customer_id) if stripe_customer_id
+  end
 end
--- a/spec/models/pro_account_spec.rb
+++ b/spec/models/pro_account_spec.rb
@@ -64,16 +64,14 @@ describe ProAccount do
 
     it 'returns true if there is an active subscription' do
       Stripe::Subscription.create(customer: customer,
-                                  plan: plan.id,
-                                  tax_percent: 20.0)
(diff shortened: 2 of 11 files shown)


lizconlan:
> think about how/where to record account status changes as Stripe won't keep the history for us

I was wrong about this - Stripe will (sort of) keep the history for us via the [Events API](https://stripe.com/docs/api#events) but we'd have to interpret that and make sure site users map 1:1 to Stripe customer ids

garethrees:
> - [x] basic functionality / proof of concept

This PR

- [x] think about how/where to record account status changes as Stripe won't keep the history for us

It does, in the Customer view in the Stripe dashboard. This does mean keeping a consistent `stripe_customer_id` for a `ProAccount`.

- [x] how to record a former pro user - new flag? set `stripe_customer_id` to "cancelled"? (we might need to allow them read-only access to some of their pro data)

See above – the `ProAccount` will still be linked to the `Stripe::Customer` – it will just have no active subscriptions.

garethrees:
I think this is ready for review. It covers the mechanics of issuing a cancellation notification to Stripe, but nothing more. Things like having access to existing pro requests etc will be covered in https://github.com/mysociety/alaveteli-professional/issues/342.

app/models/pro_account.rb
```diff
@@ -13,4 +13,8 @@
 class ProAccount < ActiveRecord::Base
   belongs_to :user,
              :inverse_of => :pro_account
+
+  def stripe_customer
+    Stripe::Customer.retrieve(stripe_customer_id)
```
lizconlan: What does a failure look like? Is there always a `stripe_customer_id`? (in theory there probably should be) Can we save an API lookup if there isn't?

spec/models/pro_account_spec.rb
```diff
@@ -44,4 +44,43 @@
...
+
+    it 'returns true if there is an expiring subscription' do
+      subscription =
+        Stripe::Subscription.create(customer: customer,
+                                    plan: plan.id,
+                                    tax_percent: 20.0)
+      subscription.delete(at_period_end: true)
+      expect(subject.active?).to eq(true)
+    end
+
```
lizconlan: I think an example of an instant cancellation (unlikely but we might end up banning someone) would be useful here - also demonstrates that someone with a fully cancelled subscription would be marked as not active

PR #4247, opened 2017-09-22
Add a note about using an organisational domain

crowbot:
Prompted by
https://medium.freecodecamp.org/how-i-hacked-hundreds-of-companies-through-their-helpdesk-b7680ddc2d4c

### DIFF
--- a/docs/running/server.md
+++ b/docs/running/server.md
@@ -61,6 +61,13 @@ these settings:
 You should consider running the admin part of the site over HTTPS. This can be
 achieved with rewrite rules that redirect URLs beginning with `/admin`.
 
+Additionally, the `INCOMING_EMAIL_DOMAIN` should not be the one that you use for your organisational email.
+Mail sent to Alaveteli request addresses is published on the site, and using an organisational email domain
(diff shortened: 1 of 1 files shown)


docs/running/server.md
```diff
@@ -61,6 +61,12 @@ these settings:
 You should consider running the admin part of the site over HTTPS. This can be
 achieved with rewrite rules that redirect URLs beginning with `/admin`.
 
+Additionally, the `INCOMING_EMAIL_DOMAIN` should not be the one that you use for your organisational email.
+Mail sent to Alaveteli request addresses is published on the site, and using an organisational email domain
+would leave you vulnerable to attacks that sign up for your internal tools using these addresses and use
```
garethrees: I think I'd link to the blog post here

crowbot: Yeah, I thought about this but was wondering about it's longevity. I guess we can always remove or update later. Will add.

garethrees: Yeah, and there's always archive.org if the link disappears.

PR #4232, opened 2017-09-15
Remove PurgeRequest

crowbot:
Fixes #4025

PR #4252, opened 2017-09-22
Deprecate the use of Varnish purging

crowbot:
We don't use it as it wasn't stable on testing


garethrees:
👍

garethrees:
Removed in https://github.com/mysociety/alaveteli/pull/4232

PR #4256, opened 2017-09-27
Add * to the end of Disallow entries

garethrees:
Baiduspider seems to support `*` and `$` [1] but we were seeing lots of
hits to `/similar/` on AskTheEU.

There were no hits in the access logs for the existing rules that had a
trailing `*`, so our best guess is that it needs the explicit trailing
wildcard to follow the rule.

[1] http://www.baidu.com/search/robots_english.html


garethrees:
If we're happy to try, we should cherry pick this on to a new branch `hotfix/0.25.0.26` and try it out on AskTheEU for a week or so.

garethrees:
Deployed on AskTheEU via https://github.com/mysociety/alaveteli/tree/hotfix/0.25.0.26.

Will give this a week or so to see if it makes any difference.

garethrees:
No change in frequency, so I'm not sure this has made a difference:

```
$ cat logs/https_access_log | g 'Baiduspider' | g similar | cut -d[ -f2 | cut -d] -f1 | awk -F: '{print $2":00"}' | sort -n | uniq -c
    105 00:00
    480 01:00
    489 02:00
    500 03:00
    517 04:00
    479 05:00
    494 06:00
    494 07:00
    471 08:00
    518 09:00
    480 10:00
    511 11:00
     97 12:00
```

garethrees:
Much better now:

```
$ cat logs/https_access_log.2017-10-09 | g 'Baiduspider' | g similar | cut -d[ -f2 | cut -d] -f1 | awk -F: '{print $2":00"}' | sort -n | uniq -c
     93 00:00
     80 01:00
     94 02:00
     64 03:00
     91 04:00
     83 05:00
     97 06:00
     88 07:00
     97 08:00
    129 09:00
    118 10:00
    102 11:00
    112 12:00
     92 13:00
     91 14:00
    225 15:00
     73 16:00
     83 17:00
     97 18:00
     91 19:00
     77 20:00
     76 21:00
     88 22:00
     89 23:00
```

PR #4257, opened 2017-09-27
Update newrelic_rpm to latest version

crowbot:
This version removes `developer_mode`, more information at
https://discuss.newrelic.com/t/feedback-on-the-ruby-agent-s-developer-mode/46957


lizconlan:
I quite like ScoutApp as an alternative but they've introduced a bug at the moment which stops it from showing the full call breakdown info - they're aware and working on a solution

PR #4263, opened 2017-09-29
[#Pro372] Cancelled view requests

garethrees:
Work on https://github.com/mysociety/alaveteli-professional/issues/372.

This PR covers the display of private requests on the user profile / "my requests" page.

screen shot 2017-09-29 at 16 34 40

Here's the difference between the display of the request game:

**Before:**

screen shot 2017-09-29 at 16 36 11

**After:**

screen shot 2017-09-29 at 16 35 30

### DIFF
--- a/app/controllers/user_controller.rb
+++ b/app/controllers/user_controller.rb
@@ -22,7 +22,28 @@ class UserController < ApplicationController
 
     set_show_requests if @show_requests
 
+    @private_requests = []
+
     if @is_you
+      private_requests =
+        @display_user.
+          info_requests.
+          visible_to_requester.
+          embargoed
+
+      if params[:user_query]
+        private_requests = private_requests.
+          where("info_requests.title ILIKE :q", q: "%#{ params[:user_query] }%")
+      end
+
+      unless params[:request_latest_status].blank?
+        private_requests = private_requests.
+          where(described_state: params[:request_latest_status])
+      end
+
+      @private_requests =
+        private_requests.page(params[:page]).per_page(@per_page)
+
       # All tracks for the user
       @track_things = TrackThing.
         where(:tracking_user_id => @display_user, :track_medium => 'email_daily').
--- a/app/models/info_request.rb
+++ b/app/models/info_request.rb
@@ -129,6 +129,7 @@ class InfoRequest < ActiveRecord::Base
   scope :embargoed, Prominence::EmbargoedQuery.new
   scope :not_embargoed, Prominence::NotEmbargoedQuery.new
   scope :embargo_expiring, Prominence::EmbargoExpiringQuery.new
+  scope :visible_to_requester, Prominence::VisibleToRequesterQuery.new
 
   scope :awaiting_response, State::AwaitingResponseQuery.new
   scope :response_received, State::ResponseReceivedQuery.new
--- a/app/views/user/show/_show_requests.html.erb
+++ b/app/views/user/show/_show_requests.html.erb
@@ -3,7 +3,7 @@
     <div>
       <%= text_field_tag :user_query,
                          params[:user_query],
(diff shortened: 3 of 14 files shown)


app/views/user/show/_show_requests.html.erb
```diff
@@ -95,6 +95,25 @@
     <% end %>
   <% end %>
 
+  <% if @private_requests.any? %>
```
garethrees: > TODO: Check that the `request_path` helper is okay to use in this
context.

Yes, the link on the page is `/request/example_title_80_2`, which redirects to `http://10.10.10.30:3000/alaveteli_pro/info_requests/example_title_80_2` when clicked. This might change depending on how we allow access to embargoed requests when a user no longer has a pro role, but works for now. Will reword on merge

spec/controllers/user_controller_spec.rb
```diff
@@ -109,6 +109,14 @@ def make_request
           not_to match(/change password, subscriptions and more/)
       end
 
+      it 'does not show private requests' do
```
lizconlan: This is where the certainty of 'logged in viewing another user's requests' would be nice. Possibly also what would an admin see? (I'm guessing a standard `:admin` wouldn't see the private requests but a `:pro_admin` would?)

app/models/info_request.rb
```diff
@@ -129,6 +129,7 @@ class InfoRequest < ActiveRecord::Base
   scope :embargoed, Prominence::EmbargoedQuery.new
   scope :not_embargoed, Prominence::NotEmbargoedQuery.new
   scope :embargo_expiring, Prominence::EmbargoExpiringQuery.new
+  scope :visible_to_requester, Prominence::VisibleToRequesterQuery.new
```
lizconlan: this bugs me slightly as it only does what it says when applied to a user (I have no alternative suggestion)

spec/controllers/user_controller_spec.rb
```diff
@@ -76,7 +76,13 @@ def make_request
 
     end
 
-    context 'when the user being shown is logged in' do
+    context 'when logged in viewing your own requests' do
```
lizconlan: Do we need an equivalent 'when logged in viewing another user's requests'? I can only see this and 'not logged in'

app/controllers/user_controller.rb
```diff
@@ -22,7 +22,28 @@ def show
 
     set_show_requests if @show_requests
 
+    @private_requests = []
```
lizconlan: tiny nitpick - typo of "requests" in the extended commit message for 78fa5bd3a2!

spec/controllers/user_controller_spec.rb
garethrees: Currently, only the actual user will see requests (`@is_you == true`).

This is the same for the other "personal" things – `track_things` and `undescribed_requests`.

I did think about this, but at the moment there's a _lot_ of duplication. I think we need to [review private request access permissions](https://github.com/mysociety/alaveteli-professional/issues/213) and stub out `can?` calls so that we can use the `Ability` system for this kind of testing.

garethrees: I've pushed some extra specs for viewing other users though, to cover the most important case.

app/models/info_request.rb
garethrees: I'd read it as "all requests in a state that would allow the request owner to view them" – not quite as snappy for a scope name though 🤣

garethrees: I've added that description to the inline class documentation.

app/controllers/user_controller.rb
garethrees: Will try to remember to reword in the squash :)

PR #4264, opened 2017-10-03
[#pro372] Allow ex-pro users to view embargoed requests

garethrees:
Work on https://github.com/mysociety/alaveteli-professional/issues/372.
Also fixes https://github.com/mysociety/alaveteli-professional/issues/367.

* Removes the `is_pro?` filter from the redirect to the pro path. We
  will perform this redirect for all embargoed requests.
* Sets `@in_pro_area` for all embargoed requests so that its clearer
  that the request is still embargoed.

### DIFF
--- a/app/controllers/request_controller.rb
+++ b/app/controllers/request_controller.rb
@@ -95,6 +95,12 @@ class RequestController < ApplicationController
         return render_hidden
       end
 
+      # Always show the pro livery if a request is embargoed. This makes it
+      # clear to admins and ex-pro users that the `InfoRequest` is still
+      # private. Users who are not permitted to view the request are redirected
+      # so we don't need to consider the `current_user` here.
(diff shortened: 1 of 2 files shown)


lizconlan:
(Note - exposes the "Keep private for a further:" dropdown that allows the embargo to be extended)

garethrees:
>  exposes the "Keep private for a further:" dropdown that allows the embargo to be extended

Yes, the actual blocking of that should get tackled in https://github.com/mysociety/alaveteli-professional/issues/374

lizconlan:
> > exposes the "Keep private for a further:" dropdown that allows the embargo to be extended
>
> Yes, the actual blocking of that should get tackled in mysociety/alaveteli-professional#374

👍 just realised it was weird if the review assumed this was fine and didn't even mention it

PR #4265, opened 2017-10-03
Bump Rails version

garethrees:
* Fix regression in behavior of `normalize_path`.
* `Relation#joins` is no longer affected by the target model’s
  `current_scope`, with the exception of `unscoped`.

http://weblog.rubyonrails.org/2017/9/27/Rails-4-2-10-released/


Gemfile.lock
```diff
@@ -498,3 +500,6 @@ DEPENDENCIES
   xapian-full-alaveteli (~> 1.2.21.1)
   xml-simple (~> 1.1.0)
   zip (~> 2.0.0)
+
+BUNDLED WITH
```
lizconlan: shame we can't configure bundler not to do that

garethrees: Sorry, I keep missing that.

garethrees:
Merged in 8e9241724

PR #4271, opened 2017-10-05
[pro#374] restrict access to embargo editing

lizconlan:
Closes mysociety/alaveteli-professional#374

## Embargoes
- [x] create new destroy embargo ability and apply
   - [ ] ~~apply to batch embargoes~~
- [x] alter the update embargo ability to check for an active subscription
- [x] only show update embargo options other than "Publish Now" if update ability is available

### DIFF
--- a/app/models/ability.rb
+++ b/app/models/ability.rb
@@ -84,6 +84,10 @@ class Ability
         user && (user == embargo.info_request.user || user.is_pro_admin?)
       end
 
+      # Removing embargoes
+      can :destroy, AlaveteliPro::Embargo do |embargo|
+        user && (user == embargo.info_request.user || user.is_pro_admin?)
+      end
     end
 
     can :admin, AlaveteliPro::Embargo if user && user.is_pro_admin?
--- a/app/models/ability.rb
+++ b/app/models/ability.rb
@@ -74,9 +74,21 @@ class Ability
 
       # Creating embargoes
       can :create_embargo, InfoRequest do |info_request|
-        user && info_request.user.is_pro? && !info_request.embargo &&
-          (user == info_request.user || user.is_pro_admin?) &&
-          info_request.info_request_batch_id.nil?
+        if feature_enabled? :pro_pricing
+          user && (info_request.user.is_pro? &&
+                   !info_request.embargo &&
+                   info_request.info_request_batch_id.nil? &&
+                   ((info_request.is_actual_owning_user?(user) &&
+                     self.class.active_pro_subscription?(info_request.user)) ||
+                    user.is_pro_admin?)
+                  )
+        else
+          user && (info_request.user.is_pro? &&
+                   !info_request.embargo &&
+                   info_request.info_request_batch_id.nil? &&
+                   (info_request.is_actual_owning_user?(user) ||
+                    user.is_pro_admin?))
+        end
       end
 
       # Extending embargoes
--- a/app/models/ability.rb
+++ b/app/models/ability.rb
@@ -81,7 +81,16 @@ class Ability
 
       # Extending embargoes
       can :update, AlaveteliPro::Embargo do |embargo|
-        user && (user == embargo.info_request.user || user.is_pro_admin?)
+        if feature_enabled? :pro_pricing
+          user && (user.is_pro_admin? ||
+                   user == embargo.info_request.user &&
+                   (user.is_pro? &&
+                    self.class.active_pro_subscription?(user)))
+        else
+          user && (user.is_pro_admin? ||
+                   user == embargo.info_request.user &&
+                   user.is_pro?)
+        end
       end
 
       # Removing embargoes
@@ -131,6 +140,10 @@ class Ability
   end
 
   private
+
+  def self.active_pro_subscription?(user)
+    user.pro_account && user.pro_account.active?
+  end
 
   def self.can_update_request_state?(user, request)
     (user && request.is_old_unclassified?) || request.is_owning_user?(user)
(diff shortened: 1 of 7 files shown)


spec/models/pro_account_spec.rb
```diff
@@ -101,35 +103,53 @@
       subject.update!(stripe_customer_id: customer.id)
     end
 
+
```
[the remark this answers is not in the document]
lizconlan: gah, tired!

lizconlan:
The new CanCan logic could be DRY-ed up a bit more but putting this up for review as-is (sorry reviewer)

spec/integration/alaveteli_pro/view_request_spec.rb
```diff
@@ -88,6 +88,26 @@
     end
   end
 
+  context 'the user does not have an active subscription' do
+
+    before do
+      pro_user.pro_account.update!(stripe_customer_id: nil)
```
garethrees: As mentioned above, this is not what happens. The removal of the role is what we care about here.

garethrees: More importantly, I think this should contain a spec to assert that it _doesn't_ allow a user to extend an embargo.

spec/models/ability_spec.rb
```diff
@@ -555,10 +555,27 @@
...
+    it 'allows the info request owner to remove it' do
+      with_feature_enabled(:alaveteli_pro) do
+        ability = Ability.new(embargo.info_request.user)
+        expect(ability).to be_able_to(:destroy, embargo)
+      end
+    end
+
+    it 'allows a non-pro info request owner to remove it' do
+      with_feature_enabled(:alaveteli_pro) do
+        allow(embargo.info_request.user).to receive(:is_pro?).and_return(false)
```
garethrees: `embargo.info_request.user.remove_role(:pro)` would be better here?

spec/models/ability_spec.rb
```diff
@@ -555,10 +555,27 @@
...
+        ability = Ability.new(embargo.info_request.user)
+        expect(ability).to be_able_to(:destroy, embargo)
+      end
+    end
```
garethrees: I think we should also test that:

- A pro admin can destroy
- A normal admin cannot destroy
- Another user cannot destroy
- Anonymous users cannot destroy

garethrees: "Destroying Embargoes" deserves its own `describe` block.

app/models/ability.rb
```diff
@@ -132,6 +141,10 @@ def initialize(user)
 
   private
 
+  def self.active_pro_subscription?(user)
+    user.pro_account && user.pro_account.active?
```
garethrees: I think we can just look for the `pro` role here, otherwise we'll be hammering Stripe with API calls.

Why do we need to check the `pro_account` status?

* When you become a pro, you get the `pro` role applied.
* When your subscription expires, you get the `pro` role removed (this isn't implemented yet, but in progress in https://github.com/mysociety/alaveteli-professional/issues/371).

lizconlan: I based this on the same logic that we're currently using to determine whether you can see the dashboard which is to check `pro_account.active?` (which looks for the `stripe_customer` and tests for active Stripe subscriptions - obviously this was written before we realised we might have a clash with other apps - is this not correct?

As an interesting counter example - my main WDTK user has the pro role but not dashboard access

lizconlan: wait, that's not quite right - I have dashboard but get bounced out of `SubscriptionsController#index`

garethrees: I've had a look and can't find any use of `ProAccount#active?` to control access to the dashboard.

We use it in two places, where we're directly making use of Stripe objects:

* [PlansController](https://github.com/mysociety/alaveteli/blob/dacbdd3/app/controllers/alaveteli_pro/plans_controller.rb#L38)
* [SubscriptionsController](https://github.com/mysociety/alaveteli/blob/dacbdd3/app/controllers/alaveteli_pro/subscriptions_controller.rb#L146)

These were added because at the time, some users had the `pro` role, but no `Stripe::Subscription`. This was a slightly different use case to the normal permission access, as in these two situations we were letting someone modify their access (either by signing up for an account, or landing on the page that allowed them to cancel). Most pro features rely on the `pro` role being assigned.

Now that _all_ users with the `pro` role have subscriptions, I don't think we need to check against Stripe in either of these cases.

garethrees: > I have dashboard but get bounced out of SubscriptionsController#index
 
Your user doesn't have a `pro_account`, so you wouldn't have been picked up by the [backloading script](https://github.com/mysociety/alaveteli-professional/issues/382#issuecomment-339703490):

```ruby
User.pro.map { |u| u.name if u.pro_account.nil? }.compact
# => ["Liz Conlan"]
```

You'll need to follow the [stripped down script](https://github.com/mysociety/alaveteli-professional/wiki/Pro-Account-Request-Response-Script) to create yourself a `pro_account` record and a stripe customer.

lizconlan: cool, I'll fix my account eventually (my fault tbh), will rewrite (simplify) the checks based on the pro role

garethrees: Feel free to force-push these. I haven't reviewed enough for it to matter, and it will be easier to just assess with the updated logic given it will be a pretty big change.

garethrees: Or maybe just make a new PR – whichever works :)

lizconlan:
Closed in favour of #4409

PR #4272, opened 2017-10-06
Update Stripe Gem

garethrees:
=== 3.4.1 2017-10-05

* [586] Log query strings as well as form bodies with STRIPE_LOG
* [588] Require minimum Faraday 0.10 for bug fix in parameter encoding

=== 3.4.0 2017-09-20

* Mark legacy Bitcoin API as deprecated, and remove corresponding tests
* Mark recipients API as deprecated, and remove recipient card tests

=== 3.3.2 2017-09-20

* Correct minimum required Ruby version in gemspec (it's 2.0.0)

https://github.com/stripe/stripe-ruby/blob/v3.4.1/History.txt#L1


Gemfile.lock
```diff
@@ -510,3 +510,6 @@ DEPENDENCIES
   xapian-full-alaveteli (~> 1.2.21.1)
   xml-simple (~> 1.1.0)
   zip (~> 2.0.0)
+
+BUNDLED WITH
```
garethrees: sigh, will fix

lizconlan: ta

PR #4273, opened 2017-10-06
[pro#380] Fix changing the recipient of a draft

garethrees:
Fixes https://github.com/mysociety/alaveteli-professional/issues/380

The error when clicking the authority search box was:

    Uncaught TypeError: Failed to execute 'appendChild' on 'Node':
    parameter 1 is not of type 'Node'.

This took us to the `groups[optgroup].appendChild(option_html);` line. I
jammed a `console.log(option_html);` above it to see what the value was:
`undefined`.

Okay so then try performing the operation only `if(option_html)`… it
works!

That's about the limit of my understanding of this fix, so I think we
might want to report upstream in
https://github.com/selectize/selectize.js/issues/1188 with our proposed
fix to see what they think about it. It might well have unforseen
consequences.

### DIFF
--- a/spec/helpers/alaveteli_pro/public_bodies_helper_spec.rb
+++ b/spec/helpers/alaveteli_pro/public_bodies_helper_spec.rb
@@ -1,0 +1,48 @@
+# -*- encoding : utf-8 -*-
+require 'spec_helper'
+
+describe AlaveteliPro::PublicBodiesHelper, type: :helper do
+  let(:public_body) { FactoryGirl.create(:public_body) }
+
+  describe '#public_body_search_attributes' do
+    let(:html) { double(:html) }
+    let(:expected) do
+      {
+        id: public_body.id,
+        name: public_body.name,
+        short_name: public_body.short_name,
+        notes: public_body.notes,
+        info_requests_visible_count: public_body.info_requests_visible_count,
+        about: _('About {{public_body_name}}',
+                 public_body_name: public_body.name),
+        html: html
+      }
+    end
+
+    before do
+      # Stubbing this so we can test both branches of this conditional -
+      # In reality from within an view, `render` calls seems to internally uses
+      # `render_to_string` but we just can't call it directly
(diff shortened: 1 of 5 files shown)


gbp:
Guess I should describe what I've changed...

Instead of editing an upstream dependency I was able to fix this error by making a change to the Alaveteli code. All we needed to do is pass a `html` attribute for the currently selected recipient.  

Before this change the selectize instance wouldn't of had the HTML of the current option after reload as its only be loaded by the search results AJAX response. I have refactored to both the initial page load the search results use the same helper so attributes should be kept in 'sync' from now on.

This change it fixes the `Uncaught TypeError` and also allows the previous selected item to be reselected (say if it was accidentally removed) which wasn't possible previously.

spec/helpers/alaveteli_pro/public_bodies_helper_spec.rb
```diff
@@ -0,0 +1,46 @@
+# -*- encoding : utf-8 -*-
```
garethrees: This commit can be squashed in to c7e567b06b4667d36dfff381794b81aa40723a6b.

spec/helpers/alaveteli_pro/public_bodies_helper_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+        html: html
+      }
+    end
+
+    before do
+      # Stubbing this so we can test both branches of this conditional -
+      # In reality from within an view, `render` calls seems to internally uses
+      # `render_to_string` but we just can't call it directly
+      allow(helper).to receive(:respond_to?).with(:render_to_string)
+        .and_return(in_controller)
```
garethrees: We use trailing dot for multi-line method chaining

gbp: We should enabled Rubocop's `Layout/DotPosition` with `trailing` style then

PR #4279, opened 2017-10-10
Prevent ProAccount#active? throwing an error without stripe customer id

### LINKED ISSUE #4278: Pro signup page throws an error that prevents some users clicking through
Possibly a bit of an edge case but one that people with existing test data are likely to hit.

The signup process is halted at the Plans#show controller with an "undefined method `subscriptions` for nil::NilClass" error (attempting to call `active?` for `currrent_user.pro_account`) if the user has a `pro_account` record but no stored Stripe customer id (may also apply if the customer id doesn't match a Stripe account)

lizconlan:
Fixes #4278

### DIFF
--- a/app/models/pro_account.rb
+++ b/app/models/pro_account.rb
@@ -15,10 +15,14 @@ class ProAccount < ActiveRecord::Base
              :inverse_of => :pro_account
 
   def active?
-    stripe_customer.subscriptions.any?
+    stripe_customer.present? && stripe_customer.subscriptions.any?
   end
 
   def stripe_customer
+    @stripe_customer ||= stripe_customer!
+  end
+
+  def stripe_customer!
     Stripe::Customer.retrieve(stripe_customer_id) if stripe_customer_id
   end
 end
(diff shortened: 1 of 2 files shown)


app/models/pro_account.rb
```diff
@@ -15,10 +15,12 @@ class ProAccount < ActiveRecord::Base
...
   def active?
-    stripe_customer.subscriptions.any?
+    stripe_customer.present? && stripe_customer.subscriptions.any?
   end
 
   def stripe_customer
-    Stripe::Customer.retrieve(stripe_customer_id) if stripe_customer_id
+    if stripe_customer_id
+      @stripe_customer ||= Stripe::Customer.retrieve(stripe_customer_id)
+    end
```
garethrees: +1 to memoization, but you've introduced it without tests.

Usually I'd define a non-memoized version, and then memoize the result of that:

```ruby
  def stripe_customer
    @stripe_customer ||= stripe_customer!
  end

  def stripe_customer!
    Stripe::Customer.retrieve(stripe_customer_id) if stripe_customer_id
  end
```

Its also worth explaining _why_ an error is caused in the commit message. All it needs to say is that "we're calling `#subscriptions` on `nil`"

PR #4282, opened 2017-10-11
[pro#368] Basic design for /profile/subscriptions page

wrightmartin:
Fixes https://github.com/mysociety/alaveteli-professional/issues/368

screen shot 2017-10-11 at 12 41 42

### DIFF
--- a/app/views/alaveteli_pro/subscriptions/_card.html.erb
+++ b/app/views/alaveteli_pro/subscriptions/_card.html.erb
@@ -1,30 +1,42 @@
-<%= card.brand %>
-<%= card.last4 %>
-<%= card.exp_month %>/<%= card.exp_year %>
-<%= card_expiry_message(card.exp_month, card.exp_year) %>
+<div class="settings__item">
+  <div class="settings__item__primary">
+    <span class="card-desc">
+      <%= _('<strong>{{brand}}</strong> ending in {{last4}}, expiry ' \
+            '{{exp_month}}/{{exp_year}}',
+            brand: card.brand,
+            last4: card.last4,
+            exp_month: card.exp_month,
+            exp_year: card.exp_year)
+        %>
+    </span>
 
-<%= form_tag(update_subscription_payment_path) do %>
+    <%= card_expiry_message(card.exp_month, card.exp_year) %>
+  </div>
 
-  <%= hidden_field_tag 'old_card_id', card.id %>
+  <div class="settings__item__secondary">
+    <%= form_tag(update_subscription_payment_path) do %>
+      <%= hidden_field_tag 'old_card_id', card.id %>
 
-  <script src="https://checkout.stripe.com/checkout.js" class="stripe-button"
-          data-key="<%= AlaveteliConfiguration.stripe_publishable_key %>"
-          data-name="<%= AlaveteliConfiguration.pro_site_name %>"
-          data-panel-label="Update Card Details"
-          data-label="Update Card Details"
-          data-email="<%= current_user.email %>"
-          data-locale="auto"
-          data-allow-remember-me=false
-          data-image="https://s3.amazonaws.com/stripe-uploads/acct_19EbqNIbP0iBLddtmerchant-icon-1479145884111-mysociety-wheel-logo.png"
-          data-zip-code=false>
-  </script>
+      <script src="https://checkout.stripe.com/checkout.js" class="stripe-button"
+              data-key="<%= AlaveteliConfiguration.stripe_publishable_key %>"
+              data-name="<%= AlaveteliConfiguration.pro_site_name %>"
+              data-panel-label="Update Card Details"
+              data-label="Update Card Details"
+              data-email="<%= current_user.email %>"
+              data-locale="auto"
+              data-allow-remember-me=false
+              data-image="https://s3.amazonaws.com/stripe-uploads/acct_19EbqNIbP0iBLddtmerchant-icon-1479145884111-mysociety-wheel-logo.png"
+              data-zip-code=false>
+      </script>
 
-  <noscript>
-    <div id="error">
-      <p>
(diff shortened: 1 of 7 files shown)


wrightmartin:
```£35.99 Billed: Monthly```

I couldn't quite figure out how to get this to say

```£35.99 per month```

File: app/views/alaveteli_pro/subscriptions/_card.html.erb [hunk 1]
```diff
@@ -1,30 +1,35 @@
...
+    <noscript>
+      <div id="error">
+        <p>
+          <%= _('Updating your payment card requires your browser to have ' \
+                'JavaScript enabled.') %>
+        </p>
+      </div>
+    </noscript>
+  </div>
+<%= card_expiry_message(card.exp_month, card.exp_year) %>
```
garethrees: I can't tell if this has been put here intentionally or missed.

I'd say it should probably be bold & red?

screen shot 2017-10-12 at 15 00 38

app/views/alaveteli_pro/subscriptions/_subscription.html.erb
```diff
@@ -1,41 +1,66 @@
...
-                          :format => :text)) %>
-<% else %>
-  <%= _('Your next payment will be taken on {{date}}',
-        date: simple_date(Time.zone.at(subscription.current_period_end),
-                          :format => :text)) %>
-
-  <br>
+</div>
 
+<div class="cancel-sub-section">
```
garethrees: This still displays the "cancel" link even if the subscription is cancelled. There was previously a conditional

screen shot 2017-10-12 at 14 52 52

app/views/alaveteli_pro/subscriptions/show.html.erb
```diff
@@ -1,15 +1,37 @@
...
+          <% end %>
+        </div>
+
+      </div>  <!-- .settings-right-column -->
+
+    </div>  <!-- .row -->
+
+  </div>  <!-- .inner-canvas-body -->
+
+  </div>  <!-- .inner-canvas -->
```
garethrees: This closing tag is indented 2 spaces too many

app/views/alaveteli_pro/subscriptions/_card.html.erb
```diff
@@ -1,30 +1,35 @@
...
 
-  <noscript>
-    <div id="error">
-      <p>
-        <%= _('Updating your payment card requires your browser to have ' \
-              'JavaScript enabled.') %>
-      </p>
-    </div>
-  </noscript>
 <% end %>
```
garethrees: Oh I've also just noticed that the closing tag of this form looks like its going to wrap the outer div, so that'll need fixing

garethrees: I think the form should end just after the `</script>`

wrightmartin: oh, cheers will fix

app/views/alaveteli_pro/subscriptions/_card.html.erb
wrightmartin: It could be better.

Is there a way to check for card expiry in the template without replicating the ```card_expiry_message``` function in the template? Eg can I just say ```if (card_expiry_message(...,...))```?

wrightmartin:
Pushed again with your feedback address @garethrees

app/views/alaveteli_pro/subscriptions/_card.html.erb
lizconlan: ah, that would be my fault, I should probably have done...

```diff
- def card_expiry_message(month, year)
-  if month == Date.today.month && year == Date.today.year
-    _('Expires soon')
-  end
- end

+ def card_expiry_message(card)
+   if card_expiring?(card)
+     _('Expires soon')
+   end
+ end

+ def card_expiring?(card)
+   if card.exp_month == Date.today.month && card.exp_year == Date.today.year
+     true
+   else
+     false
+   end
+ end
```

...then you could just do `card_expiring?(@card)` in the view (I think), and change the call to `card_expiry_message` to pass in `card` instead of the `card.exp_month, card.exp_year`

lizconlan: ...or ditch `card_expiry_message` of course, might not be simpler to just check for `card_expiring?` in the view

garethrees: Oops sorry, should have pasted our IRC chat here:

```irc
[16:15:33]  <gareth>	hey what do you mean by https://github.com/mysociety/alaveteli/pull/4282#discussion_r144319914?
[16:15:58]  <gareth>	I guess the question I'm not sure of is why you'd need the conditional?
[16:16:20]  <martin>	oh - so the function for the card expiring is just ``` def card_expiry_message(month, year)
[16:16:20]  <martin>	    if month == Date.today.month && year == Date.today.year
[16:16:20]  <martin>	      _('Expires soon')
[16:16:20]  <martin>	    end
[16:16:20]  <martin>	  end```
[16:16:45]  <martin>	I want to wrap that in something for presention, but I only want the wrapping code to appear when the message does
[16:16:52]  <gareth>	ahh
[16:16:54]  <gareth>	with you
[16:16:58]  <martin>	don't want to put HTML in there^
[16:17:10]  <gareth>	You can put HTML in a helper, thats fine
[16:17:16]  <gareth>	I can help with that
[16:17:25]  <martin>	oh ok, didn't realise
[16:17:25]  <gareth>	what's the HTML you want to end up with?
[16:18:01]  <martin>	```<p class="card__expiring">...</p>``` would do it
[16:18:52]  <gareth>	```content_tag(:p, _('Expires soon'), class: 'card__expiring')```
[16:18:53]  <gareth>	:)
```

The solution in 2457345 looks like its resolved the issue.

PR #4283, opened 2017-10-11
[Pro#383] add missing high res image (pro status icons)

wrightmartin:
Fixes https://github.com/mysociety/alaveteli-professional/issues/383


garethrees:
Looking at https://github.com/mysociety/alaveteli/tree/issues/383-need-hi-res-images/app/assets/images/alaveteli-pro, it looks like:

`arrow-pointing-down.png` and the `screenshot-*` images don't have an `@2` version.

I don't know if the screenshots should, but `arrow-pointing-down.png` almost certainly should?

PR #4284, opened 2017-10-11
[pro#208] Limit embargo extensions

lizconlan:
Closes mysociety/alaveteli-professional#208

Prevents requests under embargo from receiving an embargo extension unless the existing embargo is nearing expiry. Currently places admins under the same restriction which is the easiest thing to do but may prove problematic (the restriction applies at view and controller levels but there are no such restriction on the models so it can be circumvented on the console if required).

### DIFF
--- a/app/models/alaveteli_pro/embargo.rb
+++ b/app/models/alaveteli_pro/embargo.rb
@@ -73,6 +73,11 @@ module AlaveteliPro
       save
     end
 
+    def expiring_soon?
+      (Time.zone.now >= calculate_expiring_notification_at &&
+       Time.zone.now < publish_at)
+    end
+
     def calculate_expiring_notification_at
       self.publish_at - 1.week
     end
--- a/app/models/alaveteli_pro/embargo.rb
+++ b/app/models/alaveteli_pro/embargo.rb
@@ -78,6 +78,12 @@ module AlaveteliPro
        Time.zone.now < publish_at)
     end
 
+    # embargoes should be deleted once they've expired so this is for edge
+    # cases where the deletion task has not yet completed
+    def expired?
+      Time.zone.now >= publish_at
+    end
+
     def calculate_expiring_notification_at
       self.publish_at - 1.week
     end
--- a/app/models/info_request.rb
+++ b/app/models/info_request.rb
@@ -1568,6 +1568,13 @@ class InfoRequest < ActiveRecord::Base
     embargo ? embargo.expiring_soon? : false
   end
 
+  # Is the attached embargo still present but has reached its publication date
+  #
+  # Returns boolean
+  def embargo_pending_expiry?
+    embargo ? embargo.expired? : false
+  end
+
   # @see RequestSummaries#should_summarise?
   def should_summarise?
     self.info_request_batch_id.blank?
--- a/app/models/info_request.rb
+++ b/app/models/info_request.rb
@@ -1565,11 +1565,7 @@ class InfoRequest < ActiveRecord::Base
   #
   # Returns boolean
   def embargo_expiring?
-    if self.embargo
-      self.embargo.publish_at <= AlaveteliPro::Embargo.expiring_soon_time
-    else
-      false
-    end
+    embargo ? embargo.expiring_soon? : false
   end
 
   # @see RequestSummaries#should_summarise?
--- a/app/views/alaveteli_pro/info_requests/_embargo_form.html.erb
+++ b/app/views/alaveteli_pro/info_requests/_embargo_form.html.erb
@@ -11,27 +11,36 @@
   <label class="houdini-label" for="input1"><%= _("Change privacy") %></label>
   <input class="houdini-input" type="checkbox" id="input1">
   <div class="houdini-target extend-embargo-sidebar">
-    <%= form_for(
-          [AlaveteliPro::EmbargoExtension.new(embargo: info_request.embargo)],
-          html: {class: 'js-embargo-form'}) do |f| %>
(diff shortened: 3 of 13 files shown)


lizconlan:
Hmm, I have subsequently realised there is already an `embargo_expiring?` method on `InfoRequest` which I could have used. However I quite like having the extra check for the expiration date having passed a (presumably rare) edge case outside the test suite when the date has rolled over but the embargo has not yet been removed by the task runner.

app/models/alaveteli_pro/embargo.rb
```diff
@@ -71,6 +71,11 @@ def extend(extension)
       save
     end
 
+    def expiring_soon?
```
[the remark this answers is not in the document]
lizconlan: Hmm, despite coming to the conclusion that "is the publication date less than a week away?" (`InfoRequest#embargo_expiring?`) is essentially the same question as "is the current time between a week from publication and the publication deadline?" (`Embargo#expiring_soon?`), I've added an extra check to `embargo_expiring?` to handle possible edge cases where the intention is that the embargo has expired but for some reason the embargo record has not been removed - in the case of it being really early on the publication day, that could be a matter of timing, after that date presumably some kind of error

Will happily stand corrected if this does not sound right

app/views/alaveteli_pro/info_requests/_embargo_form.html.erb
```diff
@@ -11,27 +11,29 @@
...
   <input class="houdini-input" type="checkbox" id="input1">
   <div class="houdini-target extend-embargo-sidebar">
-    <%= form_for(
-          [AlaveteliPro::EmbargoExtension.new(embargo: info_request.embargo)],
-          html: {class: 'js-embargo-form'}) do |f| %>
-      <%= render partial: "alaveteli_pro/info_requests/embargo_info",
+    <%= render partial: "alaveteli_pro/info_requests/embargo_info",
                  locals: {embargo: info_request.embargo, tense: :present} %>
-      <%= f.hidden_field :embargo_id %>
+    <% if info_request.embargo.expiring_soon? %>
```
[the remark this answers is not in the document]
lizconlan: Adding a message that indicates when they will be able to update it as that feels less frustrating

app/views/alaveteli_pro/info_requests/_embargo_form.html.erb
```diff
@@ -34,6 +34,13 @@
                  class="embargo__submit js-embargo-submit">
         </p>
       <% end %>
+    <% else %>
+      <p>
+        <%= _("If you need to keep your request private for longer, you will " \
+              "be able to extend this privacy period on " \
+              "{{embargo_extend_from}}",
```
garethrees: Maybe make this a bit more concise?

```erb
<%= _("You will be able to extend this privacy period from {{embargo_extend_from}}.",
      embargo_extend_from: embargo_extend_from(info_request.embargo)) %>
```

app/models/info_request.rb
```diff
@@ -1577,15 +1577,12 @@ def self.log_very_overdue_events
...
+  # has already occurred but the embargo has not yet been deleted)
   #
   # Returns boolean
   def embargo_expiring?
-    if self.embargo
-      self.embargo.publish_at <= AlaveteliPro::Embargo.expiring_soon_time
-    else
-      false
-    end
+    !!embargo && (embargo.expiring_soon? || Time.zone.now >= embargo.publish_at)
```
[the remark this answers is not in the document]
lizconlan: no, wait - the opposite of that, we want `embargo_ending_expiry?`

app/models/info_request.rb
```diff
@@ -1584,6 +1584,13 @@ def embargo_expiring?
     embargo ? embargo.expiring_soon? : false
   end
 
+  # Is the attached embargo still present but has reached its publication date
+  #
+  # Returns boolean
+  def embargo_pending_expiry?
```
garethrees: Its not the end of the world, but not sure what the differing method name adds here. `embargo_expired?` is a bit clearer in that its more obviously delegating to `embargo.expired?`.

spec/models/info_request_spec.rb
```diff
@@ -3515,6 +3515,48 @@ def email_and_raw_email(opts = {})
...
+          expect(info_request.reload.embargo_pending_expiry?).to be true
+        end
+      end
+
+    end
+
+    context 'when there is no embargo' do
+
+      it 'returns true' do
+        expect(info_request.embargo_pending_expiry?).to be false
```
garethrees: The spec description and assertion differ here.

lizconlan: /o\ (thank you)

app/models/info_request.rb
lizconlan: there's a different (and more accurate) `embargo_expired?` [over in pro#208](https://github.com/mysociety/alaveteli/pull/4296/files#diff-5f7dfe1185cac97cbc9f9dc356abce7cR1583) which is intended to only return true if the request has been published (and there was once an embargo), happy to come up with a better name here though - this describes an edge case which hopefully will only be true between midnight and the scheduled script run which deletes pending embargoes and causes the requests to be reindexed

garethrees: Ah right yeah, ISWYM.

I think this is fine in that case, and we can look at the `embargo_*` method names together when all merged to check each makes sense in context.

lizconlan: is it worth me combining the 3 overlapping PRs into one once I've done the fixups so that we can review/check pre-merge?

PR #4287, opened 2017-10-12
[pro#132] Allow pros to add an embargo to an existing request

lizconlan:
Fixes mysociety/alaveteli-professional#132

ToDo: WDTK overrides the sidebar template so will need tweaking to include this addition

### DIFF
--- a/app/models/ability.rb
+++ b/app/models/ability.rb
@@ -72,6 +72,13 @@ class Ability
         can :access, :alaveteli_pro
       end
 
+      # Creating embargoes
+      can :create_embargo, InfoRequest do |info_request|
+        user && info_request.user.is_pro? && !info_request.embargo &&
+          (user == info_request.user || user.is_pro_admin?) &&
+          info_request.info_request_batch_id.nil?
+      end
+
       # Extending embargoes
       can :update, AlaveteliPro::Embargo do |embargo|
         user && (user == embargo.info_request.user || user.is_pro_admin?)
--- a/app/views/alaveteli_pro/info_requests/_embargo_form.html.erb
+++ b/app/views/alaveteli_pro/info_requests/_embargo_form.html.erb
@@ -11,29 +11,8 @@
   <label class="houdini-label" for="input1"><%= _("Change privacy") %></label>
   <input class="houdini-input" type="checkbox" id="input1">
   <div class="houdini-target extend-embargo-sidebar">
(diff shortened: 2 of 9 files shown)


lizconlan:
May require template overrides in the theme

app/views/alaveteli_pro/info_requests/_embargo_form.html.erb
```diff
@@ -1,45 +1,32 @@
...
+      <%= render partial: "alaveteli_pro/info_requests/embargo_extension_form",
+                 locals: {info_request: info_request} %>
+      <%= button_to _("Publish request"),
+                    alaveteli_pro_embargo_path(info_request.embargo),
+                    method: :delete,
+                    data: {
+                      confirm: _("Are you sure you want to publish " \
+                                 "this request?")
+                    } %>
+    <% else %>
```
gbp: This `else` conditional will render the create form on batched requests - but when the form is submitted it raises `PermissionDenied` exception [here](https://github.com/mysociety/alaveteli/pull/4287/files#diff-6b4fa2fa47d11a529cdf2a0b0ff63004R15).

While we could prevent render here using an `elsif` and checking to see if `info_request.info_request_batch_id.nil?` I think a nicer solution would be to move this check into the `:create_embargo` ability

app/models/ability.rb
```diff
@@ -72,6 +72,12 @@ def initialize(user)
         can :access, :alaveteli_pro
       end
 
+      # Creating embargoes
+      can :create_embargo, InfoRequest do |info_request|
+        user && info_request.user.is_pro? && !info_request.embargo &&
+          (user == info_request.user || user.is_pro_admin?)
```
gbp: Could move checking for whether or not `info_request_batch_id.nil?` here. Would remove need to limit the create form remove being outputted and remove the check in the controller action too.

app/views/alaveteli_pro/info_requests/_embargo_form.html.erb
lizconlan: good catch, cheers!

app/models/ability.rb
lizconlan: have done - I _think_ we can do that without breaking a future requirement to update batches as that will be for `can :create_embargo, InfoRequestBatch`. I think.

PR #4218, opened 2017-09-12
Add stretch support to installer and vagrant scripts

crowbot:
Closes #4216

PR #4290, opened 2017-10-13
Update packages for internal move to stretch

garethrees:
This is required for our move to Debian Stretch internally.

It shouldn't affect re-user installs.

We can't merge this to develop until after migration as that would make WDTK undeployable on its existing host.

I think we'll want to also make these changes in https://github.com/mysociety/alaveteli/pull/4218.

### DIFF
--- a/config/packages
+++ b/config/packages
@@ -17,7 +17,7 @@ libmagickwand-dev
 libpq-dev
 libsqlite3-dev
 libxml2-dev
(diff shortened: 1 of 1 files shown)


garethrees:
Merged in 0ef96f6c215bcd999e34089dec5f59cd4803f80d and 56a6ee4f2bb2646d2e8796791674b009742d6c78

PR #4292, opened 2017-10-13
[#4288] Don't run service command as app user

### LINKED ISSUE #4288: compact-xapian-database fails to restart alaveteli under systemd
The `compact-xapian-database` script includes the following line:
```
commonlib/bin/output-on-error /usr/sbin/service "$DAEMON_NAME" restart
```
This fails on Debian Stretch (and probably any other systemd-based OS) when run as an app user as the `service` command just passes this to `systemctl` which needs root privileges.

For the purposes of this script, either running `/etc/init.d/${DAEMON_NAME} restart` directly (as the script does nothing that the app user can't do) or even simply touching the `alaveteli/tmp/restart.txt` file would probably be enough to resolve this.

Happy to submit a patch and PR, but thought we might want to discuss the preferred option and any other potential places this might bite.

garethrees:
Fixes #4288

This fails on Debian Stretch (and probably any other systemd-based OS)
when run as an app user as the service command just passes this to
systemctl which needs root privileges.

For now, just call the script directly as its the easiest thing to do to
fix the problem, and we do similar in the Capistrano deployment config.


garethrees:
Merged in 663bb16df

PR #4296, opened 2017-10-17
[pro#141] Send alert mail when an embargo expires

lizconlan:
Notes: assumes the `delete-expired-embargoes` script runs before the daily notifications run

It looks as though instant notifications will currently create and send an email for every expiring and expired request in a batch and the embargo mailer (where used) will send an email reporting the number of expired requests with a very long list of links (same for expiring) - see mysociety/alaveteli-professional#394

Closes mysociety/alaveteli-professional#141


app/models/notification.rb
```diff
@@ -76,6 +76,12 @@ def embargo_expiring_expired
     !self.info_request_event.info_request.embargo_expiring?
   end
 
+   def expire_embargo_expired
+    # If someone has changed the embargo date on the request, or published it,
+    # they might not need this notification any more.
+    !self.info_request_event.info_request.embargo_expired?
```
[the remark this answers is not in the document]
lizconlan: fixed, removing comment

app/models/info_request/prominence/embargo_expired_today_query.rb
```diff
@@ -0,0 +1,23 @@
...
+      def call
+        @relation.
+          includes(:embargo).
+          joins(:info_request_events).
+          where('embargoes.info_request_id IS NULL').
+          where([
+            'info_request_events.created_at >= ?',
+            Time.zone.now.beginning_of_day
+          ]).
+          where('info_request_events.event_type' => 'expire_embargo').
```
lizconlan: can't be `where('info_request_events.event_type': 'expire_embargo').` below Ruby 2.3, see: https://travis-ci.org/mysociety/alaveteli/builds/289446372

garethrees: I think you can do `where(info_request_events: { event_type: 'expire_embargo' })`

lizconlan: ah, yes - that looks better, cheers!

lizconlan:
ToDo: add missing `script/alert-embargoes-expired` (and crontab)

spec/models/info_request/prominence/embargo_expired_today_query_spec.rb
```diff
@@ -0,0 +1,40 @@
+# -*- encoding : utf-8 -*-
+require 'spec_helper.rb'
+
+describe InfoRequest::Prominence::EmbargoExpiredTodayQuery do
+
+  describe '#call' do
+    let(:info_request) { info_request = FactoryGirl.create(:info_request) }
+
+    it 'excludes requests that have an embargo' do
```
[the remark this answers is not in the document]
lizconlan: oh, yes - good point

lizconlan:
> Do need to make sure that this works with mysociety/alaveteli-professional#132, but it might be easier to ticket that as a bug after merging both if there is a problem.

Temporarily merged pro#132 into my local branch and successfully re-ran the test suite so I think it's ok

lizconlan:
- [ ] Check in browser (with pro#132)
- [ ] Consider checking that the event returned within `Embargo#notify_expiration` is for the current embargo (although the id is not stored so.... compare timestamps?)

The point where it might be wrong is if there was a previous embargo which expired followed by one which was manually removed by the user then `info_request.last_embargo_expire_event` may return a result which predates the current embargo (presumably a user would not want to be notified that an embargo "expired" if they removed it themselves)

lizconlan:
Works as expected and creating a new embargo directly after deleting the old one removes the request from the scope of `EmbargoMailer#alert_expired`'s query and marks the created Notification as expired if the user (and the request) are signed up for notifications.

app/models/notification.rb
```diff
@@ -76,6 +76,12 @@ def embargo_expiring_expired
     !self.info_request_event.info_request.embargo_expiring?
   end
 
+  def expire_embargo_expired
+    # If someone has added a new embargo, they might not need this notification
+    # any more.
+    !info_request_event.info_request.embargo_expired?
```
lizconlan: this needs to change once mysociety/alaveteli-professional#208 is merged

lizconlan:
- [ ] Revisit once pro#208 is ready (changes Notifications)

garethrees:
Looks good. I think it would be worth adding a ticket to describe the overlap of this and https://github.com/mysociety/alaveteli-professional/issues/208, and what we need to check/review once they're both merged.

PR #4307, opened 2017-10-20
Keep the lib/themes directory in version control

garethrees:
You need this directory in place for the theme switcher to work. If you
install manually, it doesn't get created so you get errors when setting
up theme switching.

The install script installs the default alavetelitheme, so the directory
gets created for you, which is probably why this has gone unnoticed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,7 +16,8 @@
 .vagrant.yml
 *#*#
 TAGS
-/lib/themes
+/lib/themes/*
+!/lib/themes/.keep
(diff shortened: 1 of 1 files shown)


garethrees:
@gbp is this something you needed to do when setting up alaveteli for development?

gbp:
@garethrees yes - setup throw an error because the directory didn't exist

garethrees:
Cool – happy for me to merge this?

gbp:
Yeah good to merge

PR #4321, opened 2017-10-26
[pro#396] Use Xapian results to set page count

garethrees:
`info_requests.size` exposes a rough idea of the amount of private
requests a pro user has made.

Since private requests are not indexed by Xapian, the
`matches_estimated` will provide the correct pagination.

Fixes https://github.com/mysociety/alaveteli-professional/issues/396.

---

Didn't add specs because we don't have any existing view specs, and this area is due an overhaul soon.

To test in the browser, you basically want to create a pro user with 1 public request, and many private requests:

```ruby
require 'rspec/rails'
require 'factory_girl'
require Rails.root.join('spec', 'support', 'load_file_fixtures')
require Rails.root.join('spec', 'support', 'email_helpers')

RSpec.configure do |config|
  config.fixture_path = Rails.root.join('spec', 'fixtures')
end

user = FactoryGirl.create(:pro_user, email: 'some-unique-email@localhost')
bodies = PublicBody.all

FactoryGirl.create(:successful_request, user: user, public_body: bodies.sample)

50.times do
  FactoryGirl.create(:embargoed_request, user: user, public_body: bodies.sample)
end

# Run script/update-xapian-index afterwards, so that the public request gets found

# Then navigate to the user's show page (not signed in).
# /user/:url_name
puts user.url_name
# => …
```

### DIFF
--- a/app/views/user/show/_show_requests.html.erb
+++ b/app/views/user/show/_show_requests.html.erb
@@ -51,13 +51,15 @@
     <% else %>
       <h2 class="foi_results foi_requests" id="foi_requests">
         <% if @is_you %>
-          <%= n_('Your {{count}} Freedom of Information request',
-                 'Your {{count}} Freedom of Information requests',
+          <%= n_('Your Freedom of Information request',
+                 'Your Freedom of Information requests ' \
+                 '(approximately {{count}})',
                  @xapian_requests.matches_estimated,
                  :count => @xapian_requests.matches_estimated) %>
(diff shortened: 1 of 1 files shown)


garethrees:
> This looks a bit funny when there's only a small number of requests

Yeah I wasn't totally happy with the wording. Really we should only show that if there's more than one page, but didn't want to make things too complicated and then need to update the other lists to match.

PR #4323, opened 2017-10-26
[pro#205] Don't stop followups if a request has never been published

gbp:
Fixes https://github.com/mysociety/alaveteli-professional/issues/205

### DIFF
--- a/spec/models/info_request/prominence/been_published_query_spec.rb
+++ b/spec/models/info_request/prominence/been_published_query_spec.rb
@@ -1,0 +1,19 @@
+require 'spec_helper'
+
+describe InfoRequest::Prominence::BeenPublishedQuery do
+
+  describe '#call' do
+    subject { described_class.new.call }
+
+    it 'limits the requests to those that do not have embargoes or whose embargoes have been expired in the past' do
+      info_request = FactoryGirl.create(:info_request)
+      embargoed_request = FactoryGirl.create(:embargoed_request)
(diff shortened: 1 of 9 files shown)


spec/models/info_request/prominence/been_published_query_spec.rb
```diff
@@ -0,0 +1,20 @@
+# -*- encoding : utf-8 -*-
+require File.expand_path(File.dirname(__FILE__) + '/../../../spec_helper')
+
+describe InfoRequest::Prominence::BeenPublishedQuery do
+
+  describe '#call' do
+    subject { described_class.new.call }
+
+    it 'limits the requests to those that do not have embargoes or embargoes has been expired in the pass' do
```
crowbot: The phrasing here is a little unclear - we could amend it to: 
```diff
- it 'limits the requests to those that do not have embargoes or embargoes has been expired in the pass' do
+ it 'limits the requests to those that do not have embargoes or whose embargoes have been expired in the past' do
```

spec/models/info_request/prominence/been_published_query_spec.rb
```diff
@@ -0,0 +1,20 @@
+# -*- encoding : utf-8 -*-
+require File.expand_path(File.dirname(__FILE__) + '/../../../spec_helper')
```
crowbot: I think we can just use the simpler `require spec_helper.rb` here as in `spec/models/info_request/prominence/embargo_expiring_query_spec.rb`

PR #4324, opened 2017-10-26
Issues/alaveteli pro 346/add ts and cs

crowbot:
Work on https://github.com/mysociety/alaveteli-professional/issues/346
Complement to https://github.com/mysociety/whatdotheyknow-theme/pull/422


config/routes.rb
```diff
@@ -608,6 +608,7 @@
   constraints FeatureConstraint.new(:pro_pricing) do
     namespace :alaveteli_pro, path: 'pro', as: 'pro' do
       resources :plans, only: [:index], path: 'pricing'
+      resources :pages, only: [:show], path: 'pages'
```
garethrees: Don't need to specify the path here as it's "pages" anyway.

PR #4338, opened 2017-11-03
[#pro288] Pass-through signin form if already logged in

garethrees:
We shouldn't render the signin form if a user is already logged in. Its
confusing and results in weird behaviour.

For example, if a user is signed in, receives a "You have a new response
" email and follows the URL within the email, they shouldn't need to
sign in again before being redirected to their new response, as the
session already exists.

Fixes https://github.com/mysociety/alaveteli-professional/issues/288
(and also fixes this for non-pro users).

### DIFF
--- a/spec/integration/sign_in_with_redirect_spec.rb
+++ b/spec/integration/sign_in_with_redirect_spec.rb
@@ -1,0 +1,93 @@
+# -*- encoding : utf-8 -*-
+require File.expand_path(File.dirname(__FILE__) + '/../spec_helper')
+require File.expand_path(File.dirname(__FILE__) + '/alaveteli_dsl')
+
+describe 'Signing in with a redirect parameter' do
+
+  before { set_consider_all_requests_local(true) }
+  after { restore_consider_all_requests_local }
+
+  context 'when not logged in' do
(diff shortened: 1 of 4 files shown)


garethrees:
I've left the squash/fixup commits in so that whoever reviews can see what I _intended_ to do, and how I had to work around https://github.com/mysociety/alaveteli/issues/4353.

spec/integration/sign_in_with_redirect_spec.rb
```diff
@@ -4,6 +4,9 @@
 
 describe 'Signing in with a redirect parameter' do
```
garethrees: Maybe the commit message here is misleading – perhaps the previous pass on 2.3 was just lucky that the spec ordering didn't cause a problem? I'll note that when I squash.

spec/integration/sign_in_with_redirect_spec.rb
```diff
@@ -24,7 +24,7 @@
       expect(response.status).to eq(404)
     end
 
-    it 'does not redirect to external URLs' do
+    pending 'does not redirect to external URLs' do
```
lizconlan: FWIW I think this is what I was looking at when I saw pending specs fail < 2.3

PR #4340, opened 2017-11-06
[Pro#382] Support trials on profile subscription page

gbp:
Fixes https://github.com/mysociety/alaveteli-professional/issues/382

Page looks almost the same when a coupon is used, only difference is with the next payment date - this is because stripe won't create new subscriptions each month for trials.

Trial:
trial

Coupon:
coupon


app/models/alaveteli_pro/subscription_with_discount.rb
```diff
@@ -50,7 +42,27 @@ def free?
 
   private
 
+  def coupon?
+    coupon
```
lizconlan: I'd expect this to return true or false rather than the object - I think that would be what we'd test for if this were a public method

PR #4348, opened 2017-11-09
Use pry-byebug instead of pry-debugger

lizconlan:
Now that we've dropped Ruby 1.9 support

(Initial PR to see if it works ok, needs a bit more thought about dot files/key bindings)


garethrees:
> dot files/key bindings

IMO these should not be committed to the project (e.g. https://github.com/mysociety/alaveteli/blob/develop/.pryrc) – rather they should be set in `~/.pryrc`.

lizconlan:
yeah, there's a committed `.pryrc` with PryDebugger config in it - the right thing to do is probably deleting it

Gemfile
```diff
@@ -174,7 +174,7 @@ group :test, :development do
   gem 'rspec-activemodel-mocks', '~> 1.0.0'
   gem 'rspec-rails', '~> 3.5.0'
   gem 'pry', '~> 0.10.0'
-  gem 'pry-debugger', '~> 0.2.0', :platforms => :ruby_19
+  gem 'pry-byebug', '~> 3.4.0'
```
garethrees: > Locked to 3.4.x to retain support for Ruby 2.0 and 2.1 (removed from
> byebug in version 9.1.0)

9.1.0 a typo?

lizconlan: no, I think that's correct, just confusingly stated because of the similar names:
https://github.com/mysociety/alaveteli/pull/4348/files#diff-e79a60dc6b85309ae70a6ea8261eaf95R256

\*tries to think of better phrasing\*

lizconlan: "(removed from the underlying byebug dependency in version 9.1.0)"? (or similar)

lizconlan: and/or a link to their ticket for that github.com/deivid-rodriguez/pry-byebug/issues/123

lizconlan:
ToDo - fix the commit message and merge

Issue #4342, opened 2017-11-06
Front-end performance investigation


garethrees:
Spend some time investigating and fixing.

Also related https://github.com/mysociety/alaveteli/issues/3845.

wrightmartin:
### Results from PageSpeed Insights
 (https://developers.google.com/speed/pagespeed/insights/?url=https%3A%2F%2Fwww.whatdotheyknow.com%2F&tab=mobile)

- Eliminate render-blocking JavaScript and CSS in above-the-fold content (no filesize saving but big improvements to first paint time)
- Enable compression (16.5KiB (66% reduction))
- Minify HTML (3.4KiB (14% reduction))

zarino:
@garethrees How come the whatdotheyknow.com server isn’t gzipping its HTML responses?

wrightmartin:
### Results from ySlow

Grade | B | Overall performance score 85 | Ruleset applied: YSlow(V2) | URL: https://www.whatdotheyknow.com/

- Grade C on Make fewer HTTP requests 
This page has 4 external Javascript scripts. Try combining them into one.
This page has 3 external stylesheets. Try combining them into one.
This page has 13 external background images. Try combining them with CSS sprites.
- Grade F on Use cookie-free domains

zarino:
Doesn’t really affect performance, but I’ve noticed there’s a gap at the bottom of all the WDTK pages. Leads me to suspect there’s an unclosed HTML element somewhere? Possibly in the header/banner?

![screen shot 2017-11-07 at 10 02 27](https://user-images.githubusercontent.com/739624/32487894-dab7643c-c3a2-11e7-9a81-7c4cea81fb4c.png)

I think it’s happening on all the sites that use a theme forked from [alavetelitheme](https://github.com/mysociety/alavetelitheme): https://www.whatdotheyknow.com, https://www.queremosdatos.co, https://www.asktheeu.org…

wrightmartin:
### General thoughts and observations
- 266.6K of JS
- 241.2K of CSS

- We serve pro styles to non pro users, should we only send these to pro users? Does the filesize of the CSS decrease enough to justify another request (sometimes non-pro users will see the pro styles, such as in the marketing pages)

- As above, but with JS

garethrees:
Note its also worth checking a request page rather than the homepage, as they account for something like 80% of page views.

wrightmartin:
### Spriting

[...]

PR #4351, opened 2017-11-09
[#4342] Replace many single icons with sprites

wrightmartin:
Work on #4342


gbp:
Merged as part of #4355

PR #4359, opened 2017-11-13
[pro#389] Prevent default public authority search action

gbp:
The default action is not needed as selectize will load results via AJAX. In fact sometimes the action event is triggered without selectize capturing the text inputted so this fixes https://github.com/mysociety/alaveteli-professional/issues/389


lizconlan:
> I do think we want something sensible to happen

Agreed, as this isn't a quick fix and the behaviour wasn't introduced here I've made a separate ticket for it with an example gif https://github.com/mysociety/alaveteli-professional/issues/410

PR #4368, opened 2017-11-16
[pro#363-1] Routing and controller improvements

gbp:
Fixes the following parts of https://github.com/mysociety/alaveteli-professional/issues/363

> - [x] SubscriptionController#show should be SubscriptionController#index I think – show should have an id param, which we don't set.
> - [x] In general the routes are starting to get messy – lots of match. We should circle back and clean these up once we have the full flow in place.

Have gone through the pro routes and remove almost all `match` routes instead we now using resourceful routes with standard CRUD actions.

This does means some URLs have changes, the main one as I see it is the entry to the batch request authority search which might have been given out to some beta testers? So we might need to redirect from the old `new` action to the `index`?


gbp:
@garethrees have redone old commit aeaeb7b to include redirect for existing URL (new commit b20786c). Have also expanded my commit comments.

PR #4242, opened 2017-09-20
[WIP][pro#339] Pro Signup

lizconlan:
Connects to mysociety/alaveteli-professional#338
Connects to mysociety/alaveteli-professional#339
Connects to mysociety/alaveteli-professional#340

Closes mysociety/alaveteli-professional#343 (webhook for failed payment to be implemented later)

pro-signup-demo

[previous demo](https://user-images.githubusercontent.com/27760/30656065-b77fa90e-9e2a-11e7-8bde-7f95eb43c946.gif)

PR #4369, opened 2017-11-16
[pro#363-2] Currency, Stripe and Pro signup improvements

gbp:
Fixes the following parts of https://github.com/mysociety/alaveteli-professional/issues/363

> - [x] Better handling of number_to_currency unit symbol
> - [x] Update card form does not ask for postcode

And these related parts have already been improved on `develop`

> - [x] Make ProAccount#stripe_customer handle fetching the Stripe::Customer object.
> - [x] Disable the "Pay with Card" / "Update card" buttons after the first click so you can't bring up the Stripe panel a second time before the page redirect completes (see the example video in mysociety/alaveteli#4242)
> - [x] maybe some code to format customer input to match

There is now a new `ISO_CURRENCY_CODE` configuration which is used to display the currency in the correct format but not linked to the Stripe plan yet.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -140,6 +140,7 @@ gem 'locale', '~> 2.0.0', '< 2.1.0'
 gem 'routing-filter', '~> 0.5.0', '< 0.6.0'
 gem 'unicode', '~> 0.4.0'
 gem 'unidecoder', '~> 1.1.0'
+gem 'money', '~> 6.10.0'
 
 # mime-types 3.0.0 requires Ruby 2.0.0, and _something_ is trying to update it
 gem 'mime-types', '< 3.0.0', require: false
--- a/config/general.yml-example
+++ b/config/general.yml-example
@@ -59,6 +59,23 @@ FORCE_SSL: true
 # ---
 ISO_COUNTRY_CODE: GB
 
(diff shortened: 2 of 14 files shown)


config/general.yml-example
```diff
@@ -59,6 +59,23 @@ FORCE_SSL: true
 # ---
 ISO_COUNTRY_CODE: GB
 
+# ISO currency code of the currency Alaveteli Professional cost should be
+# displayed in.
+# (https://en.wikipedia.org/wiki/ISO_4217#Active_codes)
+#
+# Doesn't effect what currency the plans are setup in so this should match what
```
lizconlan: nitpick - affect rather than effect (I get this wrong all the time)

Gemfile
```diff
@@ -140,6 +140,7 @@ gem 'locale', '~> 2.0.0', '< 2.1.0'
 gem 'routing-filter', '~> 0.5.0', '< 0.6.0'
 gem 'unicode', '~> 0.4.0'
 gem 'unidecoder', '~> 1.1.0'
+gem 'money', '~> 6.10.0'
```
lizconlan: nice but worth noting what it's for in the commit comment - it's obvious now but nice to have a note of the decision for far future people (or me in 3 weeks' time!)

PR #4373, opened 2017-11-16
Handle anchor in redirects

garethrees:


lib/safe_redirect.rb
```diff
@@ -0,0 +1,17 @@
...
+class SafeRedirect
+  attr_reader :redirect_parameter
+
+  def initialize(redirect_parameter)
+    @redirect_parameter = redirect_parameter
+  end
+
+  def path
+    uri = URI.parse(redirect_parameter)
+    URI::Generic.build(path: uri.path, fragment: uri.fragment).to_s
```
gbp: `redirect_to` can take a URI object to calling `#to_s` is only needed here because we are appending the `emergency` query param in one place. Could think about passing in query params as part of the `initialize` method to be used when building the URI.

gbp:
If understand this change correctly (from reading the specs) then can be tested be being logged in and going to `http://alaveteli.test/profile/sign_in?r=http://alaveteli.test/select_authority?evil=1#good` which should redirect `http://alaveteli.test/select_authority#good`

garethrees:
> then can be tested be being logged in 

It should work either already logged in, or when you sign in and get redirected afterwards.

> `http://alaveteli.test/profile/sign_in?r=http://alaveteli.test/select_authority?evil=1#good` which should redirect `http://alaveteli.test/select_authority#good`

Yep!

garethrees:
> redirect_to can take a URI object to calling #to_s is only needed here because we are appending the emergency query param in one place.

I return a `URI::Generic`, which `redirect_to` wasn't too happy about, so I couldn't get this working nicely.

> Could think about passing in query params as part of the initialize method to be used when building the URI.

Passed it in to `path` instead, as I think its nicer to hold as little state as possible in the object.

gbp:
Looks good to merge

PR #4374, opened 2017-11-16
[pro#124] Add virtual pageview for pro signup funnel

lizconlan:
Connects to mysociety/alaveteli-professional#124

The Goal and funnel steps for WDTK have been created but left switched off for now:

screen shot 2017-11-16 at 16 47 14


lizconlan:
>> The Goal and funnel steps for WDTK have been created but left switched off for now:
>
> This can be turned on after merge?

Done!

Will merge once Travis passes

PR #4382, opened 2017-11-21
[pro#413] expand new signup welcome message

lizconlan:
screen shot 2017-11-21 at 16 06 36

Closes mysociety/alaveteli-professional#413

### DIFF
--- a/app/views/alaveteli_pro/subscriptions/_signup_message.html.erb
+++ b/app/views/alaveteli_pro/subscriptions/_signup_message.html.erb
@@ -1,3 +1,6 @@
-<%= _('Welcome to {{pro_site_name}}!',
-      pro_site_name: AlaveteliConfiguration.pro_site_name) %>
-
+<p><%= _('Welcome to {{pro_site_name}}!',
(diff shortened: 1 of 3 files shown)


lizconlan:
ToDo - break the message into 2 strings so as not to have HTML in translation strings.

garethrees:
Oops hang on – just tried with WDTK theme and got:

screen shot 2017-11-22 at 17 08 37

app/views/alaveteli_pro/subscriptions/_signup_message.html.erb
```diff
@@ -0,0 +1,7 @@
+<p><%= _('Welcome to {{pro_site_name}}!',
+         pro_site_name: AlaveteliConfiguration.pro_site_name) %>
+%></p>
```
garethrees: Double closing erb tag here

lizconlan: ugh, copy and paste fail - thank you!

Issue #1324, opened 2014-01-13
Warn users some links in emails automatically log them in so should not be shared


RichardTaylor:
A user on WhatDoTheyKnow appears to have shared one of the links given in some emails the system sends to users and been surprised when the person who was sent the link has been logged into their account when they used the link. 

Certain links in email messages from Alaveteli site do automatically login those who use them. 

These are the links of the form:

https://www.whatdotheyknow.com/c/[long alphanumeric code]

That these links should not be shared with others is perhaps something we should make more clear to our users. 

This is the first time this has been raised as a problem; so this could be a one off report of a problem which is not widespread.

kingqueen3065:
This just happened to me; I used the batch request feature for the first time and noticed that the link sent to me automatically logged me in. I assumed it was a bug. I agree that where such links are sent out, the software should make sure that the user is entirely clear that the link automatically logs in anybody who clicks on it. Otherwise it's a big security hole.

garethrees:
This came up in AlaveteliCon 2015. The consensus seemed to be that the existing behaviour is pretty dangerous.

crowbot:
Comments from @FOIMonkey: 



We've just had an instance of a user accidentally making an FOI request whilst logged in as, and so in the name of, another user. This was likely caused by a user forwarding an email received about their request to another person, who was then logged into the first users account automatically without the second person realising by clicking on the link in the email. We have had problems in the past with users copying these links out of emails and pasting them into social media posts, which hands control of their account to the world at large.

Questions/suggestions

    Are auto login links helpful and necessary?
    Should these work only one time/be strictly time limited?
    Is there a potential for this to happen to an admin account?

kingqueen3065:
I can't work out if this thread is open or not. But anyway, 

I think users will not generally expect to be automatically logged in from a link in an email; this is rare in other websites, I've not come across it anywhere else. I think the benefit of this is questionable, as generally people will be logged in due to a retained cookie or will expect to have to log in, which isn't a big hassle. I think the risk of having such links is substantial, as has been demonstrated by the small number of cases in which a user has accidentally made FOI requests using somebody else's account. I think that it would be best to remove this functionality, i.e. to just supply a link to the request, and that in the meantime the templates for emails should be amended to make it totally clear that the link automatically logs the clicker in.

kingqueen3065:
We have just had reported to us another example of this security problem.

[...]

PR #4383, opened 2017-11-22
[#Pro325] Reduce usage of auto login links

garethrees:
Auto-login URLs are a security problem, as users sometimes forward
emails or tweet the confirmation URLs, so others can then log in as the
user.

We opted to remove these for Pro users given the sensitivity of
requests, and given the confusion it creates for regular users (and
additional maintenance for developers), this removes them entirely.

In some cases, we don't automatically get a sign-in form if we try to
complete the action, so instead we pass through the sign-in form. If the
user is already signed in, they'll get redirected without noticing. If
they need to sign in, they'll get redirect after that.

Some actions still generate auto-login links:

* Confirming your account
* Changing your password
* Changing your email address
* Authority upload URLs

At this point in time, I'm not sure what to do with these, but there's
much less likelihood that users will forward these emails, so we can
deal with these later.

`NotificationMailer#daily_summary` makes use of the hacky
`batch_request_id_mappings` as we can't specify the IDs of the requests
belonging to a batch.

* Remove auto-login from RequestMailer#overdue_alert
* Remove auto-login from RequestMailer#very_overdue_alert
* Remove auto-login from RequestMailer#new_response_reminder_alert
* Remove auto-login from RequestMailer#not_clarified_alert
* Remove auto-login from InfoRequestBatchMailer#batch_sent
* Remove auto-login from NotificationMailer#overdue_notification
* Remove auto-login from NotificationMailer#very_overdue_notification
* Remove auto-login from TrackMailer#event_digest
* Remove auto-login from NotificationMailer#daily_summary

Fixes https://github.com/mysociety/alaveteli-professional/issues/325
Fixes https://github.com/mysociety/alaveteli/issues/1324

---

Also fixes a spec copy-paste error.

---

Stupid GitHub commit ordering… should read (newest at top):

```
e35154702 Reduce usage of auto-login links from mailers
58152cfe6 Set IDs for factory data
56d1a1ac9 Fix spec copy-paste error
```

---

TODO:

- [x] Add changelog note
- [x] Check what the remaining `PostRedirect` calls are doing
- [x] Mention that they haven't been removed for account confirmation, or email & password changes. Account confirmation is probably the one valid case for auto-login. Others can be dealt with separately.
- [x] ~~Make account confirmation a 1-time token~~ Ticketed this in https://github.com/mysociety/alaveteli/issues/4386 instead.

### DIFF
--- a/app/mailers/info_request_batch_mailer.rb
+++ b/app/mailers/info_request_batch_mailer.rb
@@ -9,13 +9,7 @@ class InfoRequestBatchMailer < ApplicationMailer
 
   def batch_sent(info_request_batch, unrequestable, user)
     @info_request_batch, @unrequestable = info_request_batch, unrequestable
-
-    # Make a link going to the info request batch page, which logs the user in.
-    post_redirect = PostRedirect.new(
-      :uri => info_request_batch_url(@info_request_batch),
-      :user_id => info_request_batch.user_id)
-    post_redirect.save!
-    @url = confirm_url(:email_token => post_redirect.email_token)
+    @url = info_request_batch_url(@info_request_batch)
 
     set_reply_to_headers(user)
 
--- a/app/mailers/notification_mailer.rb
+++ b/app/mailers/notification_mailer.rb
@@ -149,12 +149,8 @@ class NotificationMailer < ApplicationMailer
 
   def overdue_notification(notification)
     @info_request = notification.info_request_event.info_request
-
-    post_redirect = PostRedirect.new(
-      :uri => respond_to_last_url(@info_request, :anchor => "followup"),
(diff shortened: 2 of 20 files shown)


garethrees:
> BUT if it's an older request or you're logged in as admin, you could end up classifying the response from the "wrong" account.

Funnily enough, I did exactly this the other day with my own FOI request. Clicked the link in my personal email to classify – was already logged in as my admin account – then classified the request and got a "someone else has classified your request email". So, I don't think this is a huge problem given you can already do this in some cases.

> Although this is definitely better than accidental login sharing.

Yeah, in any case, I'm happy with the tradeoff.

PR #4384, opened 2017-11-22
[pro#417] avoid duplicate payments

lizconlan:
Closes mysociety/alaveteli-professional#417


lizconlan:
> Unable to reproduce the issue locally

Yeah, might be something about my local (Safari) setup not expiring forms that it's treating as local or something

PR #4387, opened 2017-11-23
[#pro418] Hide /plans/show FAQs

garethrees:
Fixes https://github.com/mysociety/alaveteli-professional/issues/418

Control them with a feature flag until we have some content.


garethrees:
Yep, very much a development aid rather than a flag we'd use in production :)

PR #4405, opened 2017-11-30
[pro#425] Implement basic webhook listener

lizconlan:
Closes mysociety/alaveteli-professional#425

### DIFF
--- a/app/controllers/alaveteli_pro/stripe_webhooks_controller.rb
+++ b/app/controllers/alaveteli_pro/stripe_webhooks_controller.rb
@@ -1,0 +1,11 @@
+# -*- encoding : utf-8 -*-
+# Does not inherit from AlaveteliPro::BaseController because it doesn't need to
+class AlaveteliPro::StripeWebhooksController < ApplicationController
+
+  def receive
+    # send a 200 ok to acknowlege receipt of the webhook
+    # https://stripe.com/docs/webhooks#responding-to-a-webhook
+    head :ok
+  end
+
+end
--- a/app/controllers/alaveteli_pro/stripe_webhooks_controller.rb
+++ b/app/controllers/alaveteli_pro/stripe_webhooks_controller.rb
@@ -4,7 +4,22 @@ class AlaveteliPro::StripeWebhooksController < ApplicationController
 
   before_action :read_event_notification
 
+  class UnhandledStripeWebhookError < StandardError ; end
+
   def receive
+    begin
+      case @stripe_event.type
+      when 'invoice.payment_failed'
+        # ToDo: add specific handler code here, but for now just raise an
+        # UnhandledStripeWebhookError
+        raise UnhandledStripeWebhookError.new(@stripe_event.type)
+      else
+        raise UnhandledStripeWebhookError.new(@stripe_event.type)
+      end
+    rescue UnhandledStripeWebhookError => e
+      notify_exception(e)
+    end
+
     # send a 200 ok to acknowlege receipt of the webhook
     # https://stripe.com/docs/webhooks#responding-to-a-webhook
     render json: {}, status: 200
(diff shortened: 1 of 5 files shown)


app/controllers/alaveteli_pro/stripe_webhooks_controller.rb
```diff
@@ -2,10 +2,37 @@
 # Does not inherit from AlaveteliPro::BaseController because it doesn't need to
 class AlaveteliPro::StripeWebhooksController < ApplicationController
 
+  before_action :read_event_notification
+
   def receive
+    return unless @stripe_event
```
gbp: From my testing this line doesn't do anything as `@stripe_event` is never empty because Stripe gem always returns a `Stripe::Event` object

app/controllers/alaveteli_pro/stripe_webhooks_controller.rb
```diff
@@ -4,9 +4,26 @@ class AlaveteliPro::StripeWebhooksController < ApplicationController
 
   before_action :read_event_notification
 
+  class UnhandledStripeWebhookError < StandardError ; end
+
   def receive
     return unless @stripe_event
 
+    begin
+      case @stripe_event.type
```
gbp: I'm able to get an exception here by not passing a `type` JSON field in the payload. Would be good to catch this.

app/controllers/alaveteli_pro/stripe_webhooks_controller.rb
```diff
@@ -2,10 +2,37 @@
...
+      @stripe_event = Stripe::Webhook.construct_event(
+        payload, sig_header, endpoint_secret
+      )
+    rescue JSON::ParserError => e
+      # Invalid payload, reject the webohook
+      render json: {}, status: 400
+      return
+    rescue Stripe::SignatureVerificationError => e
+      # Invalid signature, reject the webhook
+      render json: {}, status: 401
```
gbp: We could output the error message here E.G. `{ error: e.message }` this helped me when testing. Also do we want to pass these exceptions onto `ExceptionNotifier`?

app/controllers/alaveteli_pro/stripe_webhooks_controller.rb
```diff
@@ -2,10 +2,37 @@
...
+    endpoint_secret = AlaveteliConfiguration.stripe_webhook_secret
+    @stripe_event = nil
+
+    begin
+      @stripe_event = Stripe::Webhook.construct_event(
+        payload, sig_header, endpoint_secret
+      )
+    rescue JSON::ParserError => e
+      # Invalid payload, reject the webohook
+      render json: {}, status: 400
```
gbp: See below

gbp:
Testing by running:

```
❯ date +%s
1512478725

❯ ruby -r 'openssl' -e "puts OpenSSL::HMAC.hexdigest(OpenSSL::Digest.new('sha256'), 'whsec_123', '1512478725.{}')"
f2ff4bb838f0f14bbacc542bad5702ae15c3ea40dfafdea6e190e3a94b7df076

❯ curl -X POST -H 'Stripe-Signature: t=1512478725,v1=f2ff4bb838f0f14bbacc542bad5702ae15c3ea40dfafdea6e190e3a94b7df076' -H 'Content-Type: application/json' -d '{}' http://alaveteli.test/pro/subscriptions/stripe-webhook
```

app/controllers/alaveteli_pro/stripe_webhooks_controller.rb
lizconlan: ugh, it was intended to be blank if the sig didn't match - will double-check (sounds like I made an assumption)

lizconlan: (d'oh - the `before_action` returns and that code is never reached - good catch, thanks!)

lizconlan:
(Shout if you'd prefer me to rebase it for re-review!)

app/controllers/alaveteli_pro/stripe_webhooks_controller.rb
```diff
@@ -18,6 +18,11 @@ def receive
       end
     rescue UnhandledStripeWebhookError => e
       notify_exception(e)
+    rescue NoMethodError => e
```
gbp: This rescue will also pickup any `NoMethodError` exceptions raise within a matching case/when statement - not just for when `@stripe_event.type` doesn't exist. I think this could be confusing, do we want to be more explicit here?

PR #4409, opened 2017-12-01
[pro#374] restrict access to embargo settings for non-pro users

lizconlan:
~~Closes~~ ~~Connects to~~ Closes mysociety/alaveteli-professional#374

Tightens up access controls for existing embargo functionality.

The outstanding issue is adding the ability to retrospectively add an embargo to an existing batch which should be subject to the same checks (as it doesn't yet exist, this PR is excused from restricting it) - split out to mysociety/alaveteli-professional#447

### DIFF
--- a/app/views/alaveteli_pro/info_requests/_embargo_form.html.erb
+++ b/app/views/alaveteli_pro/info_requests/_embargo_form.html.erb
@@ -13,20 +13,30 @@
   <label class="houdini-label" for="input1"><%= _("Change privacy") %></label>
   <input class="houdini-input" type="checkbox" id="input1">
   <div class="houdini-target extend-embargo-sidebar">
-    <% if info_request.embargo %>
-      <%= render partial: "alaveteli_pro/info_requests/embargo_extension_form",
-                 locals: {info_request: info_request} %>
+  <% if info_request.embargo && can?(:update, info_request.embargo) %>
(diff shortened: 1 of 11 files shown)


lizconlan:
Does not allow retrospective embargoing of batches, I think I'd rather tackle that separately

app/views/alaveteli_pro/info_requests/_embargo_form.html.erb
```diff
@@ -13,20 +13,30 @@
...
                       confirm: _("Are you sure you want to publish " \
-                                 "this request?")
-                    } %>
-    <% else %>
-      <%= render partial: "alaveteli_pro/info_requests/embargo_create_form",
-                 locals: {info_request: info_request} %>
+                                 "this request?") } %>
     <% end %>
+    <%= render partial: "alaveteli_pro/info_requests/embargo_info",
+               locals: {embargo: info_request.embargo, tense: :present} %>
```
gbp: To be consistent with the batch views, this partial should be rendered above the 'Publish request' button above

lizconlan: Great, thanks!

lizconlan:
ToDo:
- [ ] double check in browser
- [ ] get clean Travis run

PR #4415, opened 2017-12-05
[pro#406] Show number of replies to requests on dashboard

gbp:
Fixes https://github.com/mysociety/alaveteli-professional/issues/406
 
Add number of incoming_messages to the pro dashboard for both normal and batch requests

### DIFF
--- a/app/views/alaveteli_pro/info_requests/_info_request.html.erb
+++ b/app/views/alaveteli_pro/info_requests/_info_request.html.erb
@@ -40,6 +40,11 @@
       <p><%= info_request.date_response_required_by.strftime('%d-%m-%Y') %></p>
     </div>
 
+    <div class="request__status">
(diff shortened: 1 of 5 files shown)


app/views/alaveteli_pro/info_requests/_info_request.html.erb
```diff
@@ -40,6 +40,11 @@
       <p><%= info_request.date_response_required_by.strftime('%d-%m-%Y') %></p>
     </div>
 
+    <div class="request__status">
```
garethrees: I think this needs a separate class as you've done for batches above because

* Its not aligned correctly on smaller screens
* It doesn't make sense semantically

screen shot 2017-12-06 at 10 28 36

gbp:
Updated to look better at different screen sizes. Also found/ fixed a broken status icon image.

large
medium
small

PR #4421, opened 2017-12-08
Include webhook data in error notifications

lizconlan:
Connects to mysociety/alaveteli-professional#425

### DIFF
--- a/app/controllers/alaveteli_pro/stripe_webhooks_controller.rb
+++ b/app/controllers/alaveteli_pro/stripe_webhooks_controller.rb
@@ -12,12 +12,12 @@ class AlaveteliPro::StripeWebhooksController < ApplicationController
       when 'invoice.payment_failed'
         # ToDo: add specific handler code here, but for now just raise an
         # UnhandledStripeWebhookError
-        raise UnhandledStripeWebhookError.new(@stripe_event.type)
+        raise UnhandledStripeWebhookError.new(@stripe_event)
       else
-        raise UnhandledStripeWebhookError.new(@stripe_event.type)
+        raise UnhandledStripeWebhookError.new(@stripe_event)
       end
     rescue UnhandledStripeWebhookError => e
-      notify_exception(e)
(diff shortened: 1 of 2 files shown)


app/controllers/alaveteli_pro/stripe_webhooks_controller.rb
```diff
@@ -55,15 +55,22 @@ def check_for_event_type
...
   end
 
-  def notify_exception(error)
+  def notify_exception(error, data = nil)
     if send_exception_notifications?
-      ExceptionNotifier.notify_exception(error, :env => request.env)
+      if data
+        ExceptionNotifier.notify_exception(
+          error,
+          data: { webhook_data:  data }
```
garethrees: Quick Q: What's the reason for not including `request.env` in this case?

lizconlan: oh, er - oversight I think

although I'm not convinced there'll be anything useful in `request.env` in this case (but as the whole point is to have as much information as possible about what might have gone wrong...)

garethrees: Okay – I'll give this an initial review but I don't see the point of adding extra conditionals here when we want the `request.env` anyway.

lizconlan: agreed, will fix (simplify) post review - thanks!

rough sketch:

```diff
- def notify_exception(error, data = nil)
+ def notify_exception(error, data = {})
...
# remove conditional, always send `data` and `request.env` to notifier
```

lizconlan:
Withdrawing as now that the initial emails are coming through it's obvious that the data _is_ being included

I found this hard to test conclusively without seeing what an actual webhook event looked like and thought it was safer to be sure it was there rather than spending the same amount of time trying to replicate the live environment. In hindsight, taking the time to add an integration test and passing in a mock  event from StripeMock would have been the best approach here.

PR #4425, opened 2017-12-12
[pro#303] Announcements

gbp:
Fixes https://github.com/mysociety/alaveteli-professional/issues/303


app/models/announcement.rb
```diff
@@ -0,0 +1,16 @@
...
+  default_scope -> { order(created_at: :asc) }
+  scope :for_user, -> (user) {
+    where(
+      'announcements.id NOT IN (SELECT announcement_id FROM ' \
+      'announcement_dismissals WHERE user_id = :user_id)',
+      user_id: user
+    )
+  }
+
+  validates :title, :content, presence: true
```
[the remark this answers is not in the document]
garethrees: Also, these will need to be translated attributes. See `PublicBodyCategory` (and related models) to get a feel for how that works.

spec/models/announcement_dismissal_spec.rb
```diff
@@ -0,0 +1,13 @@
+require 'spec_helper'
+
+RSpec.describe AnnouncementDismissal, type: :model do
```
garethrees:
```diff
- Spec.describe AnnouncementDismissal, type: :model do
+ describe AnnouncementDismissal do
```

app/controllers/announcement_controller.rb
```diff
@@ -0,0 +1,22 @@
+class AnnouncementController < ApplicationController
+  before_action :authenticate
+
+  def destroy
+    AnnouncementDismissal.find_or_create_by(
```
garethrees: Should handle failure here I think?

app/models/announcement.rb
```diff
@@ -0,0 +1,16 @@
+class Announcement < ActiveRecord::Base
```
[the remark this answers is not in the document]
gbp: Done

spec/models/announcement_dismissal_spec.rb
gbp: Happy to change this but we should consider switching to the newer syntax: https://github.com/rspec/rspec-rails/commit/ca0d249858903949052e06884e8e7f9d596cdc79

garethrees: Good Q. Lets go with `describe` for now and make a ticket for https://github.com/rspec/rspec-rails/commit/ca0d249858903949052e06884e8e7f9d596cdc79 and we can just make one commit for the change if we decide to do it.

gbp: Have changed this in 542ed06 & 04c27ba

gbp:
> Site-wide should only show most recent; not spawn more as you close them?

@garethrees do you mean when dismissing an announcement it should also hide all previous ones? But only for site-wide announcements?

garethrees:
> @garethrees do you mean when dismissing an announcement it should also hide all previous ones? But only for site-wide announcements?

Well, the _problem_ (IMO) is that lets say there are 5 "active" (as in, not deleted) site-wide announcements.

* You come to the site, not logged in.
* You see the first announcement
* Close the announcement
* Visit a new page
* At this point you see the _next_ announcement
* Close it; visit a new page
* Another announcement appears
* etc…

Its not a huge problem, but doesn't strike me as hugely user-friendly.

I think when we're showing a site-wide, we only ever want to show the most recent one. Once a user closes it, they shouldn't see more until an admin creates a new site-wide announcement.

I haven't thought about how tricky this would be to actually implement – so we might have to compromise somewhere  – but I think we should do what we can to not "spam" users with lots of popups.

gbp:
Guess I was under the assumption site-wide announcements would be time limited and removed manually by an admin. Would mean this wouldn't be an issue but obviously this feature might not be used like that so yes I agree its not the best way. I'm sure I can change make some changes to the scopes to change this.

garethrees:
> I'm sure I can change make some changes to the scopes to change this.

Yeah, don't worry if its a huge pain. I was thinking if its something trivial like adding `.last` somewhere then worth doing.

PR #4445, opened 2018-01-03
[pro#335] Add pro batch feature flag

lizconlan:
Up for ~~early~~ review ~~requesting feedback on how linking through to the batch form should work.~~ (Probably not like this... although it's bearable as an extension of the current beta functionality. **Edit:** after discussion, we've decided that this is an adequate solution for the **beta** functionality - need to capture the concerns somewhere so we can work out what the full release version needs to do.)

I think the first bit (adding a filter to check for the feature flag and redirecting without it) is correct **but** the link across to the batch request form needs some more thought.

At the moment it looks like this...

screen shot 2018-01-04 at 17 19 59

If you click the link, you get the batch form (great). If you click the link after filling in some data, the data is lost (not ideal). Also, the wording is a placeholder but that's the simplest thing to fix.

Probably not relevant to this ticket, but the batch form should probably also have a way back to the standard request form (might come up as part of mysociety/alaveteli-professional#404)

Closes mysociety/alaveteli-professional#335

### DIFF
--- a/app/assets/stylesheets/responsive/alaveteli_pro/_new_request_layout.scss
+++ b/app/assets/stylesheets/responsive/alaveteli_pro/_new_request_layout.scss
@@ -19,3 +19,9 @@
 .request_embargo {
   margin-top: 2em;
 }
+
+.pro_batch_link {
+  display: block;
+  margin-top: 0.25em;
+  margin-bottom: -1em;
+}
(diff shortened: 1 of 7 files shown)


gbp:
When this is deployed we will need to set the feature flags for the relevant users who've already been given access.

lizconlan:
> When this is deployed we will need to set the feature flags for the relevant users who've already been given access.

Good point - note to self, do not close the ticket yet

PR #4446, opened 2018-01-04
[sec#5] Change password encryption to use bcrypt

### LINKED ISSUE #3404: Enforce password length

gbp:
Closes https://github.com/mysociety/alaveteli_security/issues/5 & Closes #3404

### DIFF
--- a/spec/models/user_spec.rb
+++ b/spec/models/user_spec.rb
@@ -167,6 +167,35 @@ describe User, 'password hashing algorithms' do
       expect(found_user).to eq(user)
     end
 
+    it 'updates hashed password with bcrypt version' do
+      expect(found_user.hashed_password).to match(/^\$2[ayb]\$.{56}$/)
+    end
+
+  end
+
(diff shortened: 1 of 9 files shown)


spec/models/user_spec.rb
```diff
@@ -138,108 +138,204 @@
...
-    @user.email = "silly@localhost"
-    @user.valid?
-    expect(@user.errors[:hashed_password].size).to eq(1)
+  it 'should not save with no password' do
+    @user.password = ''
+    expect(@user).to_not be_valid
+    expect(@user.errors[:password].size).to eq(1)
+  end
+
+  it 'should not save with too short password' do
```
gbp: > some specs to show the new length constraints in action would be good

This spec and the one below are testing password length - are these okay?

gbp:
I deliberately didn't update the users fixtures - my thinking is that leaving them as they are tests the greatest amount of code and it would become obvious if ever anything happened to the encryption upgrades. 

Think I have changed all the remaining issues.

PR #4449, opened 2018-01-04
Remove spring preloader

garethrees:
This seems to cause more problems than it solves. I personally disable
it, and it repeatedly catches people out when they don't.


lizconlan:
You might need to de-springify some of the `/bin` files as well - `rails`, `rake` and `rspec` are likely to be affected, and possibly others

gbp:
Changed my thinking here after looking at how `rails new --no-spring` generates the files

gbp:
@garethrees can I double check you are happy with this?

garethrees:
Looks Good To Me!

PR #4450, opened 2018-01-05
WIP: Browse authorities by category in Batch builder

zarino:
**Not ready for review**

Connects to https://github.com/mysociety/alaveteli-professional/issues/76
Connects to https://github.com/mysociety/alaveteli-professional/issues/347

TO DO:

- [x] Styling and markup for "mode" tabs (search vs browse) – @zarino
- [x] Styling and markup for browse "tree" (parent category with children) – @zarino
- [x] Make the tabs work – @gbp and @zarino
   - JavaScript-side, we might have Foundation tabs component available to us, if that’s useful?
- [x] Generate the browse tree properly with real data – @gbp 
   - Might need pagination?
   - JavaScript enhancement to collapse children until parent category clicked? @zarino

screen shot 2018-01-05 at 12 30 46


zarino:
> unless I'm missing something, the markup doesn't seem to support having more than two levels

Good spot. I can add that.

> Finally can you confirm the location of the 'Write request' button

In Search mode, the submit button fits nicely at the top, to the right of the search input. But in Browse mode, there’s no input for it to line up next to, so either we have some dead space up there (a bit weird) or we move it down to the bottom (also a bit weird, but maybe slightly less weird). I went for what I thought was the less weird solution. Can always change it. Does it feel really odd?

gbp:
> Does it feel really odd?

No It's fine, I just wasn't sure moving it was intentional

garethrees:
I'm going to close this and just work off the "master" PR (https://github.com/mysociety/alaveteli/pull/4480) that includes these commits.

Issue #4398, opened 2017-11-28
[ERROR] sessions#new (ActionController::BadRequest) "Invalid query parameters: invalid %-encodi...


lizconlan:
There have been a few of these today, feels like an exploit attempt

```
An ActionController::BadRequest occurred in sessions#new:

  Invalid query parameters: invalid %-encoding (%2Shfre%2Scrgre_oynpx_2%23rznvy_fhofpevcgvbaf)
  lib/strip_empty_sessions.rb:13:in `call'

-------------------------------
Request:
-------------------------------

  ERROR: Failed to generate exception summary:

  ActionView::Template::Error: Invalid query parameters: invalid %-encoding (%2Shfre%2Scrgre_oynpx_2%23rznvy_fhofpevcgvbaf)

-------------------------------
Environment:
-------------------------------

  * REQUEST_METHOD                                 : GET
  * REQUEST_URI                                    : /profile/sign_in?r=%2Shfre%2Scrgre_oynpx_2%23rznvy_fhofpevcgvbaf
```

garethrees:
We're getting a lot of these, so lets just rescue it in [`render_exception`](https://github.com/mysociety/alaveteli/blob/develop/app/controllers/application_controller.rb#L179) and return a 404.

PR #4451, opened 2018-01-05
[#4398] Ignore ActionController::BadRequest notifications

garethrees:
We keep getting notifications from these for malformed requests.

Since Rack is generating these, it doesn't look like we can rescue them
in `ApplicationController#render_exception`. [1]

We could try utf8-cleaner [2] but I'm not sure its worth adding a new
dependency.

Instead, lets just ignore notifications about this error.

I wasn't able to test this – couldn't see any indication of exception
notification sending a mail in the logs – so I'm not 100% sure without a
second pair of eyes.

Fixes https://github.com/mysociety/alaveteli/issues/4398

[1] https://stackoverflow.com/q/24807592/387558
[2] https://github.com/singlebrook/utf8-cleaner

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -98,6 +98,7 @@ Rails.application.configure do
 
   if !AlaveteliConfiguration.exception_notifications_from.blank? && !AlaveteliConfiguration.exception_notifications_to.blank?
     middleware.use ExceptionNotification::Rack,
+      :ignore_exceptions => ['ActionController::BadRequest'] + ExceptionNotifier.ignored_exceptions,
       :email => {
         :email_prefix => exception_notifier_prefix,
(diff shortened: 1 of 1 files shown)


garethrees:
> If this doesn't stop the errors from coming through then maybe we can look at adding a middleware to deal with this such as `Rack::EnforceValidEncoding`.

I _think_ this is already in place – the [code comment](https://github.com/rack/rack-contrib/blob/master/lib/rack/contrib/enforce_valid_encoding.rb#L2-L4) says that "If the validation fails, then a 400 Bad Request response is returned immediately.". This is what we're getting, but because its from middleware we can't rescue it in the app (or at least that's my shaky understanding).

> Suspect this will work but not 100% certain as I can't fully test it.

Yeah, lets give it a go and see what happens.

PR #4452, opened 2018-01-05
Fix comment in .vagrant.yml.example

garethrees:
The comment was getting included in the value instead of just '2',
breaking the VirtualBox config.


garethrees:
> From my testing having a comment on the same line doesn't affect the value. Happy for this to be merged tho

Weird; might be a local ruby version difference.

PR #4458, opened 2018-01-11
Add Flipper UI gem and constrained admin route

gbp:
Can now go to `/admin/flipper` to configure feature flags.

Not sure if this is needed/ wanted but have been sitting on this change for a while and have found it useful in development. Thoughts?


Gemfile
```diff
@@ -155,6 +155,7 @@ gem 'therubyracer', '~> 0.12.0'
 
 # Feature flags
 gem 'alaveteli_features', :path => 'gems/alaveteli_features'
+gem 'flipper-ui'
```
garethrees: Have a read of the Gemfile documentation above. Will want to add a pessimistic constraint.

config/routes.rb
```diff
@@ -13,6 +13,8 @@
 end
 
 Rails.application.routes.draw do
+  admin_constraint = lambda { |request| request.session[:using_admin] }
```
garethrees: There's a little gotcha here, but I don't think its worth doing anything about at this point. Just highlighting for reference.

* Log out
* Go to homepage
* Sign in through sign-in form
* Get redirected back to homepage
* Go directly to `/admin/flipper`
* RouteNotFound error happens

This is because you need to actually go to an`/admin/` page to get `:using_admin` set.

No big deal, but just thought I'd point it out.

config/routes.rb
```diff
@@ -411,6 +413,9 @@
 
   #### AdminPublicBody controller
   scope '/admin', :as => 'admin' do
+    constraints admin_constraint do
+      mount Flipper::UI.app(AlaveteliFeatures.backend) => '/flipper'
+    end
```
garethrees: add a newline after this end to break up the different routes.

config/routes.rb
gbp: Had noticed this too. My think was it wasn't a big enough issue to fix especially as it would introduce more complexity in the admin constraint, such as loading users from the DB and checking roles

garethrees: yeah 👍  – :shipit:

PR #4461, opened 2018-01-12
[pro#212] Make the blank slate call to action a button

wrightmartin:
also fixes broken hover state

Now looks like...

screen shot 2018-01-11 at 14 25 25

Fixes https://github.com/mysociety/alaveteli-professional/issues/212

### DIFF
--- a/app/assets/stylesheets/responsive/alaveteli_pro/_requests_style.scss
+++ b/app/assets/stylesheets/responsive/alaveteli_pro/_requests_style.scss
@@ -125,12 +125,6 @@ $complete: rgb(112, 203, 99);
   }
 }
 
-.new-request-button,
-.new-request-button:visited,
-.new-request-button:hover {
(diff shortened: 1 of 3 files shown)


app/assets/stylesheets/responsive/alaveteli_pro/_requests_style.scss
```diff
@@ -126,7 +126,8 @@ $complete: rgb(112, 203, 99);
 }
 
 .new-request-button,
-.new-request-button:visited {
+.new-request-button:visited,
+.new-request-button:hover {
```
gbp: I think these styles aren't being used at all anymore as the class of the button been changed to just `button`. Can these be removed?

wrightmartin: It appears so - will remove

PR #4462, opened 2018-01-12
Remove wheezy support

lizconlan:


gbp:
Missed reference to wheezy in `.vagrant.yml.example`

PR #4465, opened 2018-01-16
Added Binstubs for supported vagrant OSs

gbp:
Found these useful while testing vagrant box builds.

The `config.vm.define` setting allows us to have more than one vagrant box per repo without using snapshots or destroying the VM. In the binstubs this is assigning by using an ENV variable so we can call `./bin/vagrant-stretch up` & `./bin/vagrant-jessie up` at the same time

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -87,6 +87,7 @@ DEFAULTS = {
   'memory' => 1536,
   'themes_dir' => '../alaveteli-themes',
   'os' => 'jessie64',
+  'name' => 'default',
   'use_nfs' => false,
   'show_settings' => false,
   'cpus' => cpu_count
@@ -139,6 +140,7 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
   else
     box
   end
+  config.vm.define SETTINGS['name']
   config.vm.box_url = box_url
   config.vm.network :private_network, ip: SETTINGS['ip']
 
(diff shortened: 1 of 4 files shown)


Vagrantfile
```diff
@@ -139,6 +140,7 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
   else
     box
   end
+  config.vm.define SETTINGS['name']
```
garethrees: Could we use `SETTINGS['os']` here so that we don't need to introduce a new key?

gbp: A problem I can see with that is because `config.vm.define` defaults to `default`. Using `SETTINGS['os']` as the default would mean existing `default` boxes wouldn't be loaded so `vagrant up` would force rebuild on everyone

garethrees: Ah that makes sense.

Pushed a couple of commits in https://github.com/mysociety/alaveteli/commits/vagrant-os-binstubs-gr as a slightly alternative approach. Have a look and see what you think :)

garethrees: Haven't extensively tested btw, but can do if it looks promising.

gbp: Your approach seems like is could work, there is one issue I've noticed tho...

As the VM name is being set by the last vagrant cmd argument this isn't being added to either `ENV` or `SETTINGS` hash so the wrong branch of this conditional is being run, it will always run as `jessie64`

https://github.com/mysociety/alaveteli/blob/24adde5001d2a1cb7db02a167acfc8ddb52d768b/Vagrantfile#L204

Personally I feel using separate keys is the least complex option here, especially considering this line from multi-machine docs: 
> Other commands, such as `vagrant up`, operate on every machine by default.

Which will probably cause all the OS to get built and run at the same time!

Happy to leave this for now and come back at a later date.

garethrees: > Personally I feel using separate keys is the least complex option here

Yeah okay, lets just go with it as-is.

PR #4466, opened 2018-01-16
[Pro#433] Make opengraph tags change for pro parts of the site

wrightmartin:
Fixes https://github.com/mysociety/alaveteli-professional/issues/433

screen shot 2018-01-16 at 15 12 33

Here's what you now see when a page is part of pro. 

This doesn't apply to the marketing pages, which I think is wrong.

Is there an easy way to override the opengraph tags for a page?


wrightmartin:
Update: These changes will apply to the pricing page, just not the top level /pro page

garethrees:
Think this is for https://github.com/mysociety/alaveteli-professional/issues/433

PR #4469, opened 2018-01-18
[WIP] Mark the public body name as html_safe before translating the salutation

lizconlan:
**initial version, PR created for visibility and to get an initial spec run - needs a new spec to confirm the fix**

To prevent part of the public body name from being displayed with HTML
escape sequences - seems to mostly affect single quotes when substituted
into the initial request body copy on the new request form.

Connects to mysociety/belgium-theme#72

### DIFF
--- a/app/models/outgoing_message/template/initial_request.rb
+++ b/app/models/outgoing_message/template/initial_request.rb
@@ -20,7 +20,9 @@ class OutgoingMessage
 
       def salutation(replacements = {})
         if replacements[:public_body_name]
-          _("Dear {{public_body_name}},", replacements)
+          replacements[:public_body_name] =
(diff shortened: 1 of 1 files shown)


lizconlan:
I think this is the wrong approach as although the public body name is supposed to be controlled by responsible admins, it's possible we could accidentally end up with some included HTML. As only the straight single quote seems to be affected, a literal gsub either to switch the straight quote for a curly one or to remove the encoding is probably a safer alternative.

Have advised the site admin to use a curly quote instead

garethrees:
Didn't get chance to look at this during the current sprint, and unlikely we'll get to it this sprint either, so just going to close this as you're not convinced at the implementation. We can take a look at it again if we need to revisit the issue.

PR #4471, opened 2018-01-23
Remove Ruby 1.9 references

lizconlan:
Part of #4209

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -104,7 +104,6 @@ gem 'holidays', '~> 2.2.0', '< 3.0.0'
 gem 'iso_country_codes', '~> 0.7.0'
 gem 'mahoro', '~> 0.4'
 gem 'newrelic_rpm'
-gem 'net-http-local', '~> 0.1.0', :platforms => [:ruby_19]
 gem 'nokogiri', '~> 1.6.0', '< 1.7'
 gem 'open4', '~> 1.3.0'
 gem 'rack', '~> 1.6.0'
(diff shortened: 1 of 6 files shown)


Gemfile
```diff
@@ -104,7 +104,6 @@ gem 'holidays', '~> 2.2.0', '< 3.0.0'
 gem 'iso_country_codes', '~> 0.7.0'
 gem 'mahoro', '~> 0.4'
 gem 'newrelic_rpm'
-gem 'net-http-local', '~> 0.1.0', :platforms => [:ruby_19]
```
garethrees: This commit should also remove the code that used the gem:

* https://github.com/mysociety/alaveteli/blob/develop/lib/quiet_opener.rb#L3-L5
* https://github.com/mysociety/alaveteli/blob/develop/lib/quiet_opener.rb#L46-L50

lib/tasks/config_files.rake
```diff
@@ -78,7 +78,7 @@ namespace :config_files do
               'VCSPATH=alaveteli ' \
               'SITE=alaveteli ' \
               'SCRIPT_FILE=config/alert-tracks-debian.example ' \
-              'RUBY_VERSION=1.9.1 '
+              'RUBY_VERSION=2.1.5 '
```
garethrees: Why 2.1.5? I assume because this is the default included with Debian Jessie – our current default Vagrant VM?

Commit message should explain this.

lizconlan: Good point - thanks. 2.1.5 is the most commonly available Ruby at this point in time (off the top of my head - will check for the purposes of the commit message - Stretch adds 2.3.x)

lizconlan:
Closed in favour of #4209

PR #4472, opened 2018-01-23
Add a specific note about overriding the signup form template

lizconlan:
Found this when trying to use the Belgium theme against 0.31, thought it was worth its own note in case that error is raised in production, particularly as it affects the signup/signin page.


lizconlan:
I'll close it in that case - I imagine this will come up in Google searches in the future anyway if the error looks baffling to future implements (e.g. future me)

PR #4474, opened 2018-01-24
[#Pro449] Update zero charge invoice

garethrees:
Handle an Invoice with no charge 

Trial customers receive invoices with no attached charge, so we're
trying to update a `nil` charge in some cases.

Here, we only update the charge if one exists.

Fixes mysociety/alaveteli-professional#449.

Also includes some refactoring.


garethrees:
Tested locally and happy that a `invoice.payment_succeeded` event where there's no charge attached to an invoice doesn't return a 500.

I dismissed a CodeClimate check saying `AlaveteliPro::StripeWebhooksController#receive` is too complex, because it was out of scope for this bugfix.

PR #4476, opened 2018-01-24
Minor change to user drop down positioning 

wrightmartin:
Ensure the whole button behaves better when the navigation menu needs to overflow

Previously, due to the way the arrow icon was positioned it didn't follow the button when it overflowed to a new line

Themes may need a tweak for this fix if they have overridden the `right`  value of `.js-loaded .account-link:after`

Connects to https://github.com/mysociety/belgium-theme/issues/74

### DIFF
--- a/app/assets/stylesheets/responsive/_header_layout.scss
+++ b/app/assets/stylesheets/responsive/_header_layout.scss
@@ -350,15 +350,14 @@
 .js-loaded {
   .account-link {
     @include respond-min( $main_menu-mobile_menu_cutoff ) {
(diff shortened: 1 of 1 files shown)


lizconlan:
Problem confirmed as reviewer's browser settings \*headdesk\*

screen shot 2018-01-25 at 11 54 46

vs

screen shot 2018-01-25 at 11 42 51

screen shot 2018-01-25 at 11 56 09

lizconlan:
(merged and commit message tweaked slightly)

PR #4481, opened 2018-01-25
[#Pro450] Support annual billing

garethrees:
Fixes https://github.com/mysociety/alaveteli-professional/issues/450


garethrees:
Ignoring the codeclimate warning of the spec file being too long.

I agree, but the current (complex) controller code makes it necessary.

Really, `AlaveteliPro::StripeWebhooksController#receive` should hand off the webhook to dedicated handlers where the thorough testing happens:

```ruby
class AlaveteliPro::StripeWebhooksController < ApplicationController
  def receive
    stripe_event = build_event(request.body.read)
    handler = StripeWebhookHandler.for(stripe_event.type)
    result = handler.process
    render json: { message: result.message }, status: result.status
  end
end

PR #4485, opened 2018-01-26
Refactor public body queries gr

garethrees:
Tweaked https://github.com/mysociety/alaveteli/pull/4457 so that specs are with new methods and all commits pass the controller specs.


gbp:
Reviewed and as I needed to reword a commit I've rebased onto the original `refactor-public-body-queries` here: #4457

PR #4513, opened 2018-02-08
Add Pull Request template

garethrees:
I've been meaning to do this for a while but didn't get around to
thinking of a good format. I saw this in an EP repo [1] and thought it
looked like a great starting point.

[1]
https://github.com/everypolitician/legislative-explorer/blob/master/.github/PULL_REQUEST_TEMPLATE.md


garethrees:
Changed this to be more of a "plz try to include these" rather than a rigid template. Lets see how it goes.

PR #4518, opened 2018-02-09
[#pro356] Show public body notes in the batch builder

garethrees:
Fixes https://github.com/mysociety/alaveteli-professional/issues/356.

MVP of adding important notes to the batch builder. Here we just show
any notes if the authority has any notes. We want to be more selective
about which notes we show, but that will come in a future commit.

I think this is a good-enough implementation of this, given we lost a bunch of time this sprint.

**Initial load:**

screen shot 2018-02-09 at 16 25 16

**Open notes:**

screen shot 2018-02-09 at 16 25 22

**Long notes:**

screen shot 2018-02-09 at 16 33 12

**Added to the batch:**

screen shot 2018-02-09 at 16 40 10

**Read & closed the notes**

screen shot 2018-02-09 at 16 39 36

> The notes are closed by default in the search results pane. When a user
adds them, the notes are automatically expanded in the selected
authorities pane.

This sounded like a good idea, but the "cheap" wait of adding the `open` attribute to the `<details>`  tag means that _all_ notes get opened when you add an authority. Looks a bit annoying?

auto-open-notes-opens-all-notes

## Not done:

* Style the notes with the blue background – we [decided](https://github.com/mysociety/alaveteli-professional/issues/356#issuecomment-360437191) not to do this
* Adding notes below the selection panes when you add an authority with notes. I think we need another solution here, because https://github.com/mysociety/alaveteli/pull/4480#pullrequestreview-91856336 brought up the fact that we shouldn't even have the selection panes – the page should just expand as far as necessary. If we do that, then the notes could well end up quite far down.

### DIFF
--- a/app/assets/stylesheets/responsive/_ms_edge.scss
+++ b/app/assets/stylesheets/responsive/_ms_edge.scss
@@ -1,0 +1,8 @@
+@supports (-ms-ime-align: auto) {
+  details.batch-builder__authority-list__authority__notes {
+    summary {
+      color: grey !important;
+      font-weight: bold;
+    }
+  }
+}
--- a/app/assets/stylesheets/responsive/alaveteli_pro/_batch_request_authority_search_layout.scss
+++ b/app/assets/stylesheets/responsive/alaveteli_pro/_batch_request_authority_search_layout.scss
@@ -65,11 +65,14 @@
 .batch-builder__authority-list__authority {
     padding: 1.5em 1em;
     @include flexbox();
-    @include flex-align(center);
+    @include flex-align('flex-start');
 }
 
-.batch-builder__authority-list__authority__name {
+.batch-builder__authority-list__authority__details {
     @include flex(1 0 0%);
+}
+
+.batch-builder__authority-list__authority__notes {
 }
 
 .batch-builder__authority-list__authority__action {
--- a/app/assets/stylesheets/responsive/alaveteli_pro/_batch_request_authority_search_style.scss
+++ b/app/assets/stylesheets/responsive/alaveteli_pro/_batch_request_authority_search_style.scss
@@ -52,6 +52,18 @@
     }
 }
 
+.batch-builder__authority-list__authority__notes {
+  font-size: 0.85em;
+
+  summary {
+    color: #2688dc;
+  }
+
+  p {
+    color: grey;
+  }
+}
+
 .batch-builder__authority-list__authority__action__added {
     color: green;
     font-weight: 600;
--- a/app/assets/stylesheets/responsive/all.scss
+++ b/app/assets/stylesheets/responsive/all.scss
@@ -101,4 +101,6 @@
 
 @import "responsive/alaveteli_pro/_pro_marketing";
 
+@import "responsive/_ms_edge";
+
 @import "responsive/custom";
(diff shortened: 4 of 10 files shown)


garethrees:
Also now added the special tag for showing notes.

Both authorities have notes; only silly walks is tagged with `important_notes`:

screen shot 2018-02-09 at 17 16 52

Cleaned up the special tags help in the admin form too:

screen shot 2018-02-09 at 17 36 09

garethrees:
@wrightmartin could you just have a look at the CSS/flexbox stuff to make sure it works. Would also be great to get your thoughts on some of the behaviour questions I've noted in the main PR comments.

wrightmartin:
Using the `<details>` and `<summary>` tags fails over on browsers that don't support them. The notes will already be revealed and some of the styling doesn't make sense (`<details>` acts as a heading but looks like a link).

Here's the UI on IE10
screen shot 2018-02-13 at 11 18 40

An improvement here would be to change the styles based on the browser support offered.

garethrees:
* 88be128 removes the `open` attribute so that details tags aren't auto expanded when added
* 565ec0d fixes summary for MS Edge:

screen shot 2018-02-16 at 13 33 01

gbp:
Merged with #pro347 here: https://github.com/mysociety/alaveteli/tree/pro347-pro356-batch-browsing-and-notes

garethrees:
> In Firefox we get no disclosure triangles with the `<details>` tag.

A quick Google finds me:

> The cause of the missing arrow appears to be related to the way Firefox handles `<details>` elements. From the [Mozilla Developer Network](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details):
>
> > For cross browser compatible styling hide Firefox's marker by setting summary { display: block }
>
> _– https://github.com/timarney/react-faq/issues/29_

The linked MDN page doesn't actually say that any more.

The `display: block` comes from the vendored normalize.scss in foundation:

```css
/* HTML5 display definitions
   ========================================================================== */
/**
 * Correct `block` display not defined for any HTML5 element in IE 8/9.
 * Correct `block` display not defined for `details` or `summary` in IE 10/11
 * and Firefox.
 * Correct `block` display not defined for `main` in IE 11.
 */
/* line 34, /home/vagrant/bundle/ruby/2.3.0/gems/foundation-rails-5.5.3.2/vendor/assets/stylesheets/normalize.scss */
article,
aside,
details,
figcaption,
figure,
footer,
header,
hgroup,
main,
menu,
nav,
section,
summary {
  display: block;
}
```

I'm going to just ignore this for now as it seems like quite a pain to fix globally. @zarino @wrightmartin feel free to make a PR if this looks like an obvious fix to you.

> When adding/removing authorities any notes, belonging to added authorities, which have already been expanded will be collapsed. I'm leaning toward this being acceptable behaviour, would you agree?

Yeah, I'd assume you've already read the notes if you're going back to add another authority.

zarino:
Well well, you learn something new every day. Firefox’s default styles:

    details > summary:first-of-type {
        display: list-item;
    }

Chrome and Safari’s default styles:

    summary {
        display: block;
    }

I guess a quick fix in our case, would be:

    batch-builder__authority-list__authority__notes > summary {
        display: initial;
    }

Which will set the `summary` to `list-item` in Firefox, and `block` in what looks like most other browsers.

PR #4520, opened 2018-02-12
Use low BCrypt cost in specs

garethrees:
We don't need to do really expensive hashing while running specs, so
this should reduce the spec time a bit.

The controller specs seemed to go from ~17 minutes to ~12 minutes with
this change when I tried it, but I was doing other things at the time so
it wasn't a totally scientific test.

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -128,6 +128,8 @@ RSpec.configure do |config|
         File.write(oink_log, '')
       end
     end
+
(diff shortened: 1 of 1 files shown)


garethrees:
Looks like a similar reduction of time from a [recent build](https://travis-ci.org/mysociety/alaveteli/builds/339566704).

Ruby 2.3.3: 26 mins down to 19.

PR #4522, opened 2018-02-13
Ignore hostname resolution error during setup

zarino:
For some reason, when `script/install-as-user` is run on my machine (as part of the Vagrant provisioning), the hostname check failed, resulting in the setup being exited.

I had to comment out the `exit 1` line here, and also in the copy of `install-site.sh` in `commonlib`, just to get Alaveteli installed.

It could be some weird bug in my local network setup. But it did make us wonder *why* this hostname check is even performed. Could it just be removed entirely?


zarino:
See also https://github.com/mysociety/commonlib/pull/60 for the change to commonlib.

dracos:
It was removed in FMS in mysociety/fixmystreet@ffbec97ee1b821119646df0ddd696c6317b9fa4b

PR #4528, opened 2018-02-20
Refactor find by tag

garethrees:
Adds specs for `.find_by_tag` and then refactors so that it uses less Ruby and returns a chainable `ActiveRecord::Relation`. See commit messages for more details.

### DIFF
--- a/lib/has_tag_string/has_tag_string.rb
+++ b/lib/has_tag_string/has_tag_string.rb
@@ -143,11 +143,28 @@ module HasTagString
   module ClassMethods
     # Find all public bodies with a particular tag
     def find_by_tag(tag_as_string)
-      return HasTagStringTag.where(:name => tag_as_string,
-                                   :model => self.to_s).
-                                map { |t| t.tagged_model }.
-                                  sort { |a,b| a.name <=> b.name }.
-                                    uniq
+      join_sql = <<-EOF.strip_heredoc.squish
+      LEFT JOIN has_tag_string_tags
+      ON has_tag_string_tags.model = '#{ to_s }'
+      AND has_tag_string_tags.model_id = #{ table_name }.id
+      EOF
+
+      search =
+        joins(join_sql).
+        where(has_tag_string_tags: { name: tag_as_string, model: to_s }).
+        references(:has_tag_string_tags)
+
+      ordered =
+        if respond_to?(:translated_locales)
+          search.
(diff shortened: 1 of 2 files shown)


lib/has_tag_string/has_tag_string.rb
```diff
@@ -143,11 +143,17 @@ def add_tag_if_not_already_present(tag_as_string)
...
+      join_sql = <<-EOF.strip_heredoc.squish
+      LEFT JOIN has_tag_string_tags
+      ON has_tag_string_tags.model = '#{ to_s }'
+      AND has_tag_string_tags.model_id = #{ table_name }.id
+      EOF
+
+      joins(join_sql).
+      where(has_tag_string_tags: { name: tag_as_string, model: to_s }).
+      references(:has_tag_string_tags).
+      order("#{ table_name }.name ASC").
```
lizconlan: The disadvantage of moving the data sort to SQL is that this method now needs to be aware of Globalize's `_translations` columns (for e.g. `PublicBody` - which is why the model specs are failing) when `name` is a translated field.

My initial stab at fixing this is pretty clumsy (I've fallen into the trap of creating an extra variable assignment rather than duplicating the SQL query) and, arguably, the Class knows too much about the translation implementation - there may be a better way hidden in the Globalize methods.

```diff
def find_by_tag(tag_as_string)
  join_sql = <<-EOF.strip_heredoc.squish
  LEFT JOIN has_tag_string_tags
  ON has_tag_string_tags.model = '#{ to_s }'
  AND has_tag_string_tags.model_id = #{ table_name }.id
  EOF

- joins(join_sql).
+ results = joins(join_sql).
    where(has_tag_string_tags: { name: tag_as_string, model: to_s }).
    references(:has_tag_string_tags).
-   order("#{ table_name }.name ASC").
    uniq

+  if self.column_names.include?('name')
+    results.order("#{table_name}.name ASC")
+  else
+    results.
+      includes(:translations).
+      references(:translations).
+      order("#{table_name.singularize}_translations.name ASC")
+  end
end
```

garethrees:
Argh good spot. Sorry; totally should have waited for the specs to run.

Pushed some changes that check for translated models. I've reordered the commits so that they make more sense (1 fixup for the specs, 1 fixup for the implementation) so, don't rely on the GH ordering.

PR #4544, opened 2018-03-06
[WIP] Improve public authority search

### LINKED ISSUE #4426: Authority search results don't prioritise direct matches to quoted search terms
dq6n_7gx0aaodtb

lizconlan:
## What does this do?

Creates a new [**term prefix**](http://getting-started-with-xapian.readthedocs.io/en/latest/concepts/indexing/terms.html#term-prefixes) in the Xapian index which stores the public body name as a single term so that we can attempt to directly match it with (lowercased) user input, [weighting the term](https://trac.xapian.org/wiki/FAQ/ExtraWeight) in the search results via an optional subclause so that a direct match should float to the top (the 1.2 magic number - intended to read as "boost this by 20%" - is a guess but seems to have worked) without negatively affecting the original search if no match is found.

## Why was it needed?

To prevent a direct match on the public body (authority) name from being buried in the search results (as currently) without losing the ability to find e.g. a previous name for that authority in the notes field. Xapian's relevancy score is based on [**wdf**](https://xapian.org/docs/glossary.html) (Within-document frequency) of [**terms**](http://getting-started-with-xapian.readthedocs.io/en/latest/concepts/indexing/terms.html) so a search for "Cardiff council" is (crudely speaking) split into term/keyword searches for "cardiff" and "council" so without the ability to favour a direct match on name, if the notes field contains these words multiple times, it will appear higher in the search results than a user would expect. (We sort on relevancy by default.)

## Notes to reviewer
Work in progress - Among other things, currently lacks integration (probably) specs to show the direct match and weighting in action. 

Feels like a bit of a hack at the moment, hard to reuse. Initial thoughts: might be useful if the code could be extracted either somehow into `ActsAsXapian` itself or per `lib/xapian_queries.rb` to build (and join?) the query subclause, passing in the string (and prefix?) with an optional weight value with a sensible default. 

To test in the browser, you will need to update the index for either a strategic subset or all of your PublicBody records to get the new term prefix data set up then you should get improved search results in:

The select authority (start of a new request) screen: http://10.10.10.30:3000/select_authority?utf8=✓&query=cardiff+council&bodies=1&commit=Search

The general search (particularly when restricted to authorities): http://10.10.10.30:3000/search/cardiff%20council/bodies?utf8=✓&query=cardiff+council

## Related ticket(s)

Fixes #4426

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -425,7 +425,7 @@ class ApplicationController < ActionController::Base
   end
 
   # Function for search
-  def perform_search(models, query, sortby, collapse, per_page = 25, this_page = nil)
+  def perform_search(models, query, sortby, collapse, per_page = 25, this_page = nil, user_query = nil)
     @query = query
     @sortby = sortby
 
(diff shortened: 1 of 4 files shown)


garethrees:
> you will need to update the index for either a strategic subset or all of your PublicBody records

For reference, this only took 4 minutes:

```
rake xapian:rebuild_index models="PublicBody" verbose="true" texts="false" values="false"
```

garethrees:
> you will need to update the index for either a strategic subset or all of your PublicBody records

For reference, rebuilding them all (live export) only took 5 or so minutes:

```
rake xapian:rebuild_index models="PublicBody" verbose="true"

```

garethrees:
Superseded by https://github.com/mysociety/alaveteli/pull/4579

PR #4550, opened 2018-03-12
Make browse-mode categories clearer

garethrees:
Adds a border between categories so that each element is clearer.

**Before**

screen shot 2018-03-12 at 14 59 55

**After**

screen shot 2018-03-12 at 14 59 34


wrightmartin:
This is fine, helps a lot especially when everything is minimized and you just see the list of top level categories  👍

PR #4554, opened 2018-03-14
[#4426] Prioritise direct name matches

### LINKED ISSUE #4426: Authority search results don't prioritise direct matches to quoted search terms

garethrees:
Prioritise Xapian N terms for PublicBody

This boosts the WDF increment for terms with the "N" prefix – namely
`PublicBody#name`, prioritising direct matches when a user searches for
a public body by name.

Xapian gives the option of doing this either at index time or at search
time [1].

@lizconlan attempted the latter [2], but it resulted in having to jam
lots of extra logic in to our search methods. Given indexing WDTK's
public bodies only takes around 5 minutes (for ~23K records) it seems
more efficient to do this at index time. It also results in much less
code changed.

The actual weighting is quite strange. The example [1] seems to suggest
that you want a lower termpos and a higher WDF increment for a
prioritised term. I'm not entirely sure what the interplay between the
term position and WDF increment is. For future reference, I'm including
links to `add_posting` [3] and `add_term` [4].

Note that the specs assert on `PublicBody#name` instead of comparing
object equality because the rspec diff isn't very helpful now that name
isn't an attribute on `PublicBody` directly (its on
`PublicBody::Translation`).

Aside: Ideally we'd have used the standard `S` term prefix for
"subject/title" [5], but that prefix is already in use in
`InfoRequestEvent`. We can create our own prefix, but using a standard
would help reduce the complexity of what we give weight to if we applied
weighting to other indexed models:

    # InfoRequestEvent
    acts_as_xapian terms: [ [:request_title, 'S', 'title'] ]

    # User
    acts_as_xapian terms: [ [:name, 'S', 'name'] ]

[1] https://trac.xapian.org/wiki/FAQ/ExtraWeight
[2] #4544
[3]
https://xapian.org/docs/apidoc/html/classXapian_1_1Document.html#a161943bb133fea0da4a21fa37621764b
[4]
https://xapian.org/docs/apidoc/html/classXapian_1_1Document.html#a730bb59dda98f19c61fe67360f0adb3a
[5] https://xapian.org/docs/omega/termprefixes.html

Fixes #4426

### DIFF
--- a/doc/CHANGES.md
+++ b/doc/CHANGES.md
@@ -2,6 +2,8 @@
 
 ## Highlighted Features
 
+* Prioritise direct matches on `PublicBody#name` in search results (Liz Conlan,
+  Gareth Rees)
 * Log an `InfoRequestEvent` when updating response handling attributes in
   `InfoRequest.stop_new_responses_on_old_requests` (Gareth Rees)
 * Show that a request is part of a batch on the request page in the admin
--- a/doc/CHANGES.md
+++ b/doc/CHANGES.md
@@ -31,7 +31,7 @@
   use `InfoRequest.last_event`.
 * You'll need to reindex your public bodies to benefit from the improved direct
   match results:
-  `bundle exec xapian:rebuild_index models="PublicBody" verbose="true"`
+  `bundle exec rake xapian:rebuild_index models="PublicBody" verbose="true"`
 
 ### Changed Templates
 
--- a/doc/CHANGES.md
+++ b/doc/CHANGES.md
@@ -29,6 +29,9 @@
 * We've removed the spring preloader as it didn't seem to provide much benefit.
 * `InfoRequest.get_last_event` is deprecated and will be removed in 0.33. Please
   use `InfoRequest.last_event`.
+* You'll need to reindex your public bodies to benefit from the improved direct
+  match results:
+  `bundle exec xapian:rebuild_index models="PublicBody" verbose="true"`
 
 ### Changed Templates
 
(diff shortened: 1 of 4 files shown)


doc/CHANGES.md
```diff
@@ -29,6 +29,9 @@
 * We've removed the spring preloader as it didn't seem to provide much benefit.
 * `InfoRequest.get_last_event` is deprecated and will be removed in 0.33. Please
   use `InfoRequest.last_event`.
+* You'll need to reindex your public bodies to benefit from the improved direct
+  match results:
+  `bundle exec xapian:rebuild_index models="PublicBody" verbose="true"`
```
gbp: Missing `rake` in the command

lib/acts_as_xapian/acts_as_xapian.rb
```diff
@@ -936,9 +936,17 @@ def xapian_index(terms = true, values = true, texts = true)
         if value.kind_of?(Array)
           for v in value
             doc.add_term(term[1] + v)
+
+            if term[1] == 'N'
+              doc.add_posting(term[1] + v, 1, 50)
+            end
```
gbp: Feel a bit uneasy about hard coding the `N` and `50` here. Its probably okay as this library is included in the repo. I took a look at how we could do this differently and came up with: https://gist.github.com/gbp/800309a5349ebbdcfba7ef1d58cb795b which seems to work. Thoughts?

garethrees:
Superseded by https://github.com/mysociety/alaveteli/pull/4579

PR #4579, opened 2018-03-22
4426 prioritise direct name matches 2

garethrees:


garethrees:
Merged in 010a2f1bb3250c795bf92abf4d85c0d1f8c08384

PR #4580, opened 2018-03-22
TEMP: Ignore any errors in idexing

garethrees:
Just output the problems to STDERR – can tee that to a log then and deal
with later

### DIFF
--- a/lib/acts_as_xapian/acts_as_xapian.rb
+++ b/lib/acts_as_xapian/acts_as_xapian.rb
@@ -680,6 +680,9 @@ module ActsAsXapian
       # this can happen if the record was hand deleted in the database
       job.action = 'destroy'
       retry
+    rescue Exception => e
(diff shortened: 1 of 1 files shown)


lib/acts_as_xapian/acts_as_xapian.rb
```diff
@@ -680,6 +680,9 @@ def self.run_job(job, flush, verbose)
       # this can happen if the record was hand deleted in the database
       job.action = 'destroy'
       retry
+    rescue Exception => e
+      STDERR.puts("ERROR: ActsAsXapian.rebuild_index      #{model_class} #{model.id}")
```
garethrees: `model_class` doesn't exist here

garethrees:
```diff
- STDERR.puts("ERROR: ActsAsXapian.rebuild_index      #{model_class} #{model.id}")
+ STDERR.puts("ERROR: ActsAsXapian.rebuild_index      #{job.model} #{model.id}")
```

PR #4581, opened 2018-03-23
Rename and warn about dangerous Xapian commands

garethrees:
I had thought `rebuild_index` was effectively saying “mark each record
for reindex and then reindex it” rather than “destroy the index, then
reindex everything”.

This should be _really_ clear at the point of running the command.

### DIFF
--- a/config/deploy.rb
+++ b/config/deploy.rb
@@ -42,9 +42,9 @@ end
 
 # Not in the rake namespace because we're also specifying app-specific arguments here
 namespace :xapian do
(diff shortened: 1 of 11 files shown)


garethrees:
> We'll probably want to fix the Changelog where we're promoting the old command to be run.

Its expected that you upgrade Alaveteli releases one at a time, so those commands make sense under those versions of Alaveteli. You'll need to use the old name of the command if you're running any version before this gets released, so I think leaving it as is is better. Good spot though 👍

PR #4582, opened 2018-03-23
Add task to add all model records to reindex queue

garethrees:
Sometimes we want to reindex everything. This allows you to add all
records of specified models to the queue for reindexing.


garethrees:
Superseded by https://github.com/mysociety/alaveteli/pull/4585

PR #4591, opened 2018-03-27
Rescue all errors when indexing

garethrees:
When we hit a record that can't be indexed becauase of an error, we:

1. Get cron output every 5 minutes as the next run of the indexing
   script continuously tries to index the record again.
2. Prevent indexing of other records, because the script is exiting at
   the point of error.

This commit fixes point 2.

We rescue from `StandardError` so that we just output any problems to
STDERR – we'll get a notification message for records that error, but
other records will get indexed with no problem.

We'll continue to receive cron error output every 5 minutes as the
script will run over the job still left in the queue. We can remove the
job from the indexing queue if we want to silence the errors until the
next time it gets marked for reindexing.

Originally I'd resuced from `Exception`, but this is a code smell.
[1][2]

We were debating keeping this as there's not much we don't want to
rescue from, but `StandardError` _seems_ to be high enough in the
hierarchy to catch most things, so lets give it a go like this first.

[1] https://robots.thoughtbot.com/rescue-standarderror-not-exception
[2] https://stackoverflow.com/a/10048406/387558

### DIFF
--- a/lib/acts_as_xapian/acts_as_xapian.rb
+++ b/lib/acts_as_xapian/acts_as_xapian.rb
@@ -680,6 +680,10 @@ module ActsAsXapian
       # this can happen if the record was hand deleted in the database
       job.action = 'destroy'
       retry
+    rescue StandardError => e
+      method = 'ActsAsXapian.run_job'
+      STDERR.puts("ERROR: #{method}      #{job.model} #{model.id}")
+      STDERR.puts(e)
     end
     if flush
       ActsAsXapian.writable_db.flush
@@ -727,7 +731,13 @@ module ActsAsXapian
         STDOUT.puts("ActsAsXapian.rebuild_index: Rebuilding #{model_class.to_s}") if verbose
         model_class.find_each do |model|
           STDOUT.puts("ActsAsXapian.rebuild_index      #{model_class} #{model.id}") if verbose
-          model.xapian_index(terms, values, texts)
+          begin
+            model.xapian_index(terms, values, texts)
+          rescue StandardError => e
+            method = 'ActsAsXapian.rebuild_index'
+            STDERR.puts("ERROR: #{method}      #{model_class} #{model.id}")
+            STDERR.puts(e)
+          end
         end
       end
       ActsAsXapian.writable_db.flush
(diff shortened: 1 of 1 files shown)


garethrees:
`ActsAsXapian` [suggests](https://github.com/mysociety/alaveteli/blob/develop/lib/acts_as_xapian/acts_as_xapian.rb#L653-L654) it already has this behaviour. This might be something to look in to. Kinda hard given it has no tests though :(

garethrees:
Maybe we just want this:

```diff
diff --git a/lib/acts_as_xapian/acts_as_xapian.rb b/lib/acts_as_xapian/acts_as_xapian.rb
index 9ac1b60d3..7c4bff0bc 100644
--- a/lib/acts_as_xapian/acts_as_xapian.rb
+++ b/lib/acts_as_xapian/acts_as_xapian.rb
@@ -637,18 +637,17 @@ module ActsAsXapian
       job = nil
       begin
         ActiveRecord::Base.transaction do
-          begin
-            job = ActsAsXapianJob.lock(true).find(id)
-          rescue ActiveRecord::RecordNotFound => e
-            # This could happen if while we are working the model
-            # was updated a second time by another process. In that case
-            # ActsAsXapianJob.delete_all in xapian_mark_needs_index below
-            # might have removed the first job record while we are working on it.
-            #STDERR.puts("job with #{id} vanished under foot") if verbose
-            next
-          end
+          job = ActsAsXapianJob.lock(true).find(id)
           run_job(job, flush, verbose)
         end
+      rescue ActiveRecord::RecordNotFound => e
+        # This could happen if while we are working the model
+        # was updated a second time by another process. In that case
+        # ActsAsXapianJob.delete_all in xapian_mark_needs_index below
+        # might have removed the first job record while we are working on it.
+
+        #STDERR.puts("job with #{id} vanished under foot") if verbose
+        next
       rescue => detail
         # print any error, and carry on so other things are indexed
         STDERR.puts(detail.backtrace.join("\n") + "\nFAILED ActsAsXapian.update_index job #{id} #{$!} " + (job.nil? ? "" : "model " + job.model + " id " + job.model_id.to_s))
```

garethrees:
I think this is superseded by https://github.com/mysociety/alaveteli/pull/4592, so I'm going to close it unless we decide otherwise.

PR #4592, opened 2018-03-27
Improved handling of failed Xapian indexing

garethrees:
* Add spec demonstrating failed xapian indexing
* Minor refactoring
* Prevent continuous retries of failing index jobs

As far as I can tell, an indexing error _does not_ block indexing for all other jobs. It just means that we retry the failing job over and over again. I think what we should be doing in response to this (when we get a flood of cron emails) is just removing the job from the `acts_as_xapian_jobs` queue, so that its not indexed any more. It might get re-added to the queue later (another response, an edit, etc), but we might have fixed the underlying problem by then. If not, we just remove and bump the ticket that we surely created describing the bug.  Rather than doing this manually, I've introduced the job removal here, so that we only get one email per failed job run, rather than a continuous stream.

See commit messages for details.

### DIFF
--- a/lib/acts_as_xapian/acts_as_xapian.rb
+++ b/lib/acts_as_xapian/acts_as_xapian.rb
@@ -649,7 +649,7 @@ module ActsAsXapian
           end
           run_job(job, flush, verbose)
         end
-      rescue => detail
+      rescue StandardError => detail
         # print any error, and carry on so other things are indexed
         STDERR.puts(detail.backtrace.join("\n") + "\nFAILED ActsAsXapian.update_index job #{id} #{$!} " + (job.nil? ? "" : "model " + job.model + " id " + job.model_id.to_s))
       end
--- a/lib/acts_as_xapian/acts_as_xapian.rb
+++ b/lib/acts_as_xapian/acts_as_xapian.rb
@@ -660,6 +660,13 @@ module ActsAsXapian
             ''
           end
         STDERR.puts("#{backtrace}\n#{msg} #{job_info}")
+      ensure
+        # We never want to reprocess existing jobs.
+        # If it succeeded the first time, it should already be destroyed.
+        # If it failed, then we don't want to keep trying to process it every
+        # cron run – we should create an issue and investigate it, and requeue
+        # the record once there's a fix in place.
+        job.try(:destroy)
       end
     end
     # We close the database when we're finished to remove the lock file. Since writable_init
--- a/lib/acts_as_xapian/acts_as_xapian.rb
+++ b/lib/acts_as_xapian/acts_as_xapian.rb
@@ -651,7 +651,15 @@ module ActsAsXapian
         end
       rescue StandardError => error
         # print any error, and carry on so other things are indexed
-        STDERR.puts(error.backtrace.join("\n") + "\nFAILED ActsAsXapian.update_index job #{id} #{$!} " + (job.nil? ? "" : "model " + job.model + " id " + job.model_id.to_s))
+        backtrace = error.backtrace.join("\n")
+        msg = "FAILED ActsAsXapian.update_index job #{id} #{$!}"
+        job_info =
+          if job
+            "model #{job.model} id #{job.model_id}"
+          else
+            ''
+          end
+        STDERR.puts("#{backtrace}\n#{msg} #{job_info}")
       end
     end
     # We close the database when we're finished to remove the lock file. Since writable_init
--- a/lib/acts_as_xapian/acts_as_xapian.rb
+++ b/lib/acts_as_xapian/acts_as_xapian.rb
@@ -649,9 +649,9 @@ module ActsAsXapian
           end
           run_job(job, flush, verbose)
         end
-      rescue StandardError => detail
+      rescue StandardError => error
         # print any error, and carry on so other things are indexed
-        STDERR.puts(detail.backtrace.join("\n") + "\nFAILED ActsAsXapian.update_index job #{id} #{$!} " + (job.nil? ? "" : "model " + job.model + " id " + job.model_id.to_s))
+        STDERR.puts(error.backtrace.join("\n") + "\nFAILED ActsAsXapian.update_index job #{id} #{$!} " + (job.nil? ? "" : "model " + job.model + " id " + job.model_id.to_s))
       end
     end
     # We close the database when we're finished to remove the lock file. Since writable_init
--- a/lib/acts_as_xapian/acts_as_xapian.rb
+++ b/lib/acts_as_xapian/acts_as_xapian.rb
@@ -611,6 +611,75 @@ module ActsAsXapian
   class ActsAsXapianJob < ActiveRecord::Base
   end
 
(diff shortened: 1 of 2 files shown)


lib/acts_as_xapian/acts_as_xapian.rb
```diff
@@ -611,6 +611,69 @@ def log_description
...
+    def job_info
+      msg = "FAILED ActsAsXapian.update_index job #{ job_id } #{ error.class }"
+      msg += " model #{ job_model }" if job_model
+      msg += " id #{ job_model_id }" if job_model_id
+      msg += '.'
+      msg
+    end
+
+    def error_backtrace
+      error.backtrace.join("\n")
```
gbp: For some strange reason I'm not getting backtraces for exceptions thrown on line [749](https://github.com/mysociety/alaveteli/blob/c5f54072f7fdbdeed2efed22beed23d17824fa30/lib/acts_as_xapian/acts_as_xapian.rb#L749) so this line blows up which is reeeally hard to track down, the full error and stack given was:

```
NoMethodError: undefined method `join' for nil:NilClass
Traceback (most recent call last):
./bin/rake: undefined method `reject' for nil:NilClass (NoMethodError)
```

Probably best to change this line to something like:

`error.backtrace.try(:join, "\n").to_s`

garethrees: Hmm strange, I get:

```
$ script/update-xapian-index
FAILED ActsAsXapian.update_index job 643 RuntimeError model InfoRequestEvent id 900.

This job will be removed from the queue. Once the underlying problem is fixed, manually re-index the model record.

You can do this in a rails console with `InfoRequestEvent.find(900).xapian_mark_needs_index`.

---------
Backtrace
---------

/home/vagrant/alaveteli/lib/acts_as_xapian/acts_as_xapian.rb:749:in `run_job'
/home/vagrant/alaveteli/lib/acts_as_xapian/acts_as_xapian.rb:713:in `block (2 levels) in update_index'
/home/vagrant/bundle/ruby/2.3.0/gems/activerecord-4.2.10/lib/active_record/connection_adapters/abstract/database_statements.rb:213:in `block in transaction'
/home/vagrant/bundle/ruby/2.3.0/gems/activerecord-4.2.10/lib/active_record/connection_adapters/abstract/transaction.rb:184:in `within_new_transaction'
/home/vagrant/bundle/ruby/2.3.0/gems/activerecord-4.2.10/lib/active_record/connection_adapters/abstract/database_statements.rb:213:in `transaction'
/home/vagrant/bundle/ruby/2.3.0/gems/activerecord-4.2.10/lib/active_record/transactions.rb:220:in `transaction'
/home/vagrant/alaveteli/lib/acts_as_xapian/acts_as_xapian.rb:702:in `block in update_index'
/home/vagrant/alaveteli/lib/acts_as_xapian/acts_as_xapian.rb:699:in `each'
/home/vagrant/alaveteli/lib/acts_as_xapian/acts_as_xapian.rb:699:in `update_index'
/home/vagrant/alaveteli/lib/acts_as_xapian/tasks/xapian.rake:12:in `block (2 levels) in <top (required)>'
/home/vagrant/bundle/ruby/2.3.0/gems/rake-12.1.0/lib/rake/task.rb:251:in `block in execute'
/home/vagrant/bundle/ruby/2.3.0/gems/rake-12.1.0/lib/rake/task.rb:251:in `each'
/home/vagrant/bundle/ruby/2.3.0/gems/rake-12.1.0/lib/rake/task.rb:251:in `execute'
/home/vagrant/bundle/ruby/2.3.0/gems/rake-12.1.0/lib/rake/task.rb:195:in `block in invoke_with_call_chain'
/usr/lib/ruby/2.3.0/monitor.rb:214:in `mon_synchronize'
/home/vagrant/bundle/ruby/2.3.0/gems/rake-12.1.0/lib/rake/task.rb:188:in `invoke_with_call_chain'
/home/vagrant/bundle/ruby/2.3.0/gems/rake-12.1.0/lib/rake/task.rb:181:in `invoke'
/home/vagrant/bundle/ruby/2.3.0/gems/rake-12.1.0/lib/rake/application.rb:153:in `invoke_task'
/home/vagrant/bundle/ruby/2.3.0/gems/rake-12.1.0/lib/rake/application.rb:109:in `block (2 levels) in top_level'
/home/vagrant/bundle/ruby/2.3.0/gems/rake-12.1.0/lib/rake/application.rb:109:in `each'
/home/vagrant/bundle/ruby/2.3.0/gems/rake-12.1.0/lib/rake/application.rb:109:in `block in top_level'
/home/vagrant/bundle/ruby/2.3.0/gems/rake-12.1.0/lib/rake/application.rb:118:in `run_with_threads'
/home/vagrant/bundle/ruby/2.3.0/gems/rake-12.1.0/lib/rake/application.rb:103:in `top_level'
/home/vagrant/bundle/ruby/2.3.0/gems/rake-12.1.0/lib/rake/application.rb:81:in `block in run'
/home/vagrant/bundle/ruby/2.3.0/gems/rake-12.1.0/lib/rake/application.rb:179:in `standard_exception_handling'
/home/vagrant/bundle/ruby/2.3.0/gems/rake-12.1.0/lib/rake/application.rb:78:in `run'
/home/vagrant/bundle/ruby/2.3.0/gems/rake-12.1.0/exe/rake:27:in `<top (required)>'
/home/vagrant/bundle/ruby/2.3.0/bin/rake:22:in `load'
/home/vagrant/bundle/ruby/2.3.0/bin/rake:22:in `<main>'
```

To get that, I did:

```sh
bin/rails runner "ActsAsXapian::ActsAsXapianJob.create(model: 'InfoRequestEvent', model_id: InfoRequestEvent.last.id, action: 'wjheebwc')"
script/update-xapian-index
```

gbp: Difference seems to be cause by Ruby versions; 2.3.x I get the same as you, 2.5.x I get what I did yesterday.

Only relevant change in 2.5 I know about is exception backtraces are now printed in reverse order - doesn't seem like that would cause this problem.

Installing 2.4 now.

gbp: 2.4.x has a bracktrace

garethrees: Yeah, seems like [`backtrace`](http://ruby-doc.org/core-2.5.0/Exception.html#method-i-backtrace) is still a method on `Exception `.

Also, looks like you can [`set_backtrace`](http://ruby-doc.org/core-2.0.0/Exception.html#method-i-set_backtrace) rather than stubbing it out. TIL.

gbp: So seems to be an issue with 2.5, with Rails 4.2:

I've got a reduced test case proving an issue:

```ruby
namespace :temp do
  task :foo do
    raise 'foo'
  end
end
```

This throws a similar error.  But if in `Rakefile`, if I remove the Rails' `require` and its `load_tasks` the task works.

PR #4594, opened 2018-03-28
[pro#501] Destroy embargoes with their info request

garethrees:
There's no reason to keep an `Embargo` around after its `InfoRequest`
has been destroyed. This came up when writing code to reindex requests
with an attached `Embargo`. I had to add `try` in case the `Embargo`
returned `nil` when asking for the `info_request`:

    AlaveteliPro::Embargo.find_each do |embargo|
      embargo.info_request.try(:reindex_request_events)
    end

I'd expect there to always be an `InfoRequest` when doing something like
this.

Fixes https://github.com/mysociety/alaveteli-professional/issues/501

### DIFF
--- a/spec/models/info_request_spec.rb
+++ b/spec/models/info_request_spec.rb
@@ -1296,6 +1296,13 @@ describe InfoRequest do
       info_request.destroy
       expect(UserInfoRequestSentAlert.where(:info_request_id => info_request.id)).to be_empty
     end
+
+    it 'destroys associated embargoes' do
+      AlaveteliPro::Embargo.destroy_all
+      FactoryGirl.create(:embargo, info_request: info_request)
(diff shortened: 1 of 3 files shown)


spec/models/info_request_spec.rb
```diff
@@ -1296,6 +1296,13 @@
       info_request.destroy
       expect(UserInfoRequestSentAlert.where(:info_request_id => info_request.id)).to be_empty
     end
+
+    it 'destroys associated embargoes' do
+      FactoryGirl.create(:embargo, info_request: info_request)
+      info_request.destroy
```
gbp: This isn't actually testing the record is removed from the DB. Could rewrite as:

`expect { info_request.destroy }.to change(AlaveteliPro::Embargo, :count).by(-1)`

Issue #4541, opened 2018-03-06
Drop support for Ruby 2.0


garethrees:
Official support ended on 2016-02-24

https://www.ruby-lang.org/en/news/2016/02/24/support-plan-of-ruby-2-0-0-and-2-1/

PR #4598, opened 2018-04-04
Faster application booting

gbp:
* What does this do?
Using the Bootsnap gem to improve initial app loading by ~50%

* Why was this needed?
Moar speed and faster tests runs

* Notes to reviewer
Not tested to see if this works on Vagrant or not. Can't see why it wouldn't work tho.


gbp:
Looks like Bootsnap is incompatible with Ruby 2.0. We need wait for #4541 first.

PR #4605, opened 2018-04-11
[#4586] Add tasks to restore from Xapian backup

### LINKED ISSUE #4586: Compile various scripts used to sync older xapian index to current state into reusable form
Could be useful for any larger Alaveteli instance that loses its index and needs to sync a backup

garethrees:

* Relevant issue(s)

Fixes #4586

* What does this do?

Adds some rake tasks that have been ported from the [scripts](https://gist.github.com/garethrees/e9d436f1c506ab801fa622a27917799e#file-remove_deleted-rb) we used to restore from a Xapian database backup.

* Why was this needed?
We should

* Implementation notes
* Screenshots
* Notes to reviewer


garethrees:
@gbp and I spoke about this yesterday and decided its not worth a load of effort extracting in to classes with tests, so instead I added some info about it being beta.

gbp:
@garethrees I've pushed some commits to [4586-restore-from-xapian-backup-gbp](https://github.com/mysociety/alaveteli/tree/4586-restore-from-xapian-backup-gbp) which adds a basic spec which can be used to test this out.

I attempted to implement a scenario where a Public Body name has been changed but the specs fail and I think that is due to the rake task not reindexing based on the `updated_at` timestamp. Do you think this might be an issue?

garethrees:
> @garethrees I've pushed some commits to 4586-restore-from-xapian-backup-gbp which adds a basic spec which can be used to test this out.

Nice

> I attempted to implement a scenario where a Public Body name has been changed but the specs fail and I think that is due to the rake task not reindexing based on the updated_at timestamp. Do you think this might be an issue?

We didn't use `updated_at` because of https://github.com/mysociety/alaveteli/issues/4637. I think I've figured out fixing this, so might be able to make the switch here.

garethrees:

PR #4606, opened 2018-04-11
Add updated at to all tables

garethrees:
Part 1 of https://github.com/mysociety/alaveteli/issues/4614.

Going to ignore the Hound comments here:

* Missing top-level class documentation comment. – Don't really need these on migrations
* Line is too long. – Its a temp task, and I think it looks cleaner in this case where we've got lots of sequential blocks.

- [x] Update migration timestamps now that the `db:rollback` fixes have been merged, as these appear _before_ that migration.

### DIFF
--- a/db/migrate/20171207140915_create_announcements.rb
+++ b/db/migrate/20171207140915_create_announcements.rb
@@ -1,3 +1,4 @@
+# -*- encoding : utf-8 -*-
 class CreateAnnouncements < ActiveRecord::Migration
   def change
     create_table :announcements do |t|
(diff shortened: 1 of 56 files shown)


db/migrate/20171207140915_create_announcements.rb
```diff
@@ -1,3 +1,4 @@
+# -*- encoding : utf-8 -*-
```
gbp: Since [Ruby 2.0](https://www.ruby-lang.org/en/news/2013/02/24/ruby-2-0-0-p0-is-released/) these lines aren't needed:

> The UTF-8 default encoding, which make many magic comments omissible

We should consider enabling the `Style/Encoding` cop which now warns when these magic comments are present.

PR #4608, opened 2018-04-12
Make it quicker to ban users in admin interface

garethrees:
Adds a controller to handle account suspensions. It sets a default
suspension message, but the main use right now is to ban for spamming.

I've added buttons to the user list and user show page to quickly set a
"Banned for spamming" message.

This could later be extended with an
`AdminUsersAccountSuspensionsController#destroy` to unban users (reset
`User#ban_text` to `''`), or further buttons could be added with other
common suspension reasons.

This was prompted by an influx of "tech support" spam accounts and
having to manually copy/paste the ban text in each.

* Implementation notes

I wrote this quickly before tea… it might not be quite up to usual standards.

* Screenshots

quicker-banning-spam-users

### DIFF
--- a/app/controllers/admin_users_account_suspensions_controller.rb
+++ b/app/controllers/admin_users_account_suspensions_controller.rb
@@ -1,0 +1,31 @@
+# -*- encoding : utf-8 -*-
+# app/controllers/admin_users_account_suspensions_controller.rb:
+# Controller for suspending user accounts
+#
+# Copyright (c) 2018 UK Citizens Online Democracy. All rights reserved.
+# Email: hello@mysociety.org; WWW: http://www.mysociety.org/
+class AdminUsersAccountSuspensionsController < AdminController
+  before_filter :set_banned_user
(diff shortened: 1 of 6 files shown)


app/controllers/admin_users_account_suspensions_controller.rb
```diff
@@ -0,0 +1,32 @@
...
+#
+# Copyright (c) 2017 UK Citizens Online Democracy. All rights reserved.
+# Email: hello@mysociety.org; WWW: http://www.mysociety.org/
+
+class AdminUsersAccountSuspensionsController < AdminController
+  before_filter :set_banned_user
+  before_filter :set_suspension_reason
+
+  def create
+    if @banned_user.update(ban_text: @suspension_reason)
```
garethrees: Should also clear profile photo

gbp: Yeah, this is happening currently. If you want to get this merge without this for now it is fine

PR #4620, opened 2018-04-19
[pro#477] Delist unrequestable authorities from Batch

garethrees:
At the moment its possible to send a batch to authorities that no longer
exist, where FOI doesn't apply, or where we don't have an email address.
This is made worse by the new Batch Browse mode. For now, the easiest
option is to just prevent listing those authorities

* Exclude `not_apply` authorities from batch searches, as its currently
  too complicated to explain that you can make a request to them, but
  they don't have to answer.
* Extract the `exclude_tags` to a variable and switch to new-style Hash
  syntax.
* Skip rendering results if the authority is not requestable. This
  handles authorities with missing email addresses. In the normal
  interface we point this out, but the user doesn't get told about this
  until their batch sending has completed. Ideally we'd do this at
  search time, but this is currently the best option.

Fixes https://github.com/mysociety/alaveteli-professional/issues/477.

**Setup:**

screen shot 2018-04-19 at 16 31 24

**Before:**

screen shot 2018-04-19 at 16 13 52

We already do this in Browse mode:

screen shot 2018-04-19 at 16 32 43

…except that when we hit "Add all", we include unrequestables:

screen shot 2018-04-19 at 16 48 40

**After:**

screen shot 2018-04-19 at 16 23 15

screen shot 2018-04-19 at 16 25 59

screen shot 2018-04-19 at 16 50 15

### DIFF
--- a/app/controllers/alaveteli_pro/batch_request_authority_searches_controller.rb
+++ b/app/controllers/alaveteli_pro/batch_request_authority_searches_controller.rb
@@ -15,8 +15,10 @@ class AlaveteliPro::BatchRequestAuthoritySearchesController < AlaveteliPro::Base
   def search
     # perform_seach sets @query but typeahead_search doesn't
     @query = params[:authority_query] || ""
-    @search = typeahead_search(@query, { :model => PublicBody,
-                                         :exclude_tags => [ 'defunct' ] })
+    excluded_tags = %w(defunct not_apply)
+    @search = typeahead_search(@query, model: PublicBody,
+                                       exclude_tags: excluded_tags)
+
     unless @search.blank?
       @result_limit = calculate_result_limit(@search)
       check_page_limit!(@page, @per_page)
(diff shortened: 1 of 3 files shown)


gbp:
Here are some Public Body scopes which should help with this. Think I've fixed the issues we identified with this on Friday - where results would be returned if the were tagged with `defunct`/`not_apply` and any other tag.

```
  scope :joins_tags, -> {
    # provides an PG array of tag names which can be queried in SQL where clause
    joins(
      'LEFT JOIN LATERAL (' \
        'SELECT ARRAY_AGG(t."name") AS names ' \
        'FROM public_bodies AS pb ' \
        'LEFT OUTER JOIN "has_tag_string_tags" t ON t."model_id" = pb."id" ' \
          'AND t."model" = \'PublicBody\' ' \
        'WHERE pb.id = public_bodies.id' \
      ') tags ON TRUE'
    )
  }

  scope :requestable, -> {
    joins_tags.
      where.not(%q("tags"."names" @> ARRAY['defunct'])).
      where.not(%q("tags"."names" @> ARRAY['not_apply'])).
      where.not(request_email: [nil, 'blank'])
  }
```

PR #4627, opened 2018-04-23
[#4618] Add spam checking to user names

### LINKED ISSUE #4618: Add ability to check for spam name terms in UserSpamScorer
We get lots of similar signups with "X tech support" in the name.

It would be easier to catch these if we could add a value or regexp to an array that names get filtered through.

I think we need to:

* Add a `DEFAULT_SPAM_NAME_FORMATS`, or `DEFAULT_SPAM_NAME_TERMS`.
* Make this configurable as a class attribute.
* Add a method similar to `about_me_is_spam_format?` to check against the name.
* Add the check to the `DEFAULT_SCORE_MAPPINGS` – I think quite a high score for this.

I think it would be worth using a regexp rather than terms, as they make it a bit easer to catch a larger variety of spam, but I could be convinced either way.

I'd be hesitant to suggest sharing the spam terms across `about_me` and `name`. I think this might be more problematic for false positives in `about_me` if we add the simple terms that tend to appear in names.

garethrees:
* Relevant issue(s)

Fixes #4618

* What does this do?

Add spam checking to user names

* Why was this needed?

Help catch spam influxes of similar spam

* Implementation notes

Used existing pattern from `about_me` checking

### DIFF
--- a/doc/CHANGES.md
+++ b/doc/CHANGES.md
@@ -2,6 +2,7 @@
 
 ## Highlighted Features
 
+* Add user name spam checking (Gareth Rees)
 * Make it quicker to ban users for spamming in admin interface (Gareth Rees)
 * Limit the frequency that `PublicBody#updated_at` gets updated by unrelated
   changes to an associated `InfoRequest` (Gareth Rees)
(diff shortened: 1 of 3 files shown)


garethrees:
Should also add a match for bitcoin.

PR #4654, opened 2018-05-03
Remove vertical align declaration on follow buttons 

wrightmartin:
* Relevant issue(s)
https://github.com/mysociety/whatdotheyknow-theme/pull/465

* What does this do?
Changes vertical alignment of follower count. It appears it was set to `top` to make the text align on the baseline, which looks nice but is largely useless.
* Why was this needed?
Without this styling this component is unecessarily complex

### DIFF
--- a/app/assets/stylesheets/responsive/_global_layout.scss
+++ b/app/assets/stylesheets/responsive/_global_layout.scss
@@ -179,7 +179,7 @@ textarea{
 .action-bar__follow,
 .action-bar__follow-button,
 .action-bar__follower-count {
- vertical-align: top;
+ vertical-align: middle;
  font-size: 16px;
  margin: 1em 0;
  @include respond-min( 30em ){


garethrees:
I don't really understand this change I'm afraid.

Relevant issue(s) mysociety/whatdotheyknow-theme#465 is a PR, not a description of a problem.

The only difference I can make out is that when no theme is applied, there's a slight alignment difference.

**Before:**

screen shot 2018-05-17 at 13 51 02


**After:**

screen shot 2018-05-17 at 13 51 11

In the After case, "Follow" is a px or two lower.

I'm not sure how we're reducing complexity here either, given we're just changing an argument.

I'm not against merging this, but I don't understand why the original was "useless"

wrightmartin:
> I'm not against merging this, but I don't understand why the original was "useless"

With a theme on, it's pushing the border around the follow count below the button . So for a very minor gain in un-themed alaveteli it's causing issues for theming. Rather than reset the v-align in the theme it made sense to me to remove it from the core altogether

screen shot 2018-05-17 at 14 08 27

garethrees:
But https://github.com/mysociety/whatdotheyknow-theme/pull/465 seemed to fix it in the theme, or will this fix it for other themes?

wrightmartin:
> But mysociety/whatdotheyknow-theme#465 seemed to fix it in the theme, or will this fix it for other themes?

It fixed that bug yeah, but you're correct, this should fix it in other themes (or prevent it from happening in new ones)

garethrees:
Thanks, merged in 619e423d0

PR #4678, opened 2018-06-19
Add Bilmehakki to deployments page

wrightmartin:
- Change column width to five rather than six to accommodate growing number of deployments


zarino:
I’ve tested this out. Looks good! :+1:

PR #4679, opened 2018-06-19
[#4677] Correctly handle Atom feed responses with no events

### LINKED ISSUE #4677: api#body_request_events (NoMethodError) "undefined method `created_at' for nil:NilClass
```
A NoMethodError occurred in api#body_request_events:

  undefined method `created_at' for nil:NilClass
  app/views/api/request_events.atom.builder:3:in `block in _app_views_api_request_events_atom_builder__806364424503412446_69872756192820'


Request:

  * URL        : https://www.whatdotheyknow.com/api/v2/body/[snip]/request_events.atom?since_date=2018-06-15
  * HTTP Method: GET
  * Parameters : {"since_date"=>"2018-06-15", "feed_type"=>"atom", "controller"=>"api", "action"=>"body_request_events", "id"=>"[snip]"}
```

gbp:
### Relevant issue(s)

Fixes #4677 

### What does this do?

Removed the `updated` timestamp from the Atom API feed when there are no `InfoRequestEvent` to render.

### Why was this needed?

So we return successful responses for API requests. This has been causing exceptions and email alerts every ~20 minutes.

### DIFF
--- a/app/views/api/request_events.atom.builder
+++ b/app/views/api/request_events.atom.builder
@@ -1,6 +1,6 @@
 atom_feed("xmlns:alaveteli" => "http://www.alaveteli.org/API/v2/RequestEvents/Atom") do |feed|
   feed.title("Events relating to #{@public_body.name}")
-  feed.updated(@events.first.created_at)
+  feed.updated((@events.first || @public_body).created_at)
 
   for event in @events
     feed.entry(event) do |entry|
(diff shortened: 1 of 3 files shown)


app/views/api/request_events.atom.builder
```diff
@@ -1,6 +1,6 @@
 atom_feed("xmlns:alaveteli" => "http://www.alaveteli.org/API/v2/RequestEvents/Atom") do |feed|
   feed.title("Events relating to #{@public_body.name}")
-  feed.updated(@events.first.created_at)
+  feed.updated(@events.first.created_at) if @events.first
```
crowbot: Would an `updated` value of the `created_at` timestamp for the public body make more sense in the absence of any events?

PR #4705, opened 2018-07-06
Remove foundation dependencies

wrightmartin:
* What does this do?
This removes the Foundation gem and extracts the source assets to Alaveteli

* Why was this needed?
We are moving away from the Foundation framework and this is step 1. From here we will replace Foundation on a feature-by-feature basis


garethrees:
Should unpack these in to [`vendor/assets`](http://guides.rubyonrails.org/v4.2.10/asset_pipeline.html#asset-organization) really.

wrightmartin:
> Should unpack these in to vendor/assets really.

Yeah - there's a few things we could do to improve this. For now I put them exactly where they were expecting them to be as a 'change as little as possible' approach.

PR #4737, opened 2018-07-18
Migrate to factory_bot_rails

garethrees:
Followed the official guide:

https://github.com/thoughtbot/factory_bot/blob/4-9-0-stable/UPGRADE_FROM_FACTORY_GIRL.md

Closes https://github.com/mysociety/alaveteli/pull/4725


garethrees:
I've fixed some style violations, but there are too many to manually go through for such a minor change. Most of the problems already existed, but have been pointed out by hound because the lines changed. I'd love to fix them all, but I'd rather do it with a run of `rubocop -a` when we investigate that further.

Gemfile.lock
```diff
@@ -502,3 +502,6 @@ DEPENDENCIES
   will_paginate (~> 3.1.6)
   xapian-full-alaveteli (~> 1.2.21.1)
   xml-simple (~> 1.1.0)
+
```
lizconlan: nitpick - shouldn't include the BUNDLED WITH bit

spec/mailers/request_mailer_spec.rb
```diff
@@ -42,21 +42,21 @@
...
     it "puts messages with a malformed To: in the holding pen" do
-      request = FactoryGirl.create(:info_request)
+      request = FactoryBot.create(:info_request)
       receive_incoming_mail('incoming-request-plain.email', 'asdfg')
       expect(InfoRequest.holding_pen_request.incoming_messages.size).to eq(1)
     end
 
     it "puts messages with the request address in Bcc: in the holding pen" do
-      request = FactoryGirl.create(:info_request)
+      request = FactoryBit.create(:info_request)
```
gbp: This looks wrong

lizconlan: \*headdesk\* thanks!

PR #4742, opened 2018-07-20
[#3584] Improve handling of 'not_foi' and 'vexatious' requests

### LINKED ISSUE #3584: "This request has been hidden from the site" message shouldn't ever appear on published threads
Currently it is possible for a request to have a status of "not_foi" and a prominence of normal which prompts the

> "This request has been hidden from the site"

message to appear on published requests. 

Presumably this state of affairs arose as administrators made mistakes and manually changed the status without changing the prominence. 

29 requests on WhatDoTheyKnow are currently affected; which is a small enough number for manual correction by administrators.

lizconlan:
## What does this do?

1. Prevents the text "this request has been hidden" from appearing against publicly visible requests.
1. Tries to alert admins via tooltip messages that they are leaving "invalid" requests (with a `described_state` of `not_foi` or `vexatious`) visible (this may be a valid choice in some cases although in the UK, the intention is usually to hide the requests).
1. Uses an "Are you sure?" dialog to allow the admin to correct the data if the combination is a mistake. 

not_foi_edit_2

## Why was this needed?

It is not clear that setting 2 different attributes is required to hide a request when editing metadata. In the past we have had requests on WDTK that display "this request is hidden" text which then requires and admin to try to find all such requests and manually hide them.

Closes #3584

### DIFF
--- a/app/assets/javascripts/admin.js
+++ b/app/assets/javascripts/admin.js
@@ -6,6 +6,7 @@
 //= require bootstrap-collapse
 //= require bootstrap-tab
 //= require bootstrap-dropdown
+//= require bootstrap-tooltip
 //= require admin/admin
 //= require admin/category-order
 //= require admin/holidays
--- a/app/assets/javascripts/admin/admin.coffee
+++ b/app/assets/javascripts/admin/admin.coffee
@@ -49,9 +49,14 @@ jQuery ->
         attr('title',
              'Warning! You are about to save this request without hiding it!')
       submit_button.tooltip()
+      submit_button.
+        data('confirm',
+             'You have set this request to "' + this.value + '" but not' +
+             ' hidden it using prominence. Are you sure you want to continue?')
     else
       $('#info_request_prominence').removeAttr('title')
       $('#info_request_prominence').tooltip('destroy')
+      submit_button.removeData('confirm')
       submit_button.removeAttr('title')
       submit_button.tooltip('destroy')
   )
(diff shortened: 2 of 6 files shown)


garethrees:
I haven't run this, so take with a bag of salt, but I'm not sure this is going to be obvious enough?

Chances are you're going to click the button before really noticing the popup. On an iPad, you wouldn't get the message at all.

I think we probably want a [tooltip](http://getbootstrap.com/2.3.2/javascript.html#tooltips) to appear on the prominence field when you've got "the bad combination" selected, and maybe also a confirm dialog so that the user can back out if they click the button without noticing.

lizconlan:
> not sure this is going to be obvious enough?

Yeah, I'm not convinced either. Plus unless you're using Edge it can appear slowly enough that the button can be pressed before the message appears.

> we probably want a tooltip to appear on the prominence field

As that might be off the top of the screen I'm inclined to add some sort of indicator on the just changed described state field as well as that's guaranteed to still be in vision.

garethrees:
> As that might be off the top of the screen (on my 13" MBP I had to position things carefully to have everything included in the screenshot) I'm inclined to add some sort of indicator on the just changed described state field as well as that's guaranteed to still be in vision

Or we could just move the `prominence` field so that it's grouped with `described_state`?

screen shot 2018-07-20 at 10 43 31

lizconlan:
> Or we could just move the `prominence` field so that it's grouped with `described_state`?

Genius! Makes it easier to see whether there's a contradiction even without an extra prompt (still going to have the extra prompt tho). Thanks!

lizconlan:
That makes sense. Done! (Just switched the if clause from `if value is 'normal'` to `if (value is 'normal' or value is 'backpage')` so that the behaviour is identical for both.

PR #4748, opened 2018-07-23
[pro#539] Fix losing draft on pagination

garethrees:
Adding an authority after paginating the batch authority search loses
the current authority selection. This is caused by the search results
losing the `value` of the `js-draft-id` hidden field after loading a new
page of results.

This ensures that element is updated after every refresh of the results
list.

Fixes https://github.com/mysociety/alaveteli-professional/issues/539

---

I'm not sure I've got the `.js-draft-id` scoped correctly, but I tried `$($results, '.js-draft-id').val(DraftBatchSummary.draftId);` and that didn't work.

### DIFF
--- a/app/assets/javascripts/alaveteli_pro/batch_authority_search/results.js
+++ b/app/assets/javascripts/alaveteli_pro/batch_authority_search/results.js
@@ -39,6 +39,11 @@
     }
   };
 
+  // Update the draft ID of each result
+  var updateDraftId = function updateDraftId() {
+    $('.js-draft-id', $results).val(DraftBatchSummary.draftId);
+  };
+
   // Main render method
   var render = function render() {
     var content = html;
@@ -51,6 +56,7 @@
 
     if ($results.html() !== content) {
       $results.html(content);
+      updateDraftId();
       $search.trigger(SearchEvents.rendered);
     }
   };


zarino:
> I'm not sure I've got the .js-draft-id scoped correctly, but I tried $($results, '.js-draft-id').val(DraftBatchSummary.draftId); and that didn't work.

You want either `$( '.js-draft-id', $results ).val(…)` or `$results.find('.js-draft-id').val(…)` 🙂

(Don’t worry, took the best part of a whole morning to spot that mistake myself)

app/assets/javascripts/alaveteli_pro/batch_authority_search/results.js
```diff
@@ -66,6 +70,7 @@
 
     $search.on(SearchEvents.domUpdated, update);
     $search.on(SearchEvents.loadingSuccess, updateResults);
+    $search.on(SearchEvents.loadingSuccess, updateDraftId);
```
zarino: I have a feeling the event you should be listening to is `SearchEvents.rendered` – because that’s the event emitted when `render()` has updated the HTML content of `$results`.

In fact, it makes me wonder, why not just add `$results.find('.js-draft-id').val(DraftBatchSummary.draftId);` in near the end of render(), perhaps on line 54, before the `$search.trigger(SearchEvents.rendered);` ?

garethrees: Thanks, both useful comments.

I quite like the named method for updating the draft ID, so I think:

```js
  var updateDraftId = function updateDraftId() {
    $('.js-draft-id').val(DraftBatchSummary.draftId);
  };

  var render = function render() {
    // …
    if ($results.html() !== content) {
      $results.html(content);
      updateDraftId();
      $search.trigger(SearchEvents.rendered);
    }
  };
```

That look okay?

zarino: Sure! :+1:

PR #4750, opened 2018-07-25
Remove mentions of chat from alaveteli.org/community

garethrees:
See commit messages for details/reasoning.

c063505 is up for discussion, but I feel pretty strongly in favour of merging it.


garethrees:
We said we were happy with this.

PR #4752, opened 2018-07-26
[#4499] Prevent requests from authorities tagged as not subject to legislation from showing "By law..." in status text

### LINKED ISSUE #295: Don't Say Authorities Not Subject to Access to Info Laws Are In Breach of Them
WhatDoTheyKnow tags those bodies which are not subject to FOI as foi_no

(I suspect other installations will want to do similar)

At the moment after the relevant period we display:

"By law, under all circumstances,[Body] should have responded by now (details)."

We shouldn't say this in respect of bodies which are not formally subject to FOI

We should probably just cut the sentence and leave the description at: "Response to this request is long overdue." 

Or we could say "If [Body] had been subject to the Freedom of Information Act, ,[Body] should have responded by now (details)."

In the case of eir_only tagged bodies we can still include the "By law..." sentence. (Though that depends on the content of the request!)

### LINKED ISSUE #4499: foi_no doesn't prevent message "By law, under all circumstances, AUTHORITY should have responded by now"
An example, brought by our attention by the Church Commissioners. 

They are tagged foi_no yet for requests that haven't been responded to, e.g. [this one](https://www.whatdotheyknow.com/request/removal_of_maidstone_borough_cou_10), Alaveteli still states "By law, under all circumstances, Church Commissioners should have responded by now".

That's evidently incorrect: bodies not subject to the FOI Act aren't bound by the section of the FOI act obliging them to respond by a certain date. I'd suggest changing this.

Of course, the moral obligation is a different issue.

lizconlan:
## What does this do?

1. Adds a `PublicBody#not_subject_to_law?` method which allows us to identify authorities which are not subject to FOI law but are still considered eligible to receive information requests (returns true always if `authority_must_respond` has been set to false in config)
1. Alters the status text for "waiting_response_overdue" and "waiting_response_very_overdue" requests from authorities not subject to the law to remove text that states their legal obligation (and link to the #authorities section on the help page instead of #quickly_response)
1. Alters the text for messages shown after classifying a request
1. Alters the text - and help links - that appears when sending a followup message for overdue requests
1. Alters the text in overdue request mailer alerts
1. Alters the text in overdue notification mailer alerts
1. Updates the help section on the authority edit page for admins to describe the `foi_no` tag and to warn that `not_apply` prevents further requests
1. Adds a note to the end of the request header subtitle section of the request page if the authority is not subject to FOI

screen shot 2018-08-02 at 18 32 17

## Why was this needed?

We were implying a legal obligation in all cases regardless of whether the authority fell under FOI legislation or not.

## Notes to reviewer

* As the simplest solution, I've chosen to adopt the `foi_no` tag as used by WDTK in the core code but am happy to hear other suggestions for this. (I avoided using `not_apply` as this is currently used to prevent further requests from being sent.) An alternative approach would be to use - and recommend in documentation - a different tag with international re-users in mind whose legislation may not be called "FOI" and add an override of the new method to in the WDTK theme to accept either tag.

* Is there a better (less clunky) way of saying `not_subject_to_law?`? In its favour, it matches the phrasing on the ICO and Scottish Information Commissioner websites and is clearly distinct from `not_apply?` while following the same general convention.

### Overdue (standard)

screen shot 2018-07-26 at 16 30 00

### Overdue (not subject to law)

screen shot 2018-07-26 at 16 29 27

### Long overdue (standard)

screen shot 2018-07-26 at 15 45 12

### Long overdue (not subject to law)

screen shot 2018-07-26 at 15 45 22

Closes #4499 
Closes #295

### DIFF
--- a/app/helpers/info_request_helper.rb
+++ b/app/helpers/info_request_helper.rb
@@ -53,7 +53,7 @@ module InfoRequestHelper
                        simple_date(info_request.date_response_required_by))
     str += ' '
     str += "("
-    str += link_to _("details"), "/help/requesting#quickly_response"
+    str += details_help_link(info_request.public_body)
     str += ")."
   end
 
@@ -76,8 +76,7 @@ module InfoRequestHelper
                        simple_date(info_request.date_response_required_by))
     str += ' '
     str += "("
-    str += link_to _('details'),
-                   help_requesting_path(:anchor => 'quickly_response')
+    str += details_help_link(info_request.public_body)
     str += ")"
   end
 
@@ -95,8 +94,7 @@ module InfoRequestHelper
     end
     str += ' '
     str += "("
-    str += link_to _('details'),
-                   help_requesting_path(:anchor => 'quickly_response')
+    str += details_help_link(info_request.public_body)
     str += ")."
 
     unless info_request.is_external?
@@ -273,6 +271,12 @@ module InfoRequestHelper
     else
       get_attachment_path(attach_params)
     end
+  end
+
+  def details_help_link(public_body)
+    anchor =
+      public_body.not_subject_to_law? ? 'authorities' : 'quickly_response'
+    link_to _('details'), help_requesting_path(:anchor => anchor)
   end
 
   private
--- a/app/helpers/info_request_helper.rb
+++ b/app/helpers/info_request_helper.rb
@@ -42,7 +42,7 @@ module InfoRequestHelper
 
   def status_text_waiting_response(info_request, opts = {})
     str = _('Currently <strong>waiting for a response</strong> from ' \
-            '{{public_body_link}}, they must respond promptly and',
+            '{{public_body_link}}, they should respond promptly and',
             :public_body_link => public_body_link(info_request.public_body))
     str += ' '
     str += _('normally')
--- a/app/helpers/info_request_helper.rb
+++ b/app/helpers/info_request_helper.rb
@@ -60,12 +60,18 @@ module InfoRequestHelper
   def status_text_waiting_response_overdue(info_request, opts = {})
     str = _('Response to this request is <strong>delayed</strong>.')
     str += ' '
-    str += _('By law, {{public_body_link}} should normally have responded ' \
-             '<strong>promptly</strong> and',
-             :public_body_link => public_body_link(info_request.public_body))
-    str += ' '
-    str += _('by')
-    str += ' '
+    if info_request.public_body.not_subject_to_law?
+      str += _('Although not legally required to do so, we would have ' \
+               'expected {{public_body_link}} to have responded by ',
+               :public_body_link => public_body_link(info_request.public_body))
+    else
+      str += _('By law, {{public_body_link}} should normally have responded ' \
+               '<strong>promptly</strong> and',
+               :public_body_link => public_body_link(info_request.public_body))
+      str += ' '
+      str += _('by')
+      str += ' '
+    end
     str += content_tag(:strong,
                        simple_date(info_request.date_response_required_by))
     str += ' '
@@ -78,9 +84,15 @@ module InfoRequestHelper
   def status_text_waiting_response_very_overdue(info_request, opts = {})
     str = _('Response to this request is <strong>long overdue</strong>.')
     str += ' '
-    str += _('By law, under all circumstances, {{public_body_link}} should ' \
-            'have responded by now',
-            :public_body_link => public_body_link(info_request.public_body))
+    if info_request.public_body.not_subject_to_law?
+      str += _('Although not legally required to do so, we would have ' \
+               'expected {{public_body_link}} to have responded by now',
+               :public_body_link => public_body_link(info_request.public_body))
+    else
+      str += _('By law, under all circumstances, {{public_body_link}} should ' \
+               'have responded by now',
+               :public_body_link => public_body_link(info_request.public_body))
+    end
     str += ' '
     str += "("
     str += link_to _('details'),
--- a/app/views/followups/_followup.html.erb
+++ b/app/views/followups/_followup.html.erb
@@ -84,14 +84,16 @@
               '<strong>{{date}}</strong>',
               :date=>simple_date(@info_request.date_response_required_by)) %>
 
-        (<%= link_to _('details'), "#{help_requesting_path}#quickly_response" %>).
+        (<%= link_to _('details'),
+                     help_requesting_path(anchor: 'quickly_response') %>).
       </p>
     <% elsif status == 'waiting_response_very_overdue' %>
       <p>
         <%= _('The response to your request is <strong>long overdue</strong>. ' \
               'You can say that, by law, under all circumstances, the ' \
               'authority should have responded by now') %>
-        (<%= link_to _('details'), "#{help_requesting_path}#quickly_response" %>).
+        (<%= link_to _('details'),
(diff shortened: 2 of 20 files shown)


garethrees:
I think first step here is finding all the places its mentioned so that we can see the full picture. It may be that we try to use something like "must _normally_ have responded by now" everywhere other than the main request page to keep things simpler.

spec/integration/request_controller_spec.rb
```diff
@@ -350,37 +373,55 @@
...
+
+      include_examples 'authority is subject to FOI law'
+
     end
 
     context 'marking very overdue request as waiting_responses' do
 
       let(:classification) { 'waiting_response1' }
 
+      before do
```
garethrees: Why have these been extracted?

As far as I can tell the behaviour is the same

app/views/followups/_followup.html.erb
```diff
@@ -78,20 +78,43 @@
...
-              '<strong>{{date}}</strong>',
-              :date=>simple_date(@info_request.date_response_required_by)) %>
+        <% if @info_request.public_body.not_subject_to_law? %>
+          <%= _('The response to your request has been <strong>delayed' \
+                '</strong>. Althouth the authority has no legal obligation ' \
+                'to reply, they should normally have responded by ' \
+                '<strong>{{date}}</strong>',
+                :date=>simple_date(@info_request.date_response_required_by)) %>
 
-        (<%= link_to _('details'), "#{help_requesting_path}#quickly_response" %>).
```
garethrees: It would have been nice to make a commit that clean this up before propagating it further…

```diff
# Fix incorrect use of linking to anchor
- (<%= link_to _('details'), "#{help_requesting_path}#quickly_response" %>).
+ (<%= link_to _('details'), help_requesting_path(anchor: 'quickly_response') %>).
```

app/views/followups/_followup.html.erb
```diff
@@ -78,20 +78,43 @@
...
     <% if status == 'waiting_response_overdue' %>
       <p>
-        <%= _('The response to your request has been <strong>delayed</strong>. ' \
-              'You can say that, by law, the authority should normally have ' \
-              'responded <strong>promptly</strong> and by ' \
-              '<strong>{{date}}</strong>',
-              :date=>simple_date(@info_request.date_response_required_by)) %>
+        <% if @info_request.public_body.not_subject_to_law? %>
+          <%= _('The response to your request has been <strong>delayed' \
+                '</strong>. Althouth the authority has no legal obligation ' \
```
garethrees: s/Althouth/Although/

app/views/request_mailer/overdue_alert.text.erb
```diff
@@ -1,9 +1,16 @@
 <%= raw @info_request.public_body.name %> <%= _('have delayed.') %>
 
+<% if @info_request.public_body.not_subject_to_law? %>
+<%= _('They have not replied to your {{law_used_short}} request {{title}} ' \
+      'promptly',
+      :law_used_short => @info_request.law_used_human(:short).html_safe,
+      :title => @info_request.title.html_safe) %>.
+<% else %>
```
garethrees: I think this is adding complexity for no benefit. The `else` says "normally required by law", which would cover the `not_subject_to_law?` branch.

We can either:

1. Remove the conditional and keep the original wording, as that covers both cases
2. Be more forceful in the `else` branch and remove `"normally"`.

I'd vote for option 1.

app/views/request_mailer/very_overdue_alert.text.erb
```diff
@@ -1,9 +1,15 @@
 <%= raw @info_request.public_body.name %> <%= _('are long overdue.')%>
 
+<% if @info_request.public_body.not_subject_to_law? %>
```
garethrees: As per the comment on `app/views/request_mailer/overdue_alert.text.erb`. I'd remove the conditional and use "as _normally_ required by law".

app/views/notification_mailer/overdue_notification.text.erb
```diff
@@ -1,9 +1,16 @@
 <%= raw @info_request.public_body.name %> <%= _('have delayed.') %>
 
+<% if @info_request.public_body.not_subject_to_law? %>
+<%= _('They have not replied to your {{law_used_short}} request {{title}} ' \
```
garethrees: I think we can remove the conditionals on this and use "normally required by law" for both situations.

app/views/notification_mailer/very_overdue_notification.text.erb
```diff
@@ -1,9 +1,16 @@
 <%= raw @info_request.public_body.name %> <%= _('are long overdue.')%>
 
+<% if @info_request.public_body.not_subject_to_law? %>
+<%= _('They have still not replied to your {{law_used_short}} request ' \
```
garethrees: I think we can remove the conditionals on this and use "normally required by law" for both situations.

app/helpers/info_request_helper.rb
```diff
@@ -275,6 +273,14 @@ def attachment_path(incoming_message, attachment, options = {})
     end
   end
 
+  def details_help_link(public_body)
```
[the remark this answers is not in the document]
lizconlan: 👍

spec/integration/request_controller_spec.rb
lizconlan: Hmm, I could have explained that better in the commit comment. It's to separate out the time travel setup from the process of classifying to allow us to use shared examples. I agree it's pretty clumsy to have it in the `before`/`after` block but I couldn't work out how else to do it.

app/views/request_mailer/overdue_alert.text.erb
lizconlan: _selects option 1_

spec/integration/request_controller_spec.rb
lizconlan: (going to move this into its own commit step for easier explanation/consideration)

app/helpers/info_request_helper.rb
```diff
@@ -78,9 +84,15 @@ def status_text_waiting_response_overdue(info_request, opts = {})
...
+               'expected {{public_body_link}} to have responded by now',
+               :public_body_link => public_body_link(info_request.public_body))
+    else
+      str += _('By law, under all circumstances, {{public_body_link}} should ' \
+               'have responded by now',
+               :public_body_link => public_body_link(info_request.public_body))
+    end
     str += ' '
     str += "("
     str += link_to _('details'),
```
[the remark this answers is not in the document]
garethrees: Resolved in 3e5c6ee323a21bd810f5ae367c6b7c8886a3073b

app/views/followups/_followup.html.erb
```diff
@@ -78,22 +78,43 @@
...
-                     help_requesting_path(anchor: 'quickly_response') %>).
+        <% if @info_request.public_body.not_subject_to_law? %>
+          <%= _('The response to your request has been <strong>delayed' \
+                '</strong>. Although the authority has no legal obligation ' \
+                'to reply, they should normally have responded by ' \
+                '<strong>{{date}}</strong>',
+                :date=>simple_date(@info_request.date_response_required_by)) %>
+
+          (<%= link_to _('details'),
+                       "#{help_requesting_path}#authorities" %>).
```
garethrees: This re-implements the broken `link_to` fixed in 8c23cceef88bf49d71ef6f90a1d13f888a0f4cb0.

PR #4775, opened 2018-08-06
[#4664] Use full spam check at signup

### LINKED ISSUE #4664: Use full User spam checks in UserController#signup
I added extra spam checking for user names in https://github.com/mysociety/alaveteli/pull/4627 to try to help identify spam User accounts on signup.

After wondering why we weren't catching any spam, I noticed that we're only checking for [spam email domains](https://github.com/mysociety/alaveteli/blob/619e423/app/controllers/user_controller.rb#L156) on signup, rather than running the full spam checker.

I can't see a reason not to apply all checks, or if we find in practice its too much we can always configure a custom set of `score_mappings`.

lizconlan:
## What does this do?

Uses the full spam check rather than just looking for a known spam email domain when a new user signs up.

## Why was this needed?

We've noticed accounts signing up with spam in the name field and wanted to catch those as well.

Closes #4664


lizconlan:
Happy to do those but as this is for new users, the only data we have at this point is the name and email address, they'll not have supplied the `about_me` text or been marked as `confirmed_not_spam` yet as the User doesn't yet exist in the database

lizconlan:
Hmm...

```rb
u = User.new(name: 'Not Spam', email: 'ns@example.com')
u.about_me_already_exists?
```

Yeah, that's not going to work as expected

garethrees:
> or been marked as `confirmed_not_spam`

That was just for illustration that this check would have potentially blocked a bunch of genuine signups

> they'll not have supplied the about_me text

Oh durr – my bad. Yeah, it doesn't matter what we score those, so they can just be set to `1` in the signup config

lizconlan:
I'm inclined to fix #4784 here as it's probably quicker to fix it than patch it

lizconlan:
>> or been marked as `confirmed_not_spam`
>
> That was just for illustration that this check would have potentially blocked a bunch of genuine signups

doh! misread, sorry!

app/models/user.rb
```diff
@@ -531,6 +531,7 @@ def set_profile_photo(new_profile_photo)
   end
 
   def about_me_already_exists?
+    return false if about_me.blank?
```
garethrees: 👍

app/models/user.rb
```diff
@@ -532,7 +532,7 @@ def set_profile_photo(new_profile_photo)
 
   def about_me_already_exists?
     return false if about_me.blank?
-    self.class.where(:about_me => about_me).any?
+    self.class.where(:about_me => about_me).where.not(id: id).any?
```
garethrees: 👍

app/controllers/user_controller.rb
```diff
@@ -633,4 +633,25 @@ def handle_spam_signup(user_signup)
...
+      score_mappings: {
+        name_is_all_lowercase?: 1,
+        name_is_one_word?: 1,
+        name_includes_non_alpha_characters?: 1,
+        name_is_garbled?: 1,
+        email_from_suspicious_domain?: 10,
+        email_from_spam_domain?: 13,
+        email_from_spam_tld?: 1,
+        name_is_spam_format?: 10,
+        about_me_includes_currency_symbol?: 1,
```
garethrees: I think 👎 to weighting `about_me`. We don't plan to introduce it on signup at all, so this just communicates that these checks have significance in this case, but in reality they don't do a thing. I'd set all the `about_me*` keys to `0`.

garethrees: p.s. sorry for the initial confusion I added https://github.com/mysociety/alaveteli/pull/4775#issuecomment-411784007

lizconlan: fixed (and edited the commit comment to match)

PR #4781, opened 2018-08-07
[pro#543] Subscription page referral code

garethrees:
## Relevant issue(s)

Fixes https://github.com/mysociety/alaveteli-professional/issues/543.

## What does this do?

Allows developers to configure a referral code that's been set up in Stripe to be shown on `/profile/subscriptions` so that existing pro users can give friends or colleagues a signup discount.

## Why was this needed?

We want to encourage our existing users to share Pro with their networks.

## Implementation notes

* This doesn't perform much error checking that the coupon code has been set up correctly, but I think this is a very beta feature at the moment so I think that's fine?
* The translation string may get a little clunky depending on what discount the site applies. I haven't thought through all the permutations ("1 month free", "first month free", "£5 off", etc), but again, I think this is one with us to live with and adapt to change if it becomes clear where the problems are.
* ~~Ideally I'd have done some nice highlighting around the coupon code itself, but that would make the translation string more complicated and probably require some design skills. I think its fine as though, tbh.~~

## Screenshots

screen shot 2018-08-07 at 17 38 52

### DIFF
--- a/app/views/alaveteli_pro/subscriptions/index.html.erb
+++ b/app/views/alaveteli_pro/subscriptions/index.html.erb
@@ -39,6 +39,18 @@
           <% end %>
         </div>
 
+        <% if @discount_code && @discount_terms %>
+          <div class="settings-section">
+            <h3><%= _('Refer a friend') %></h3>
+            <p class="settings__item">
+              <%= _('Share the code {{discount_code}} with a friend to give ' \
+                    'them {{discount_terms}} on signup.',
+                    discount_code: content_tag(:strong, @discount_code),
+                    discount_terms: @discount_terms) %>
+            </p>
+          </div>
+        <% end %>
+
       </div>  <!-- .settings-right-column -->
 
     </div>  <!-- .row -->
--- a/app/views/alaveteli_pro/subscriptions/index.html.erb
+++ b/app/views/alaveteli_pro/subscriptions/index.html.erb
@@ -57,4 +57,4 @@
 
   </div>  <!-- .inner-canvas-body -->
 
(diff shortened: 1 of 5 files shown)


app/views/alaveteli_pro/subscriptions/index.html.erb
```diff
@@ -39,6 +39,18 @@
           <% end %>
         </div>
 
+        <% if @referral_coupon %>
+          <div class="settings-section">
+            <h3><%= _('Refer a friend') %></h3>
+            <p class="settings__item">
+              <%= _('Share the code {{discount_code}} with a friend to give ' \
+                    'them {{discount_terms}} on signup.',
+                    discount_code: @referral_coupon.id,
```
lizconlan: I think this is fine as-is, but we could consider something like this to make the coupon code bold as we're passing in the value:

```rb
discount_code: "<strong>#{ h(@referral_coupon.id) }</strong>".html_safe,
```

(maybe by way of a helper method as that's a bit untidy)

garethrees: Ah, nice. I did mention this…

> * Ideally I'd have done some nice highlighting around the coupon code itself, but that would make the translation string more complicated and probably require some design skills. I think its fine as though, tbh.

…but didn't think to try it on that side of the translation string.

I'll give this a go.

lizconlan: I did see that - probably why it was on my mind while testing -but failed to quote it when making the comment, sorry!

garethrees:
Just needs a quick once over to account for the changes to handle Stripe namespace.

Issue #4505, opened 2018-02-05
Add a sticky element to make the 'action' button and other commonly used elements easier to find and use


wrightmartin:
Feedback from Lucas:

> I find it hard to navigate this page. I often find myself scrolling and scanning the page for info, and losing my way, and then scrolling back up (and down again) to find the status button.

wrightmartin:
This and other issues could be addressed with a 'sticky' request page header, which appears when you scroll past a certain point on the request page. A concept design is below

This does a few things:

- Action button is always on screen
- Context of the request is given with the title and authority always visible
- The first request text is available to view when needed for reference when comparing to a request response.

![sticky header](https://user-images.githubusercontent.com/2292925/35803544-0ef9d83c-0a6c-11e8-9be4-86d36379e193.png)

garethrees:
I think this is a good start, but has some issues:

* Its going to take up a load of screen real estate on a laptop at the current size; It should be much shorter
* Most initial correspondence will be too big for that bubble – a jump link is a cool idea though
* The authority name is a bit small; I think we should only show this on _your_ request, so we could drop "You" and increase the size of the authority name
* Maybe this could include forward/backward links to jump up and down correspondence bubbles nice and quickly.
* It should include the status updater

dracos:
I would have liked such a header (or sticky sidebar?) on requests that aren't mine too, most commonly for e.g. "this response is boilerplate, I want to skip to next thing".

garethrees:
I guess I was thinking that the sticky header would be mostly "request management" tools, but it could be easy to show a stripped down one for other viewers.

wrightmartin:
> Its going to take up a load of screen real estate on a laptop at the current size; It should be much shorter

Yep agree, just a quick mockup but you're right. We can use some media queries on the vertical height to adjust this too

> Most initial correspondence will be too big for that bubble – a jump link is a cool idea though

Yeah, I think in most cases the first request truncated will be largely useless information, will rethink

> Maybe this could include forward/backward links to jump up and down correspondence bubbles nice and quickly.

Hadn't considered that but what a great idea

> It should include the status updater

Is that not covered by the 'action' button?

> I guess I was thinking that the sticky header would be mostly "request management" tools, but it could be easy to show a stripped down one for other viewers.

My first thoughts too - but in the IPV feedback I think they said something about being linked directly into the guts of a request with no context at all? This would help that

garethrees:
> Is that not covered by the 'action' button?

[...]

PR #4783, opened 2018-08-08
[pro#547] Make the pro request sidebar sticky

wrightmartin:
* Relevant issue(s)

Fixes https://github.com/mysociety/alaveteli-professional/issues/547

* What does this do?
On desktop this makes the pro request sidebar sticky on scroll 

* Why was this needed?
Part of improvements to the pro request experience https://github.com/mysociety/alaveteli/issues/4505

Precursor to https://github.com/mysociety/alaveteli-professional/issues/548

* Implementation notes
This implementation has limited support - especially for Internet Explorer 11 and below (https://caniuse.com/#feat=css-sticky). Support for older browsers will require a polyfill or JavaScript solution, suggest we discuss this.

### DIFF
--- a/app/assets/stylesheets/responsive/_global_layout.scss
+++ b/app/assets/stylesheets/responsive/_global_layout.scss
@@ -140,6 +140,26 @@ textarea{
   }
 }
 
+.sidebar--sticky {  
+  @supports (position: sticky) or (position: -webkit-sticky){
+    $sidebar-vertical-margin: 20px;
+    $adminbar-height: 41px;
+    @include respond-min( $main_menu-mobile_menu_cutoff ){
+      position: -webkit-sticky;
+      position: sticky;
+      top: $sidebar-vertical-margin;
+      //If the sticky sidebar is taller than the screen height, make the contents scrollable
+      max-height: calc(100vh - #{$sidebar-vertical-margin});
+      overflow: auto;
+      .admin + * & {
+        // If you're an admin your sticky sidebar needs more space to account for the sticky admin bar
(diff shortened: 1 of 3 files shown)


app/assets/stylesheets/responsive/_global_layout.scss
```diff
@@ -140,6 +140,26 @@ textarea{
...
+.sidebar--sticky {  
+  @supports (position: sticky) or (position: -webkit-sticky){
+    $sidebar-vertical-margin: 20px;
+    $adminbar-height: 41px;
+    @include respond-min( $main_menu-mobile_menu_cutoff ){
+      position: -webkit-sticky;
+      position: sticky;
+      top: $sidebar-vertical-margin;
+      //If the sticky sidebar is taller than the screen height, make the contents scrollable
+      max-height: calc(100vh - #{$sidebar-vertical-margin});
```
garethrees: I got a scrollbar anyway:

sticky-sidebar-scrollbar

garethrees: Just noticed that I can't seem to replicate this with whatdotheyknow-theme, so worth checking what differences there are between the sidebar CSS there and in core.

wrightmartin: Fixed

wrightmartin:
It turns out the bug was due to the houdini controls, `.houdini-input` is supposed to be hidden, but it still had a calculated height which, due to the way it was positioned, was throwing the calculation off.

Adding the houdini styles to core fixes this

PR #4786, opened 2018-08-08
Improves theme reconciliation

gbp:
### Relevant issue(s)

none!

### What does this do?

Stops reconciling files if the file the repo either:
* doesn't exist
* is empty
* contains a placeholder

### Why was this needed?

Prevents opening some diffs where the left hand side is empty and saves unneeded `:qa` in vim

### Implementation notes

Might be a better way to do this - my shell scripting is a bit rusty


gbp:
I'm going to close this. While I still think we could do with an improved reconciliation process I haven't used these changes in ages and I don't see myself doing any additional work to in the near future on this.

PR #4767, opened 2018-08-03
WIP: Update InfoRequestBatch#sent_at after all requests sent

garethrees:
I wouldn't call a Batch "sent" until all requests are sent, so move the
update.

Just trying this to see what specs do.

What if script/send-batches is called at the same time as a batch is
being sent?

Please include as many as the following as possible to help the reviewer and future readers:

* Relevant issue(s)
* What does this do?
* Why was this needed?
* Implementation notes
* Screenshots
* Notes to reviewer

Issue #4749, opened 2018-07-24
[ERROR] script/send-batch-requests (PG::UniqueViolation) "ERROR: duplicate key value violates unique constraint


garethrees:
> `PG::UniqueViolation: ERROR: duplicate key value violates unique constraint "index_request_summaries_summary_categories_unique" (ActiveRecord::RecordNotUnique)
DETAIL: Key (request_summary_id, request_summary_category_id)=(490925, 5) already exists.`

```
/data/vhost/www.whatdotheyknow.com/alaveteli-2018-07-16T13-52-20/vendor/bundle/ruby/2.3.0/gems/activerecord-4.2.10/lib/active_record/connection_adapters/postgresql_adapter.rb:602:in `exec_prepared': PG::UniqueViolation: ERROR: duplicate key value violates unique constraint "index_request_summaries_summary_categories_unique" (ActiveRecord::RecordNotUnique)
DETAIL: Key (request_summary_id, request_summary_category_id)=(490925, 5) already exists.
: INSERT INTO "request_summaries_summary_categories" ("request_summary_id", "request_summary_category_id") VALUES ($1, $2)
from /data/vhost/www.whatdotheyknow.com/alaveteli-2018-07-16T13-52-20/vendor/bundle/ruby/2.3.0/gems/activerecord-4.2.10/lib/active_record/connection_adapters/postgresql_adapter.rb:602:in `block in exec_cache'
from /data/vhost/www.whatdotheyknow.com/alaveteli-2018-07-16T13-52-20/vendor/bundle/ruby/2.3.0/gems/activerecord-4.2.10/lib/active_record/connection_adapters/abstract_adapter.rb:484:in `block in log'
from /data/vhost/www.whatdotheyknow.com/alaveteli-2018-07-16T13-52-20/vendor/bundle/ruby/2.3.0/gems/activesupport-4.2.10/lib/active_support/notifications/instrumenter.rb:20:in `instrument'
from /data/vhost/www.whatdotheyknow.com/alaveteli-2018-07-16T13-52-20/vendor/bundle/ruby/2.3.0/gems/activerecord-4.2.10/lib/active_record/connection_adapters/abstract_adapter.rb:478:in `log'
from /data/vhost/www.whatdotheyknow.com/alaveteli-2018-07-16T13-52-20/vendor/bundle/ruby/2.3.0/gems/activerecord-4.2.10/lib/active_record/connection_adapters/postgresql_adapter.rb:601:in `exec_cache'
[...]

PR #4789, opened 2018-08-09
[#4749] Rework batch sending

gbp:
### Relevant issue(s)

Supplemental work on #4749
Depends on #4779 
Closes #4767 

### What does this do?

* Improves logic around determine if a batch has completely sent
* Makes it so batch sending so it will be resumed if it exits early (by moving when `sent_at` is set).

### Why was this needed?

Occasional exceptions (#4749) meant some batch requests were not sent straight away.

### DIFF
--- a/spec/integration/alaveteli_pro/request_list_spec.rb
+++ b/spec/integration/alaveteli_pro/request_list_spec.rb
@@ -35,6 +35,7 @@ describe "pro request list" do
     TestAfterCommit.with_commits(true) do
       batch_requests[0..3].each do |batch|
         batch.create_batch!
+        batch.update(sent_at: Time.zone.now)
         batch.reload
       end
     end
--- a/spec/models/info_request_batch_spec.rb
+++ b/spec/models/info_request_batch_spec.rb
@@ -101,23 +101,23 @@ describe InfoRequestBatch do
       end
     end
 
-    it 'should send requests to requestable public bodies, and return a list of unrequestable ones' do
-      allow(first_public_body).to receive(:is_requestable?).and_return(false)
-      unrequestable = info_request_batch.create_batch!
-      expect(unrequestable).to eq([first_public_body])
-      expect(info_request_batch.info_requests.size).to eq(1)
+    it 'does not resend requests to public bodies that have already received the request' do
+      expect(info_request_batch).to receive(:requestable_public_bodies).
+        and_return([first_public_body])
+      expect { info_request_batch.create_batch! }.to(
+        change(info_request_batch.info_requests, :count).by(1)
+      )
       request = info_request_batch.info_requests.first
       expect(request.outgoing_messages.first.status).to eq('sent')
     end
 
     it 'should not only send requests to public bodies if already sent' do
-      allow(second_public_body).to receive(:is_requestable?).and_return(false)
       request = info_request_batch.info_requests = [
         FactoryBot.create(:info_request, public_body: first_public_body)
       ]
-      unrequestable = info_request_batch.create_batch!
-      expect(unrequestable).to eq([second_public_body])
-      expect(info_request_batch.info_requests.size).to eq(1)
+      expect { info_request_batch.create_batch! }.to(
+        change(info_request_batch.info_requests, :count).by(1)
+      )
     end
 
     it 'should set the sent_at value of the info request batch' do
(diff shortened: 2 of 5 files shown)


spec/models/info_request_batch_spec.rb
```diff
@@ -101,23 +101,21 @@
...
+    it 'should send requests to requestable public bodies' do
+      expect(info_request_batch).to receive(:requestable_public_bodies).
+        and_return([first_public_body])
+      info_request_batch.create_batch!
       expect(info_request_batch.info_requests.size).to eq(1)
       request = info_request_batch.info_requests.first
       expect(request.outgoing_messages.first.status).to eq('sent')
     end
 
     it 'should not only send requests to public bodies if already sent' do
```
garethrees: This spec description is hard to read (added in https://github.com/mysociety/alaveteli/pull/4666/commits/0ef4cfc8b4bc4f24b2d05ee739ac9a7ee5cf0834).

I _think_ this should read:

```diff
- it 'should not only send requests to public bodies if already sent' do
+ it 'does not resend requests to public bodies that have already received the request' do
```

spec/models/info_request_batch_spec.rb
```diff
@@ -101,23 +101,21 @@
...
-      allow(second_public_body).to receive(:is_requestable?).and_return(false)
       request = info_request_batch.info_requests = [
         FactoryGirl.create(:info_request, public_body: first_public_body)
       ]
-      unrequestable = info_request_batch.create_batch!
-      expect(unrequestable).to eq([second_public_body])
-      expect(info_request_batch.info_requests.size).to eq(1)
+      info_request_batch.create_batch!
+      expect(info_request_batch.info_requests.size).to eq(2)
```
garethrees: The expected value here has changed from `1` to `2`. Shouldn't it remain as `1` in this case?

To me, it looks like we start off with a Batch with a request sent to a PublicBody. We then call `create_batch!` again, presumably to check that the PublicBody doesn't receive the request again.

This spec implies that it _should_ send twice?

garethrees: Ah, so what's happening here is that we're now saying is:

Given the batch already has 1 request created,
When I send the batch
Then I'll have a _total_ of 2 requests sent (rather than 3, if we'd resent to the body of the existing request).

garethrees: I think this is kinda confusing.

I'd rewrite this spec entirely as:

```ruby
it 'does not resend requests to public bodies that have already received the request' do
  info_request_batch = FactoryBot.create(:info_request_batch, :sent)
  info_request_batch.create_batch!
  expect(info_request_batch.info_requests.count).to eq(1)
end
```

Requires https://github.com/mysociety/alaveteli/pull/4712

gbp: To make it even clearer I'm now checking `info_request_batch.info_requests.count` increases by 1. Changed in 590e1e9826eb6c365903ef6f1c3fa91f9809f444

spec/models/info_request_batch_spec.rb
gbp: Changed in 590e1e9826eb6c365903ef6f1c3fa91f9809f444

spec/integration/alaveteli_pro/request_list_spec.rb
```diff
@@ -35,6 +35,7 @@
     TestAfterCommit.with_commits(true) do
       batch_requests[0..3].each do |batch|
         batch.create_batch!
+        batch.update(sent_at: Time.now)
```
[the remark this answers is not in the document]
gbp: Changed in b5597f8fadcb334ae068bd21c5f9e67379f4a56d

spec/models/info_request_batch_spec.rb
```diff
@@ -443,6 +443,30 @@
     it { is_expected.to_not include(unrequestable_body) }
   end
 
+  describe '#unrequestable_public_bodies?' do
```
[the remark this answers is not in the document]
gbp: Changed in b10a3f9088d4337299e7c03f297ed0a96b45ff5f and fixed similar issue in 71f644fbce02d388fee6084f712e723d96a87080

gbp:
Noting that CI hasn't run as the PR isn't targeting `develop`. Once #4779 is merged this should be rebased and the base branch set to develop.

garethrees:
> Noting that CI hasn't run as the PR isn't targeting develop. Once #4779 is merged this should be rebased and the base branch set to develop.

Will review after this

Issue #75, opened 2011-07-07
Add voluntary moderation queue for people making FOI requests


skenaja:
Some people are not confident that the request they are making is well written. To combat this we would add an optional feature where your request is put in a public place for expert commentary and improvement before it is finalised and sent off.

The easiest way to build this system is probably not to try for a wiki approach, but instead simply allow comments that are attached to the letter and then mailed to the author.

(this was a WDTK enhancement feature, but would be useful in other installations).

https://github.com/mysociety/whatdotheyknow/issues/19
https://github.com/mysociety/whatdotheyknow/issues/12

brfairless:
This is something that Right to Know would find useful, both before and during the request process - https://github.com/mysociety/alaveteli/issues/2155

RichardTaylor:
Also known as draft requests. 

Bringing this feature into the site could be better than the option of directing users to a forum / Google group.

HelenWDTK:
This issue has been automatically closed due to a lack of discussion or resolution for over 12 months.
Should we decide to revisit this issue in the future, it can be reopened.

PR #4790, opened 2018-08-09
[WIP] Per authority pre-moderation of requests

gbp:
This was a spike to quickly see how much work it would be to add this sort of feature. Whether or not we want to add a feature like this and continue working on this is up for discussion.

## What does this do?

Adds per authority pre-moderation of requests. This applies only to authorities tagged with `review_requests`.

## Why was this needed?

Abuse and sockpuppet accounts.

## Screenshots

screen shot 2018-08-09 at 15 02 34
screen shot 2018-08-09 at 15 02 40

### DIFF
--- a/app/controllers/admin_outgoing_message_controller.rb
+++ b/app/controllers/admin_outgoing_message_controller.rb
@@ -1,7 +1,7 @@
 # -*- encoding : utf-8 -*-
 class AdminOutgoingMessageController < AdminController
 
-  before_filter :set_outgoing_message, :only => [:edit, :destroy, :update, :resend]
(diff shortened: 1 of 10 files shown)


garethrees:
Thanks for having a go at this.

We discussed this on the [volunteers call](https://docs.google.com/document/d/1NTEpl_8KRE-4ugSxMluz_UdOUT650eyi40BMM17oaTY/edit) last night, and came to the conclusion that its not really a desirable approach.

The main problem is that by taking responsibility for publishing in this way, we increase our risk in defamation and other legal cases.

Its also problematic because the user targets some popular authorities, so it would massively increase workload.

RichardTaylor:
Possible link to #75 (Add voluntary moderation queue for people making FOI requests)

Issue #3074, opened 2016-03-02
automate the delete user operation


davewhiteland:
Current process:

http://alaveteli.org/docs/running/admin_manual/#deleting-a-user

This is complicated, when really it's just clearing the user's data with a string like "[redundant account]" and setting the password to a random string. Should be a one-button operation in the admin. Or maybe a "deny login" setting if we want to get really fancy instead of the random password thing?

garethrees:
The `:delete => :nullify` association options would be worth looking at here

crowbot:
![](https://github.trello.services/images/mini-trello-icon.png) [Make it easier to delete a user from WDTK](https://trello.com/c/xPBFZZz6/35-make-it-easier-to-delete-a-user-from-wdtk)

crowbot:
See also https://github.com/mysociety/alaveteli/issues/444

garethrees:
Just a thought, maybe this should be "anonymise and close" rather than "close and anonymise" as the former is the order in which things are happening. Whatever, though :)

RichardTaylor:
If I'm interpreting the proposal correctly a censor rule for a user name of "Fred Bloggs won't catch a response addressed to "Mr Bloggs"; ideally the censor rule would be smart enough to catch some common variations of a name such as this. (This is discussed on #444 which has been linked already). 

On a separate point I note the new feature retains the original user name - both as a censor rule, and in the replacement for the email address - as per the initial ticket and the admin manual. This will be appropriate in the majority of cases where administrators want to anonomise an account, it may not though be sufficient where individuals exercise a right of erasure under data protection law. 

It might be appropriate to note in the admin interface next to the new "close and anonomise" button that the user name is retained. It needs to be clear to admins what the button does, and doesn't do.  This could be achieved with an in-line note, an information tip which appears on mouse-over, or a link to the relevant section of the admin manual.

RichardTaylor:
Something else to consider for a warning / tip - as with existing censor rules, admins should check - if required - to see if censor rules created via this automatic process have had the desired effect.

garethrees:
> On a separate point I note the new feature retains the original user name - both as a censor rule, and in the replacement for the email address

Hmm, yeah that would be correct for the name (not email though).

> ideally the censor rule would be smart enough to catch some common variations of a name such as this

https://github.com/basecamp/name_of_person might give us some nice quick wins there.

> It needs to be clear to admins what the button does, and doesn't do. This could be achieved with an in-line note…

[...]

PR #4806, opened 2018-08-14
[#3074] User anonymisation

gbp:
## Relevant issue(s)

Resolves initial and main part of #3074 

## What does this do?

- Adds a `closed_at` timestamp to the `User` model
- Update the `User#active?` to return false if the user account has been closed
- Adds `Close and anonymise` button in the admin UI for against active, non-pro, users.
- Creates censor rule before anonymising the user so that obvious uses of the user's name are removed automatically
- Adds missing functionality of hiding requests on profile and wall

## Why was this needed?

To make it easier to handle GDPR requests.

## Implementation notes

This copies the same general approach as banning users. We might want to consider hiding these buttons under an ability check and prevent `User#close_and_anonymise` calls if the user doesn't have the correct permissions.

The original issue (#3074) said to 404 when viewing a closed user's profile, but this felt weird in use. We show the account name as a link, and then you click it and get a 404, which just feels broken. It ended up being easier to hide the requests from view instead.

### DIFF
--- a/app/controllers/admin_users_account_suspensions_controller.rb
+++ b/app/controllers/admin_users_account_suspensions_controller.rb
@@ -5,23 +5,23 @@
 # Copyright (c) 2018 UK Citizens Online Democracy. All rights reserved.
 # Email: hello@mysociety.org; WWW: http://www.mysociety.org/
 class AdminUsersAccountSuspensionsController < AdminController
-  before_filter :set_banned_user
+  before_filter :set_suspended_user
   before_filter :set_suspension_reason
 
   def create
-    if @banned_user.update(ban_text: @suspension_reason)
+    if @suspended_user.update(ban_text: @suspension_reason)
       flash[:notice] = 'The user was suspended.'
     else
       flash[:error] = 'Something went wrong. The user could not be suspended.'
     end
 
-    redirect_to admin_user_path(@banned_user)
+    redirect_to admin_user_path(@suspended_user)
   end
 
   private
 
-  def set_banned_user
-    @banned_user = User.find(params[:user_id])
+  def set_suspended_user
+    @suspended_user = User.find(params[:user_id])
   end
 
   def set_suspension_reason
--- a/app/controllers/admin_users_account_suspensions_controller.rb
+++ b/app/controllers/admin_users_account_suspensions_controller.rb
@@ -9,7 +9,7 @@ class AdminUsersAccountSuspensionsController < AdminController
   before_filter :set_suspension_reason
 
   def create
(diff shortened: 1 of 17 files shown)


app/models/user.rb
```diff
@@ -442,8 +445,25 @@ def banned?
     !ban_text.empty?
   end
 
+  def closed?
+    closed_at.present?
+  end
+
+  def close_and_anonymise
+    sha = Digest::SHA1.hexdigest(rand.to_s)
+    update(
```
garethrees: We'll need to create the censor rule before we anonymise, otherwise we'll lose the original name that we need to redact.

app/controllers/admin_users_account_suspensions_controller.rb
```diff
@@ -5,23 +5,23 @@
...
-    redirect_to admin_user_path(@banned_user)
+    redirect_to admin_user_path(@user)
   end
 
   private
 
-  def set_banned_user
-    @banned_user = User.find(params[:user_id])
+  def set_user
+    @user = User.find(params[:user_id])
```
garethrees: `@user` is reserved for the currently logged in user, which is a bit of an Alaveteli gotcha.

I suggest renaming to `@suspended_user`; easiest to do this with a few edits when rebasing instead of fixups, given later changes.

app/assets/javascripts/admin/admin.coffee
```diff
@@ -89,4 +89,6 @@ jQuery ->
     console.log 'click'
     parent = $(this).parents(".#{$(this).data('dismiss')}")
     parent.hide('slow')
-
+  $('.btn.disabled').on 'click', (e) ->
```
garethrees: Implemented this in a better way in 9b78319fa3ea572f2c44cd8a7c9e2bf2b538ef42 and adb38369c54df805ecfdee4f0047d000329c7dae

spec/helpers/admin_users_helper_spec.rb
```diff
@@ -18,14 +18,22 @@
...
+    it 'adds a closed label if the user is banned' do
+      user =  FactoryBot.create(:user, closed_at: Time.zone.now)
+      html = %q(<span class="label label-warning">closed</span>)
+      expect(user_labels(user)).to eq(html)
+    end
+
     it 'adds labels for all noteworthy attributes' do
-      user = FactoryBot.create(:admin_user, :ban_text => 'Banned')
+      user = FactoryBot.create(:admin_user, ban_text: 'Banned',
+                               closed_at: Time.zone.now)
```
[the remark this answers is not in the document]
garethrees: Fixed in 3d6c67ba8

PR #4807, opened 2018-08-14
[pro#549] apostrophe encoding

### LINKED ISSUE #4793: Encoding issues with user and body names
Users with names containing an apostrophe have their names displayed with an encoding error in a number of circumstances including the listing on /admin/users and in the draft messages used when an admin hides a request. 

There are user facing issues too, and encoded strings are ending up in correspondence, as noted at: 

https://github.com/mysociety/alaveteli-professional/issues/549

(I thought this issue should be tracked publicly as well as where it impacts Pro users)

screen shot 2018-08-12 at 20 28 43

lizconlan:
## What does this do?

1. Prevents appending text to the outgoing message body from triggering HTML escaping of the existing text (only applies to logged in users) screen shot 2018-08-07 at 18 41 33 

2. Alters the preview templates on the pro side to bring them in line with the standard ones (the HTML was not being escaped before display before)
screen shot 2018-08-17 at 16 04 08

3. Prevent apostrophe's being displayed as `&#39;` (and the whole message being wrapped in an HTML template) when an admin is sending an email as part of marking a request as vexatious/not_foii screen shot 2018-08-12 at 20 28 43

4. Fixes the display of user names with apostrophes in the admins' user list screen shot 2018-08-17 at 11 08 23

## Why was this needed?

Fixes some of the inconsistencies of how apostrophes (and other special characters such as ampersands) are displayed across the site.

## Notes to reviewer

* Are the pro request preview changes correct or was the behaviour difference intentional?
* Is removing the HTML wrapper from the admin email textarea correct? (4410611)

## Relevant issue(s)

Fixes mysociety/alaveteli-professional#549
Fixes #4793


app/views/admin_request/hidden_user_explanation.html.erb
```diff
@@ -1,4 +1,4 @@
-<%= _("Dear {{name}},", :name => name_to) %>
+<%= _("Dear {{name}},", :name => name_to.html_safe) %>
```
garethrees: We're going to try making this a text template rather than html.

app/models/outgoing_message.rb
```diff
@@ -94,7 +94,7 @@ def get_default_message
   def set_signature_name(name)
     # TODO: We use raw_body here to get unstripped one
     if raw_body == get_default_message
-      self.body = raw_body + name
+      self.body = "#{raw_body}#{name}".html_safe
```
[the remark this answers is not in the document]
garethrees: We think a better solution is:

```diff
diff --git a/app/models/outgoing_message.rb b/app/models/outgoing_message.rb
index 560a23a59..1cf33c71d 100644
--- a/app/models/outgoing_message.rb
+++ b/app/models/outgoing_message.rb
@@ -94,7 +94,7 @@ class OutgoingMessage < ActiveRecord::Base
   def set_signature_name(name)
     # TODO: We use raw_body here to get unstripped one
     if raw_body == get_default_message
-      self.body = raw_body + name
+      self.body = get_default_message + name
     end
   end
```

app/controllers/admin_request_controller.rb
```diff
@@ -40,8 +40,10 @@ def show
                             :name_to => @info_request.user_name,
                             :name_from => AlaveteliConfiguration::contact_name,
                             :info_request_url => request_url(@info_request, :only_path => false)}
-    @request_hidden_user_explanation = render_to_string(:template => "admin_request/hidden_user_explanation",
-                                                        :locals => vars_for_explanation)
+    @request_hidden_user_explanation =
+      render_to_string(template: "admin_request/hidden_user_explanation",
```
garethrees: If we make the template a text template rather than html we probably won't need this change.

app/views/admin_request/hidden_user_explanation.html.erb
lizconlan: Waiting to see if the changes I've made so far break anything I'd not thought about. Then I think the best place to test the textarea/email to user to say their request has been hidden as vexatious content is the `ServicesController` spec

app/views/admin_user/_user_table.html.erb
```diff
@@ -7,7 +7,7 @@
         <span class="user-labels">
           <%= user_labels(user) %>
         </span>
-        <%= link_to("#{ h(user.name) }", admin_user_path(user)) %>
+        <%= link_to("#{ user.name }", admin_user_path(user)) %>
```
garethrees: This doesn't need to be interpolated in to a string.

spec/views/admin_user/_user_table.html.erb_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+    allow(user_array).to receive(:total_pages).and_return(1)
+    user_array
+  end
+
+  it 'does not double escape apostrophes' do
+    allow(controller).to receive(:current_user).and_return(admin_user)
+    render partial: 'admin_user/user_table.html.erb',
+           locals: { users: users,
+                     banned_column: false }
+    expect(rendered).not_to match("O&amp;#39;Toole")
```
garethrees: Would be clearer and more robust to `expect(rendered).to match("O'Toole")`. That way the spec would fail if the behaviour was incorrect, with a diff of expected vs actual (and the actual would be the incorrect double escaped string).

spec/controllers/services_controller_spec.rb
```diff
@@ -1,83 +1,94 @@
-# -*- encoding : utf-8 -*-
-require File.expand_path(File.dirname(__FILE__) + '/../spec_helper')
+require 'spec_helper'
```
garethrees: 1-line commit messages are rarely useful, unless perhaps the change is a _really_ simple change.

This should explain _what_ has been refactored and _why_.

I have no idea what I'm looking for when reviewing this 😞

spec/controllers/services_controller_spec.rb
```diff
@@ -1,83 +1,94 @@
-# -*- encoding : utf-8 -*-
```
garethrees: why has this been removed?

spec/controllers/services_controller_spec.rb
garethrees: After figuring out what's changed, here's what I'd have written for this commit message:

```
* Simplify `require` of spec helper; can just reference the file by name as its already in the load path
* Make the top level `describe` block describe only the class we're testing
* Group specs under `describe '#method_name' do` block
* Remove line breaks from spec descriptions
* Reduce line lengths
```

app/controllers/services_controller.rb
```diff
@@ -55,14 +55,15 @@ def other_country_message
...
-    render :template => "admin_request/hidden_user_explanation",
-      :content_type => "text/plain",
-      :layout => false,
-      :locals => {:name_to => info_request.user_name,
-                  :name_from => AlaveteliConfiguration.contact_name,
-                  :info_request => info_request, :reason => params[:reason],
-                  :info_request_url => 'http://' + AlaveteliConfiguration.domain + request_path(info_request),
-                  :site_name => site_name}
+    render template: "admin_request/hidden_user_explanation",
+           formats: [:text],
```
garethrees: Why has this changed?

app/controllers/services_controller.rb
```diff
@@ -55,14 +55,14 @@ def other_country_message
...
+           layout: false,
+           locals: {
+             name_to: info_request.user_name.html_safe,
+             info_request: info_request, :reason => params[:reason],
+             info_request_url: 'http://' + AlaveteliConfiguration.domain + request_path(info_request),
```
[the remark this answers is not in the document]
garethrees: Should be able to change this to `request_url(info_request)` to clean this up and fix the line-length issue. I'd make a new commit for that.

app/controllers/services_controller.rb
```diff
@@ -55,14 +55,15 @@ def other_country_message
...
-                  :info_request => info_request, :reason => params[:reason],
-                  :info_request_url => 'http://' + AlaveteliConfiguration.domain + request_path(info_request),
-                  :site_name => site_name}
+    render template: "admin_request/hidden_user_explanation",
+           formats: [:text],
+           layout: false,
+           locals: {
+             name_to: info_request.user_name,
+             name_from: AlaveteliConfiguration.contact_name,
+             info_request: info_request, :reason => params[:reason],
```
garethrees: You missed `:reason => params[:reason]` – should drop that on to a new line and make it a 1.9-style hash

app/controllers/services_controller.rb
```diff
@@ -59,11 +59,11 @@ def hidden_user_explanation
            formats: [:text],
            layout: false,
            locals: {
-             name_to: info_request.user_name,
+             name_to: info_request.user_name.html_safe,
```
garethrees: I think we should qualify (in the commit message) that as this is rendering `text/plain`, malicious user input won't be rendered as HTML so `<script>` tags can't be injected.

app/controllers/services_controller.rb
```diff
@@ -115,4 +115,14 @@ def ask_the_eu_link
     %q(<a href="http://asktheeu.org">Ask The EU</a>)
   end
 
+  def request_link(request_path)
```
garethrees: 👎 – pretty sure the following would fix this:

```diff
diff --git a/app/controllers/services_controller.rb b/app/controllers/services_controller.rb
index 079f69d7f..76771a6c0 100644
--- a/app/controllers/services_controller.rb
+++ b/app/controllers/services_controller.rb
@@ -61,7 +61,7 @@ class ServicesController < ApplicationController
            locals: {
              name_to: info_request.user_name.html_safe,
              info_request: info_request, :reason => params[:reason],
-             info_request_url: request_link(request_path(info_request)),
+             info_request_url: request_url(info_request),
              site_name: site_name.html_safe }
   end
```

The specs do fail, but that's because `config.force_ssl` is  set in the environment initialiser so can't be stubbed.

Its worth checking that I'm not talking rubbish here, but I'm pretty sure the helpers take protocol in to account…

spec/controllers/services_controller_spec.rb
lizconlan: sorry 😞 - I'll belatedly squash that in as penance (also practise)

app/controllers/services_controller.rb
lizconlan: Gah, another missed commit comment _tracks down additional detail_

Without it, it defaults to `:formats=>[:html]` and doesn't find the renamed text template (raising an `ActionView::MissingTemplate` error). Prior to renaming, it rendered it as HTML but would have produced plain text output when the original code was written (for Rails 2.3) .

The `formats` option is described in  [the current documentation](https://guides.rubyonrails.org/layouts_and_rendering.html#the-formats-option) but isn't mentioned for Rails 4.2, despite apparently having been there from Rails 4.0. Before then, `content_type` would have done the trick as the format took its cue from the MIME type - https://github.com/rails/rails/commit/67b2404cf9b2c7e0f86bf0294571ef97391a6dcd

[This space left blank while I try to work out whether I need to set them both]

spec/controllers/services_controller_spec.rb
gbp: These aren't needed in Ruby 2.0+ and as #4209 is closed I think its fine to get rid of these. One of RuboCop checks I was going to suggest we enable would enforce getting rid of these lines.

PR #4812, opened 2018-08-15
Allow users to switch status with keyboard

wrightmartin:

## Relevant issue(s)

Resolves some issues with keyboard operation (see https://github.com/mysociety/alaveteli-professional/issues/555 for more details)

## What does this do?
Allows the user to switch status with keyboard (on the pro request page)

## Notes to reviewer
While working on this I've found a bug with the houdini component we're using for controls like these. I'll open a ticket with details, but the tldr is that it's got some issues with keyboard control that we need to look into

### DIFF
--- a/app/assets/stylesheets/responsive/_global_style.scss
+++ b/app/assets/stylesheets/responsive/_global_style.scss
@@ -464,6 +464,11 @@ button,
 .houdini-label {
   color: #2688dc;
   text-decoration: underline;
+    &:hover,
+    &:active,
+    &:focus {
+      color: #000;
+    }
 }
 
 .houdini-target {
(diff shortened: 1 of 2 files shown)


wrightmartin:
> In Chrome there's no visual sign that either of the sidebar options are selected and after a bit of guesswork I realised it opened with the spacebar instead of the enter key. (Safari - note to self Alt + Tab - just skips them unless I hack the view to add tabindex=0, but then it won't do anything with the link.)

> The status elements are now selectable - and clearly indicated - but I haven't managed to choose any of them (space scrolls the screen, enter is ignored) without cheating and using the mouse.

@lizconlan you're correct, I discovered this while working on it and ticketed it here https://github.com/mysociety/alaveteli/issues/4813. This PR does what it says on the tin for now (changes the status options to be keyboard selectable) but your point stands - you can't actually get to them with a keyboard without guessing, which is sadly a bigger issue with Houdini

lizconlan:
> it says on the tin for now (changes the status options to be keyboard selectable) but your point stands - you can't actually get to them with a keyboard without guessing, which is sadly a bigger issue with Houdini

They're highlightable but not actually selectable as you can't choose them

PR #4835, opened 2018-08-30
[pro#497] Distinguish requests vs batches in pro request list

garethrees:
## Relevant issue(s)

Fixes https://github.com/mysociety/alaveteli-professional/issues/497.

## What does this do?

This commit wraps each item in the list in a bounding box, making it
really clear that items are distinct.

## Why was this needed?

When an individual request appears below a batch request in the pro
request list, the individual request visually appears to be _part of the
batch_. This is made even more extreme when the batch is public and the
individual request is private, due to the padlock icon indenting the
individual request title.

## Implementation notes

Does not address https://github.com/mysociety/alaveteli-professional/issues/561, which will be
addressed separately and make the spacing look better here.

### Before

screen shot 2018-08-30 at 17 20 38

### After

screen shot 2018-08-30 at 17 21 22

## Notes to reviewer

A thicker border-bottom looked messy, as it conflicted with the progress
bar of the batch requests.

### DIFF
--- a/app/assets/stylesheets/responsive/alaveteli_pro/_requests_style.scss
+++ b/app/assets/stylesheets/responsive/alaveteli_pro/_requests_style.scss
@@ -10,7 +10,9 @@ $complete: rgb(112, 203, 99);
 }
 
 .request {
-  border-bottom: 1px solid #ddd;
+  border: 3px solid #e6e4df;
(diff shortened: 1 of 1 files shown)


gbp:
Oooh box shadow looks nice

PR #4854, opened 2018-09-14
Improve loading sample data for non 'en' default locales

lizconlan:
## What does this do?

Loading the sample data with the site's default locale not set to 'en'
should now copy the default (English) names for PublicBody,
PublicBodyCategory and PublicBodyHeading to the translation tables for
the current default so that the search indexing scripts and demo site
work as expected.

## Why was this needed?

When trying to replicate a live site config (e.g. Belgium's) rather than lazily setting English as the default language for my own convenience, and running `be ./script/load-sample-data`, the result is a test site without any valid authority or request data.

Found the cause by accident when attempting to run (destructive, don't run against a search index you'd like to keep) `rake xapian:destroy_and_rebuild_index models="PublicBody"` which crashed out as with a non-English default locale, all the public body names are set to nil.


lizconlan:
Sorry - was trying to do too many things at once and messed up the rebase :(

garethrees:

PR #4866, opened 2018-09-26
[#4160] duplicate requests

### LINKED ISSUE #4160: Uncaught duplicate requests
This request: https://www.whatdotheyknow.com/request/use_of_public_cloud_4 and this one: https://www.whatdotheyknow.com/request/use_of_public_cloud_3 appear to be identical. They are from the same user, to the same authority, with the same subject line and with (so far as I can determine including via a file compare utility) exactly the same content (sent 14 minutes apart.) I think that technical measures to prevent such may have failed.

gbp:
## Relevant issue(s)

Fixes #4160

## What does this do?

* Extract existing duplication detection (that already takes into account whitespace differences) to a new `OutgoingMessage.with_body` scope
* Use this new scope in `InfoRequest.find_existing`.

## Why was this needed?

Help prevent the same duplicate requests from being sent the authorities.

## Implementation notes

As with the original implementation this only strips whitespace when using Postgres.


spec/models/outgoing_message_spec.rb
```diff
@@ -36,25 +36,32 @@
...
-      if ActiveRecord::Base.connection.adapter_name != 'PostgreSQL'
-        skip 'These tests only work for PostgreSQL'
-      end
-    end
-
-    it 'should match message body when whitespace is ignored' do
-      expect(OutgoingMessage.with_body('foobar')).to include(message)
-      expect(OutgoingMessage.with_body('foo bar')).to include(message)
-      expect(OutgoingMessage.with_body("foo\nbar")).to include(message)
+    it 'should mreturn message if body matches exactly' do
```
garethrees: minor typo `mreturn`

PR #4872, opened 2018-09-28
[#4871] Handle nil track_medium param in TrackController#update

### LINKED ISSUE #4871: [ERROR] track#update (TypeError) "no implicit conversion of nil into String"
```
A TypeError occurred in track#update:

no implicit conversion of nil into String
app/controllers/track_controller.rb:212:in `+'


Request:

* URL : https://www.whatdotheyknow.com/track/update/81443
* HTTP Method: GET
* IP address : REDACTED
* Parameters : {"controller"=>"track", "action"=>"update", "track_id"=>"81443"}
* Timestamp : 2018-09-27 17:31:46 +0100
* Rails root : /REDACTED/alaveteli-2018-09-05T08-32-15

Backtrace:

app/controllers/track_controller.rb:212:in `+'
app/controllers/track_controller.rb:212:in `update'
```

garethrees:
## Relevant issue(s)

Fixes #4871 

## What does this do?

* Minor refactoring
* Handle a missing param rather than only an invalid param

## Why was this needed?

I expect what's happened in https://github.com/mysociety/alaveteli/issues/4871 is someone has mistakenly "paused" a form submission (clicking "stop" or "back"?) and ended up on https://www.whatdotheyknow.com/track/update/81443. They've then hit reload and as the action responds to GET it submits with no params.

We don't handle that error very well – we try to concatenate `nil` in to a `String` which raises a `TypeError`. This handles that more gracefully.

We'll still get an error if someone does this again, but it will be much more descriptive. To prevent us getting errors at all, we should remove the ability for the action to respond to `GET`.

### DIFF
--- a/app/controllers/track_controller.rb
+++ b/app/controllers/track_controller.rb
@@ -211,7 +211,7 @@ class TrackController < ApplicationController
     else
       msg =
         if new_medium
-          "new medium not handled #{ new_medium }"
+          "Given track_medium not handled: #{ new_medium }"
         else
           'No track_medium supplied'
         end
--- a/app/controllers/track_controller.rb
+++ b/app/controllers/track_controller.rb
@@ -209,7 +209,14 @@ class TrackController < ApplicationController
       flash[:notice] = view_context.unsubscribe_notice(track_thing)
       redirect_to SafeRedirect.new(params[:r]).path
     else
-      raise "new medium not handled " + new_medium
+      msg =
+        if new_medium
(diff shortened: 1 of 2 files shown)


garethrees:
I was going to fix the codeclimate complexity warning here. The "fix" is to refactor it in to a `destroy` method and remove all the junky `track_medium` stuff, as that's the only thing its doing. I've made a stab at this but there are a few complicated bits, so I'm going to merge this as-is and push a new branch in continuation.

PR #4880, opened 2018-10-03
Use sanitize to strip malicious code from admin announcements

### LINKED ISSUE #4877: Fix potential cross-site scripting vulnerability
Found by brakeman during release prep:

```
{
      "warning_type": "Cross-Site Scripting",
      "warning_code": 2,
      "fingerprint": "8eea4b2cd9e4ca9afcca3bb6b26cbbea72c9bdcbc6216e5212da68b2e9b11714",
      "check_name": "CrossSiteScripting",
      "message": "Unescaped model attribute",
      "file": "app/views/admin_general/_announcement.html.erb",
      "line": 8,
      "link": "https://brakemanscanner.org/docs/warning_types/cross_site_scripting",
      "code": "Announcement.new.content",
      "render_path": [
        {
          "type": "controller",
          "class": "AdminGeneralController",
          "method": "index",
          "line": 70,
          "file": "app/controllers/admin_general_controller.rb"
        },
        {
          "type": "template",
          "name": "admin_general/index",
          "line": 1,
          "file": "app/views/admin_general/index.html.erb"
        }
      ],
      "location": {
        "type": "template",
        "template": "admin_general/_announcement"
      },
      "user_input": null,
      "confidence": "High"
    }
```

Required by #4615

lizconlan:
## Relevant issue(s)

Fixes #4877 

## What does this do?

Uses sanitize when rendering the announcement content rather than putting raw HTML onscreen unfiltered.

## Why was this needed?

To guard against a potential cross-site scripting attack, as notified by brakeman.

## Implementation notes

The standard set of `sanitized_allowed_tags` from the action_view config should be fine here as we're mostly screening out iframes & script tags.

### DIFF
--- a/spec/views/admin_general/_announcement.html.erb_spec.rb
+++ b/spec/views/admin_general/_announcement.html.erb_spec.rb
@@ -1,0 +1,29 @@
+require 'spec_helper'
+
+describe 'admin_general/_announcement.html.erb' do
+
+  let(:announcement) do
+    FactoryBot.
+      create(:announcement,
+             visibility: 'admin',
+             content: '<b>Exciting news!</b> <script>alert("!")</script>')
+  end
+
+  describe 'displaying an announcement' do
+
+    it 'shows the announcement' do
+      render template: 'admin_general/_announcement',
+             locals: { announcement: announcement }
+      expect(rendered).to include('Introducing projects')
+      expect(rendered).to include('<b>Exciting news!</b>')
+    end
+
+    it 'strips out dangerous tags' do
+      render template: 'admin_general/_announcement',
+             locals: { announcement: announcement }
+      expect(rendered).not_to include('<script>')
+    end
+
+  end
+
+end
--- a/app/views/admin_general/_announcement.html.erb
+++ b/app/views/admin_general/_announcement.html.erb
@@ -5,7 +5,7 @@
         <span aria-hidden="true">&times;</span>
       <% end %>
       <% if announcement.title.present? %><strong><%= announcement.title %> - </strong><% end %>
-      <%= raw announcement.content %>
+      <%= sanitize announcement.content %>
     </div>
   </div>
 </div>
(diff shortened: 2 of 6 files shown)


spec/views/admin_general/_announcement.html.erb_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+             locals: { announcement: announcement }
+      expect(rendered).to match('Introducing projects')
+    end
+
+    it 'strips out dangerous tags' do
+      render template: 'admin_general/_announcement',
+             locals: { announcement: announcement }
+      expect(rendered).to match('<b>Exciting news!</b>')
+      expect(rendered).not_to match('<script>')
+      expect(rendered).not_to match('alert("!")')
```
garethrees: Not sure why this is passing:

screen shot 2018-10-05 at 11 43 31

lizconlan: ah... I used `match` which take a regex so I'd have needed to say something like `not_to match(/alert\(\"!\"\)/)`\\

so it now reads: `not to include('alert("!")')` as that's easier to make sense of

spec/views/admin_general/_announcement.html.erb_spec.rb
```diff
@@ -2,10 +2,6 @@
 
 describe 'admin_general/_announcement.html.erb' do
 
-  def render_view
-    render :partial => 'admin_general/announcement'
-  end
-
```
lizconlan: Why did I do that?

lizconlan:
I thought the CI fails were potentially instructive as they'd managed to hit 2 different types of transient in a single run

garethrees:
> I thought the CI fails were potentially instructive as they'd managed to hit 2 different types of transient in a single run

Oh, I didn't realise sorry. Would have been worth mentioning that on the issue or documenting them elsewhere before assigning for review.

lizconlan:
> > I thought the CI fails were potentially instructive as they'd managed to hit 2 different types of transient in a single run
> 
> Oh, I didn't realise sorry. Would have been worth mentioning that on the issue or documenting them elsewhere before assigning for review.

Probably not that interesting tbh, I'm pretty sure we've seen both types before. Xapian things and missing id numbers, nothing new.

PR #4883, opened 2018-10-04
Update links to fonts.googleapis.com to use HTTPS

sagepe:
As we switch to enforcing HTTPS, these generate mixed-content warnings.

The only thing flagged as failing to load due to mixed content in my console were these links, so I think that's all that's needed. 

Fixes https://github.com/mysociety/sysadmin/issues/1067

### DIFF
--- a/_layouts/es/default.html
+++ b/_layouts/es/default.html
@@ -7,7 +7,7 @@
     <link href="/atom.xml" rel="alternate" title="Alaveteli" type="application/atom+xml">
     <link rel="icon" type="image/png" href="/assets/img/favicon.png" />
     <meta name="viewport" content="initial-scale=1">
-    <link href='http://fonts.googleapis.com/css?family=Source+Sans+Pro:400,600,700,900,400italic&subset=latin,latin-ext' rel='stylesheet' type='text/css'>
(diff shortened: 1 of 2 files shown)


_layouts/es/default.html
```diff
@@ -7,7 +7,7 @@
     <link href="/atom.xml" rel="alternate" title="Alaveteli" type="application/atom+xml">
     <link rel="icon" type="image/png" href="/assets/img/favicon.png" />
     <meta name="viewport" content="initial-scale=1">
-    <link href='http://fonts.googleapis.com/css?family=Source+Sans+Pro:400,600,700,900,400italic&subset=latin,latin-ext' rel='stylesheet' type='text/css'>
+    <link href='https://fonts.googleapis.com/css?family=Source+Sans+Pro:400,600,700,900,400italic&subset=latin,latin-ext' rel='stylesheet' type='text/css'>
```
garethrees: I _think_ we should be able to do `<link href='//fonts…` as we've done for `<script src="//ajax.googleapis.com/ajax/libs/jquery/1.11.1/jquery.min.js"></script>` a few lines below?

garethrees: Actually its now considered bad practice to do this https://webhint.io/docs/user-guide/hints/hint-no-protocol-relative-urls/#why-is-this-important, so I think this is fine.

PR #4889, opened 2018-10-09
Update reCAPTCHA key info

lizconlan:
## What does this do?

* Alters the config documentation to reflect the changed name of the reCAPTCHA config keys
* Replaces the outdated link to recaptcha.net

## Why was this needed?

With release 0.32 we've introduced a new version of the `recaptcha` gem which uses different terminology of the reCAPTCHA public and private keys (now 'site' and 'secret'); we updated our code to match so the docs are out of date.

Also while editing noticed that the recaptcha.net page we were linking to no longer exists.

## Notes to reviewer

Should we add a sentence "previously called..." to the docs for re-users on older versions?

### DIFF
--- a/docs/customising/config.md
+++ b/docs/customising/config.md
@@ -1179,10 +1179,12 @@ href="#smtp_mailer_enable_starttls_auto">SMTP_MAILER_ENABLE_STARTTLS_AUTO</a>.
       <p>Examples:</p>
       <ul class="examples">
         <li>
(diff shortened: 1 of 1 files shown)


lizconlan:
> > Should we add a sentence "previously called..." to the docs for re-users on older versions?
>
> I don't think its vital, but feel free if you have something quick and easy in mind.

I'm attempting to be in writing rather that coding mode this morning so I'll see what I can come up with. (Otherwise I'll merge it as-is.)

Issue #3148, opened 2016-04-04
Review and restrict what banned users are able to do


crowbot:

RichardTaylor:
An administrator could change an account password when banning; this might have the desired effect without needing a coding change?

Sometimes on WhatDoTheyKnow we edit the email address field ourselves to prevent suspended users logging in. 

It's akin to two levels of suspension/ban; one is the standard effect of a suspension to ban a user making requests/annotations but letting them log in to manage alerts, view any material that's generally hidden but accessible to them etc. then the next level, if we change the email/password on an account is a total ban on being able to log in at all.

RichardTaylor:
Banned users can also currently classify requests; we might want a ban to prevent that too.

garethrees:
> An administrator could change an account password when banning; this might have the desired effect without needing a coding change?

Yes, but they could easily reset their password via "Forgotten your password?".

garethrees:
Some initial thoughts on this. I'll flesh this out at some point, but I have some immediate things I'm going to do before thinking too much about it.

---

There are two reasons why we might want to limit access to Alaveteli.

**Banning:** Temporarily limiting new content creation while we correspond with a user

**Closing:** All correspondence has failed, so we want to permanently deactivate the account (or its just spam).

Currently, we can only **ban** users.

We should reflect these two tiers in the codebase.

## Banning

* Stops users creating public content
* Should also prevent classification, but we don't

## Closing

* `User` now has a `closed_at` attribute
* Prevent _any_ action by closed users. No sign in; stop sending emails, etc

garethrees:
I've updated the comment above to reflect what I think each of these should do

RichardTaylor:
>they could easily reset their password via "Forgotten your password?"

Should banned users be prevented from changing their password then?

garethrees:
> Should banned users be prevented from changing their password then?

Well, perhaps. Right now we still allow banned users to log in and perform some actions. I think we're converging on a [two tier system](https://github.com/mysociety/alaveteli/issues/3148#issuecomment-428232578).

HelenWDTK:
This issue has been automatically closed due to a lack of discussion or resolution for over 12 months.
Should we decide to revisit this issue in the future, it can be reopened.

PR #4890, opened 2018-10-09
[#3148] Prevent closed users signing in

garethrees:
## Relevant issue(s)

https://github.com/mysociety/alaveteli/issues/3148

## What does this do?

* Refactoring
* Prevents users with a populated`closed_at` attribute signing in

## Why was this needed?

Quickly enable us to prevent a few specific users from signing in.

## Implementation notes

`User.authenticate_from_form` is too complex, so needs refactoring, but refactoring the specs was a first step at least.

## Screenshots

screen shot 2018-10-09 at 16 45 35

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -179,6 +179,7 @@ class User < ActiveRecord::Base
       end
 
       if user.has_this_password?(params[:password]) && user.closed?
+        logger.info "Closed user attempted login: #{ params[:email] }"
         user.errors.add(:base, _('This account has been closed.'))
       end
     else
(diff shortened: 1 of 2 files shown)


app/models/user.rb
```diff
@@ -177,6 +177,10 @@ def self.authenticate_from_form(params, specific_user_login = false)
       unless user.has_this_password?(params[:password])
         user.errors.add(:base, auth_fail_message)
       end
+
+      if user.closed?
+        user.errors.add(:base, _('This account has been closed.'))
```
gbp: Not sure this message should be shown without the correct password being entered for the user account. Could combine with the conditional above to prevent this.

garethrees: +1, Fixed in f8cbe27

Issue #3969, opened 2017-05-10
Upgrade to Rails 5.0


garethrees:

garethrees:
https://gemnasium.com/blog/how-to-deal-with-major-ruby-on-rails-upgrades/

garethrees:
Update column default values https://stackoverflow.com/a/44115924/387558

garethrees:
https://shopifyengineering.myshopify.com/blogs/engineering/upgrading-shopify-to-rails-5-0

garethrees:
https://kickstarter.engineering/upgrading-kickstarter-to-rails-5-e8203f93df55

garethrees:
Mail gem is now updated https://github.com/mysociety/alaveteli/pull/4422

gbp:
Query objects (E.G: `ProQuery.new` & `Prominence::PublicQuery.new`) don't seem to be chainable (E.G: `InfoRequest.pro.is_public`) in Rails 5 so will need to be rewritten

garethrees:
https://www.ombulabs.com/blog/rails/upgrades/upgrade-rails-from-4-2-to-5-0.html

gbp:
Depends on #4542 as:


```
Gem::InstallError: activesupport requires Ruby version >= 2.2.2.
An error occurred while installing activesupport (5.0.7), and Bundler cannot continue.
Make sure that `gem install activesupport -v '5.0.7'` succeeds before bundling.
```

garethrees:
> There is still one deprecation warning for `#table_exists?` coming from the `acts_as_versioned` gem - we might need patch this or look at alternative versioning gems.
>
> – https://github.com/mysociety/alaveteli/pull/5208#issue-282003915

PR #4893, opened 2018-10-11
Rails 5/gemfile and ci

gbp:
## Relevant issue(s)

Connects to #3969 

## What does this do?

Adds a new Rails 5 Gemfile and setup Travis to test (with failures allowed). Commands can be run under Rails 5 by setting `BUNDLE_GEMFILE=Gemfile.rails5`. Currently the Rails 5 doesn't boot.

## Why was this needed?

To make a start upgrading Rails.


lizconlan:
> Coverage decreased (-4.0%) to 88.442%

Not the end of the world but if there's a quick fix... (although a very quick look suggests there may not be)

gbp:
> > Coverage decreased (-4.0%) to 88.442%  ([Details](https://coveralls.io/builds/19473997))
> 
> Not the end of the world but if there's a quick fix... (although a very quick look suggests there may not be)

Huh - that is strange. Not sure why it has dropped so much but I think it could be due less builds passing on Travis than there were previously.

lizconlan:
> > > Coverage decreased (-4.0%) to 88.442% (Details)
> >
> > Not the end of the world but if there's a quick fix... (although a very quick look suggests there may not be)
> 
> Huh - that is strange. Not sure why it has dropped so much but I think it could be due less builds passing on Travis than there were previously.

I don't think this is about failing tests, I wonder if it's treating the `Gemfile`s as testable code and including it in the lines of "code" without test coverage.

Running a quick test over lunch to see if appending `,Gemfile*` to `spec/rspec.opts` is useful here

lizconlan:
Yep...

spec/rcov.opts
```diff
- --exclude "spec/*,gems/*"
+ --exclude "spec/*,gems/*,Gemfile*"
```

Should do the trick :) (I'll delete mine, not convinced by my rebase onto develop)

lizconlan:
Note to self (as reviewer) - double-check whether this causes Gemfile.lock issues (or problems associated with conditionals in Gemfiles, there might be pre-existing tickets)

PR #4895, opened 2018-10-12
[Rails 5] Stub TestAfterCommit

gbp:
## Relevant issue(s)

Connects to #3969 
Requires #4893 

## What does this do?

Stubs TestAfterCommit for Rails 5 to just call the inner block regardless. This allows us to keep the existing spec code for Rails 4 compatibility. Once Rails 5 is merged we'll need to remove the stub and go through the spec suite to remove TestAfterCommit blocks.

## Why was this needed?

Rails 5 doesn't need TestAfterCommit as `after_commit` callbacks are automatically run.

## Notes to reviewer

Unused if having this in the initializer directory is the best places for these stubs. Happy to move if there is a better place.


gbp:
CI hasn't run yet as this isn't targeting `develop` branch yet.

PR #4899, opened 2018-10-12
[WIP][#4871] Add TrackController#destroy

garethrees:
## Relevant issue(s)

Initiated by #4871
Continuation of https://github.com/mysociety/alaveteli/pull/4872

## What does this do?

Add and use `Add TrackController#destroy` for unfollowing.

## Why was this needed?

https://github.com/mysociety/alaveteli/pull/4872 increased the code complexity of `TrackController#update` to an unacceptable level. Also, `update` should not be used for destroying records.

## Implementation notes

Still WIP because testing that unfollowing via a widget needs some further checking.

### DIFF
--- a/app/assets/stylesheets/responsive/_user_layout.scss
+++ b/app/assets/stylesheets/responsive/_user_layout.scss
@@ -9,3 +9,13 @@
     width: 300px;
   }
 }
+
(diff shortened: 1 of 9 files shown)


garethrees:

PR #4900, opened 2018-10-12
[Rails 5] Update functional test requests calls

gbp:
## Relevant issue(s)

Connects to #3969 
Requires #4893 

## What does this do?

Updates functional test requests calls to keyword argument style.

## Why was this needed?

So many deprecation warnings without this its was impossible to see what is really broken.


lizconlan:
Closed in favour of #4929

PR #4901, opened 2018-10-15
[#4760] Don't apply censor rules to outgoing messages on the admin page

### LINKED ISSUE #4760: Don't apply censor rules to OutgoingMessage#body in AdminOutgoingMessagesController#edit
Censor rules are being applied to message text on outgoing message admin pages such as:

https://www.whatdotheyknow.com/admin/outgoing_messages/449151/edit

A consequence of this is that if an administrator:

1. Implements a censor rule which impacts an outgoing message
2. Changes the prominence of that message via the message admin page

that results in the outgoing message text with the censor rule applied being saved as the outgoing message.  The record of the edit shows the version with the censor rule applied as both the before, and after, and there is no record accessible of the request prior to the censor rule. 

Is applying censor rules to message text on outgoing message admin pages new? Is it a bug?

The lack of a record of the original message text in these circumstances is an issue which needs considering.

lizconlan:
## Relevant issue(s)

Closes #4760 

## ToDo

- [ ] Add changelog entry

## What does this do?

* Uses `OutgoingMessage#raw_body` as the source of the editable message on the admin screen for editing outgoing messages so that the version of the text without the censor rules applied is shown.
* Appends `raw_body` data to the accordion view of the outgoing message on the `admin_request#show` page so that admins can see the original message data
* Sets the `raw_body` data as if it were the `body` data `InfoRequestEvent#params` when an update is made so that the original version of the text will appear in the event log going forward

Includes some minor code tidying to improve readability re param indenting and line length.

## Why was this needed?

Admins were unable to edit outgoing messages with applicable censor rules without permanently applying redactions to the stored message. They were also unable to see the unreacted message in full as it currently stands or in the edit history.

## Screenshots

The edit page previously looked like this:

screen shot 2018-10-15 at 13 47 21

Now this:

screen shot 2018-10-18 at 10 00 27

### On the admin_request#show page

After an edit has been made, the admin should see this - with the `raw_body` substituted for the `body` field so that they can see the unredacted versions of the text:

screen shot 2018-10-19 at 16 4 03

([Previously](https://user-images.githubusercontent.com/27760/46962119-335bdf00-d09a-11e8-979c-334a31568bcc.png))

The Outgoing Message section shows the message as it will appear on the public site in the summary and appends the `raw_body` (unredacted) text at the end:

screen shot 2018-10-19 at 15 00 00

([Previously](https://user-images.githubusercontent.com/27760/46962290-abc2a000-d09a-11e8-8986-956b10e04bc0.png))

### DIFF
--- a/app/controllers/admin_outgoing_message_controller.rb
+++ b/app/controllers/admin_outgoing_message_controller.rb
@@ -28,15 +28,17 @@ class AdminOutgoingMessageController < AdminController
     old_prominence = @outgoing_message.prominence
     old_prominence_reason = @outgoing_message.prominence_reason
     if @outgoing_message.update_attributes(outgoing_message_params)
-      @outgoing_message.info_request.log_event("edit_outgoing",
-                                               { :outgoing_message_id => @outgoing_message.id,
-                                                 :editor => admin_current_user,
-                                                 :old_body => old_body,
-                                                 :body => @outgoing_message.raw_body,
-                                                 :old_prominence => old_prominence,
-                                                 :old_prominence_reason => old_prominence_reason,
-                                                 :prominence => @outgoing_message.prominence,
-                                                 :prominence_reason => @outgoing_message.prominence_reason })
+      @outgoing_message.
+        info_request.
+          log_event("edit_outgoing",
+                    { outgoing_message_id: @outgoing_message.id,
+                      editor: admin_current_user,
+                      old_body: old_body,
+                      body: @outgoing_message.raw_body,
+                      old_prominence: old_prominence,
+                      old_prominence_reason: old_prominence_reason,
+                      prominence: @outgoing_message.prominence,
+                      prominence_reason: @outgoing_message.prominence_reason })
       flash[:notice] = 'Outgoing message successfully updated.'
       @outgoing_message.info_request.expire
       redirect_to admin_request_url(@outgoing_message.info_request)
--- a/app/controllers/admin_outgoing_message_controller.rb
+++ b/app/controllers/admin_outgoing_message_controller.rb
@@ -24,7 +24,7 @@ class AdminOutgoingMessageController < AdminController
   end
 
   def update
-    old_body = @outgoing_message.body
+    old_body = @outgoing_message.raw_body
     old_prominence = @outgoing_message.prominence
     old_prominence_reason = @outgoing_message.prominence_reason
     if @outgoing_message.update_attributes(outgoing_message_params)
@@ -32,7 +32,7 @@ class AdminOutgoingMessageController < AdminController
                                                { :outgoing_message_id => @outgoing_message.id,
                                                  :editor => admin_current_user,
                                                  :old_body => old_body,
-                                                 :body => @outgoing_message.body,
+                                                 :body => @outgoing_message.raw_body,
                                                  :old_prominence => old_prominence,
                                                  :old_prominence_reason => old_prominence_reason,
                                                  :prominence => @outgoing_message.prominence,
--- a/app/views/admin_request/show.html.erb
+++ b/app/views/admin_request/show.html.erb
@@ -299,13 +299,13 @@
             <% outgoing_message.for_admin_column do |name, value, type, column_name| %>
               <tr>
                 <td class="span3">
-                  <b><%=name%></b>
+                  <b><%= name %></b>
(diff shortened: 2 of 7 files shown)


app/controllers/admin_outgoing_message_controller.rb
```diff
@@ -33,6 +34,8 @@ def update
                                                  :editor => admin_current_user,
                                                  :old_body => old_body,
                                                  :body => @outgoing_message.body,
+                                                 :old_raw_body => old_raw_body,
+                                                 :raw_body => @outgoing_message.raw_body,
```
garethrees: I'm not sure we should introduce a concept of `raw_body` here.

I think a less complex solution (that essentially mirrors the change to the form) is:

```diff
diff --git a/app/controllers/admin_outgoing_message_controller.rb b/app/controllers/admin_outgoing_message_controller.rb
index c852a7c29..d7629e977 100644
--- a/app/controllers/admin_outgoing_message_controller.rb
+++ b/app/controllers/admin_outgoing_message_controller.rb
@@ -24,7 +24,7 @@ class AdminOutgoingMessageController < AdminController
   end

   def update
-    old_body = @outgoing_message.body
+    old_body = @outgoing_message.raw_body
     old_prominence = @outgoing_message.prominence
     old_prominence_reason = @outgoing_message.prominence_reason
     if @outgoing_message.update_attributes(outgoing_message_params)
@@ -32,7 +32,7 @@ class AdminOutgoingMessageController < AdminController
                                                { :outgoing_message_id => @outgoing_message.id,
                                                  :editor => admin_current_user,
                                                  :old_body => old_body,
-                                                 :body => @outgoing_message.body,
+                                                 :body => @outgoing_message.raw_body,
                                                  :old_prominence => old_prominence,
                                                  :old_prominence_reason => old_prominence_reason,
                                                  :prominence => @outgoing_message.prominence,
```

app/views/admin_request/show.html.erb
```diff
@@ -280,7 +280,7 @@
           #<%= outgoing_message.id %> -- <%= outgoing_message.status.humanize %> <%= outgoing_message.message_type.humanize %>
       <% end %>
       <blockquote>
-        <%= truncate(outgoing_message.body, :length => 400) %>
+        <%= truncate(outgoing_message.raw_body, :length => 400) %>
```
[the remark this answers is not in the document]
lizconlan: The simplest/laziest way of resolving this - without messing with `for_admin_column` - would look like this...

screen shot 2018-10-19 at 15 00 00

(Although **Raw body** could just as easily go at the top but ... although easier for comparison purposes that doesn't feel right somehow. Also as this is being added manually, we don't have to stick with "Raw body" if something like "Unredacted" or "Uncensored" would be clearer.)

app/controllers/admin_outgoing_message_controller.rb
lizconlan: That's nicer to use, thanks!

lizconlan:
Those changes seem to pass CI, I'll rebase it as these fixups won't apply cleanly

app/views/admin_request/show.html.erb
```diff
@@ -312,6 +312,18 @@
...
+            <tr>
+              <td class="span3">
+                <b>Raw body</b>
+              <td>
+                <% truncated_value = truncate(h(outgoing_message.raw_body), length: 400, omission: '') { link_to '…', '#', title: 'Toggle hidden', class: 'toggle-hidden' } %>
+                    <%= simple_format(truncated_value) %>
+                    <div style="display:none;"><%= simple_format(outgoing_message.raw_body) %></div>
+                <div style="display:none;">
+                  <%= simple_format(outgoing_message.raw_body) %>
+                </div>
```
garethrees: This includes the full text twice, so clicking `toggle-hidden` results in two copies:

screen shot 2018-10-24 at 13 04 27

The indenting here was really confusing at first glance.

I _think_ the intent was for the final code to be:

```erb
<% truncated_value = truncate(h(outgoing_message.raw_body), length: 400, omission: '') { link_to '…', '#', title: 'Toggle hidden', class: 'toggle-hidden' } %>
<%= simple_format(truncated_value) %>
<div style="display:none;">
  <%= simple_format(outgoing_message.raw_body) %>
</div>
```

PR #4909, opened 2018-10-18
[#4865] Enable optional reCAPTCHA on contact form

### LINKED ISSUE #4865: Add RECAPTCHA to HelpController#contact
WDTK volunteers report an increase of contact form spam.

While we are catching span in the [honeypot](https://github.com/mysociety/alaveteli/blob/3fb7b936e8bdd3669f26e56aa718b8674054c583/app/controllers/help_controller.rb#L78-L84), some are starting to avoid it.

Initially we should only show the recaptcha if the IP is from a foreign country, so that we don't disrupt genuine users.

lizconlan:
## Relevant issue(s)

Closes #4865 

## What does this do?

* Adds a `CONTACT_FORM_RECAPTCHA` config key so that the help controller knows whether to attempt to valid a reCAPTCHA on form submit
* Adds a `HelpController#recaptcha_required?` method that a theme can override if there are extra conditions - such as checking the originating IP of the request for a particular site

## Why was this needed?

WDTK has started getting a high level of spam through the contact form so we wanted to add reCAPTCHA as a deterrent but without forcing it on to every Alaveteli instance.

## Notes to reviewer

Does this need a config key? Is that clearer than just having a `recaptcha_required` method in the controller or just more work for a reuser? (Some work required in all cases as there's no contact form in Alaveteli core, it's always implemented by the theme.)

## ToDo

Add a description of the new config key to the docs (if it stays in)

### DIFF
--- a/app/controllers/help_controller.rb
+++ b/app/controllers/help_controller.rb
@@ -11,6 +11,7 @@ class HelpController < ApplicationController
 
   before_action :long_cache
   before_action :catch_spam, :only => [:contact]
+  before_action :set_recaptcha_required, :only => [:contact]
 
   def index
     redirect_to help_about_path
@@ -52,7 +53,13 @@ class HelpController < ApplicationController
         params[:contact][:name] = @user.name
       end
       @contact = ContactValidator.new(params[:contact])
-      if @contact.valid? && !params[:remove]
+
+      if (@recaptcha_required &&
+          !params[:remove] &&
+          !verify_recaptcha)
+        flash.now[:error] = _('There was an error with the reCAPTCHA. ' \
+                              'Please try again.')
+      elsif @contact.valid? && !params[:remove]
         ContactMailer.to_admin_message(
           params[:contact][:name],
           params[:contact][:email],
@@ -81,6 +88,10 @@ class HelpController < ApplicationController
         redirect_to frontpage_url
       end
     end
+  end
+
+  def set_recaptcha_required
+    @recaptcha_required = AlaveteliConfiguration.contact_form_recaptcha
   end
 
 end
--- a/config/general.yml-example
+++ b/config/general.yml-example
@@ -1018,6 +1018,21 @@ RESTRICTED_COUNTRIES: ''
 # ---
 NEW_REQUEST_RECAPTCHA: false
 
+# Show a reCAPTCHA on the contact form to discourage spammers
+#
+# Add the following code to `help/_contact_form.html.erb` in your theme (just
+# above the submit button works best) and a reCAPTCHA will be shown if
+# CONTACT_FORM_RECAPTCHA is set to true.
+#
+#     <% if @recaptcha_required %>
+#       <%= recaptcha_tags %><br />
+#     <% end %>
+#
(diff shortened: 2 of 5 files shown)


config/general.yml-example
```diff
@@ -1018,6 +1018,13 @@ RESTRICTED_COUNTRIES: ''
 # ---
 NEW_REQUEST_RECAPTCHA: false
 
+# Show a reCAPTCHA on the contact form to discourage spammers
+#
+# CONTACT_FORM_RECAPTCHA – Boolean (default: false)
+#
+# ---
+CONTACT_FORM_RECAPTCHA: false
```
garethrees: I think a config key is a good option, given you need to explicitly account for it in the theme.

app/controllers/help_controller.rb
```diff
@@ -89,4 +89,8 @@ def catch_spam
     end
   end
 
+  def recaptcha_required?
```
garethrees: I'm not sure about this implementation. [The work required on the theme end](https://github.com/mysociety/whatdotheyknow-theme/pull/521/files#diff-c38892457e3c9bd7f3c03c9d1a066150R20) is quite complex.

The [filter](https://github.com/mysociety/whatdotheyknow-theme/pull/521/files#diff-c38892457e3c9bd7f3c03c9d1a066150R20) should exist in core, and we should use `@recaptcha_required` to decide whether to show and validate the recaptcha. At the moment, the theme has to care about setting [`@recaptcha_required`](https://github.com/mysociety/whatdotheyknow-theme/pull/521/files#diff-c38892457e3c9bd7f3c03c9d1a066150R34) _and_ overriding [`recaptcha_required?`](https://github.com/mysociety/whatdotheyknow-theme/pull/521/files#diff-c38892457e3c9bd7f3c03c9d1a066150R37).

```ruby
# Core
class HelpController < ApplicationController
  before_filter :set_recaptcha_required, only: [:contact]

  def contact
    # …
    if (@recaptcha_required && !params[:remove] && !verify_recaptcha)
      flash.now[:error] = _('There was an error with the reCAPTCHA. ' \
                            'Please try again.')
    elsif @contact.valid? && !params[:remove]
    # …
  end

  private

  def set_recaptcha_required
    @recaptcha_required = AlaveteliConfiguration.contact_form_recaptcha
  end
end
```

That way, the only thing a theme owner needs to do is:

* Set `CONTACT_FORM_RECAPTCHA` to `true`
* Add recaptcha to the form:

```erb
<% if @recaptcha_required %>
  <%= recaptcha_tags %>
<% end %>
```

If a theme owner wanted to only show for requests from foreign IPs, the only extra thing they need to do is set `@recaptcha_required` appropriately, without having to worry about adding filters etc.

```ruby
# Additional checks in theme
HelpController.class_eval do
  def set_recaptcha_required
    @recaptcha_required =
      AlaveteliConfiguration.contact_form_recaptcha &&
      request_from_foreign_country?
  end

  def request_from_foreign_country?
    country_from_ip != AlaveteliConfiguration.iso_country_code
  end
end
```

lizconlan: Good point - I was too focused on this exact case rather than the simplest possible override scenario. Thanks!

Issue #4919, opened 2018-10-23
Slow parsing of IncomingMessage belonging to a Batch


garethrees:
```ruby
IncomingMessage.where(last_parsed: nil).joins(:info_request).merge(InfoRequest.pro).find_each do |msg|
  bm = Benchmark.measure { msg.parse_raw_email! }
  puts "Parsed #{msg.id} in #{ bm.real.round(6) }"
end
Parsed 1179170 in 40.009797
Parsed 1179171 in 40.549561
Parsed 1179172 in 40.747063
Parsed 1179173 in 40.0386
Parsed 1179174 in 39.418158
```

My guess here is that saving an `IncomingMessage` saves an `InfoRequest`, which updates the `RequestSummary`, and and so for an `IncomingMessage` that “belongs” to a Batch, we have to update a `RequestSummary` for _every request in the batch_ on save.

garethrees:
This also causes problems in the admin interface. Clicking save after editing an `InfoRequest` belonging to an `InfoRequestBatch` can cause timeouts.

garethrees:
https://github.com/mysociety/alaveteli/pull/4951 is now merged, and is running every hour via cron on WDTK.

garethrees:
I just ran this again (albeit for only one incoming message, but for a request in a batch of ~100) and it was fast:

```ruby
IncomingMessage.where(last_parsed: nil).joins(:info_request).merge(InfoRequest.pro).find_each do |msg|
  bm = Benchmark.measure { msg.parse_raw_email! }
  puts "Parsed #{msg.id} in #{ bm.real.round(6) }"
end
Parsed 1546239 in 0.765475
```

I'm going to remove `script/parse-unparsed-pro` and then after a few days, give the above a run to make sure everything seems fast. If all that's good, I'll get rid of `parse-unparsed-pro` altogether.

garethrees:
Another quick run before the cron task got a chance to…

```ruby
IncomingMessage.where(last_parsed: nil).joins(:info_request).merge(InfoRequest.pro).find_each do |msg|
  bm = Benchmark.measure { msg.parse_raw_email! }
  puts "Parsed #{msg.id} in #{ bm.real.round(6) }"
end
Parsed 1546244 in 0.889128
Parsed 1546246 in 0.249953
Parsed 1546249 in 0.267352
Parsed 1546250 in 0.220896
Parsed 1546253 in 0.45271
Parsed 1546257 in 0.48046
[...]

PR #4920, opened 2018-10-23
[#4919] Add incoming_messages:parse_unparsed_pro rake task

garethrees:

## Relevant issue(s)

https://github.com/mysociety/alaveteli/issues/4919

## What does this do?

Add incoming_messages:parse_unparsed_pro rake task

Allows us to run this as a cron task so that Pro users don't suffer lag
while parsing new messages during the request cycle.

## Why was this needed?

https://github.com/mysociety/alaveteli/issues/4919 describes the problem of pro incoming messages being slow to parse. Looks like its related to large batches because each `save` call requires updates for all requests in a batch. Running this rake task through cron reduces the chances that a user will experience a slow parse.

### DIFF
--- a/app/models/incoming_message.rb
+++ b/app/models/incoming_message.rb
@@ -71,6 +71,8 @@ class IncomingMessage < ActiveRecord::Base
   after_update :update_request
   before_destroy :destroy_email_file
 
+  scope :unparsed, -> { where(last_parsed: nil) }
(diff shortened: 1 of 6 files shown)


garethrees:
Merged in 8d40211e4d134a8309ef7d0a1141eb5361155bb0

PR #4914, opened 2018-10-22
Link to parent batch from request page

zarino:
Fixes mysociety/alaveteli-professional#435.

If a request is part of a batch, a link to the main batch page is now shown in the header, and also in the sidebar.

The pro sidebar style is slightly different (icons, shorter headings) so this also includes a modified version of the template for pro (ie: private) requests.

While I was in the area, I also added a commit to improve the spacing around the pro sidebar icons, because they felt a bit squished up next to their accompanying text.

# Screenshots

A _public_ request that’s part of a batch:

image

A _private_ request that’s part of a batch:

image

PR #4923, opened 2018-10-24
[pro#435] Link to parent batch from request page [GR]

garethrees:
Fixups to https://github.com/mysociety/alaveteli/pull/4914 that were easier to just do rather than explain.


garethrees:
Merged in to https://github.com/mysociety/alaveteli/pull/4914

Issue #516, opened 2012-07-01
Change the "(requires_admin)" and/or "(attention_requested)" subject lines in admin emails


hsenag:
Even several weeks after they were introduced, I find it really hard to remember the difference between these subjects.

How about "(requester_reported)" and "(moderation_needed)"?

garethrees:
There are three related states here:

* `error_message`
* `requires_admin`
* `attention_requested`

garethrees:
Linking to #6931 as I'd like to separate out the "reporting" (via `Reportable`) from the request state.

HelenWDTK:
This issue has been automatically closed due to a lack of discussion or resolution for over 12 months.
Should we decide to revisit this issue in the future, it can be reopened.

PR #4961, opened 2018-11-20
[#1764] Improve email threading when responding to change requests

### LINKED ISSUE #1764: Thread emails to users related to request for admin attention or public body addresses with original email to admins
When the admin team receives a request for a public body address update or a request is drawn to their attention, any follow up email they send from the admin interface to the user gets a different subject, so isn't added to the same correspondence thread in email clients. One question to resolve here is whether there is some way that will be recognized by most email clients to indicate that a new message should be threaded with an existing email. An alternative approach would be to send an email to the admin team with the same subject as the previous admin email - this would then presumably be threaded by subject line.

lizconlan:

## Relevant issue(s)

Connects to #1764

## What does this do?

* Uses the original subject line to set the suggested subject for admin replies to requests for updates/additions to authorities
* Moves the construction of the subject lines from `ContactMailer` to `PublicBodyChangeRequest`
* Simplifies `ContactMailer` to use a single method to deal with sending response to both types of change request
* Minor spec tidy up

## Why was this needed?

Using unrelated subject lines for the initial request and the response was making it harder for admins to follow a change request conversation over email as the they did not appear in a single conversation thread.

## Notes to reviewer

The ticket mentions improving threading for requests for admin attention as well as change requests but I have been unable to find a specific admin feature for replying to admin attention requests so this PR does not address this.

Requesting admin attention sends an email with a subject `FOI response requires admin (attention_requested) - [request title]` (via "Report this request") or `FOI response requires admin (requires_admin) - [request title]` (updating the status to "requires administrator attention"),  and the admin summary just presents a list of flagged requests for review/editing so presumably the simplest way to contact the user will be to reply to the initial email. (I have not checked whether the site functionality has changed since the original ticket was created. And will happily stand corrected - and open a new PR - if I've missed something!)

### DIFF
--- a/app/mailers/contact_mailer.rb
+++ b/app/mailers/contact_mailer.rb
@@ -61,9 +61,7 @@ class ContactMailer < ApplicationMailer
                     blackhole_email
                   ),
          :to => contact_from_name_and_email,
-         :subject => _('Add authority - {{public_body_name}}',
-                       :public_body_name => @change_request.
-                                              get_public_body_name.html_safe))
+         :subject => @change_request.request_subject)
   end
 
   # Send a request to the administrator to update an authority email address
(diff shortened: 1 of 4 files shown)


garethrees:
> The ticket mentions improving threading for requests for admin attention  

May be related to https://github.com/mysociety/alaveteli/issues/516, but yeah, one thing at a time! 👍

PR #4966, opened 2018-11-21
[#426] Only show alternative contacts where this is helpful

### LINKED ISSUE #426: Only show alternative followup addresses if there are many
When clicking "send a follow up" on my request at http://www.whatdotheyknow.com/request/electronic_case_file_sharing_cps I got the following which doesn't make any sense:

> Don't want to address your message to the main FOI contact at Ministry of Justice? You can also write to:
> 
>
> * the main FOI contact address for Ministry of Justice

lizconlan:

## Relevant issue(s)

Closes #426

## What does this do?

* Prevents alternative contacts from being displayed where there are no authority responses
* Prevents duplicate contacts from earlier responses inadvertently being included
* Adds some extra specs to describe the behaviour of `InfoRequest#who_can_followup_to`

## Why was this needed?

We occasionally show "alternative contacts" when there is only a single option, which is confusing.

### DIFF
--- a/app/views/followups/_followup.html.erb
+++ b/app/views/followups/_followup.html.erb
@@ -20,20 +20,19 @@
                          name_for_followup: name_for_followup } %>
   <% end %>
 
-  <% if @info_request.who_can_followup_to(incoming_message).count > 0 %>
+  <% if incoming_message &&
+        @info_request.who_can_followup_to(incoming_message).any? %>
     <div id="other_recipients" class="box other_recipients">
       <%= _("Don't want to address your message to {{person_or_body}}?  You can also write to:", :person_or_body => name_for_followup) %>
 
       <ul>
         <% @info_request.who_can_followup_to(incoming_message).each do |name, email, id|  %>
-          <% if id.nil? && !incoming_message.nil? && incoming_message.valid_to_reply_to? %>
+          <% if id.nil? && incoming_message.valid_to_reply_to? %>
             <li><%= link_to(_("the main FOI contact address for {{public_body}}", :public_body => name), new_request_followup_path(:request_id => @info_request.id)) %></li>
           <% else %>
-            <% if !id.nil? %>
+            <% if id.present? %>
               <% if @info_request.public_body.request_email == email %>
-                <% if !incoming_message.nil? %>
-                  <li><%= link_to(_("the main FOI contact address for {{public_body}}", :public_body => name), new_request_followup_path(:request_id => @info_request.id)) %></li>
-                <% end %>
+                <li><%= link_to(_("the main FOI contact address for {{public_body}}", :public_body => name), new_request_followup_path(:request_id => @info_request.id)) %></li>
               <% else %>
                 <li><%= link_to name, new_request_incoming_followup_path(:request_id => @info_request.id, :incoming_message_id => id)%></li>
               <% end %>
--- a/spec/controllers/followups_controller_spec.rb
+++ b/spec/controllers/followups_controller_spec.rb
@@ -93,24 +93,34 @@ describe FollowupsController do
         expect(response).to render_template('new')
       end
 
-      it "offers the opportunity to reply to the main address" do
-        get :new, params: { :request_id => request.id,
-                            :incoming_message_id => message_id }
-        expect(response.body).
-          to have_css("div#other_recipients ul li", :text => "the main FOI contact address for")
-      end
+      context 'the request has responses' do
+        let(:message_id) { request.incoming_messages[0].id }
 
-      it "offers an opportunity to reply to another address" do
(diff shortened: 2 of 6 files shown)


spec/controllers/followups_controller_spec.rb
```diff
@@ -97,7 +97,7 @@
         get :new, params: { :request_id => request.id,
                             :incoming_message_id => message_id }
         expect(response.body).
-          to have_css("div#other_recipients ul li", :text => "the main FOI contact address for")
+          to have_css("div#followup", :text => "the main FOI contact at")
```
garethrees: Not sure about this change as the assertion has now changed its meaning.

Originally we were checking that _if_ there are alternative reply addresses available, then we should _also_ list the main contact address for selection.

Now its just asserting that somewhere we say "you're replying to the main contact address".

The change here should be in the spec setup, with the original assertion.

spec/views/followups/_followup.html.erb_spec.rb
```diff
@@ -97,4 +97,25 @@
...
+
+  describe 'displaying followup contact options' do
+
+    context 'with an incoming message' do
+
+      let(:incoming) do
+        FactoryBot.create(:plain_incoming_message, info_request: info_request)
+      end
+
+      it 'does not offer the same contact option twice' do
```
garethrees: I don't think the template is responsible for deciding this? At least, I can't work out where it does deduplication – that happens in the model layer as of this commit, so I don't think a view spec is necessary here?

app/views/followups/_followup.html.erb
```diff
@@ -20,13 +20,13 @@
                          name_for_followup: name_for_followup } %>
   <% end %>
 
-  <% if @info_request.who_can_followup_to(incoming_message).count > 0 %>
+  <% if incoming_message && @info_request.who_can_followup_to(incoming_message).count > 0 %>
```
garethrees: `s/count > 0/any?` while we're here.

spec/views/followups/_followup.html.erb_spec.rb
```diff
@@ -100,12 +100,39 @@
 
   describe 'displaying followup contact options' do
 
+    context 'without an incoming message' do
+
+      it 'does not show the "other options" panel' do
+        render partial: "followups/followup", locals: { incoming_message: nil }
+        expect(rendered).
+            to_not have_content 'You can also write to'
```
garethrees: Doesn't need to be split over two lines

spec/views/followups/_followup.html.erb_spec.rb
```diff
@@ -100,12 +100,39 @@
...
 
       let(:incoming) do
         FactoryBot.create(:plain_incoming_message, info_request: info_request)
       end
 
+      it 'shows the "other options" panel if the incoming message is not from the main contact address' do
+        render partial: "followups/followup",
+               locals: { incoming_message: incoming }
+        expect(rendered).
+          to have_content 'You can also write to'
```
garethrees: Doesn't need to be split over two lines

app/views/followups/_followup.html.erb
lizconlan: 👍

spec/views/followups/_followup.html.erb_spec.rb
lizconlan: True - it just iterates over what it's given

app/views/followups/_followup.html.erb
```diff
@@ -20,13 +20,13 @@
...
     <div id="other_recipients" class="box other_recipients">
       <%= _("Don't want to address your message to {{person_or_body}}?  You can also write to:", :person_or_body => name_for_followup) %>
 
       <ul>
         <% @info_request.who_can_followup_to(incoming_message).each do |name, email, id|  %>
-          <% if id.nil? && !incoming_message.nil? && incoming_message.valid_to_reply_to? %>
+          <% if id.nil? && incoming_message.valid_to_reply_to? %>
```
lizconlan: of course in doing this, I suddenly realise that we are checking the `incoming_message` is valid to reply to in order to show the main contact address, which makes my head hurt

app/views/followups/_followup.html.erb
```diff
@@ -30,11 +30,9 @@
           <% if id.nil? && incoming_message.valid_to_reply_to? %>
             <li><%= link_to(_("the main FOI contact address for {{public_body}}", :public_body => name), new_request_followup_path(:request_id => @info_request.id)) %></li>
           <% else %>
-            <% if !id.nil? %>
+            <% if id.present? %>
               <% if @info_request.public_body.request_email == email %>
-                <% if !incoming_message.nil? %>
```
lizconlan: we now always have an`incoming_message` at this point so we no longer need to check for it

Issue #4603, opened 2018-04-10
Ability to unsubscribe from all transaction emails


garethrees:
- [ ] Review all emails that we send out
- [ ] Add ability to unsubscribe from any that you can't unsubscribe from that aren't sent in response to a direct user action
  * e.g. new response, new response reminder, late, etc
  * password resets are fine as the user will have clicked something like "send me the email" and they're one-time

Related to #1971.

RichardTaylor:
Just to note that I recall a discussion years ago about if we had a duty to email people to let them know they'd had a response to their request, and as I recall that wasn't something we then thought it was right to allow/enable people to opt out of. 

I think times have changed and people do expect more control over what emails they get. I'm not objecting to this proposal just flagging that's it's not entirely obvious and clear cut that we should allow people to opt-out of getting an email alert to tell them they've had a response to a request they've made; it has been a contentious subject in the past.

garethrees:
First step with this is adding a "turn off email alerts" link to emails that the user _already has control over_, as per https://github.com/mysociety/alaveteli/issues/4551.

garethrees:
Being able to "unsubscribe" users from emails would help with https://github.com/mysociety/alaveteli/issues/5090

RichardTaylor:
This feature would be desirable to enable both a user-ban (based on their email address) and a suppression of emails to banned users where such a suppression is requested.

RichardTaylor:
A WhatDoTheyKnow user today has replied to a "Please update the status of your request" email asking to be unsubscribed from "these emails".

garethrees:
A [user has asked](https://groups.google.com/a/mysociety.org/g/alaveteli-dev-private/c/aLpPnVo7ADc/m/mLJmwn2tAgAJ) to be unsubscribed from post request survey emails (https://github.com/mysociety/alaveteli-project/issues/127).

garethrees:
https://github.com/stevepolitodesign/unsubscribe might be some nice inspiration here.

garethrees:
Linking to #4910. Not 100% sure if "closing" an account would enable us to solve this from the admin interface (in that closed users shouldn't receive email, but would need to actually check it stops everything).

FOIMonkey:
+1 This would have helped with a case where an org responded to an old request belonging to a banned user, who then contacted us to ask why they are still getting emails when they are banned.

RichardTaylor:
[...]

PR #4967, opened 2018-11-23
[#4551] One-click unsubscribe from Email Alerts

### LINKED ISSUE #4551: Should be able to unsubscribe from track emails without signing in
Currently there is an "Alter your subscription" link, which is a very high barrier to entry for an email subscription; you should not have to log in in order to stop getting an email.

garethrees:
## Relevant issue(s)

Fixes #4551

## What does this do?

Adds "one-click-unsubscribe" to `TrackMailer#event_digest` emails (sent by the alert-tracks daemon).

## Why was this needed?

Requiring users to sign in to unsubscribe from email alerts adds unnecessary effort, and encourages users to mark genuine emails as spam. This affects our reputation with email providers.

## Implementation notes

`User::EmailAlerts` adds advanced capability for managing a `User`'s email alerts, namely adding a token generator and verifier so that we can look up a `User` by a unique token and then change their email preferences.

Note that `User#receive_email_alerts` only manages notifications from `TrackMailer#event_digest`, so I've added messaging to that effect on the unsubscribe confirmation template.

#4603 covers adding additional notification management options.

## Screenshots

The unsubscribe link:

screen shot 2018-11-23 at 17 34 59

Successfully unsubscribing:

screen shot 2018-11-23 at 17 21 38

Passing an invalid token:

screen shot 2018-11-23 at 17 48 19

## Notes to reviewer

This was a quick way to setup and send a basic `event_digest` email:

```ruby
user = User.find(3)

track_thing = TrackThing.first

xapian_object = ActsAsXapian::Search.new([InfoRequestEvent], 'chicken',
                                                 :sort_by_prefix => 'described_at',
                                                 :sort_by_ascending => true,
                                                 :collapse_by_prefix => nil,
                                                 :limit => 100)


alert_results = []
alert_results.push(xapian_object.results.first)                                               
email_about_things = []
email_about_things.push([track_thing, alert_results, xapian_object])

TrackMailer.event_digest(user, email_about_things).deliver_now
```

This should work if you have the fixture data around, but I'm sure if you tweak the user ID / search term you can find something workable.


app/views/users/email_alerts/destroy.html.erb
```diff
@@ -0,0 +1,17 @@
+<h1><%= _('Email Alerts Disabled') %></h1>
+
+<p>
+  <%= _('Email alerts for things you follow have now been disabled.') %>
+</p>
+
+<p>
+  <%= _('<a href="{{sign_in_link}}">Sign in</a> to get updates on things ' \
```
lizconlan: Seeing this when signed in is a little odd, maybe a link to the wall instead for signed in users?

app/views/track_mailer/event_digest.text.erb
```diff
@@ -75,4 +75,10 @@
 <% _('Please click on the link below to cancel or alter these emails.') %>
 <%= @unsubscribe_url %>
 
+<%= _('Disable Email Alerts') %>
```
lizconlan: Maybe we should call this "Unsubscribe From Email Alerts"? I'm not entirely sure on this but it is the language people seem to use/look for when they're trying to make the emails go away.

Or maybe make a ticket to review the wording in general 🤔

garethrees: I did think about this, but I initially thought that "Disable Email Alerts" makes a better distinction between the two actions. On reflection I think you're right – Unsubscribe is what you look for, and is clear enough. Fixed in cdf9513d844afed5b5d97c3bba4c3f1935441293.

app/views/users/email_alerts/destroy.html.erb
garethrees: Ah yes! Fixed in 220e08c8c.

I did want to take people directly to the wall, but the route requires a `url_name` param, so doing the redirect via signin gets really messy.

I did just try to add a route for `/profile/wall` to go to the current user's wall, but the controller doesn't handle that without modification, so will have to wait until another day.

app/views/track_mailer/event_digest.text.erb
lizconlan: I also wasn't sure about this. The intent of disable is clearer but if I'm in a rush and become sufficiently irritated by something I don't remember signing up for to be considering the "Mark as Spam" button, I'll probably be skim-reading looking for "Unsubscribe"

PR #4979, opened 2018-11-28
[#4977] incoming message counter cache

### LINKED ISSUE #4977: Batch dashboard loads incoming messages count twice for each request
Load happens: https://github.com/mysociety/alaveteli/blob/f93d003/app/views/alaveteli_pro/info_request_batches/_info_request.html.erb#L16-L17

This should be cached on a new `InfoRequest#incoming_messages_count` column.

gbp:
## Relevant issue(s)

Fixes #4977 

## What does this do?

Adds a counter cache for the number of incoming messages received for each info request.

## Why was this needed?

To reduce the number to database queries when loading the pro dashboard.

## Implementation notes

When deloying we need to migrate the database and run `rake incoming_messages:update_counter_cache`


app/views/admin_request/show.html.erb
```diff
@@ -375,7 +375,7 @@
       </div>
     </div>
   <% end %>
-  <% if @info_request.incoming_messages.count > 0 %>
+  <% if @info_request.incoming_messages_count > 0 %>
```
garethrees: Pretty sure `.count` uses the counter cache value?

gbp: I don't believe so:
```
[1] pry(main)> request = InfoRequest.first; p request.incoming_messages_count; p request.incoming_messages.count
  InfoRequest Load (8.0ms)  SELECT  "info_requests".* FROM "info_requests"  ORDER BY "info_requests"."id" ASC LIMIT 1
1
   (0.4ms)  SELECT COUNT(*) FROM "incoming_messages" WHERE "incoming_messages"."info_request_id" = $1  [["info_request_id", 1]]
1
```

garethrees: Oh, lame.

Its `#size` that picks the cached column or falls back to `count` if it doesn't exist:

> Tip: Although #count sounds like the more intuitive choice for counting the number of responses, this example uses #size, as #count will always do a COUNT query, while #size will skip the query if the responses are already loaded.
>
> – https://blog.appsignal.com/2018/06/19/activerecords-counter-cache.html

db/migrate/20181128160243_add_incoming_messages_count_to_info_requests.rb
```diff
@@ -0,0 +1,5 @@
+class AddIncomingMessagesCountToInfoRequests < ActiveRecord::Migration
```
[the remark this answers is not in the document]
gbp: Will be resolved by d08714138ff28276273caebd3a178150a3704d9b

lib/tasks/incoming_messages.rake
```diff
@@ -9,4 +9,9 @@ namespace :incoming_messages do
       puts "Parsed #{msg.id} in #{ bm.real.round(6) }" if verbose
     end
   end
+
+  desc 'Update InfoRequest#incoming_messages_count counter cache'
+  task update_counter_cache: :environment do
+    InfoRequest.update_all('incoming_messages_count = (SELECT COUNT(*) FROM "incoming_messages" WHERE "incoming_messages"."info_request_id" = "info_requests"."id")')
```
garethrees: Any way to estimate how long this will take to run in prod?

spec/controllers/request_controller_spec.rb
```diff
@@ -2752,7 +2752,7 @@ def expect_redirect(status, redirect_path)
 
   it "should prevent uploads if you are not a requester" do
     @ir = info_requests(:fancy_dog_request)
-    incoming_before = @ir.incoming_messages.size
+    incoming_before = @ir.incoming_messages.count
```
garethrees: How come all these changed?

The specs still appear to pass with `#size`.

Worth explaining _why_ these changed in the commit message.

garethrees: Oh, looks like it works around isolation issues when running the full suite?

gbp: Yea, if these were left as `size` then it would read the counter cache which might be incorrect. Will update the commit message with details.

lib/tasks/incoming_messages.rake
gbp: Not very long: https://gist.github.com/gbp/538369487c20a1369aeb42c2622cd987

PR #4980, opened 2018-11-29
Rubocop and hound config

gbp:
## What does this do?

Update the configuration of Rubocop and Hound CI, by:

- Update cops which have been renamed/ removed
- Update HoundCI config so JS is checked again
- Add Rubocop to the Gemfile
- Specify the Rubocop version for Hound to use
- Skip documentation for migrations.


.ruby-style.yml
```diff
@@ -1,23 +1,22 @@
 AllCops:
+  RubyInterpreters:
```
garethrees: Commit msg should explain why the previous was wrong.

Looks like it was just _assuming_ all ruby files were included, which I expect was valid at some point in the past, but now rubocop is expecting you you to explicitly include them?

PR #4986, opened 2018-11-30
[#4975] Refactor URL title generation

### LINKED ISSUE #4975: When sending large batches URL title generation runs many SQL queries
`InfoRequest#update_url_title` could be rewritten to do a single query to get the next unique URL title.

gbp:
## Relevant issue(s)

Fixes #4975

## What does this do?

This now does a single DB query to work out the next number suffix when
requests have the same title - EG batch requests.

### Sample data

```rb
InfoRequest.where('url_title LIKE ?', 'batch_test%').map(&:url_title)
=> ["batch_test", "batch_test_2"]
```

### Before
```rb
InfoRequest.new(title: 'batch test')
  (0.2ms)  SELECT COUNT(*) FROM "info_requests" WHERE "info_requests"."url_title" = $1  [["url_title", "batch_test"]]
  (0.2ms)  SELECT COUNT(*) FROM "info_requests" WHERE "info_requests"."url_title" = $1  [["url_title", "batch_test_2"]]
  (0.3ms)  SELECT COUNT(*) FROM "info_requests" WHERE "info_requests"."url_title" = $1  [["url_title", "batch_test_3"]]
=> #<InfoRequest … url_title: "batch_test_3">
```

### After
```rb
InfoRequest.new(title: 'batch test')
  (0.4ms)  SELECT "info_requests"."url_title" FROM "info_requests" WHERE (url_title ~ '^batch_test(_\d+)?$')
=> #<InfoRequest … url_title: "batch_test_3">
```

### DIFF
--- a/app/models/info_request/sluggable.rb
+++ b/app/models/info_request/sluggable.rb
@@ -10,10 +10,11 @@ module InfoRequest::Sluggable
                 if: proc { |request| request.title_changed? }
   end
 
-  # When name is changed, also change the url name
+  # When title is changed, also change the URL title
   def title=(title)
-    write_attribute(:title, title)
-    update_url_title
+    super.tap do
+      update_url_title
+    end
   end
 
   # Public: url_title attribute reader
--- a/app/models/info_request/sluggable.rb
+++ b/app/models/info_request/sluggable.rb
@@ -29,6 +29,8 @@ module InfoRequest::Sluggable
     _url_title
   end
 
+  private
+
   def update_url_title
     return unless title
     url_title = MySociety::Format.simplify_url_part(title, 'request', 32)
--- a/app/models/info_request/sluggable.rb
+++ b/app/models/info_request/sluggable.rb
@@ -1,0 +1,55 @@
+# A module to generate unique URL title slugs for Info Requests
+#
+module InfoRequest::Sluggable
+  extend ActiveSupport::Concern
+
+  included do
+    # make sure the url_title is unique but don't update
+    # existing requests unless the title is being changed
+    before_save :update_url_title,
+                if: proc { |request| request.title_changed? }
+  end
+
+  # When name is changed, also change the url name
+  def title=(title)
+    write_attribute(:title, title)
+    update_url_title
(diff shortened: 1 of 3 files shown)


app/models/info_request/sluggable.rb
```diff
@@ -33,25 +33,26 @@ def url_title(opts = {})
...
+
+  def suffix_number(url_title)
+    # For request with same title as others, add on arbitrary numeric identifier
+    suffixes = existing_url_titles(url_title)
+    suffixes.last + 1 if suffixes.present?
+  end
+
+  def existing_url_titles(url_title)
+    scope = InfoRequest
+    scope = scope.where.not(id: id) if persisted?
```
lizconlan: minor nitpick, I'd be tempted to restate this as:

```rb
    scope =
      if persisted?
        InfoRequest.where.not(id: id)
      else
        InfoRequest
      end

```

arguably (as in "feel free to argue"!) clearer and might be a tiny fraction faster without assigning `scope` twice (although a quick Benchmark test suggests we're literally talking a couple of thousandths of a second so ... meh)

app/models/info_request/sluggable.rb
```diff
@@ -0,0 +1,55 @@
...
+
+  included do
+    # make sure the url_title is unique but don't update
+    # existing requests unless the title is being changed
+    before_save :update_url_title,
+                if: proc { |request| request.title_changed? }
+  end
+
+  # When name is changed, also change the url name
+  def title=(title)
```
garethrees: I don't necessarily think this comment should block this merge, but I find it pretty uncomfortable that `#title=` is redefined here. Scanning through `InfoRequest`, this behaviour wouldn't be at all obvious. Its not an issue here, right now, but what would we do if some other concern also needed to perform a method on `#title=`?

I guess what would be nice is some way to run a callback on a per-method basis:

```ruby
after_receive :title=, -> { update_url_title }
```

garethrees: 👌 nice!

PR #4992, opened 2018-12-05
Fix transient spec fail

lizconlan:
Otherwise this spec always fails if there is already data in the test db
and this is run first (or in isolation).

To reproduce:
 - Revert the fix commit
 - `be rspec ./spec/models/alaveteli_pro/request_summary_spec.rb` (ignore any fails)
 - `be rspec ./spec/models/alaveteli_pro/request_summary_spec.rb:387` (should fail)


lizconlan:
> I can't seem to get this to fail with or without the commit. Are you sure this is required and wasn't due to something on your local machine?

Hmm, thought it was genuine but now I've got a fresh box that I built while testing build scripts I can't reproduce it either 😞 

Closing as either something going wrong with my vagrant box or something that got fixed elsewhere

PR #4994, opened 2018-12-05
move JS that are from third party vendors to the vendors directory

### LINKED ISSUE #4707: Move vendor assets to vendor/assets
There are quite a few assets used by Alaveteli that should be in vendor/assets, just a quick search and I found
https://github.com/mysociety/alaveteli/blob/develop/app/assets/javascripts/dropit.js 
anything that starts with `jquery`

wrightmartin:
(only applies to core alaveteli, and not alaveteli pro)

## Relevant issue(s)
Closes #4707

## Why was this needed?
General tidy up of assets

## Implementation notes
Will tests need to be reviewed for this?


garethrees:
> only applies to core alaveteli, and not alaveteli pro

Why's that?

wrightmartin:
Sorry missed this!

> > only applies to core alaveteli, and not alaveteli pro

> Why's that?

Because so much of core is out of step with Alaveteli in general - for example core contains too much opinionated pro styles that should probably be in the themes. It felt like if I did that here it would be reworked again some time in the future anyway?

garethrees:
> Because so much of core is out of step with Alaveteli in general

The issue is about moving 3rd party code to `vendor` – so I'm not sure there's much of a difference between "core" and "pro" in this case?

wrightmartin:
> The issue is about moving 3rd party code to vendor – so I'm not sure there's much of a difference between "core" and "pro" in this case?

Probably not now I'm thinking about it properly

So if we included the pro files, would we still have them in their own `/pro` folder within `/vendor`? or does that not really matter?

garethrees:
> So if we included the pro files, would we still have them in their own /pro folder within /vendor? or does that not really matter?

I _think_ the directory tree should be the same, just under `vendor` instead of `app`. https://guides.rubyonrails.org/v4.2.10/asset_pipeline.html should have some info on that though

garethrees:
@wrightmartin is this ready for review now?

Looks like we can also move https://github.com/mysociety/alaveteli/blob/issues/4707-move-assets-to-vendor-assets/app/assets/javascripts/ba-throttle-debounce.js

wrightmartin:
@garethrees now ready for review

lizconlan:
> screen shot 2019-01-04 at 11 10 00

Hmm, I should probably ask codeclimate to overlook `/vendor` as by definition that's not our app code

Issue #3080, opened 2016-03-04
Reopen requests to allow new responses from anyone when a request is resent


RichardTaylor:
Reset "Allow new responses from:"  to "anybody" when an initial request or other outgoing message on a thread is resent. 

Currently only admins can re-send requests but even experienced admins can forget to check and update this setting when re-sending an old message.

Sometimes there are good reasons for re-sending an old message eg. a public body spots an old request they failed to respond to, lets us know, and wants to send a reply.

Closely related to #863 and explicitly [mentioned in the comments](https://github.com/mysociety/alaveteli/issues/1335#issuecomment-50912154) at #1335

RichardTaylor:
At 
https://www.whatdotheyknow.com/admin/requests/385168

the public body became confused when they replied to the internal review request and their reply didn't get through. 

I think what this ticket probably should suggest is re-setting the countdown to closing a request when there's any activity on a thread.

PR #4995, opened 2018-12-06
[#863] Keep FOI requests with recent followups open to new responses

### LINKED ISSUE #863: When a requester sends a followup in a request marked to allow responses from authority_only, reset it to anybody
In http://www.whatdotheyknow.com/request/school_applications_broken_by_fa

the requester sent the request in March, then an internal review request in December., by which time the request had become "Allow new responses from authority_only". One of the school governors then tried to reply in January from his own email address, which wasn't associated with the school.

I believe the current logic stops new outgoing messages from being sent from "old" requests that are completely closed: in this case it should either do the same thing, or the request should be reset to allow responses from "anybody". I'd favour the latter.

lizconlan:
## Relevant issue(s)

Fixes #863 

## What does this do?

Allows the query that determines whether a request should be treated as old (or very old) and closed to new responses to take the creation date of outgoing_messages into account when calculating the request's "age" .

Introduces a Query Object (and specs) for the reworked `updated_before` scope.

## Why was this needed?

Volunteers are finding that the request automatically closing to new responses can be problematic in cases where the user has been asked for clarification or has opened an internal review; in some cases the authority has tried to respond to the internal review and their email has bounced, causing extra work for admins and authorities alike.

## Implementation notes

This turned out to be about "resetting the clock" on the age of the request - or changing how we calculate that - (keeping it open rather than "reopening" it), but I'm no longer sure this is the right (or simplest) way to do it. I like the idea of using the data we already have to work out what to do but the query logic's turned out to be quite hard to follow.

The alternative approach would be to touch the related request and update its `updated_at` attribute whenever a followup is sent (which will probably be how we reset the clock in the related #3080 where we want to reopen requests when resending the request email).

### DIFF
--- a/app/models/info_request.rb
+++ b/app/models/info_request.rb
@@ -151,7 +151,9 @@ class InfoRequest < ActiveRecord::Base
     alias_method :in_progress, :awaiting_response
   end
   scope :action_needed, State::ActionNeededQuery.new
-  scope :updated_before, ->(ts) { where('"info_requests"."updated_at" < ?', ts) }
+  scope :updated_before, ->(ts) do
+    State::UpdatedBeforeQuery.new(timestamp: ts).call
+  end
 
   # user described state (also update in info_request_event, admin_request/edit.rhtml)
   validate :must_be_valid_state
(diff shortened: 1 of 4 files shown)


lizconlan:
This was interesting but not the right approach. Closing.

Issue #4896, opened 2018-10-12
Upgrade Mail to 2.7.x


garethrees:
[2.7.0](https://rubygems.org/gems/mail/versions/2.7.0) is out.

We attempted in https://github.com/mysociety/alaveteli/pull/4422, but looks like there were some issues to fix.

Required by https://github.com/mysociety/alaveteli/issues/3907
Possibly required by https://github.com/mysociety/alaveteli/issues/3465

garethrees:
Target release 0.35 #5229

gbp:
Tested with #5452 - still the same issues as previously <details>
<summary>15 failures</summary>

```
$ rspec spec/mailers/application_mailer_spec.rb spec/mailers/request_mailer_spec.rb spec/script/handle-mail-replies_spec.rb spec/mailers/alaveteli_pro/subscription_mailer_spec.rb spec/lib/mail_handler/mail_handler_spec.rb spec/lib/alaveteli_mail_poller_spec.rb

Randomized with seed 9228
.....F....F..F...F...F.F..F...Non US-ASCII detected and no charset defined.
Defaulting to UTF-8, set your own if this is incorrect.
.........*...Non US-ASCII detected and no charset defined.
Defaulting to UTF-8, set your own if this is incorrect.
........................F...............................................F................F......F....F.F..F........F..................

Pending: (Failures listed here are expected and do not affect your suite's status)

  1) when getting attachment attributes should ignore anything beyond the final MIME boundary
     # No reason given
     # ./spec/lib/mail_handler/mail_handler_spec.rb:432


Failures:

  1) when creating a mail object from raw data should not fail on invalid byte sequence in content-disposition header
     Failure/Error: part = Mail::Part.new("Content-Disposition: inline; filename=a\xB8z\r\n\r\nThis is the body text.")

     ArgumentError:
       invalid byte sequence in UTF-8
     # ./spec/lib/mail_handler/mail_handler_spec.rb:26:in `new'
     # ./spec/lib/mail_handler/mail_handler_spec.rb:26:in `block (2 levels) in <top (required)>'

  2) when creating a mail object from raw data should not be confused by subject lines with malformed UTF-8 at the end
     Failure/Error: expect(mail.subject).to eq('hello')

       expected: "hello"
            got: "hello�"

       (compared using ==)
     # ./spec/lib/mail_handler/mail_handler_spec.rb:52:in `block (2 levels) in <top (required)>'

[...]

PR #4998, opened 2018-12-07
[Rails 5] Lock Mail gem to 2.6.x

gbp:
## Relevant issue(s)

Connects to #3969

## What does this do?

Lock Mail gem to 2.6.x

## Why was this needed?

There are breaking changes in 2.7 which will require specs to be updated.


garethrees:
Linking to #4896

PR #5001, opened 2018-12-07
[#4410] Use the url_name instead of the id on the user contact form

### LINKED ISSUE #4410: Use User#url_name instead of User#id in UserController#contact
https://mysociety.slack.com/files/U026043K1/F7TRFM33L/screen_shot_2017-11-02_at_11.38.35.png

Are spammers incrementing the ID to send spam?

If so, this would help make it more difficult.

Worth checking the logs to see if this actually happens.

lizconlan:
## Relevant issue(s)

Closes #4410

## What does this do?

* Extracts the user contact code to its own controller
* Tidies up the spec code
* Cleans up the `#contact` action code by extracting most of the logic into new methods/filters

## Why was this needed?

We wanted to make it harder for people to script sending messages to other users by guessing numeric ids. Also a good opportunity to refactor the code.

## Implementation detail

The new controller is called `Messages` rather than `Contacts` as we're currently following the convention of pluralising the controller name and "contacts" makes me think of an address book rather than a way of sending messages to people.

## Note to reviewer

I think this breaks the WDTK theme override for enabling reCAPTCHA (mysociety/whatdotheyknow-theme#536) so that will need to a fixup/double-check when this PR is ready for merge.


app/controllers/users/messages_controller.rb
```diff
@@ -0,0 +1,77 @@
...
+    # between the two users)
+    #
+    # "authenticated?" has done the redirect to signin page for us
+    return unless authenticated?(
+        :web => _("To send a message to {{user_name}}",
+                  :user_name => CGI.escapeHTML(@recipient_user.name)),
+        :email => _("Then you can send a message to {{user_name}}.",
+                    :user_name => @recipient_user.name),
+        :email_subject => _("Send a message to {{user_name}}",
+                            :user_name => @recipient_user.name)
```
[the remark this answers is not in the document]
lizconlan: gah, I thought I found all of these!

app/controllers/users/messages_controller.rb
```diff
@@ -26,7 +26,7 @@ def contact
   private
 
   def set_recipient
-    @recipient_user = User.find(params[:id])
+    @recipient_user = User.find_by(url_name: params[:url_name])
```
garethrees: The behaviour is different here:

```ruby
User.find(91919191919)
# ActiveRecord::RecordNotFound: Couldn't find User with an out of range value for 'id'
# from /home/vagrant/bundle/ruby/2.3.0/gems/activerecord-4.2.10/lib/active_record/core.rb:159:in `rescue in find'

User.find_by(url_name: '91919191919')
# => nil
```

This leads to a 500 error (`undefined method `name' for nil:NilClass`) instead of a 404 if a `url_name` is mistyped.

app/controllers/users/messages_controller.rb
```diff
@@ -70,6 +61,19 @@ def check_logged_in
...
 
+  def set_contact
+    if params[:submitted_contact_form]
+      params[:contact][:name] = @user.name
+      params[:contact][:email] = @user.email
+      @contact = ContactValidator.new(params[:contact])
+    else
+      @contact = ContactValidator.new(
+        message: '' + @recipient_user.name +
+                 _(',\n\n\n\nYours,\n\n{{user_name}}', user_name: @user.name)
```
garethrees: This is broken.

screen shot 2018-12-10 at 18 29 39

lizconlan: Fixed in 2692084

app/controllers/users/messages_controller.rb
lizconlan: Fixed in bc37e877f16b39f8a01b820b2e760cfd05465c7a (with new spec to check for the expected behaviour)

app/controllers/users/messages_controller.rb
```diff
@@ -27,6 +27,7 @@ def contact
 
   def set_recipient
     @recipient_user = User.find_by(url_name: params[:url_name])
+    raise ActiveRecord::RecordNotFound unless @recipient_user
```
garethrees: https://api.rubyonrails.org/v4.2.10/classes/ActiveRecord/FinderMethods.html#method-i-find_by-21 is a better choice for doing this.

PR #5013, opened 2018-12-14
[#3515] Deter personal requests

### LINKED ISSUE #3515: Help people make requests for personal information direct to commonly requested public bodies
There are a relatively small number of public bodies that commonly get requests for personal information. In Alaveteli we should be able to flag these bodies. When the flag is set the form for making a request should force the requester to answer a question. See this [great example](https://www.privacy.org.nz/further-resources/aboutme-request-my-info-tool/):

selection_001

So on Alaveteli it would ask if you are requesting personal information. If you select yes then the form disappears and instead shows a message about how you can make a request directly to the body. It could also include the body's email address and even a mailto: with the subject/body prefilled from the form. This means the person can just click a link to start making a request for personal information by email.

gbp:
## Relevant issue(s)

Fixes #3515 
Connects to ...

## What does this do?

Adds ability to ask questions on the new request form about the type of request which is going to be made. Allowing us to limit requests for personal information.

## Why was this needed?

To help prevent users from accidentally releasing personal information and to ease WDTK volunteer work load.

## Implementation notes

This should work with JS enabled. Without JS the user shouldn't see any difference to what they see now - EG no questions asked.


app/views/request/new.html.erb
```diff
@@ -216,7 +231,30 @@
...
+            </label>
+          <% end %>
+        </div>
+
+        <% @info_request.public_body.questions.each do |question| %>
+          <div class="request_form_response request_form_response--<%= question.action %>" id="request_form_response--<%= question.key %>">
+            <% if question.action == :allow %>
+              <%= render 'form', f: f %>
+            <% else %>
+              <div class="request_deny_response"><%= _(question.response, public_body_name: @info_request.public_body.name) %></div>
```
garethrees: Hmm this clearly works, in that the response is rendered with the interpolated name, but I'm not sure that this will end up getting translated.

You might need to do something like I did in the `OutgoingMessage::Template` classes https://github.com/mysociety/alaveteli/blob/develop/app/models/outgoing_message/template/initial_request.rb#L21-L26.

Can probably create an issue for this as there's probably more value in getting it out early.

app/models/public_body_question.rb
```diff
@@ -0,0 +1,41 @@
...
+# )
+#
+# PublicBodyQuestion.build(
+#   public_body: home_office,
+#   key: :foi,
+#   question: 'Making an actual FOI request?',
+#   response: :allow
+# )
+#
+class PublicBodyQuestion
```
garethrees: Could do with a couple of specs to avoid regressions.

app/views/request/new.html.erb
```diff
@@ -216,52 +231,29 @@
...
-        </p>
-
-        <div class="form_item_note">
-          <%= _("A one line summary of the information you are requesting, e.g.") %>
-          <%= render :partial => "summary_suggestion" %>
+      <% if @batch || @info_request.public_body.questions.blank? %>
+        <%= render 'form', f: f %>
+      <% else %>
+        <div id="request_form_questions">
+          <p><%= _('Are you asking for <strong>personal information</strong> that should be confidential?') %></p>
```
garethrees: Phrasing suggestion from volunteers:

> Are you asking about your, or others', personal information or personal situation?

gbp: Updated in ce811f4

app/views/request/new.html.erb
gbp: Lets ditch the localisation here and bump to the theme for now. Removed in ce811f4

app/views/request/new.html.erb
```diff
@@ -216,7 +231,30 @@
...
+
+          <% @info_request.public_body.questions.each do |question| %>
+            <label for="<%= question.key %>">
+              <input type="radio" name="request_form_response" id="<%= question.key %>" value="<%= question.key %>">
+              <%= _(question.text) %>
```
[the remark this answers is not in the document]
gbp: Lets ditch the localisation here and bump to the theme. Removed in ce811f4

app/models/public_body_question.rb
```diff
@@ -0,0 +1,41 @@
...
+  end
+
+  def self.fetch(public_body)
+    @all.select { |q| q.public_body == public_body } if @all
+  end
+
+  attr_reader :public_body, :key, :response, :text
+
+  def initialize(options = {})
+    @public_body = options[:public_body]
```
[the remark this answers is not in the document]
gbp: Changed in 4913cd2

app/assets/stylesheets/responsive/_new_request_style.scss
```diff
@@ -121,6 +121,10 @@ input#reedit_button {
   border: none;
 }
 
+.request_deny_response {
+  background-color: #FFF;
+}
+
```
lizconlan: Headache-based nitpick!

I found the results - particularly with the longer question in place - a bit hard on the eyes when it came to seeing where the question ends and the form begins so could we add something like this...

```
.js-loaded {
  div.request_form_response--allow {
    div#request_subject {
      margin-top: 1em;
      padding-top: 1em;
      border-top: 1px solid #ccc;
      border-color: rgba(0, 0, 0, 0.1);
    }
  }
}
```

... to add a divider between the sections (assumes that subject is the first input in the form, only applies when JS is loaded and the form is nested inside a `request_form_response--allow` block)

Before:

screen shot 2019-01-23 at 17 13 25

After:

screen shot 2019-01-23 at 17 12 39

gbp:
> With JS on, if you submit the form without filling in all the info correctly, the radio button is not selected so the form is hidden when the page redraws.

Changed in 284cc68

app/assets/stylesheets/responsive/_new_request_style.scss
gbp: Added in 7ffb5b1

PR #5026, opened 2018-12-20
[#4782] Drop jessie support

### LINKED ISSUE #4782: Drop support for Debian Jessie
Jessie is now end of life (for full support):

https://wiki.debian.org/DebianReleases

We've never used Jessie internally, so I don't think we should continue to support it.

garethrees:
## Relevant issue(s)

Fixes #4782

## What does this do?

* Makes Debian Stretch the default Vagrant OS, instead of the older Debian Jessie
* Removes Debian Jessie Support

## Why was this needed?

Debian Stretch is our current OS of choice. We've never used Jessie internally, and its now end of life.


gbp:
> Since /etc/motd basically gets overwritten at every reboot, you need to instead update /etc/motd.tail 

source: https://wiki.debian.org/motd#How_to_update_your_.2Fetc.2Fmotd

lizconlan:
> > Since /etc/motd basically gets overwritten at every reboot, you need to instead update /etc/motd.tail
>
> source: https://wiki.debian.org/motd#How_to_update_your_.2Fetc.2Fmotd

It looks like something got broken with motd-update and ssh sessions in Jessie and has not been resolved for Stretch https://groups.google.com/forum/#!topic/linux.debian.bugs.dist/ke7Kaq_kHzI

garethrees:
As far as I can tell:

* Setting MOTD (other than actually overwriting `/etc/motd`, which means you lose kernal updates) is a buggy nightmare.
* This change doesn't actually _break_ it for stretch, only highlight that it was already broken due to the new default OS.

Proceeding with PR as-is.

Issue #4902, opened 2018-10-16
Release 0.33


garethrees:
Prep release before making #3969 the default.

garethrees:
Liz's draft release notes https://gist.github.com/lizconlan/65edf0d32ad19e7c5f6bcd0db99501a0

garethrees:
Planned Ruby / OS support list https://gist.github.com/garethrees/ec3c403e1ea0744fb736381785e1e788

garethrees:
Confirmed stretch install on vagrant is 👌

lizconlan:
Done! https://github.com/mysociety/alaveteli/releases/tag/0.33.0.0

Issue #4616, opened 2018-04-16
Release 0.32.1


garethrees:

garethrees:
Closing as we're making #4902.

PR #5027, opened 2018-12-20
Close unclosed tag in English translations

garethrees:
## Relevant issue(s)

Fixes https://github.com/mysociety/whatdotheyknow-theme/issues/446.

## What does this do?

Adds a missed `</small>` tag to translation files.

## Why was this needed?

The unclosed tag was causing a broken page layout (https://github.com/mysociety/whatdotheyknow-theme/issues/446).

Looks like this got missed in bce54f66.

## Implementation notes

I think we can make the change directly to the `.po` files in this case,
because:

* Transifex treats `en` as the source language, so it can't be
  translated through their web interface.
* Transifex doesn't support `en_RW` and `en_UG`, so we make changes to
  these manually.

We'll want to check that this change doesn't get undone next time we
make a release (https://github.com/mysociety/alaveteli/issues/4616 / https://github.com/mysociety/alaveteli/issues/4902).

### DIFF
--- a/locale/en/app.po
+++ b/locale/en/app.po
@@ -3007,7 +3007,7 @@ msgid "The accounts have been left as they previously were."
 msgstr "The accounts have been left as they previously were."
 
 msgid "The authority do <strong>not have</strong> the information <small>(maybe they say who does)</small>"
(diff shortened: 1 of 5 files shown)


garethrees:
Superseded by https://github.com/mysociety/alaveteli/pull/5030

PR #5035, opened 2019-01-03
[#5034] Allow Pro users to see followup form for other users' public requests

### LINKED ISSUE #5034: Pro users (including admins) unable to access followup form for requests they don't own
Code in the followups controller intended to prevent Pro users from seeing embargoed requests that are not their own inadvertently returns a 404 not found error when they try to view the followup page for a public request. (Discovered this while trying to check the behaviour of the followup form while logged in as a pro admin).

lizconlan:
## Relevant issue(s)

Fixes #5034

## What does this do?

Prevents pro users from being served a 404 when they try to access the followup/reply page for a public request they do not have ownership of.

## Why was this needed?

Provided an inconsistent experience for Pro users and made it impossible for admins with Pro status to view/check the followup page for other people's public requests.

### DIFF
--- a/app/controllers/followups_controller.rb
+++ b/app/controllers/followups_controller.rb
@@ -170,10 +170,10 @@ class FollowupsController < ApplicationController
 
   def set_info_request
     if current_user && current_user.is_pro?
-      @info_request = current_user.info_requests.find(params[:request_id].to_i)
-    else
-      @info_request = InfoRequest.not_embargoed.find(params[:request_id].to_i)
+      @info_request =
+        current_user.info_requests.find_by(id: params[:request_id].to_i)
     end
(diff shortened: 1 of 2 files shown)


app/controllers/followups_controller.rb
```diff
@@ -170,8 +170,10 @@ def set_incoming_message
 
   def set_info_request
     if current_user && current_user.is_pro?
-      @info_request = current_user.info_requests.find(params[:request_id].to_i)
-    else
+      @info_request =
+        current_user.info_requests.find_by(id: params[:request_id].to_i)
+    end
+    unless @info_request
```
gbp: `@info_request ||=`

lizconlan: 💡 of course! thanks!

PR #5051, opened 2019-01-14
[Rails 5] Don't call `each_index` directly on an ActiveRecord Relation

lizconlan:
## Relevant issue(s)

Required by 3969

## What does this do?

* Rewrites `InfoRequest#index_of_last_described_event` to avoid use of `each_index`
* Improves the specs for `InfoRequest#last_event_id_needing_description` (which relies on a call to `InfoRequest#index_of_last_described_event` so removing method stubs and mock models was directly useful here)
* Makes `InfoRequest#index_of_last_described_event` a private method as it's not called outside `InfoRequest` (and probably shouldn't be as it describes the internal state of the object)

Reduces Rails 5 spec failures from ~460 to ~280

## Why was this needed?

To accommodate Rails 5 changes to the public interface of ActiveRecord Relations, see: https://flexport.engineering/upgrading-to-rails-5-98c81b56517#3311

### DIFF
--- a/app/models/info_request.rb
+++ b/app/models/info_request.rb
@@ -1319,11 +1319,11 @@ class InfoRequest < ActiveRecord::Base
 
   # Returns index of last event which is described or nil if none described.
   def index_of_last_described_event
-    events = info_request_events
-    events.each_index do |i|
-      revi = events.size - 1 - i
-      m = events[revi]
-      return revi if m.described_state
+    info_request_events.reverse.each_with_index do |event, index|
+      if event.described_state
+        reverse_index = info_request_events.size - 1 - index
+        return reverse_index
+      end
     end
     nil
   end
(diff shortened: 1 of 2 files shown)


app/models/info_request.rb
```diff
@@ -1319,11 +1319,11 @@ def initial_request_text
...
   def index_of_last_described_event
-    events = info_request_events
-    events.each_index do |i|
-      revi = events.size - 1 - i
-      m = events[revi]
-      return revi if m.described_state
+    info_request_events.reverse.each_with_index do |event, index|
+      if event.described_state
+        reverse_index = info_request_events.size - 1 - index
+        return reverse_index
```
lizconlan: On second look _could_ just return the calculation here but I don't find the purpose of the calculation immediately obvious. An alternative - with a presumed tiny efficiency saving over assigning a variable - would be to simply add a comment above `return info_request_events.size - 1 - index` instead but I quite like that the self-describing code

lizconlan:
Part of me wants to expand 3c34b112 to end up with the same result but with smaller steps

lizconlan:
Withdrawn in favour or #5071

PR #5060, opened 2019-01-18
[Rails 5] Fix Rails 5 parameter deprecation warnings

### LINKED ISSUE #5057: [Rails 5] Fix ActionController::Parameters deprecation warnings
Required by #3969 

These are causing a lot of noise in the spec output, and might even be the cause of some of the actual failures. Either way, removing even some of these is going to make the spec output a lot easier to work with.

<details><summary>Method update is deprecated</summary>
<pre><code>DEPRECATION WARNING: Method update is deprecated and will be removed
in Rails 5.1, as `ActionController::Parameters` no longer inherits from hash.
Using this deprecated behavior exposes potential security problems. If you
continue to use this method you may be creating a security vulnerability
in your app that can be exploited. Instead, consider using one of these
documented methods which are not deprecated: 
http://api.rubyonrails.org/v5.0.7/classes/ActionController/Parameters.html
(called from check_in_post_redirect at 
./app/controllers/application_controller.rb:368)
</code></pre>
</details>

<details><summary>Method with_indifferent_access is deprecated</summary>
<pre><code>DEPRECATION WARNING: Method with_indifferent_access is deprecated
and will be removed in Rails 5.1, as `ActionController::Parameters` no
longer inherits from hash. Using this deprecated behavior exposes potential
security problems. If you continue to use this method you may be creating a
security vulnerability in your app that can be exploited. Instead, consider
using one of these documented methods which are not deprecated: 
http://api.rubyonrails.org/v5.0.7/classes/ActionController/Parameters.html
(called from new at 

./app/controllers/followups_controller.rb:130
./app/controllers/concerns/translatable_params.rb:28
</code></pre>
</details>

<details><summary>Comparing equality between `ActionController::Parameters` and a `Hash` is deprecated</summary>
<pre><code>DEPRECATION WARNING: Comparing equality between `ActionController::Parameters`
and a `Hash` is deprecated and will be removed in Rails 5.1. Please only
do comparisons between instances of `ActionController::Parameters`.
If you need to compare to a hash, first convert it using
`ActionController::Parameters#new`.
(called from block (2 levels) in <top (required)> at

./spec/controllers/public_body_controller_spec.rb:471
./spec/controllers/request_controller_spec.rb:996)
</code></pre>
</details>

<details><summary>#to_hash unexpectedly ignores parameter filtering</summary>
<pre><code>DEPRECATION WARNING: #to_hash unexpectedly ignores parameter filtering,
and will change to enforce it in Rails 5.1. Enable `raise_on_unfiltered_parameters`
to respect parameter filtering, which is the default in new applications. For
the existing deprecated behaviour, call #to_unsafe_h instead.
(called from check_in_post_redirect at
./app/controllers/application_controller.rb:369)
</code></pre>
</details>

 --- 
(Grouped together as they feel like they should be similar but may have to break something out - most likely `TranslatableParams` if they prove to be tricker than I thought)

lizconlan:
## Relevant issue(s)

Fixes #5057 

## What does this do?

Handles the new `ActionController::Parameters` using Rails 5 safe methods like `merge!` and `to_unsafe_h` rather than `update, `clone` and direct comparisons with Hash objects. Prevents a deluge of deprecation warnings from appearing in spec output.

## Why was this needed?

Rails 5 raises deprecation warnings when trying to treat parameters as if they were Hashes.

## Implementation notes

Subtlety/annoyance - blindly calling `.to_unsafe_h` on params, a) doesn't work in Rails 4, b) sometimes fails in Rails 5 as calling params can return either an instance of `ActionController::Parameters` or `{}` _sigh_

## Notes to reviewer

No, I'm not convinced that `HashableParams` is the right name for it either - would `ParamsHelper` be better or is it too easily confused with the Rails naming convention for View helpers? Suggestions welcomed.

### DIFF
--- a/app/controllers/concerns/hashable_params.rb
+++ b/app/controllers/concerns/hashable_params.rb
@@ -1,0 +1,17 @@
+# Module that supplies a helper method to ensure that parameters are converted
+# - unalterered - to a Hash whether the input is nil, a Hash (Rails 4), an
+# instance of ActionController::Parameters (Rails 5) or an empty Hash (Rails 5)
+#
+module HashableParams
+  extend ActiveSupport::Concern
+
+  def params_to_unsafe_hash(input_params)
+    return {} if input_params.blank?
+    if rails5?
+      input_params.try(:to_unsafe_h) || input_params.clone
+    else
+      input_params.clone
+    end
+  end
+
+end
--- a/app/controllers/concerns/hashable_params.rb
+++ b/app/controllers/concerns/hashable_params.rb
@@ -8,7 +8,7 @@ module HashableParams
   def params_to_unsafe_hash(input_params)
     return {} if input_params.blank?
     if rails5?
-      input_params.try(:to_unsafe_h) || input_params.clone
+      input_params.to_unsafe_h
     else
       input_params.clone
     end
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -371,7 +371,13 @@ class ApplicationController < ActionController::Base
         post_redirect =
           PostRedirect.find_by_token(session[:post_redirect_token])
         if post_redirect
-          params.update(post_redirect.post_params)
+          post_redirect_params =
+            params_to_unsafe_hash(post_redirect.post_params)
+          if rails5?
+            params.merge!(post_redirect_params)
+          else
+            params.update(post_redirect_params)
+          end
           params[:post_redirect_user] = post_redirect.user
         end
       else
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -530,4 +530,6 @@ class ApplicationController < ActionController::Base
 
   # Site-wide access to configuration settings
   include ConfigHelper
+
+  include HashableParams
 end
--- a/app/controllers/concerns/translatable_params.rb
+++ b/app/controllers/concerns/translatable_params.rb
@@ -15,6 +15,7 @@ module TranslatableParams
   #        containing the list of whitelisted keys for
   #        the base model, and for translations, respectively.
   class WhitelistedParams
+    include ::HashableParams
 
     attr_reader :keys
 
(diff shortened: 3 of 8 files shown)


app/controllers/concerns/hashable_params.rb
```diff
@@ -8,7 +8,7 @@ module HashableParams
   def params_to_hash(input_params)
     return {} if input_params.blank?
     if rails5?
-      input_params.to_unsafe_h
+      input_params.try(:to_unsafe_h) || input_params
```
lizconlan: Turns out - after running this and the `each_index` fix together - this does sometimes end up with Hash input - so this assumes that's what's happened and returns the unchanged input as the result if it can't call `to_unsafe_h`

lizconlan: I think it's advisable to deal with the possibility of it receiving a `Hash` (blank params seem to be passed as an empty Hash so there's that, not entirely sure how we end up passing it populated Hashes now and again but ... having the "just call this, it'll be fine" method deal with everything feels like the right first step).

But... you've helped me realise that we can drop the `if rails5?` check instead and just return `input_params.clone` if the input doesn't respond to `to_unsafe_h`

5918a2e collapses my initial nonsense down to make this a 2 line method

lizconlan: Ok so - less lazy answer - aside from where I've written a test that supplies a populated Hash to prove it can cope with it, it genuinely receives a Hash as input when called from `TranslatableParams::WhitelistedParams` and while I _could_ treat that as a special case it leaves me worried that there might be other exceptions involving constructed and/or nested params so I think I'll keep it as the responsibility of the method to Do The Right Thing.

Less satisfying in 43b27144

lizconlan:
I've squashed those commits and changed the method name, maybe I can make the `TranslatableParams::WhitelistedParams` code a little better...

gbp:
@lizconlan what is the state of this PR. Guessing the experimental commits aren't meant to be reviewed yet as CI fails?

lizconlan:
@gbp sorry - had intended a quick fix then forgot to switch the assignment when it wasn't

lizconlan:
Probably obvious from the dates but the last 3 commits are new

PR #5061, opened 2019-01-18
[Rails 5] Avoid Rails 5 deprecation warning for passing conditions to destroy_all

lizconlan:
Calls `destroy_all` on the results of a query instead of passing parameters to `destroy_all`  (deprecated in Rails 5.0)

Warning output:

```
DEPRECATION WARNING: Passing conditions to destroy_all is deprecated
and will be removed in Rails 5.1. To achieve the same use
where(conditions).destroy_all.
```

## Note to reviewer

My fault - sorry!

## Relevant issue(s)

Required by #3969

### DIFF
--- a/spec/models/info_request_spec.rb
+++ b/spec/models/info_request_spec.rb
@@ -1800,7 +1800,7 @@ describe InfoRequest do
     end
 
     context 'email with an id mistyped using letters and missing punctuation' do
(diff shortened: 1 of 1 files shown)


lizconlan:
> Its more useful to put the deprecation warning in the commit message than just the PR description.

Amended and repushed!

PR #5083, opened 2019-01-30
Update recaptcha site URL

garethrees:
## What does this do?

Update recaptcha site URL

## Why was this needed?

Old one was out of date – missed when doing https://github.com/mysociety/alaveteli/pull/4889

### DIFF
--- a/config/general.yml-example
+++ b/config/general.yml-example
@@ -495,7 +495,7 @@ READ_ONLY: ''
 STAGING_SITE: 1
 
 # Recaptcha, for detecting humans. Get keys here:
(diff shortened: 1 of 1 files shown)


garethrees:
> Probably should specify we use Recaptcha V2. The new V3 works different and isn't currently supported by the recaptcha gem.

Out of scope for this PR, but a good idea👍 . I think we want to upgrade to v3, so that would make a good ticket.

lizconlan:
> Out of scope for this PR

As this might be an obvious-only-to-us thing, maybe we could add a note for reusers in the comment that v2 is required

Issue #4484, opened 2018-01-26
ActionController::InvalidAuthenticityToken Errors


garethrees:
Some stats on the `ActionController::InvalidAuthenticityToken` errors we're getting through.

```csv
location,last_occurrence,total_occurrences
request#new,2018-01-25,111
comment#new,2017-12-16,18
password_changes#create,2017-12-07,11
request#describe_state,2018-01-19,10
followups#preview,2018-01-21,6
account_request#create,2017-11-14,4
general#not_found,2017-12-14,1
```

When we investigate these, we should probably create a ticket for each so that notes are collected next to the exception report.

garethrees:
I've just experienced some weirdness that may account for this.

* I'm signed in in a tab
* In a new tab, I click a WDTK URL from Twitter
* I land on WDTK, not signed in.
* I click "sign in" as I want to go to the admin page of the URL
* I get immediately redirected back, signed in, without having to enter password

In this case, if I'd submitted a form I don't think it would contain an auth token, but because I'm actually signed in we'd check it, and it would fail because its blank.

HelenWDTK:
This issue has been automatically closed due to a lack of discussion or resolution for over 12 months.
Should we decide to revisit this issue in the future, it can be reopened.

PR #5089, opened 2019-01-31
Make our custom form_authenticity_token behaviour call the Rails method

### LINKED ISSUE #5087: [Rails 5] Handle new CSRF behaviour
https://guides.rubyonrails.org/upgrading_ruby_on_rails.html#per-form-csrf-tokens

With our current code, when we call `form_tag` (and probably `form_for`) under Rails 5 we get an error:

```
ArgumentError: wrong number of arguments (given 1, expected 0)
```

As `form_authenticity_token` is getting called, passing in the `form_tag` args to a method that's not expecting any args at all.

lizconlan:
## Relevant issue(s)

Fixes #5087 

## What does this do?

* Stops Rails 5 raising a `ArgumentError (wrong number of arguments (given 0, expected 1)` error when calling the `form_tag` helper with a logged in user
* Updates our override of `form_authenticity_token` so that is works as expected with Rails 4.2 and Rails 5.0

## Why was this needed?

* Rails 5.0 wasn't working properly with forms when logged in to the site (see above)
* We were mistakenly running the old Rails 4.1 code for generating CSRF tokens against Rails 4.2+, unintentionally bypassing the newer `masked_authenticity_token` code - this may have been a contributing factor to the high number of `InvalidAuthenticityToken` errors we've been seeing in production (#4484)

## Note for reviewer

This whole override appears to be the cause of the `InvalidAuthenticityToken` problem (at least in the case of passing in a blank token). In a recent case, the user had signed in in a new tab then posted the form in the original (blank token in the POST data, set token in the session cookie ... 500 error served).

Repeating those steps with this method removed entirely from  the controller prevents the problem (for both versions of Rails). Presumably this would come at a cost to the Varnish cache ([per previous work on this](https://github.com/mysociety/alaveteli/pull/2308)).

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -256,6 +256,8 @@ class ApplicationController < ActionController::Base
     !session[:user_id].nil?
   end
 
+  # Override the Rails method to only set the CSRF form token if there is a
+  # logged in user
   def form_authenticity_token(*args)
     if user?
       if rails5?
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -256,9 +256,14 @@ class ApplicationController < ActionController::Base
     !session[:user_id].nil?
   end
 
-  def form_authenticity_token
+  def form_authenticity_token(*args)
     if user?
-      session[:_csrf_token] ||= SecureRandom.base64(32)
+      session[:_csrf_token] ||=
+        if rails5?
+          super
+        else
+          super()
+        end
     end
   end
 
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -258,12 +258,11 @@ class ApplicationController < ActionController::Base
 
   def form_authenticity_token(*args)
     if user?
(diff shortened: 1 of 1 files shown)


lizconlan:
https://www.fastly.com/blog/caching-uncacheable-csrf-security

gbp:
@lizconlan Ci has passed with my change and there is the same number of failures as the previous commit. Can you see anything wrong with it?

before: https://travis-ci.org/mysociety/alaveteli/jobs/488947714
after: https://travis-ci.org/mysociety/alaveteli/jobs/503661508

lizconlan:
@gbp I like your version better as it should solve the `InvalidAuthenticityToken` issues **but** I think - without some more work that I haven't really looked into - it's going to cause problems with the Varnish cache because it still generates the token

screen shot 2019-03-08 at 16 59 32

screen shot 2019-03-08 at 16 59 49

See: https://github.com/mysociety/alaveteli/pull/2308/commits/17882311cce273c62322959a2deb31c2dc5cbae9

PR #5094, opened 2019-02-06
[Rails 5] Prevent `non-sanitized request parameters!` errors

### LINKED ISSUE #5074: [Rails 5] Fix "non-sanitized params" errors
Required by #3969 

```
When searching should not strip quotes from quoted query
    Failure/Error: url = url_for(routing_info)

    ArgumentError:
      Attempting to generate a URL from non-sanitized request parameters! An attacker can inject malicious data into the generated URL, such as changing the host. Whitelist and sanitize passed parameters to be secure.
    # /home/vagrant/bundle/ruby/2.3.0/gems/actionpack-5.0.7/lib/action_dispatch/routing/url_for.rb:176:in `url_for'
    # ./app/helpers/link_to_helper.rb:236:in `search_url'
    # ./app/controllers/general_controller.rb:81:in `search_redirect'


    ./spec/integration/search_request_spec.rb:13
```

```
RequestController when handling prominence when the request is requester_only should not cache an attachment when showing an attachment to the requester
  Failure/Error: path = url_for(param)

  ArgumentError:
    Attempting to generate a URL from non-sanitized request parameters! An attacker can inject malicious data into the generated URL, such as changing the host. Whitelist and sanitize passed parameters to be secure.
  # /home/vagrant/bundle/ruby/2.3.0/gems/actionpack-5.0.7/lib/action_dispatch/routing/url_for.rb:176:in `url_for'
  # ./app/controllers/application_controller.rb:212:in `foi_fragment_cache_part_path'
  # ./app/controllers/application_controller.rb:220:in `foi_fragment_cache_path'
  # ./app/controllers/request_controller.rb:610:in `cache_attachments'
  # /home/vagrant/bundle/ruby/2.3.0/gems/activesupport-5.0.7/lib/active_support/callbacks.rb:382:in `block in make_lambda'

  # ./spec/controllers/request_controller_spec.rb:797
```

lizconlan:
## Relevant issue(s)

Fixes #5074

## What does this do?

Permits valid parameter key before they're passed into `url_for` 

## Why was this needed?

Avoids Rails 5 raising `ArgumentError: Attempting to generate a URL from non-sanitized request parameters!`

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -209,6 +209,11 @@ class ApplicationController < ActionController::Base
   # URL using the first three digits of the info request id, because we can't
   # have more than 32,000 entries in one directory on an ext3 filesystem.
   def foi_fragment_cache_part_path(param)
+    if rails5?
+      param =
+        param.permit(:incoming_message_id, :part, :file_name, :id,
+                     :only_path, :locale, :skip_cache)
+    end
     path = url_for(param)
     id = param['id'] || param[:id]
     first_three_digits = id.to_s[0..2]
--- a/app/helpers/link_to_helper.rb
+++ b/app/helpers/link_to_helper.rb
@@ -233,6 +233,18 @@ module LinkToHelper
     if !options.nil?
       routing_info = options.merge(routing_info)
     end
+
+    if rails5?
+      if routing_info.kind_of?(Hash)
+        routing_info = ActionController::Parameters.new(routing_info)
+      end
+
+      allowed_keys =
+        %w[query latest_status view combined only_path controller action page]
+      unpermitted = routing_info.keys - allowed_keys
+      routing_info = routing_info.reject { |k| unpermitted.include?(k) }.permit!
+    end
+
     url = url_for(routing_info)
     # Here we can't escape the slashes, as RFC 2396 doesn't allow slashes
     # within a path component. Rails is assuming when generating URLs that


app/helpers/link_to_helper.rb
```diff
@@ -294,4 +292,9 @@ def message_path(message, options = {})
     message_url(message, options.merge(:only_path => true))
   end
 
+  def sanitize_params(raw_params)
+    unsafe_keys = %w[protocol host]
+    raw_params.reject { |k| unsafe_keys.include?(k) }.permit!
+  end
```
gbp: FWIW - I am not sure we can be 100% sure this returns parameters which have been sanitised in all cases. It works in `current_path_with_locale` because we're using the `path_only` option.

lizconlan: Fair point - although I've seen this [done elsewhere, but with `port` excluded as well](https://gitlab.com/gitlab-org/gitlab-ce/merge_requests/18241)

May try to work out what the actual valid search params are and spend a bit more time trying to permit them. Came unstuck with `latest_status` for reasons which escaped me at the time but ... maybe something more like ...

```rb
allowed_keys = %w[query latest_status view combined only_path controller action page]
diff_keys = routing_info.keys - allowed_keys
routing_info.reject { |k| diff_keys.include?(k) }.permit!
```

Sample params:
```
<ActionController::Parameters {
  "query"=>"bob",
  "latest_status"=>["successful"],
  "controller"=>"general",
  "action"=>"search",
  "combined"=>"bob/requests/relevant",
  "only_path"=>true,
  "view"=>nil} permitted: false>
```

app/controllers/application_controller.rb
```diff
@@ -209,6 +209,11 @@ def render_hidden(template='request/hidden', opts = {})
   # URL using the first three digits of the info request id, because we can't
   # have more than 32,000 entries in one directory on an ext3 filesystem.
   def foi_fragment_cache_part_path(param)
+    if rails5?
+      param =
+        param.permit(:incoming_message_id, :part, :file_name, :id,
+                     :only_path, :locale, :skip_cache)
```
gbp: Unclear where these params are sourced from. If I understand correctly they are from https://github.com/mysociety/alaveteli/blob/develop/config/routes.rb#L123-L130 https://github.com/mysociety/alaveteli/blob/develop/app/controllers/request_controller.rb#L609 and https://github.com/mysociety/alaveteli/blob/develop/app/controllers/request_controller.rb#L679 maybe the commit message could be expanded.

lizconlan: I think it was finalised by setting a breakpoint and watching what was passed in as a double-check but I've expanded the commit message as follows:

```
The permitted parameter list is derived from:
  We generally expect :locale so that's included here
  The get_attachment and get_attachment_as_html routes specify
    :id, :incoming_message_id, :part, and :file_name
  RequestContoller#cache_attachments
    expects :skip_cache, :only_path
```

lizconlan:
Uh, I think i managed to merge this on GitHub by mistake by fat-fingering the buttons on HuBoard. Sorry!

PR #5098, opened 2019-02-07
[Rails 5] Small spec fixes

### LINKED ISSUE #5077: [Rails 5] Fix spec expectations - mime types
Required by #3969 

Rails 5 appears to see the full mime type string instead of the `:json` - not sure what the extent is here due to other errors causing integration specs to return the wrong content and genuinely complain of finding e.g. a pdf instead of html.

```
When searching using JSON format should return JSON formatted results
  Failure/Error: expect(response.content_type).to eq(:json)

    expected: :json
         got: "application/json"

    (compared using ==)

    Diff:
    @@ -1,2 +1,2 @@
    -:json
    +"application/json"

  # ./spec/integration/search_request_spec.rb:94
```

lizconlan:
## Relevant issue(s)

Required by #3969 
Fixes #5077

## What does this do?

Minor spec fixes (where Rails 5 changes were breaking the specs rather than there being a problem with the app code)

## Note to reviewer

~The cherry-picked commit (33195ae) to update the Rails 5 lock file is only needed to allow CI to run against Rails 5 (avoids rebasing and losing the original history)~

### DIFF
--- a/spec/controllers/users/sessions_controller_spec.rb
+++ b/spec/controllers/users/sessions_controller_spec.rb
@@ -259,6 +259,7 @@ describe Users::SessionsController do
           msg = "Attempted signup from suspected spammer, " \
                 "email: spammer@example.com, " \
                 "name: 'Download New Person 1080p!'"
+          allow(Rails.logger).to receive(:info)
           expect(Rails.logger).to receive(:info).with(msg)
 
           do_signin(user.email, 'password1234')
--- a/spec/models/incoming_message_spec.rb
+++ b/spec/models/incoming_message_spec.rb
@@ -111,6 +111,7 @@ describe IncomingMessage do
       message = FactoryBot.create(:incoming_message)
       message.raw_email.data = raw_email_data
       message.parse_raw_email!(true)
+      message.reload
       FactoryBot.create(:censor_rule,
                         :text => 'Person',
                         :info_request => message.info_request)
--- a/spec/controllers/one_time_passwords_controller_spec.rb
+++ b/spec/controllers/one_time_passwords_controller_spec.rb
@@ -164,7 +164,8 @@ describe OneTimePasswordsController do
 
     it 'renders #show on failure' do
       user = FactoryBot.create(:user, :otp_enabled => true)
-      allow_any_instance_of(User).to receive(:save).and_return(false)
+      allow_any_instance_of(User).
+        to receive(:increment!).and_return(false)
       session[:user_id] = user.id
       put :update
       expect(response).to render_template(:show)
@@ -172,7 +173,8 @@ describe OneTimePasswordsController do
 
     it 'warns the user on failure' do
       user = FactoryBot.create(:user, :otp_enabled => true)
-      allow_any_instance_of(User).to receive(:save).and_return(false)
+      allow_any_instance_of(User).
+        to receive(:increment!).and_return(false)
       session[:user_id] = user.id
       put :update
       expect(flash[:error]).
--- a/spec/controllers/user_controller_spec.rb
+++ b/spec/controllers/user_controller_spec.rb
@@ -894,6 +894,7 @@ describe UserController, "when signing up" do
         msg = "Attempted signup from suspected spammer, " \
               "email: spammer@example.com, " \
               "name: 'Download New Person 1080p!'"
+        allow(Rails.logger).to receive(:info)
         expect(Rails.logger).to receive(:info).with(msg)
 
         post :signup, params: {
--- a/spec/integration/search_request_spec.rb
+++ b/spec/integration/search_request_spec.rb
@@ -91,7 +91,7 @@ describe "When searching" do
       get '/feed/search/chicken.json'
       response_data = JSON.parse(response.body, :symbolize_names => true)
 
-      expect(response.content_type).to eq(:json)
+      expect(response.content_type).to eq('application/json')
       expect(response_data.size).to eql(1)
       expect(response_data.first[:info_request][:title]).
         to eq('How much public money is wasted on breeding naughty chickens?')
--- a/spec/models/info_request_event_spec.rb
+++ b/spec/models/info_request_event_spec.rb
@@ -30,7 +30,7 @@ describe InfoRequestEvent do
     it 'should not add an error message for described_state if it is valid' do
       ire = InfoRequestEvent.new(:described_state => 'waiting_response')
       ire.valid?
-      expect(ire.errors.messages[:described_state]).to be_nil
+      expect(ire.errors.messages[:described_state]).to be_blank
     end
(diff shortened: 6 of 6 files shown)


spec/models/incoming_message_spec.rb
```diff
@@ -115,7 +115,7 @@
                         :text => 'Person',
                         :info_request => message.info_request)
 
-      expect(message.safe_mail_from).to eq('FOI [REDACTED]')
+      expect(message.reload.safe_mail_from).to eq('FOI [REDACTED]')
```
garethrees: Would be good to include the failing spec output in the commit message, as its really difficult to understand _why_ this was failing.

spec/controllers/users/sessions_controller_spec.rb
```diff
@@ -259,9 +259,9 @@ def do_signin(email, password)
           msg = "Attempted signup from suspected spammer, " \
                 "email: spammer@example.com, " \
                 "name: 'Download New Person 1080p!'"
-          expect(Rails.logger).to receive(:info).with(msg)
-
+          allow(Rails.logger).to receive(:info)
           do_signin(user.email, 'password1234')
+          expect(Rails.logger).to have_received(:info).with(msg)
```
garethrees: Would be good to include the failing spec output in the commit message, as its really difficult to understand _why_ this was failing.

Seems fine in principle though.

spec/models/incoming_message_spec.rb
garethrees: Have you looked at whether its safe to do this in terms of it working as we'd expect?

When calling `safe_mail_from` I _think_ I'd always expect to get a fresh cache of censor rules as we haven't directly loaded the associated info request or censor rules in the spec setup (though I haven't looked through the full method call chain in detail).

```diff
diff --git a/spec/models/incoming_message_spec.rb b/spec/models/incoming_message_spec.rb
index c47e6fa3f..cf4c66a9b 100644
--- a/spec/models/incoming_message_spec.rb
+++ b/spec/models/incoming_message_spec.rb
@@ -114,6 +114,7 @@ describe IncomingMessage do
       FactoryBot.create(:censor_rule,
                         :text => 'Person',
                         :info_request => message.info_request)
+      binding.pry
 
       expect(message.safe_mail_from).to eq('FOI [REDACTED]')
     end
```

```
# Rails 4
message.info_request.censor_rules
# => [#<CensorRule:0x00000010b08fc8
  id: 875,
  info_request_id: 112,
  user_id: nil,
  public_body_id: nil,
  text: "Person",
  replacement: "[REDACTED]",
  last_edit_editor: "FactoryBot",
  last_edit_comment: "Modified by rspec",
  created_at: Tue, 12 Feb 2019 16:21:29 UTC +00:00,
  updated_at: Tue, 12 Feb 2019 16:21:29 UTC +00:00,
  regexp: nil>]

# Rails 5
message.info_request.censor_rules
# => []
```

From the info in this commit message its not clear whether we've established that in real operation we always end up with the most recently added censor rules. Basically I'm wondering whether the `reload` should be in the app code itself, rather than the spec.

spec/integration/user_profile_updates_spec.rb
```diff
@@ -13,8 +13,9 @@
       it "page displays thank you message with nudge to upload photo" do
         using_session(login(user)) do
           msg = "Thanks for changing the text about you on your " \
-                "profile.Next... You can " \
+                "profile.\nNext... You can " \
```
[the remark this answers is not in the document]
lizconlan: In the browser it's HTML - I think it's just a difference in the text representation of the page that gets passed to the `have_content` RSpec matcher. I looks like it's started putting a newline char between each block level HTML element when calling `page.text`.

(Sorry - in hindsight, "text content of the page" was not a clear description of that at all.)

Rails 4:
```
# from page.body
<p>Thanks for changing the text about you on your profile.</p><p><strong>Next...</strong> You can upload a profile photograph too.</p>

# from from page.text
Thanks for changing the text about you on your profile.Next... You can upload a profile photograph too.
```


Rails 5:
```
# from page.body
<p>Thanks for changing the text about you on your profile.</p><p><strong>Next...</strong> You can upload a profile photograph too.</p>

# from page.text
Thanks for changing the text about you on your profile.\nNext... You can upload a profile photograph too.
```

spec/models/incoming_message_spec.rb
lizconlan: Hmm... looks like it's actually `parse_raw_email!(true)` that causes the need to reload as this works:

```
      message = FactoryBot.create(:incoming_message)
      message.raw_email.data = raw_email_data
      message.parse_raw_email!(true)
      message.reload
      FactoryBot.create(:censor_rule,
                        :text => 'Person',
                        :info_request => message.info_request)

      expect(message.safe_mail_from).to eq('FOI [REDACTED]')
```

lizconlan: which makes me wonder whether I need to do this instead...

```diff
diff --git a/app/models/incoming_message.rb b/app/models/incoming_message.rb
index 52112e9d1..849c631bf 100644
--- a/app/models/incoming_message.rb
+++ b/app/models/incoming_message.rb
@@ -110,6 +110,7 @@ class IncomingMessage < ActiveRecord::Base
         self.foi_attachments.reload
         self.save!
       end
+      self.info_request.reload
     end
   end
```

spec/integration/user_profile_updates_spec.rb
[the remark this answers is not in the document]
lizconlan: Ah - we're loading a newer Capybara for Rails 5 and it's this: https://stackoverflow.com/a/54672926

spec/controllers/users/sessions_controller_spec.rb
lizconlan: turns out, it was the `allow` that was the magic bit - fixed in https://github.com/mysociety/alaveteli/pull/5098/commits/14a1c17af6fa7f5e823c310f24434d9ca2a5053f

spec/integration/user_profile_updates_spec.rb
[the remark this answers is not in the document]
lizconlan: Extended commit message reworded (no code changes) in 531df73c9

spec/models/incoming_message_spec.rb
lizconlan: Ah - found it, fixed in `parse_raw_email!` saves and a callback to update the related request inadvertently caches it before we add the censor rule. Fixed in 116e80907

gbp: @lizconlan from my understanding your analysis looks correct and what you've done in 4d02788, reloading before the censor rule is created, looks to be the simplest solution to this.

PR #5110, opened 2019-02-18
[Rails 5][#5078] Fix stripe spec errors

### LINKED ISSUE #5078: [Rails 5] Fix Stripe spec failures
Required by #3969

Not entirely sure what's happening here. `symbolize_keys` is being called on `payload` (which for the example given below is set on line 40)

```
AlaveteliPro::StripeWebhooksController#receive updating the Stripe charge description when a payment succeeds when there is no charge for an invoice does not attempt to update the nil charge
     Failure/Error: post :receive, params: payload

     NoMethodError:
       undefined method `symbolize_keys' for #<String:0x0000000004f16018>
     # /home/vagrant/bundle/ruby/2.3.0/gems/actionpack-5.0.7/lib/action_controller/test_case.rb:523:in `process'
     # /home/vagrant/bundle/ruby/2.3.0/gems/rails-controller-testing-1.0.2/lib/rails/controller/testing/template_assertions.rb:61:in `process'
     # /home/vagrant/bundle/ruby/2.3.0/gems/actionpack-5.0.7/lib/action_controller/test_case.rb:644:in `process_with_kwargs'
     # /home/vagrant/bundle/ruby/2.3.0/gems/actionpack-5.0.7/lib/action_controller/test_case.rb:397:in `post'
     # /home/vagrant/bundle/ruby/2.3.0/gems/rails-controller-testing-1.0.2/lib/rails/controller/testing/integration.rb:12:in `block (2 levels) in <module:Integration>'
     # ./spec/controllers/alaveteli_pro/stripe_webhooks_controller_spec.rb:410:in `block (4 levels) in <top (required)>'
```

lizconlan:
## Relevant issue(s)

Fixes #5078 

## What does this do?

* Refactors `stripe_webhooks_controller_spec.rb` (mostly to remove repetition)
* Uses the real Stripe method of computing signatures instead of recreating it ourselves
* Converts the mock payload into a hash rather than a string for Rails 5 tests (otherwise we get `undefined method symbolize_keys` errors - presumably a change to how controller params are handled)
* When creating a signature, converts the mock payload to a JSON representation of a string for Rails 5 (otherwise the string representation of our payload used when constructing the signature uses hashrockets and line breaks and doesn't match Stripe's expectation)
* Check against shorter versions of email subject lines as these are getting truncated under Rails 5
* Fixes 27 failing Rails 5 specs

## Why was this needed?

* Stripe specs were failing under Rails 5
* Specs were failing against newer versions of the ruby-stripe-mock gem

### DIFF
--- a/spec/controllers/alaveteli_pro/stripe_webhooks_controller_spec.rb
+++ b/spec/controllers/alaveteli_pro/stripe_webhooks_controller_spec.rb
@@ -53,7 +53,13 @@ describe AlaveteliPro::StripeWebhooksController, feature: [:alaveteli_pro, :pro_
       StripeMock.mock_webhook_event('customer.subscription.deleted')
     end
 
-    let(:payload) { stripe_event.to_s }
+    let(:payload) do
+      if rails5?
+        stripe_event.to_hash
+      else
+        stripe_event.to_s
+      end
+    end
 
     before do
       allow(AlaveteliConfiguration).to receive(:stripe_namespace).
@@ -130,9 +136,8 @@ describe AlaveteliPro::StripeWebhooksController, feature: [:alaveteli_pro, :pro_
 
     context 'receiving an unhandled notification type' do
 
-      let(:payload) do
-        stripe_event.
-          to_s.gsub!('customer.subscription.deleted', 'custom.random_event')
+      before do
+        stripe_event.type = 'custom.unhandle_event'
       end
 
       it 'sends an exception email' do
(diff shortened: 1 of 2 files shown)


spec/controllers/alaveteli_pro/stripe_webhooks_controller_spec.rb
```diff
@@ -131,8 +137,14 @@
     context 'receiving an unhandled notification type' do
 
       let(:payload) do
-        stripe_event.
-          to_s.gsub!('customer.subscription.deleted', 'custom.random_event')
+        event_string =
```
gbp: This whole let block could be replaced with just:
```ruby
      before do
        stripe_event.type = 'custom.unhandle_event'
      end
```

spec/support/stripe_helpers.rb
```diff
@@ -4,10 +4,16 @@ def signed_headers(signing_secret: nil, payload: nil, timestamp: Time.zone.now)
   raise ArgumentError, "must provide payload key" unless payload
 
   timestamp = timestamp.to_i
+  payload_data =
+    if rails5?
+      payload.to_json.to_s
```
gbp: Don't need to call `#to_s`

PR #5123, opened 2019-03-05
[Rails 5] Fix capybara specs

lizconlan:
## Relevant issue(s)

Required by #3969 

## What does this do?

* Allows for non-normalised whitespace in specs
* Removes trailing whitespace from expectations
* Allows optional (and variable) whitespace in clickable links (e.g. on the Pro dashboard)

## Why was this needed?

Our Rails 5 specs load capybara v3 (rather than v2) which [no longer normalises whitespace by default](
https://stackoverflow.com/a/54672926), which broke some of our specs.


gbp:
Further discussion on this here: https://github.com/mysociety/alaveteli/pull/5098#discussion_r256020800

PR #5125, opened 2019-03-07
[Rails 5] Routing related fixes

### LINKED ISSUE #5059: [Rails 5] Using a dynamic :action segment in a route is deprecated
Required by #3969

```
DEPRECATION WARNING: Using a dynamic :action segment in a route
is deprecated and will be removed in Rails 5.2. (called from block in
<top (required)> at ./config/routes.rb:395)
```

It's complaining about this:
```rb
match '/help/:action' => 'help#action',
  :as => :help_general,
  :via => :get
```

Which appears to have been introduced (several Rails upgrades ago) by: https://github.com/mysociety/alaveteli/commit/5acedd09b79642d58c6de5db872f82198b2f2c73

lizconlan:
## Relevant issue(s)

Fixes #5059
Required by #3969

## What does this do?

* Fixes the `destroy_batch` route (prevents Rails 5 from seeing it)
* Replace the dynamic route that causes a Rails 5 deprecation warning

## Why was this needed?

Changes in Rails 5 that caused:
* spec failures when trying to use the `destroy_batch` route (would probably have failed in the browser as well)
* an early warning that dynamic routes were going to be removed that caused spec failures because test which expected empty output were receiving a deprecation message instead

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -685,7 +685,7 @@ Rails.application.routes.draw do
       end
       resources :embargoes, :only => [:destroy, :create] do
         collection do
-          post :destroy_batch, :only => [:destroy]
+          post :destroy_batch
         end
       end
       resources :embargo_extensions, :only => [:create] do
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -365,7 +365,7 @@ Rails.application.routes.draw do
   ####
 
   #### Help controller
-  match '/help/unhappy/:url_title' => 'help#unhappy',
+  match '/help/unhappy(/:url_title)' => 'help#unhappy',
         :as => :help_unhappy,
         :via => :get
   match '/help/about' => 'help#about',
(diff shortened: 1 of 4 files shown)


config/routes.rb
```diff
@@ -685,7 +685,7 @@
       end
       resources :embargoes, :only => [:destroy, :create] do
         collection do
-          post :destroy_batch, :only => [:destroy]
+          post :destroy_batch
```
gbp: Agreed this looks like a mistake.

config/routes.rb
```diff
@@ -368,6 +368,8 @@
   match '/help/unhappy/:url_title' => 'help#unhappy',
         :as => :help_unhappy,
         :via => :get
+  match '/help/unhappy' => 'help#unhappy',
+        :via => :get
```
gbp: Not a big issue but this route could be combined with the one above as:
```ruby
match '/help/unhappy(/:url_title)' => 'help#unhappy',
```

Then where we're calling `help_general_url(:template => 'unhappy' ...` we could just call `help_unhappy_url(...`

PR #5141, opened 2019-03-26
[#5140] Only update InfoRequest#url_title if the title has changed

### LINKED ISSUE #5140: url_title incorrectly incremented after update in Batch request
Details in https://groups.google.com/a/mysociety.org/forum/#!topic/alaveteli-pro/O9wqdC4U6nA

lizconlan:
## Relevant issue(s)

Fixes #5140

## What does this do?

Avoids unintentionally assigning a new suffix number to the url of an
existing `InfoRequest` that's in the middle of a sequence (we've seen this
in production with batch requests, but it applies to any `url_title` with
a numerical suffix)

## Why was this needed?

When we update a request metadata in the admin interface, we [call `@info_request.update_attributes(info_request_params)`](https://github.com/mysociety/alaveteli/blob/d5f659e029b440727b9341fce8429389a8edd2cb/app/controllers/admin_request_controller.rb#L54) which always tries to update `InfoRequest#title` even if it hasn't changed, unintentionally renumbering mid-sequence `url_title` sequential suffixes. This results in broken URLs for the affected requests.

### DIFF
--- a/spec/models/info_request_spec.rb
+++ b/spec/models/info_request_spec.rb
@@ -2022,6 +2022,28 @@ describe InfoRequest do
 
   end
 
+  describe '#title=' do
+
+    let(:test_requests) do
+      3.times.map { FactoryBot.create(:info_request, title: 'URL test') }
+    end
+
+    let(:request) do
+      test_requests[1]
+    end
+
+    it 'updates url_title when the title is changed' do
+      request.update_attributes(title: 'Something new')
+      expect(request.url_title).to eq('something_new')
+    end
+
+    it 'does not update url_title when the same title is assigned' do
+      request.update_attributes(title: request.title.dup)
+      expect(request.url_title).to eq('url_test_2')
+    end
+
+  end
+
   describe '#last_event_id_needing_description' do
     subject { info_request.last_event_id_needing_description }
     let(:info_request) { FactoryBot.create(:successful_request) }
--- a/app/models/info_request/sluggable.rb
+++ b/app/models/info_request/sluggable.rb
@@ -13,7 +13,7 @@ module InfoRequest::Sluggable
   # When title is changed, also change the URL title
   def title=(title)
     super.tap do
-      update_url_title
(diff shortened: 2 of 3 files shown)


spec/models/info_request_spec.rb
```diff
@@ -2022,6 +2022,32 @@
 
   end
 
+  describe 'updating the URL title' do
+
+    let(:test_requests) do
+      _requests = []
```
garethrees: Can simplify this a bit:

```ruby
      3.times.map do
         FactoryBot.create(:info_request, title: 'URL test')
      end
```

spec/models/info_request_spec.rb
```diff
@@ -2022,6 +2022,32 @@
...
+    end
+
+    it 'updates url_title when the title is changed' do
+      request.update_attributes(title: 'Something new')
+      expect(request.url_title).to eq('something_new')
+    end
+
+    it 'does not update url_title when the same title is assigned' do
+      request.update_attributes(title: request.title.dup)
+      expect(request.reload.url_title).to eq('url_test_2')
```
garethrees: Don't need to call `reload` here for the spec to pass, or for the spec to fail without the code change.

spec/models/info_request_spec.rb
lizconlan: D'oh. That's much nicer - cheers!

spec/models/info_request_spec.rb
```diff
@@ -2022,6 +2022,32 @@
 
   end
 
+  describe 'updating the URL title' do
```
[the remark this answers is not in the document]
lizconlan: I went back and forth on that; as we both considered that, I'll change it

spec/models/info_request_spec.rb
lizconlan: cheers! left in by mistake while I was trying to pin down what was happening 😞

PR #5134, opened 2019-03-15
hotfix/0.25.0.27 (Rails security fixes)

lizconlan:
## What does this do?

Backports the CVE fixes in Rails 4.2.11.1 to Rails 3.2.x

## Why was this needed?

We still have 3.2 instances and 3.2 is out of support

## Note to reviewer

Comment out `require 'mime_negotiation_patch'` in `config/initializers/alaveteli.rb` to see the 2 spec fails that the patch fixes:

```
Failures:

  1) MimeNegotiation#formats ignores unregistered mimetypes
     Failure/Error: expect(response.body).to eq '[:text]'

       expected: "[:text]"
            got: "[:text, \"mime/another\"]"

       (compared using ==)
     # ./spec/lib/mime_negotiation_patch_spec.rb:45:in `block (2 levels) in <top (required)>'

  2) MimeNegotiation#formats does not allow a modified accept header to render arbitrary files
     Failure/Error: expect(response.body).to include '# Welcome to Alaveteli!'

       expected "127.0.0.1 localhost\n\n# The following lines are desirable for IPv6 capable hosts\n::1 ip6-localhost...lnet\nff00::0 ip6-mcastprefix\nff02::1 ip6-allnodes\nff02::2 ip6-allrouters\nff02::3 ip6-allhosts\n" to include "# Welcome to Alaveteli!"
       Diff:
       @@ -1,2 +1,10 @@
       -# Welcome to Alaveteli!
       +127.0.0.1 localhost
       +
       +# The following lines are desirable for IPv6 capable hosts
       +::1 ip6-localhost ip6-loopback
       +fe00::0 ip6-localnet
       +ff00::0 ip6-mcastprefix
       +ff02::1 ip6-allnodes
       +ff02::2 ip6-allrouters
       +ff02::3 ip6-allhosts

     # ./spec/lib/mime_negotiation_patch_spec.rb:52:in `block (2 levels) in <top (required)>'
```

Fixed the spec suite failure caused by overriding the routes for the patch test and not [reloading them in the teardown](https://coderwall.com/p/ksu8qw/rails-add-a-route-for-a-test). Should now run without incident (have done so twice to reduce the likelihood that I've "luckily" got a rare spec order that happens to work):

Screen Shot 2019-03-26 at 09 31 11

## Implementation notes

We should be able to cherry-pick this across to all supported 3.2 branches and just add the relevant CHANGELOG and version bump bits for each (which is why I've left  those out for now).

[...]

PR #5142, opened 2019-03-27
hotfix/0.25.0.27-ms (Rails security fixes) 

lizconlan:
## What does this do?

Backports the CVE fixes in Rails 4.2.11.1 to Rails 3.2.x for the -ms variant of the 0.25 branch

## Why was this needed?

We still have 3.2 instances and 3.2 is out of support

## Implementation notes

Uses [0.25.0.27](https://github.com/mysociety/alaveteli/pull/5134) as the base then cherrypicks the `ms` changes (single commit) on top

## Test suite output

Screen Shot 2019-03-27 at 12 25 12

Have also run `rspec ./spec/lib/mime_negotiation_patch_spec.rb` separately with and without the patch in place to double check. Seems fine!

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,6 +1,7 @@
 source 'https://rubygems.org'
 
 gem 'rails', '3.2.22.3'
+gem 'test-unit', '~> 3.1.0' # Hack for Rails 3.2 on Ruby 2.3.5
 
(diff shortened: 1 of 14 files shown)


lizconlan:
Closing. Tagged up as a hotfix release

PR #5143, opened 2019-03-27
hotfix/0.22.4.12 (Rails security fixes)

lizconlan:
## What does this do?

Backports the CVE fixes in Rails 4.2.11.1 to Rails 3.2.x for the 0.22 branch

## Why was this needed?

We still have 3.2 instances and 3.2 is out of support

## Test suite output

Screen Shot 2019-03-27 at 14 47 18

Have also run `rspec ./spec/lib/mime_negotiation_patch_spec.rb` separately with and without the patch in place to double check. Seems fine!

PR #5144, opened 2019-03-27
hotfix/0.22.4.12-ms (Rails security fixes)

lizconlan:
## What does this do?

Backports the CVE fixes in Rails 4.2.11.1 to Rails 3.2.x for the ms variant of the 0.22 branch

## Why was this needed?

We still have 3.2 instances and 3.2 is out of support

## Implementation notes

Uses [hotfix/0.22.4.12](https://github.com/mysociety/alaveteli/pull/5143) as a base then cherry-picks [the ms-specific commit](https://github.com/mysociety/alaveteli/compare/hotfix/0.22.4.11...hotfix/0.22.4.11-ms) **and** 3455f329 adds an update that pins rspec-rails to a version that works with our code without preventing the test suite running with a `fixture_path` error (2.13.2) or bring in new breaking changes (2.14.1)

I'll squash 3455f329 post review so that if there are future versions, the next implementer can just cp a single commit to get the ms fixes with passing (if noisy) tests.

## Test suite output

Screen Shot 2019-03-27 at 16 39 00

Lots of deprecation warnings, but mostly about spec syntax e.g. calling `stub!` instead of `stub`, `mock` instead of `double`


lizconlan:
Closing. Tagged as a hotfix release

PR #5152, opened 2019-04-02
Actionmailer request exceptions

lizconlan:
## What does this do?

* Demonstrates our preferred error handling technique
* Adds some specs to NZ's PR
* Adds the handling code to display the failed messages on the admin summary so that we can use it in develop (the code to pull error messages in to the todo list only exists in develop so fair that we should add it)

### AdminRequestController#show

(My additions don't alter these but this seemed like a good place to share them for now)

#### Outgoing messages section

Screen Shot 2019-03-13 at 18 18 12

#### Events section

Screen Shot 2019-03-13 at 18 20 08

### (My bit) Pull the send failure info into the Admin Summary

Screen Shot 2019-04-02 at 13 44 26

(Otherwise it would default to displaying "User message" and "None given" which ... is technically true but not particularly helpful. If you can see why I've broken the alternate stripes on the table, let me know - it's probably a "so obvious I can't see it" mistake.) Had originally intended to add this separately but as it was fairly straightforward and makes sense to add the new factory at this point, I decided to include it.

## Notes to reviewer

Do my additions look ok to use as part of the review / contributing the specs?

The final commit (happy to drop it or to squash it into df152bdd) fixes the Code Climate warnings about method length and repetition but I'm not sure whether it's a good change or if it unintentionally adds cognitive complexity when reading an already long method (x2))

### DIFF
--- a/app/controllers/followups_controller.rb
+++ b/app/controllers/followups_controller.rb
@@ -143,13 +143,27 @@ class FollowupsController < ApplicationController
     # outgoing message, save just before sending.
     @outgoing_message.save!
 
+    errors = [ EOFError,
+               IOError,
+               Timeout::Error,
+               Errno::ECONNRESET,
+               Errno::ECONNABORTED,
+               Errno::EPIPE,
+               Errno::ETIMEDOUT,
+               Net::SMTPAuthenticationError,
+               Net::SMTPServerBusy,
+               Net::SMTPSyntaxError,
+               Net::SMTPUnknownError,
+               OpenSSL::SSL::SSLError ]
+    errors << AWS::SES::ResponseError if defined?(AWS::SES::ResponseError)
+
     begin
       mail_message = OutgoingMailer.followup(
         @outgoing_message.info_request,
         @outgoing_message,
         @outgoing_message.incoming_message_followup
       ).deliver_now
-    rescue StandardError => e
+    rescue *errors => e
       @outgoing_message.record_email_failure(e.message)
     else
       @outgoing_message.record_email_delivery(
--- a/app/controllers/followups_controller.rb
+++ b/app/controllers/followups_controller.rb
@@ -150,9 +150,7 @@ class FollowupsController < ApplicationController
         @outgoing_message.incoming_message_followup
       ).deliver_now
     rescue StandardError => e
-      @outgoing_message.record_email_failure(
-        e.message
-      )
+      @outgoing_message.record_email_failure(e.message)
     else
       @outgoing_message.record_email_delivery(
         mail_message.to_addrs.join(', '),
--- a/app/controllers/followups_controller.rb
+++ b/app/controllers/followups_controller.rb
@@ -143,27 +143,13 @@ class FollowupsController < ApplicationController
     # outgoing message, save just before sending.
     @outgoing_message.save!
 
-    errors = [ EOFError,
-               IOError,
-               Timeout::Error,
-               Errno::ECONNRESET,
-               Errno::ECONNABORTED,
-               Errno::EPIPE,
-               Errno::ETIMEDOUT,
-               Net::SMTPAuthenticationError,
-               Net::SMTPServerBusy,
-               Net::SMTPSyntaxError,
-               Net::SMTPUnknownError,
-               OpenSSL::SSL::SSLError ]
-    errors << AWS::SES::ResponseError if defined?(AWS::SES::ResponseError)
-
     begin
       mail_message = OutgoingMailer.followup(
         @outgoing_message.info_request,
         @outgoing_message,
         @outgoing_message.incoming_message_followup
       ).deliver_now
-    rescue *errors => e
+    rescue *OutgoingMessage.expected_send_errors => e
       @outgoing_message.record_email_failure(e.message)
     else
       @outgoing_message.record_email_delivery(
--- a/app/controllers/followups_controller.rb
+++ b/app/controllers/followups_controller.rb
@@ -139,19 +139,32 @@ class FollowupsController < ApplicationController
   def send_followup
     @outgoing_message.sendable?
 
-    mail_message = OutgoingMailer.followup(
-      @outgoing_message.info_request,
-      @outgoing_message,
-      @outgoing_message.incoming_message_followup
-    ).deliver_now
-
-    @outgoing_message.record_email_delivery(
-      mail_message.to_addrs.join(', '),
-      mail_message.message_id
-    )
-    @outgoing_message.info_request.reopen_to_new_responses
-
+    # OutgoingMailer.followup() depends on DB id of the
+    # outgoing message, save just before sending.
     @outgoing_message.save!
+
+    begin
+      mail_message = OutgoingMailer.followup(
+        @outgoing_message.info_request,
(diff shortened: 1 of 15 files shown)


spec/models/outgoing_message_spec.rb
```diff
@@ -1789,6 +1789,29 @@
 
   end
 
+  describe '#record_email_failure' do
+    let(:outgoing_message) { FactoryBot.create(:initial_request) }
+
+    it 'sets the status to "failed"' do
+      outgoing_message.record_email_failure('test')
```
garethrees: Would be worth making this a little clearer as to what ends up being passed as an argument IRL.

Maybe `'Exception#message'`?

spec/models/outgoing_message_spec.rb
```diff
@@ -1789,6 +1789,29 @@
 
   end
 
+  describe '#record_email_failure' do
```
garethrees: It also sets `OutgoingMessage#last_sent_at`.

spec/support/shared_examples_for_network_errors_during_send.rb
```diff
@@ -0,0 +1,37 @@
...
+    it 'logs a "send_error" event' do
+      event = request.reload.info_request_events.last
+      expect(event.event_type).to eq 'send_error'
+    end
+
+    it 'stores the reason for the failure' do
+      event = request.reload.info_request_events.last
+      expect(event.params[:reason]).to eq 'Connection timed out'
+    end
+
```
garethrees: It also forces the outgoing message to get saved, so should test that. 

```
it 'ensures the outgoing message is persisted' do
  # …
  expect(outgoing_message).to be_persisted
end
```

https://github.com/mysociety/alaveteli/pull/5152/commits/1130fe8a559afa88680c31cb9e5630796ddfab4d#diff-266c99b1323db6c6172f9d5751a56bd8R163

garethrees: Ah, the above is just for followups; in request controller we force the `InfoRequest` to get saved:

https://github.com/mysociety/alaveteli/pull/5152/commits/1130fe8a559afa88680c31cb9e5630796ddfab4d#diff-1fa1a855bf4fb5aaec94a56340bcc062R428

Could put these in the `'a network error occurs while sending a followup'` context of the controller specs?

app/views/admin_general/_to_do_list.html.erb
```diff
@@ -26,7 +26,16 @@
                   <% end %>
                 <% elsif item.is_a? InfoRequest %>
                   <%= request_both_links(item) %>
-                  <% if InfoRequest.requires_admin_states.include?(item.described_state) %>
+                  <% if item.last_event.event_type == 'send_error' %>
```
garethrees: Put `send_error` in `InfoRequest.requires_admin_states` and then tweak the conditional on L50

app/controllers/followups_controller.rb
```diff
@@ -139,20 +139,33 @@ def outgoing_message_params
...
+      @outgoing_message.info_request.reopen_to_new_responses
+    ensure
+      # Ensure DB is updated to isolate potential templating issues
+      # from impacting delivery status information.
+      @outgoing_message.save!
+    end
+
     if @outgoing_message.what_doing == 'internal_review'
       flash[:notice] = _("Your internal review request has been sent on its way.")
     else
```
lizconlan: for consistency, this should move into the else branch, below `@outgoing_message.record_email_delivery`

spec/support/shared_examples_for_network_errors_during_send.rb
lizconlan: oh, actually 2 different things here:
- both should ensure that our new `outgoing_message` gets persisted so that should go into the shared spec file
- request_controller needs an additional check that the parent `info_request` was updated (there's an earlier `save!` so `persisted?` is the wrong thing here) with the new `outgoing_message` data

lizconlan: ugh, that ensure save isn't really needed as (I think) the `info_request` (and relations) get saved during the log event; updating the test to check `is_persisted` and "does the related outgoing message have the expected status?"

app/views/admin_general/_to_do_list.html.erb
lizconlan: ah, the `described_state` here is `error_message` which is already there - just need to improve the conditional

garethrees:
Just deleted hound comments to clear this up a bit.

spec/models/outgoing_message_spec.rb
```diff
@@ -1795,6 +1795,32 @@
...
+    let(:outgoing_message) { FactoryBot.build(:initial_request) }
+    subject { outgoing_message.prepare_message_for_resend }
+
+    it 'raises an error if it receives an unexpected message_type' do
+      allow(outgoing_message).to receive(:message_type).and_return('fake')
+      expect{ subject }.to raise_error(RuntimeError)
+    end
+
+    it 'raises an error if it receives an unexpected status' do
+      outgoing_message.status = 'followup'
```
garethrees: 'followup' is a message type so we think using one of [`OutgoingMessage::STATUS_TYPES`](https://github.com/mysociety/alaveteli/blob/develop/app/models/outgoing_message.rb#L33) would be a better example (the only example would be "ready" from that list)

app/models/outgoing_message.rb
```diff
@@ -75,20 +75,22 @@ class OutgoingMessage < ActiveRecord::Base
...
+      Timeout::Error,
+      Errno::ECONNRESET,
+      Errno::ECONNABORTED,
+      Errno::EPIPE,
+      Errno::ETIMEDOUT,
+      Net::SMTPAuthenticationError,
+      Net::SMTPServerBusy,
+      Net::SMTPSyntaxError,
+      Net::SMTPUnknownError,
+      OpenSSL::SSL::SSLError ].concat(additional_send_errors)
```
garethrees: All good, but this could do with a spec

app/views/admin_general/_to_do_list.html.erb
```diff
@@ -38,7 +38,8 @@
                           <% end %>
                         </td>
                       </tr>
-                      <% if item.described_state == 'attention_requested' %>
+                      <% if item.described_state == 'attention_requested' ||
+                            item.last_event.event_type == 'send_error' %>
```
garethrees: We think it would reduce complexity if we made this more like the `params[:message]` conditional:

```erb
<% if item.last_event.params[:reason] %>
  <tr>
    <td><b>Reason</b></td>
    <td>
      <%= item.last_event.params[:reason] %>
    </td>
  </tr>
<% end %>
```

spec/models/info_request_spec.rb
```diff
@@ -4050,11 +4050,15 @@ def email_and_raw_email(opts = {})
           .to eq sending_event
       end
 
-      it 'raises an error if the request has never been sent' do
```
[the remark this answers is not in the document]
lizconlan: Agreed, reinstating

lizconlan:
Closed in favour of https://github.com/nigeljonez/alaveteli/pull/1

PR #5156, opened 2019-04-05
Small copy updates for documentation site

abimysoc:
Part of wider quest to remove offers of support we aren't currently able to provide, and direct people towards community mailing lists rather than emailing mySociety directly (and the international email address in particular).


lizconlan:
Now live on the docs site

PR #5168, opened 2019-04-23
Fix cache keys for authority browse mode

gbp:
## Relevant issue(s)

https://github.com/mysociety/sysadmin/issues/1135

## What does this do?

There was issues with both the outer and inner fragment caching calls.

The outer cache wasn't working due to there being no #cache_key method for a ActiveRecord::Relation. This method was added in Rails 5 [1] so for earlier Rails versions it ended up using an internal Ruby object ID in the cache key which changes every request and means the cached value is never used.

While the inner cache wasn't completely broken, the cache_key used included the public body translation cache_key [2] meaning we weren't stopping the extra SQL queries being performed for each public body in the category and causing a performance hit.

[1] https://github.com/rails/rails/blob/5-0-stable/activerecord/lib/active_record/relation.rb#L335
[2] https://github.com/globalize/globalize/blob/dfa9fc2/lib/globalize/active_record/instance_methods.rb#L162

### DIFF
--- a/app/views/alaveteli_pro/batch_request_authority_searches/_public_bodies.html.erb
+++ b/app/views/alaveteli_pro/batch_request_authority_searches/_public_bodies.html.erb
@@ -1,8 +1,10 @@
 <ul>
   <% public_bodies = PublicBody.with_tag(category_tag) %>
-  <% cache [:batch_builder, locale, category_tag, public_bodies] do %>
+  <% key = rails5? ? public_bodies : public_bodies.maximum(:updated_at).to_i %>
+  <% cache [:batch_builder, locale, category_tag, key] do %>
     <% public_bodies.select(&:is_requestable?).each do |public_body| %>
-      <% cache [:batch_builder, locale, category_tag, public_body] do %>
+      <% body_key = "public_bodies/#{public_body.id}-#{public_body.updated_at.to_i}" %>
+      <% cache [:batch_builder, locale, category_tag, body_key] do %>
         <%= render partial: 'alaveteli_pro/batch_request_authority_searches/search_result',
                    locals: { public_body: public_body,
                              draft: draft_batch_request,


lizconlan:
> The outer cache wasn't working due to there being no #cache_key method for a ActiveRecord::Relation. This method was added in Rails 5 [1] so for earlier Rails versions it ended up using an internal Ruby object ID in the cache key which changes every request and means the cached value is never used.

It also - from my machine, admittedly (so unrelated errors may have crept in) - appears to have been generating a really long cache key which probably wouldn't have helped (mine's 9545 chars!)

app/views/alaveteli_pro/batch_request_authority_searches/_public_bodies.html.erb
```diff
@@ -1,8 +1,10 @@
 <ul>
   <% public_bodies = PublicBody.with_tag(category_tag) %>
-  <% cache [:batch_builder, locale, category_tag, public_bodies] do %>
+  <% key = rails5? ? public_bodies : public_bodies.maximum(:updated_at) %>
```
lizconlan: minor nitpick (1/2) - if we use `public_bodies.maximum(:updated_at)` we get a space in the cache key which appears to be valid but is a little odd e.g. `school/public_bodies/212-2019-04-23 16:02:29 +0100/5d6745201cce2d42dca055b44c33923d`

if we use `public_bodies.maximum(:updated_at).to_i` instead, we end up with something more like `school/public_bodies/212-1556031749/da061a7420597d3b51959a58ba240c97`

app/views/alaveteli_pro/batch_request_authority_searches/_public_bodies.html.erb
```diff
@@ -1,8 +1,10 @@
...
+  <% cache [:batch_builder, locale, category_tag, key] do %>
     <% public_bodies.select(&:is_requestable?).each do |public_body| %>
-      <% cache [:batch_builder, locale, category_tag, public_body] do %>
+      <% body_key = "public_bodies/#{public_body.id}-#{public_body.updated_at}" %>
```
lizconlan: minor nitpick (2/2) -  if we use public_bodies(:updated_at) here, we get "false" in the cache key e.g. `school/29/2/15/23/4/2019/2/113/false/UTC/5d6745201cce2d42dca055b44c33923d`

if we use `public_bodies.updated_at.to_i`, we get `school/1556031749/da061a7420597d3b51959a58ba240c97`

app/views/alaveteli_pro/batch_request_authority_searches/_public_bodies.html.erb
gbp: Ah yes, though it would've called #to_i in the string interpolation. Changed in 4c3c23c

app/views/alaveteli_pro/batch_request_authority_searches/_public_bodies.html.erb
gbp: Also changed in 4c3c23c

PR #5169, opened 2019-04-24
Batch category picker UI pagination

gbp:
## Relevant issue(s)

mysociety/sysadmin#1135
https://github.com/mysociety/alaveteli/pull/5168

## What does this do?

Add pagination to the batch category picker UI so only 100 authorities are ever load at one time.

### DIFF
--- a/app/assets/javascripts/alaveteli_pro/batch_authority_search/browse.js
+++ b/app/assets/javascripts/alaveteli_pro/batch_authority_search/browse.js
@@ -11,6 +11,14 @@
   var closedClass = 'batch-builder__list__group--closed';
   var loadingClass = 'batch-builder__list__group--loading';
 
+  var openCaret = function openCaret(group) {
+    group.removeClass(closedClass);
+  };
+
+  var closeCaret = function closeCaret(group) {
+    group.addClass(closedClass);
+  };
+
   var toggleCaret = function toogleCaret(group) {
     group.toggleClass(closedClass);
   };
@@ -26,7 +34,7 @@
       dataType: 'html',
       success: function (data) {
         group.append(data);
-        toggleCaret(group);
+        openCaret(group);
         toggleSpinner(group);
         $draft.trigger(DraftEvents.bodyAdded);
         $search.trigger(SearchEvents.domUpdated);
@@ -53,7 +61,7 @@
 
   var collapseTopLevelGroups = function collapseTopLevelGroups() {
     var groups = $('.batch-builder__list > .batch-builder__list__group');
-    toggleCaret(groups);
+    closeCaret(groups);
   };
 
   $(function(){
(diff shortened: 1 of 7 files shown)


gbp:
> However, with the outer cache removed, rendering a single page of results is actually slower on my machine with caching enabled than without.

The problem with the outer cache is that it was caching button states and the `draft.id` value. We could include the draft in the cache key if we think its worth retaining but then the cache wouldn't have affects for other users.

> With cache (cache populated)
> `Completed 200 OK in 96716ms (Views: 96675.8ms | ActiveRecord: 11.3ms)`
> 
> Without cache
> `Completed 200 OK in 62563ms (Views: 62523.7ms | ActiveRecord: 13.5ms)`

These numbers seem way too high to me. In production mode when loading the first page of the 'schools' category it renders `Completed 200 OK in 3395ms (Views: 3237.7ms | ActiveRecord: 136.4ms)` on first load & then `Completed 200 OK in 1308ms (Views: 1040.6ms | ActiveRecord: 184.5ms)` afterwards. What is your setup? I'm guessing development mode and the logging to stdout might be slowing you down.

These speeds are another reason I thought removing the outer cache would be fine to do.

> Also I found a couple of gotchas with the way the navigation currently works
> 
>     1. pressing the Next button twice causes additional server load and breaks the UI
Ah right that'll need fixing.

>     2. I'm not convinced that loading all the authorities into memory one page at a time is going to be a good approach at scale
Not sure where this is loading all authorities, it should be loading max 100 at a time. Am I missing something?

> I've had a quick look at making a more traditional Prev / Next version

I thought about doing this but then realised that this could use more server resources as users go back and forth between pages multiple times - especially without the outer cache.

lizconlan:
>  I'm guessing development mode and the logging to stdout might be slowing you down.

Yeah, had just been turning `config.action_controller.perform_caching` on and off to quickly test stuff - even allowing for the slower speed of a laptop vs a server, not real world numbers. But the difference seemed interesting (but may be accounted for by logging the extra ops for disk access).

> The problem with the outer cache is that it was caching button states and the draft.id value. We could include the draft in the cache key if we think its worth retaining but then the cache wouldn't have affects for other users.

I didn't see button state getting cached but may have missed something. Had overlooked the `draft.id` tho. Tested by abandoning my (saved) draft and creating a new one but with the same user, which I thought worked but was far from an exhaustive test so I may have got that wrong.

>> 2. I'm not convinced that loading all the authorities into memory one page at a time is going to be a good approach at scale
>>
> Not sure where this is loading all authorities, it should be loading max 100 at a time. Am I missing something?

It's loading 100 at a time but with an ever increasing amount of stuff in the browser, that's going to become difficult for the user to scroll and manage (I've not hit a specific problem with it here but I'm familiar with a general pattern where the browser runs out of memory with this kind of solution, although I'll freely admit that there are generally images - e.g. product photography - involved so we might never actually cause it).

>> I've had a quick look at making a more traditional Prev / Next version
>
> I thought about doing this but then realised that this could use more server resources as users go back and forth between pages multiple times - especially without the outer cache.

Yeah, good point - if we can't use the outer cache, that might be an awful user experience as well as a sizeable server hit.

gbp:
Closed in favour of #5172 😭

PR #5170, opened 2019-04-24
Sysadmin1135 browse pagination lc

lizconlan:
Hacky test of some tweaks to the "browse categories" pagination / caching code. Just here as an asynch way of sharing my screen, no intention that it gets merged!


gbp:
This test was superseded by discussion on #5169

PR #5181, opened 2019-05-13
Enable Layout/EmptyLines* cops

garethrees:
## What does this do?

Enable `Layout/EmptyLines*` cops

## Why was this needed?

Trying to continuously improve codebase health.


garethrees:
Hmm, yeah we should probably add those in but lets leave that to a new PR.

PR #5182, opened 2019-05-15
5175 handle mail replies spec ci fail 2

garethrees:


garethrees:
Waiting until we've made rails 5 the default

PR #5190, opened 2019-05-20
[Rails 5] Lock down Rails 5 gem versions

### LINKED ISSUE #5189: [Rails 5] Lock down Rails 5 gem versions

lizconlan:
## Relevant issue(s)

Closes #5189 

## What does this do?

Adds major/minor version contraints for the Rails 5 gems.

## Why was this needed?

We'd temporarily removed the constraints during the initial upgrade work so we need to add them back in now that we've discovered what works.


lizconlan:
Was originally a conservative gem update. Working on a second pass to see if we can do better (spoiler: looks like we can do better!)

garethrees:
conservative sounds good at this point to reduce variability of changes – this is just to get Rails 5 shipped; not to update all the gems

PR #5197, opened 2019-05-22
Gitignore tweaks

garethrees:
## What does this do?

Cleans up `.gitignore` a bit.

## Why was this needed?

I recently re-built my VM and copied over some config files across. They were showing up as unstaged changes, and I was wondering why they weren't being ignored correctly. I assumed there was a problem with `.gitignore` so had a look. Made these commits then realised I'd moved the files to the root dir, not `config/`. In any case, this cleans up the file a bit.


garethrees:
> There are entries which don't seem to be needed anymore

Removed a bunch of these

> Hard to detrimne if some lines are ignoring folders or files without file extensions - add trailing slash

+1; fixed.

> In my opinion code editor specific entries should belong in a global git ignore file for the developers own environment.

+1; fixed.

> Have you consider merging ./config/.gitignore into this file?

omg didn't even know these existed. merged and fixed.

---

There are some I'm not sure about, so I've left them in as I don't think its worth any extra time at this point really.

PR #5208, opened 2019-05-24
[Rails 5] Migrations updates

### LINKED ISSUE #5205: [Rails 5] Update existing migrations where `#add_timestamps` is used
Part of #3969

```
DEPRECATION WARNING: `#add_timestamps` was called without specifying an option 
for `null`. In Rails 5, this behavior will change to `null: false`. You should 
manually specify `null: true` to prevent the behavior of your existing 
migrations from changing.
```

### LINKED ISSUE #5207: [Rails 5] Update existing migrations for ActiveRecord changes
Part of #3969 

```
DEPRECATION WARNING: #table_exists? currently checks both tables and views.
This behavior is deprecated and will be changed with Rails 5.1 to only check tables.
Use #data_source_exists? instead.

DEPRECATION WARNING: uniq is deprecated and will be removed from Rails 5.1
(use distinct instead)
```

gbp:
## Relevant issue(s)

Fixes #5205 
Fixes #5207 

## What does this do?

Updates old migration for compatibility with Rails 5.0 and 5.1

## Notes to reviewer

There is still one deprecation warning for `#table_exists?` coming from the `acts_as_versioned` gem - we might need patch this or look at alternative versioning gems.

### DIFF
--- a/db/migrate/20170718261524_add_expiring_notification_at.rb
+++ b/db/migrate/20170718261524_add_expiring_notification_at.rb
@@ -15,12 +15,12 @@ class AddExpiringNotificationAt < !rails5? ? ActiveRecord::Migration : ActiveRec
   private
 
   def column_exists?(table, column)
-    if table_exists?(table)
+    if data_source_exists?(table)
       connection.column_exists?(table, column)
     end
   end
 
-  def table_exists?(table)
-    connection.table_exists?(table)
+  def data_source_exists?(table)
+    connection.data_source_exists?(table)
   end
 end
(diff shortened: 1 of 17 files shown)


db/migrate/20170718261524_add_expiring_notification_at.rb
```diff
@@ -15,12 +15,12 @@ def down
   private
 
   def column_exists?(table, column)
-    if table_exists?(table)
```
garethrees: Typo in the commit message here:

```diff
- Rename table_exist? to data_source_exists? in migrations
+ Rename table_exists? to data_source_exists? in migrations
```

Issue #5212, opened 2019-05-28
[Rails 5] Staging testing 


gbp:
Initial list taken from: https://github.com/mysociety/alaveteli/issues/2968#issuecomment-264157308

List spilt into separate comments due to collisions when multiple people were editing at the same time.

gbp:
Noting for `Change a profile photo` photo doesn't change without clearing the varnish cache

garethrees:
> Noting for `Change a profile photo` photo doesn't change without clearing the varnish cache

https://github.com/mysociety/alaveteli/issues/156.

lizconlan:
Could be done locally but we should also test Pro without Pricing

gbp:
### PRO BATCH CREATION (Pro user)
- [ ] pro_batch_access feature enables batch access
- [ ] Can search for authority
- [ ] Add authority to batch draft
- [ ] pro_batch_category_ui feature enables batch browse access
- [ ] Can browse authority categories
- [ ] Can add all authority in a category to batch draft

gbp:
### PRO BATCH SENDING
...

gbp:
### PRO EMBARGOS (Pro user)
- [ ] Can add embargo to pro requests
- [ ] Extend embargo
- [ ] Publish pro requests
- [ ] Requests should not be visible to public users

gbp:
### PRO DASHBOARD
...

gbp:
### PRO SIGNUPS/PAYMENTS (User)
- [ ] Can signup using a coupon
- [x] Enter payment details
- [x] Subscription created on Stripe

gbp:
### PRO WEBHOOKS
- [ ] Response 200 to all webhook
- [ ] Customer updated - ignore renewals
- [x] Customer updated - email on cancelled
- [x] Customer deleted - remove pro access
- [x] Payment succeeded - update charge description
- [ ] Payment failed - email

gbp:
### ANNOUNCEMENTS (Admin)
- [x] Create site-wide announcement
- [x] Create pro announcement
- [x] Can dismiss announcements

gbp:
### CENSOR RULES (Admin)
- [ ] View the list of censor rules (/admin/censor_rules)
- [ ] Create a global censor rule
- [ ] Create a user censor rule
- [ ] Create an authority censor rule
- [ ] Create a request censor rule
- [ ] Create a regexp censor rule
- [ ] Edit a censor rule
- [ ] Destroy a censor rule

[...]

Issue #5194, opened 2019-05-20
Release 0.34


lizconlan:
Requires https://github.com/mysociety/alaveteli/issues/4902.

lizconlan:
* Reduce OS support #4782 #4970
* Reduce Ruby versions support #4541 #4542 #4842

garethrees:
* Make Rails 5 the default #3969

garethrees:
Remove Ruby 2.6 / Rails 4.2 monkeypatch: https://github.com/mysociety/alaveteli/pull/5055/commits/0fc7e46d0d744db66074073076724ec003c9eb50

lizconlan:
* For the release / upgrade notes - "Use rails [instead of rake] for running tasks and tests"
https://edgeguides.rubyonrails.org/upgrading_ruby_on_rails.html#use-rails-for-running-tasks-and-tests

gbp:
>     * Make Rails 5 the default #3969

If this release is also droping Rails 4.2 then we should remove the `rails5?` conditionals - such as: https://github.com/mysociety/alaveteli/pull/5202#L191

garethrees:
Planned Ruby / OS support list https://gist.github.com/garethrees/ec3c403e1ea0744fb736381785e1e788

garethrees:
Should do https://github.com/mysociety/alaveteli/issues/5199#issuecomment-495588852

lizconlan:
Done! https://github.com/mysociety/alaveteli/releases/tag/0.34.0.0

PR #5213, opened 2019-05-28
[Rails 5] Update Gemfile to use Rails 5 by default

### LINKED ISSUE #3969: Upgrade to Rails 5.0

gbp:
## Relevant issue(s)

Requires #5212 
Required by #5194 
Fixes #3969

## What does this do

~This is a bit early but should cover the individual commits that we needed to make on the rollup branch to get staging deployed.~

Repurposed as a draft of release/0.34.0.0 - rebased on top of #5249
Will be the basis of the next round of Staging testing before we cut the rc branch and start live testing

* Enables Rails 5 as the default
* Removes a bunch of old Rails 4 specific code
* Renames the old `.rails5` files to `.rails_next`
* ~Sets up basic testing for Rails 5.1 (some solvable issues purposely left out as I'd rather set them aside until after the release)~ (removed after review - a step too far)

## Notes to reviewer

Feel free to suggest improvements to my CHANGELOG notes! And please ask for a correction if I've inadvertently taken credit for something I've reworked or didn't write - I feel like I've lost track a little.


lizconlan:
Rebased onto latest develop and removed a few more `rails5?` bits that were preventing a clean build (feel free to change or revert those - I left them as fixups for easy review or removal)

.travis.yml
```diff
@@ -14,13 +14,13 @@ rvm:
   - ruby-head
 gemfile:
   - Gemfile
-  - Gemfile.rails5
+  - Gemfile.rails
```
lizconlan: not sure `.rails` is as meaningful as `rails5` was ... maybe `.next` or `.upgrade` (something like `.rails_next` would do the trick also but ... I'm a lazy typist!)

lizconlan:
Ok. I've made this a bit weird now - I've reviewed the first bit (just a tiny question about the naming of the upgrade Gemfile) and now need to ask for review on the rest...

doc/CHANGES.md
```diff
@@ -1,3 +1,13 @@
+# develop
+
+## Highlighted Features
+
+* Removed support for Ubuntu 14.04 LTS (Trusty Tahr) (Liz Conlan)
+* Dropped support for Ruby 2.0 (Liz Conlan)
+* Dropped support for Ruby 2.1 and 2.2 (Liz Conlan)
+* Upgrade to Rails 5.0 (Graeme Porteous)
```
gbp: We both should be credited here:

```suggestion
* Upgrade to Rails 5.0 (Liz Conlan, Graeme Porteous)
```

gbp:
> not sure .rails is as meaningful at a glance as .rails5 was ... maybe .next or .upgrade (something like .rails_next would do the trick also but ... I'm a lazy typist!)

My thinking was we wouldn't have to keep renaming this. Happy to append next of upgrade if you prefer.

lizconlan:
> I had been thinking it might be good to split 96f55db into its own PR, I haven't tested this but looks like you have and refactored some of the code outputted by my macro to remove the TestAfterCommit blocks.

I think this is fine, I'll leave it in

> Would it be better if enabling Rails 5.1 testing and any fixes be post release 0.34?

Yeah probably - I'll take the testing bits out

> I have tested the globalize bump yet, other than that I would approve this PR if I could.

Ta, I'll revert the globalize bump as it's only 'fixing' things in the 5.1 tests

lizconlan:
> > not sure .rails is as meaningful at a glance as .rails5 was ... maybe .next or .upgrade (something like .rails_next would do the trick also but ... I'm a lazy typist!)
>
> My thinking was we wouldn't have to keep renaming this. Happy to append next of upgrade if you prefer.

I've appended '_next' just because I think it will be more obvious what it's for to anyone who's not lived through the Rails 4.2 -> Rails 5.0 upgrade process!

lizconlan:
I'm reasonably confident that Globalize 5.1.0 doesn't break anything - the tests pass and the site seems to behave normally with translations.

The commit comparison looks reasonable to me:
* I don't think the switch from `after_` callbacks to `before_` callbacks is likely to cause harm
* The automatic inclusion of the translations table when sorting is nice (`PublicBody.order(:name)` now just works)
* Interesting that they're no longer creating indexes by default for translation tables, maybe we can drop them now (in the next version, not here) https://github.com/globalize/globalize/commit/1055b556f7ec0a247adf4f361b4a192c0b3d44d0

https://github.com/globalize/globalize/compare/v5.0.1...v5.1.0

PR #5218, opened 2019-05-30
[Rails 5] Fix URL generation

### LINKED ISSUE #5214: [Rails 5] Exception on admin/timeline
Part of #3969 
Connected to #5212 

```
An ActionView::Template::Error occurred in admin_general#timeline:

  Attempting to generate a URL from non-sanitized request parameters! An attacker can inject malicious data into the generated URL, such as changing the host. Whitelist and sanitize passed parameters to be secure.
  app/views/admin_general/timeline.html.erb:12

Request:

  * URL        : https://staging.alaveteli.org/admin/timeline
  * HTTP Method: GET
```

gbp:
## Relevant issue(s)

Part of #3969 
Fixes #5214 

## What does this do?

An ArgumentError for non-sanitized parameters will be raised if
data is merged into request parameters and then used to generate
URLs with the `url_for` or `link_to` helpers.

We actually don't need to do this as these helpers already use the
current request parameters as a starting point for the URL generated.

### DIFF
--- a/app/views/layouts/default.html.erb
+++ b/app/views/layouts/default.html.erb
@@ -28,7 +28,7 @@
       <% end %>
     <% end %>
     <% if @has_json %>
-      <link rel="alternate" type="application/json" title="JSON version of this page" href="<%=h url_for(request.params.merge(:format => 'json')) %>">
+      <link rel="alternate" type="application/json" title="JSON version of this page" href="<%= current_path_as_json %>">
     <% end %>
     <% if @no_crawl %>
       <meta name="robots" content="noindex, nofollow">
--- a/app/helpers/link_to_helper.rb
+++ b/app/helpers/link_to_helper.rb
@@ -287,6 +287,12 @@ module LinkToHelper
     url_for(sanitized_params.merge(locale: locale, only_path: true))
   end
 
+  def current_path_as_json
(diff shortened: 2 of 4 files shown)


app/views/layouts/default.html.erb
```diff
@@ -28,7 +28,7 @@
       <% end %>
     <% end %>
     <% if @has_json %>
-      <link rel="alternate" type="application/json" title="JSON version of this page" href="<%=h url_for(request.params.merge(:format => 'json')) %>">
+      <link rel="alternate" type="application/json" title="JSON version of this page" href="<%= current_path_as_json %>">
```
garethrees: I had a quick go, but I couldn't replicate https://github.com/mysociety/alaveteli/issues/5214 for this.

How did you end up with the same error?

I think the commit message here is dependent on the context in 2cbd17a53393613f3ee4d4b0f8c00533c3c8bb4b, which might not be obvious in the future, so it would be worth just tweaking this so that it explains what is meant by "It was _also_ possible for non-sanitized parameters to be used here"

gbp: Ah yes, this could do with the commit message being expanded. Basically using `request.params` by-passes the ActionController::Parameters sanitization checking. Without the `request.` we would see the same error as the admin/timeline page

Issue #5025, opened 2018-12-19
Missing dependency in Ubuntu 18.04 'pdftk'


schlos:
"E: Package 'pdftk' has no installation candidate"

In Ubuntu Bionic we're missing package which is part of install script (#4971).

The pdftk package in Ubuntu (and its upstream Debian package) was dropped due to its dependency on the now deprecated GCJ runtime.

https://askubuntu.com/questions/1028522/how-can-i-install-pdftk-in-ubuntu-18-04-and-later

garethrees:
Required by https://github.com/mysociety/alaveteli/issues/4971

garethrees:
@schlos how did you resolve this for http://imamopravoznati.org? Did you end up installing from the Cosmic repo as per https://askubuntu.com/a/1028983?

Utopiah:
Quite important as it breaks installation on Windows Linux subsystem

lizconlan:
Hi @Utopiah - you've commented on a closed ticket as (we thought) we'd fixed this for the latest release! Are you saying that this is broken in [Release  0.33](https://github.com/mysociety/alaveteli/releases/tag/0.33.0.0) or were you just agreeing that this was an important issue?

If you're installing manually, you'll need to add Cosmic's package list as a low priority APT source to get 'pdftk' per our install script https://github.com/mysociety/alaveteli/pull/5232/files#diff-d5eb0653888644638720a6b98d9d9f1d

Utopiah:
Just agreeing that this was an important issue

Issue #4971, opened 2018-11-27
Add Ubuntu 18.04 LTS (Bionic) support to install scripts


garethrees:

schlos:
Hi @garethrees I saw in #3211 that you consider skipping 16.04 support and going straight to Bionic Beaver. What are your plans to support 18.04 in your install scripts? Do you have any timeline in mind?

Did anybody try to install Alaveteli on 18.04 and are there any modifications to Alaveteli necessary to run on 18.04?

Thanks!

garethrees:
No timelines at the moment sorry. If you give it a go please do note issues here or in [alaveteli-dev](https://groups.google.com/forum/#!forum/alaveteli-dev).

schlos:
Other than #5025, there was no other issues with install on Bionic. 
I have CSS/JS issue in Assets folder (CSS/JS is not there or has wrong filename) but I think it's just theme regeneration issue which I didn't have yet time to investigate.
I'll report if there are any other issues related to Bionic.

PR #5232, opened 2019-05-31
[#4971] Add Ubuntu 18.04 LTS (Bionic) Support

garethrees:
## Relevant issue(s)

Fixes https://github.com/mysociety/alaveteli/issues/4971
Fixes https://github.com/mysociety/alaveteli/issues/5025

## What does this do?

Add Ubuntu 18.04 LTS (Bionic) Support

## Implementation notes

Obvs ignoring the hound comments as its a patch that we can hopefully remove as we upgrade further.

- [x] Check whether 1f2a600 is actually needed
- [x] Test a few actions in the app
- Lots of warnings about `constant ::Fixnum is deprecated`, but we're seeing those in CI anyway


lizconlan:
> Lots of warnings about `constant ::Fixnum` is deprecated

I _think_ this will be resolved in 0.34 (drops ruby 2.0 and allows us to upgrade nokogiri to 1.8.5 which [appears to include a fix](https://github.com/sparklemotion/nokogiri/blob/e28fa4bb2ed6844c3c63f58062d034e7b99fc90c/lib/nokogiri/html/document.rb))

lizconlan:
Yep upgrading nokogiri fixes most of the deprecation warnings, but there are still some for `BigDecimal.new()` vs `BigDecimal()` which should be a quick fix. I'll make a ticket for that

Screen Shot 2019-06-05 at 09 49 01

lizconlan:
Afterthought - I did not capture the error message but I had to upgrade Vagrant to get this to build (I had an older pre v2 version)

garethrees:
> Afterthought - I did not capture the error message but I had to upgrade Vagrant to get this to build (I had an older pre v2 version)

Pre-v2 is pretty old. Although its not really written down anywhere, I've just considered that vagrant (and virtualbox) should be running on the most recent version unless there's a _really_ good version not to do so.

lizconlan:
>  vagrant (and virtualbox) should be running on the most recent version unless there's a really good version not to do so.

Yeah, more capturing this in case a reuser hits a similar problem long after I've forgotten about this

PR #5233, opened 2019-06-03
Enable some more cops

garethrees:
Nothing massively controversial here IMO.


gbp:
@garethrees moved to current sprint so you can review the commits I pushed if you wish. Again there shouldn't be anything controversial there

garethrees:
Merging as is, as we've made a net improvement, even though some of the improvements have caused warnings for other cops.

PR #5240, opened 2019-06-05
Annotate models

gbp:
## Relevant issue(s)

Connects to #4902

## Why was this needed?

Release is coming lets annotate the models so they are up-to-date

## Implementation notes

- To ensure defaults values don't change I set `TIME_ZONE=London` in config/general.yml. This matches what is set in production for WDTK
- Cleared the existing annotations with: `bundle exec annotate -d`
- Finally ran: `bundle exec annotate`

### DIFF
--- a/app/models/incoming_message_error.rb
+++ b/app/models/incoming_message_error.rb
@@ -10,8 +10,9 @@
 #  retry_at   :datetime
 #  backtrace  :text
 #
-# models/incoming_message_error.rb:
 
+# models/incoming_message_error.rb:
+#
 # Store details of errors that have been generated when trying to import
 # emails from a POP mailbox into the application. Used by AlaveteliMailPoller
 # to record errors and to determine whether to retry importing a given mail.
--- a/spec/factories/alaveteli_pro/draft_info_request_batches.rb
+++ b/spec/factories/alaveteli_pro/draft_info_request_batches.rb
@@ -1,4 +1,18 @@
 # -*- encoding : utf-8 -*-
+
+# == Schema Information
+#
+# Table name: draft_info_request_batches
+#
+#  id               :integer          not null, primary key
+#  title            :string
+#  body             :text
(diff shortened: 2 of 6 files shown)


garethrees:
Hmm, I'm convinced I'd run this myself.

I've always used `bundle exec annotate`. I noticed you used `bundle5` – I ran this as "Rails 4" as that's still the production version for release 0.33. 

I also never use `RAILS_ENV=production` – should we be doing that?

lizconlan:
> I've always used bundle exec annotate. I noticed you used bundle5 – I ran this as "Rails 4" as that's still the production version for release 0.33.

Good point - I got confused for a moment there. We don't want the Rails 5 changes in release 0.33

> I also never use RAILS_ENV=production – should we be doing that?

Shouldn't make a difference but isn't needed as the models a) should be the same either way and b) the model docs are mostly useful for develop so running as production isn't ideal

gbp:
I have force pushed a new commit based on comments. While redoing this noticed there are some difference with our production env. see: #5246

app/models/incoming_message_error.rb
```diff
@@ -10,7 +10,6 @@
 #  retry_at   :datetime
 #  backtrace  :text
 #
-# models/incoming_message_error.rb:
```
lizconlan: This looks like it's been removed in error - it wasn't originally added by annotate: https://github.com/mysociety/alaveteli/commit/eb944c6c7aebdd066ba897745155fb233335f789

gbp: Ugh, why would annotate remove this. Think I'll add a blank line to prevent this from happening again.

gbp:
> Now wondering why running annotate on my machine doesn't pick up the changes to factory code *sigh*

Hmm, maybe this also effects @garethrees as he was convinced he had already run this?

PR #5268, opened 2019-06-21
[Rails 5.1] Enable CI

lizconlan:
Just leaving this here so we don't lose the speculative WIP from the Rails 5.0 work

(Many of the suggestions and fixes here are gbp's - credit required when we get to the CHANGELOG stage. Mistakes 100% mine.)


lizconlan:
superseded by #5351 (and components)

PR #5272, opened 2019-06-27
[pro#452] Webhook digest

gbp:
## Relevant issue(s)

Connected to https://github.com/mysociety/transparency-adessium/issues/17
Fixes https://github.com/mysociety/alaveteli-professional/issues/452

## What does this do?

* Removes the unhandled webhook exception emails
* Stores Webhook in the database
* Add Webhook digest email
* Run digest email once a day

## Why was this needed?

To make in easier to other sites to start using pro pricing.

## Implementation notes

The format of the email is open for discussion, I've deliberately not included the customers real name/email (instead it shows the internal Stripe customer ID) so we aren't receiving customers personal details via email. This makes the email less useful so we might want to consider changing this, or we could link to the admin user#show page. Thoughts?

## Notes to reviewer

I've include email preview, this can be accessed from http://0.0.0.0:3000/rails/mailers


lizconlan:
Note to self:  http:/10.0.03.0:3000/rails/mailers because Vagrant

app/models/webhook.rb
```diff
@@ -7,6 +7,8 @@
 class Webhook < ApplicationRecord
   validates :params, presence: true
 
+  scope :pending_notification, -> { where(notified_at: nil) }
```
lizconlan: I _think_ this is fine, but my first instinct on seeing it in action was to misread it as a relation or list of attributes - `Webhook.pending_notifications`, probably just me being tired!

lizconlan:
>  I've deliberately not included the customers real name/email (instead it shows the internal Stripe customer ID) so we aren't receiving customers personal details via email

I had not previously considered this - although the Pro Metrics mailer is going to run into the same question - but yeah, having thought about it an email with easily identifiable details about who our customers are and whether they've missed payments ... that's not a good idea should an email ever be misdirected. Let's stick with the Stripe identifiers for now and worry about what to change it to if we find it annoying!

Secondary thought: could you add a changelog line for this and an 'Upgrade note' about this needing Postgres 9.4 or higher. Didn't think of it on Friday. (Note to self: don't review tired!)

gbp:
@garethrees are you happy with this? Specifically what I ask in the implementation notes section of the PR. And also what @lizconlan mentioned regarding the JSONB column and PG version?

garethrees:
> could you add a changelog line for this and an 'Upgrade note' about this needing Postgres 9.4 or higher

Sounds like a good plan.

lizconlan:
Reference for the Postgres change if you need it https://www.postgresql.org/docs/9.4/release-9-4.html

PR #5284, opened 2019-07-09
[docs] Fix spelling mistakes

gbp:
Replace Alavateli with Alaveteli.


gbp:
> 👨‍🏫 👍

@garethrees Did you see the additional commits here? Are they are okay too?

PR #5285, opened 2019-07-10
Add pro help page

garethrees:
## Relevant issue(s)

https://github.com/mysociety/transparency-adessium/issues/18

## What does this do?

Adds info about embargoes and batch – the most asked about features of Pro.

## Why was this needed?

We know how all this works and how to explain it, but it would

1. save us time to point to this page
2. help reusers understand the features themselves, and explain them to customers

## Implementation notes

This PR is draft because it goes against the usual convention of providing placeholders that get overridden by a theme.

In this case we definitely want to provide some defaults somewhere, as its a brand new feature. I don't know whether we want to implement in WDTK and say "copy that", or maybe alavetelitheme is a good place for these. All options seem quite clunky though, so open to thoughts. The main thing at this point was just getting some of the text down.

## Screenshots

help-pro

## Notes to reviewer

Depending on where we decide to implement this, we'll want to update the screenshots to be from the same theme.


garethrees:
After looking back at the history around this (https://github.com/mysociety/alaveteli/pull/3153) it looks like we should really add these to alavetelitheme, and port them over to specific themes where required. This does add a bit of work in this case, but that's the current system so lets just stick with that.

PR #5291, opened 2019-07-12
Add configuration setting to enable pro_pricing

gbp:
## Relevant issue(s)

Connected to #5290 
Connected to https://github.com/mysociety/alaveteli-professional/issues/585

## What does this do?

Add configuration setting to enable `pro_pricing`.

## Why was this needed?

As we're going to be enabling this for other sites soon we should have a way to enable from the configuration settings.

### DIFF
--- a/config/general.yml-example
+++ b/config/general.yml-example
@@ -1243,3 +1243,12 @@ STRIPE_WEBHOOK_SECRET: ''
 # PRO_REFERRAL_COUPON: 'PROREFERRAL'
 # ---
 PRO_REFERRAL_COUPON: ''
+
+# If ENABLE_PRO_PRICING is set to true, Alaveteli let users enter their bank
+# details and subscribe to a Stripe subscription which grants them access to the
+# pro role and all the features of Alaveteli Professional.
+#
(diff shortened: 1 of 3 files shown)


lizconlan:
> Looks like we're using :pro_pricing feature flag – we could instead infer that pricing is enabled if the stripe keys are set.

I'm not sure - it'd neaten up the config but ... is there any reason we might want to disable (or delay enabling) pricing when the keys were set - could that be a surprising thing for reusers? (although we could fix that by adding something like "WARNING: SETTING THIS WILL ENABLE PRICING" in the docs/config file comments of course)

gbp:
> Looks like we're using `:pro_pricing` feature flag – we could instead infer that pricing is enabled if the stripe keys are set.

I was thinking having the extra configuration was a good idea as it will give installers more flexibility when evaluating pro pricing. If the Stripe keys are set then pro pricing can be enabled per user via the flipper UI or Rails console and not exposed to all users until ready.

Happy to change if you definitely feel we should remove.

garethrees:
> Happy to change if you definitely feel we should remove.

I don't feel too strongly about it, so lets just go with this approach for now.

PR #5294, opened 2019-07-15
[pro#587] Make pricing page translatable

lizconlan:
## Relevant issue(s)

Closes mysociety/alaveteli-professional#587

## What does this do?

* Adds translation markup to the pro pricing page
* Extends the `CurrencyHelper#format_currency` method to allow us to display a price as e.g. '£15' rather than '£15.00' without assuming a particular pricing format
* Fetches the price from Stripe rather than hardcoding "£10" in a translatable string (encodes the assumption that the key plan id will be "pro")

## Why was this needed?

So that the pricing page can be translated without overriding the entire template page in a theme

## Implementation notes

* The sample copy for testimonials and FAQs - hidden behind feature flags - was deliberately not included
* ~~I've left `WhatDoTheyKnow<sup>Pro</sup>` as-is rather than pass it in from `pro_site_name` (which would be without the `<sup></sup>` markup) but I'm not sure that's right, although overriding the whole page for WhatDoTheyKnow to include the superscript would be pretty annoying~~

Related PR for WhatDoTheyKnow to reinstate the <sup>Pro</sup> styling here: https://github.com/mysociety/whatdotheyknow-theme/pull/584

### DIFF
--- a/app/controllers/alaveteli_pro/plans_controller.rb
+++ b/app/controllers/alaveteli_pro/plans_controller.rb
@@ -9,6 +9,7 @@ class AlaveteliPro::PlansController < AlaveteliPro::BaseController
     default_plan_name = add_stripe_namespace('pro')
     stripe_plan = Stripe::Plan.retrieve(default_plan_name)
     @plan = AlaveteliPro::WithTax.new(stripe_plan)
+    @pro_site_name = AlaveteliConfiguration.pro_site_name
   end
 
   def show
--- a/app/controllers/alaveteli_pro/plans_controller.rb
+++ b/app/controllers/alaveteli_pro/plans_controller.rb
@@ -6,6 +6,9 @@ class AlaveteliPro::PlansController < AlaveteliPro::BaseController
   before_action :authenticate, :check_existing_subscription, only: [:show]
 
   def index
+    default_plan_name = add_stripe_namespace('pro')
+    stripe_plan = Stripe::Plan.retrieve(default_plan_name)
+    @plan = AlaveteliPro::WithTax.new(stripe_plan)
   end
 
   def show
--- a/app/helpers/currency_helper.rb
+++ b/app/helpers/currency_helper.rb
@@ -1,5 +1,6 @@
 module CurrencyHelper
-  def format_currency(amount)
-    Money.new(amount, AlaveteliConfiguration.iso_currency_code).format
+  def format_currency(amount, no_cents_if_whole: false)
+    Money.new(amount, AlaveteliConfiguration.iso_currency_code).
+      format(no_cents_if_whole: no_cents_if_whole)
   end
 end
--- a/app/views/alaveteli_pro/plans/index.html.erb
+++ b/app/views/alaveteli_pro/plans/index.html.erb
@@ -1,22 +1,22 @@
-<h1 class="pricing__title centered">Pricing</h1>
+<h1 class="pricing__title centered"><%= _('Pricing') %></h1>
 
 <div class="pricing__tiers">
   <div class="pricing__tier pricing__tier--primary">
 
     <div class="pricing__tier__heading">
-      <h2>Professional</h2>
-      <p class="pricing__tier__subhead">For journalists, academics and power users</p>
+      <h2><%= _('Professional') %></h2>
+      <p class="pricing__tier__subhead"><%= _('For journalists, academics and power users') %></p>
       <p class="price-label"><span class="price-label__amount">£10</span> per user, per month</p>
-      <p class="pricing__tier__subhead pricing__tier__subhead--intro_pricing">Introductory Pricing</p>
+      <p class="pricing__tier__subhead pricing__tier__subhead--intro_pricing"><%= _('Introductory Pricing') %></p>
     </div>
 
     <div class="pricing__tier__content">
-      <h3>Featuring</h3>
+      <h3><%= _('Featuring') %></h3>
       <ul class="pricing__tier__features">
-        <li>Public requests for information</li>
-        <li>Private requests for information</li>
-        <li>Dashboard and workflow features</li>
-        <li>Friendly support</li>
+        <li><%= _('Public requests for information') %></li>
+        <li><%= _('Private requests for information') %></li>
+        <li><%= _('Dashboard and workflow features') %></li>
+        <li><%= _('Friendly support') %></li>
       </ul>
       <%= link_to _('Sign up'), plan_path('pro'), :class => 'button button-pop' %>
     </div>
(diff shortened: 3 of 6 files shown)


app/controllers/alaveteli_pro/plans_controller.rb
```diff
@@ -6,6 +6,9 @@ class AlaveteliPro::PlansController < AlaveteliPro::BaseController
   before_action :authenticate, :check_existing_subscription, only: [:show]
 
   def index
+    plan_name = add_stripe_namespace('pro')
+    stripe_plan = Stripe::Plan.retrieve(plan_name)
+    @plan = AlaveteliPro::WithTax.new(stripe_plan)
```
garethrees: Looks like there's a spec missing here?

app/views/alaveteli_pro/plans/index.html.erb
```diff
@@ -6,7 +6,8 @@
     <div class="pricing__tier__heading">
       <h2><%= _('Professional') %></h2>
       <p class="pricing__tier__subhead"><%= _('For journalists, academics and power users') %></p>
-      <p class="price-label"><span class="price-label__amount">£10</span> per user, per month</p>
+      <p class="price-label"><%= _('<span class="price-label__amount">{{monthly_price}}</span> per user, per month</p>',
```
garethrees: Doesn't look like the closing `</p>` should be in the translation string here?

app/helpers/currency_helper.rb
```diff
@@ -1,5 +1,6 @@
 module CurrencyHelper
-  def format_currency(amount)
-    Money.new(amount, AlaveteliConfiguration.iso_currency_code).format
+  def format_currency(amount, discard_zero_cents = false)
```
garethrees: We only support Ruby 2.3+ now, so a keyword argument would be compatible and make this a lot cleaner. I think it would also make sense to mirror the name the `Money` lib uses so that you don't have to translate between the two variable names.

```ruby
def format_currency(amount, no_cents_if_whole: false)
  Money.new(amount, AlaveteliConfiguration.iso_currency_code).
    format(no_cents_if_whole: no_cents_if_whole)
end
```

app/views/alaveteli_pro/plans/index.html.erb
```diff
@@ -25,72 +25,73 @@
...
     <div class="marketing__section__content-items">
       <ul class="marketing__section__feature-list">
-        <li>Up-to-date database of <strong class="marketing-highlight">contact details</strong> for 20,000 authorities</li>
-        <li>A <strong class="marketing-highlight">searchable archive</strong> of Freedom of Information requests</li>
-        <li><strong class="marketing-highlight">Delivery verification</strong> for proof of receipt</strong></li>
-        <li>A permanent, searchable, <strong class="marketing-highlight">public record</strong> of your request and the responses</li>
-        <li><strong class="marketing-highlight">Streamlined process</strong> for requesting internal reviews</li>
+        <li><%= _('Up-to-date database of <strong class="marketing-highlight">contact details</strong> for 20,000 authorities') %></li>
+        <li><%= _('A <strong class="marketing-highlight">searchable archive</strong> of Freedom of Information requests</li>
+        <li><strong class="marketing-highlight">Delivery verification</strong> for proof of receipt</strong>') %></li>
```
garethrees: No opening erb tag here

app/views/alaveteli_pro/plans/index.html.erb
```diff
@@ -25,72 +26,73 @@
...
   <div class="marketing__section__content">
     <div class="marketing__section__content-items">
       <ul class="marketing__section__feature-list">
-        <li>Up-to-date database of <strong class="marketing-highlight">contact details</strong> for 20,000 authorities</li>
-        <li>A <strong class="marketing-highlight">searchable archive</strong> of Freedom of Information requests</li>
-        <li><strong class="marketing-highlight">Delivery verification</strong> for proof of receipt</strong></li>
-        <li>A permanent, searchable, <strong class="marketing-highlight">public record</strong> of your request and the responses</li>
-        <li><strong class="marketing-highlight">Streamlined process</strong> for requesting internal reviews</li>
+        <li><%= _('Up-to-date database of <strong class="marketing-highlight">contact details</strong> for 20,000 authorities') %></li>
+        <li><%= _('A <strong class="marketing-highlight">searchable archive</strong> of Freedom of Information requests</li>
```
garethrees: No closing erb tag here

app/views/alaveteli_pro/plans/index.html.erb
```diff
@@ -25,72 +25,73 @@
...
   </div>
   <div class="marketing__section__content">
     <div class="marketing__section__content-items">
       <ul class="marketing__section__feature-list">
-        <li>Up-to-date database of <strong class="marketing-highlight">contact details</strong> for 20,000 authorities</li>
-        <li>A <strong class="marketing-highlight">searchable archive</strong> of Freedom of Information requests</li>
-        <li><strong class="marketing-highlight">Delivery verification</strong> for proof of receipt</strong></li>
-        <li>A permanent, searchable, <strong class="marketing-highlight">public record</strong> of your request and the responses</li>
-        <li><strong class="marketing-highlight">Streamlined process</strong> for requesting internal reviews</li>
+        <li><%= _('Up-to-date database of <strong class="marketing-highlight">contact details</strong> for 20,000 authorities') %></li>
```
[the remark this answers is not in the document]
lizconlan: I wasn't sure on that but yes - that sounds easier to manage

app/views/alaveteli_pro/plans/index.html.erb
lizconlan: 😞 argh - thanks!

app/controllers/alaveteli_pro/plans_controller.rb
lizconlan: split it into a controller spec that checks the assignment of `@plan` and a new view spec file that checks that the page renders the formatted currency value as expected

app/views/alaveteli_pro/plans/index.html.erb
```diff
@@ -6,8 +6,13 @@
...
       <p class="pricing__tier__subhead"><%= _('For journalists, academics and power users') %></p>
-      <p class="price-label"><%= _('<span class="price-label__amount">{{monthly_price}}</span> per user, per month</p>',
-                                   monthly_price: format_currency(@plan.amount_with_tax, true)) %>
+      <p class="price-label">
+        <%= _('<span class="price-label__amount">{{monthly_price}}</span> ' \
+              'per user, per month',
+              monthly_price: format_currency(
+                               @plan.amount_with_tax,
+                               no_cents_if_whole: true)) %>
+      </p>
```
lizconlan: Or - given the assumptions in the layout - can we split that in e.g. `£10` (no translation required as `format_currrency` is doing the work here) and `_('per user, per month')`?

garethrees: I think keeping the full context `"{{monthly_price}} per user, per month"` is good in case it makes more sense to invert the price/text in some languages (e.g. `"per user, per month {{monthly_price}}"`).

PR #5308, opened 2019-07-18
Handle `#name` and `#nickname` methods for Stripe::Plan objects

lizconlan:
## What does this do?

Attempts to access either the `Stripe::Plan#name` (if available) then gets the newer `Stripe::Plan#nickname` if it's not (where not available covers both 'invalid method' and 'this has not been set'

## Why was this needed?

Only needed if we advance to a newer API version that has removed `Stripe::Plan#name`

## Notes to reviewer

Per the commit comments, these are not necessarily the same value. Older Stripe plans may not have a nickname set, in which case the `#nickname` will return nil whereas `#name` will return the Product name; `#nickname` and `#name` are not completely interchangeable!


lizconlan:
Immediately closing this as we're locking in to an older API version. Provided for illustrative purposes only

PR #5315, opened 2019-07-23
[pro#590] Create webhook endpoint

lizconlan:
## Relevant issue(s)

Closes mysociety/alaveteli-professional#590

(Replaces #5302 and steals the reviewer suggestions)

## What does this do?

* Adds a new class method to the `Stripe` module to store the list of events we want our webhook endpoint to listen for
* Adds a rake task to:
  - create a new Stripe webhook endpoint with the right:
    * url
    * api_version
    * list of subscribed events
  - outputs the signing secret to be added to the `general.yml` file

## Why was this needed?

We can set the Stripe API version for the webhook so - although the wrapped Event data will still use the format set by the Stripe account (new accounts will take whatever the latest version is at the time they are created) - the webhook wrapper will take the expected form as set by the `api_version`.

Once the site re-users have set up a Stripe account and shared the API keys with us, we can create the webhook endpoint without asking them to do it via the Stripe admin tools. We won't have full access to their Stripe settings so a mistake here could be quite hard to find and debug.

## Screenshots

Output of the rake task when it creates a new webhook endpoint:

Screen Shot 2019-07-23 at 11 05 51

(The example shown is for a test account that's not linked to anything so it's safe to share)

### DIFF
--- a/config/initializers/stripe.rb
+++ b/config/initializers/stripe.rb
@@ -1,3 +1,14 @@
 # -*- encoding : utf-8 -*-
 Stripe.api_key = AlaveteliConfiguration.stripe_secret_key
 Stripe.api_version = '2017-01-27'
+
+module Stripe
+  ##
+  # An array of events that we want to enable for our webhook endpoint
+  def self.webhook_events
+    %w(
+      customer.subscription.deleted invoice.payment_succeeded
+      invoice.payment_failed customer.subscription.updated
+    )
+  end
+end
--- a/lib/alaveteli_pro/webhook_endpoints.rb
+++ b/lib/alaveteli_pro/webhook_endpoints.rb
@@ -1,0 +1,38 @@
+module AlaveteliPro
+  ##
+  # A class for working with the Stripe WebhookEndpoints API
+  class WebhookEndpoints
+    ##
+    # Calculates the URL of the Stripe webhook endpoint for the install
+    def self.webhook_endpoint_url
+      options = {
+        host: AlaveteliConfiguration.domain,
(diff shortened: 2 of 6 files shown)


lib/alaveteli_pro/webhook_endpoints.rb
```diff
@@ -0,0 +1,41 @@
...
+  class WebhookEndpoints
+    ##
+    # Calculates the URL of the Stripe webhook endpoint for the install
+    def self.webhook_endpoint_url
+      options = {
+        host: AlaveteliConfiguration.domain,
+        protocol: AlaveteliConfiguration.force_ssl ? 'https' : 'http'
+      }
+
+      unless AlaveteliConfiguration.include_default_locale_in_urls
```
gbp: Doubt we ever want to `locale` in the route do we need to consider this configuration option - seems to work fine without?

config/initializers/stripe.rb
```diff
@@ -1,3 +1,12 @@
 # -*- encoding : utf-8 -*-
 Stripe.api_key = AlaveteliConfiguration.stripe_secret_key
 Stripe.api_version = '2017-01-27'
+
+module Stripe
+  def self.webhook_events
```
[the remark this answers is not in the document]
lizconlan: Good point - thanks!

lib/alaveteli_pro/webhook_endpoints.rb
lizconlan: Maybe not, let's simplify it

Issue #5229, opened 2019-05-30
Release 0.35


garethrees:
Planned Ruby / OS support list https://gist.github.com/garethrees/ec3c403e1ea0744fb736381785e1e788

lizconlan:
## Essentials

- [ ] [Drop support for Xenial](https://github.com/mysociety/alaveteli/pull/5318)
- [ ] [Fix BigDecimal deprecation warning](https://github.com/mysociety/alaveteli/pull/5314)
- [ ] [Add support for Ruby 2.6](https://github.com/mysociety/alaveteli/issues/5038)
- [ ] [Enable Pro features](https://github.com/mysociety/asktheeu-theme/issues/77)

## Nice to have

- [ ] [Open GeoIP-database-contrib isn't installable](https://github.com/mysociety/alaveteli/issues/5040)
- [ ] [Upgrade mail gem](https://github.com/mysociety/alaveteli/issues/4896)

gbp:
Release in 293dde9c04badc493db295fb30462518bcffd0d0

PR #5321, opened 2019-07-31
[#5229] Release prep for 0.35.0.0

gbp:
## Relevant issue(s)

Connected to #5229 

## What does this do?

- Annotates models
- Add missing Changelog entries.

### DIFF
--- a/doc/CHANGES.md
+++ b/doc/CHANGES.md
@@ -28,6 +28,14 @@
 * There are some database structure updates so remember to run
   `bundle exec rails db:migrate`
 
+### Changed Templates
+
+    app/views/admin_general/_to_do_list.html.erb
+    app/views/alaveteli_pro/account_request/index.html.erb
+    app/views/alaveteli_pro/plans/index.html.erb
+    app/views/alaveteli_pro/plans/show.html.erb
+    app/views/request/_correspondence.html.erb
+
 # 0.34.0.0
 
 ## Highlighted Features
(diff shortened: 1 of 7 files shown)


doc/CHANGES.md
```diff
@@ -2,18 +2,30 @@
...
+* Send daily webhook digest to the Pro Admin team (Graeme Porteous)
 
 ## Upgrade Notes
 
+* You can run this release without using the Alaveteli Pro Pricing functionality
+  - by default it is switched off.
+* We have specified a Stripe API version of `2017-01-27` if you are running Pro
+  with pricing enabled you should consider re-creating the Stripe Webhook using
+  the new `bundle exec rails stripe:create_webhook_endpoint` to ensure the
+  versions are consistent
```
lizconlan: Will also need to delete the old webhook afterwards otherwise they'll get duplicate webhook events

doc/CHANGES.md
```diff
@@ -2,18 +2,30 @@
 
 ## Highlighted Features
 
+* This release rolls out Alaveteli Pro Pricing - a way to ask users to support
+  your site by paying for access to the advanced Alaveteli Pro features?
```
lizconlan: I'd take out the question mark here (typo?)

doc/CHANGES.md
garethrees: Delete or disable?

lizconlan: Good question - disabled is probably fine (less stressful for the user as the consequences of getting the wrong one is just "oh, I'll enable it again" rather than "argh! now it's gone!"). We could advise deleting disabled webhooks if it causes a problem.

PR #5328, opened 2019-08-02
Add pro up-sell to request page sidebar

garethrees:
## What does this do?

* Various code cleanup
* Add pro up-sell to request page sidebar

## Why was this needed?

Direct more people to `/pro` for £££.

## Screenshots

Screenshot 2019-08-02 at 14 20 11

### DIFF
--- a/app/assets/stylesheets/responsive/_sidebar_layout.scss
+++ b/app/assets/stylesheets/responsive/_sidebar_layout.scss
@@ -1,7 +1,6 @@
 /* Display of sidebars */
 .sidebar {
   h2 {
-    margin-top: 3.5em;
     margin-bottom: 0.875em;
   }
   .link_button_green {
@@ -9,6 +8,10 @@
   }
   .report-this-request {
     width: 100%;
+  }
+
+  .sidebar__section {
+    margin-bottom: 3.5em;
   }
 }
 
--- a/app/assets/stylesheets/responsive/_sidebar_layout.scss
+++ b/app/assets/stylesheets/responsive/_sidebar_layout.scss
@@ -12,6 +12,16 @@
   }
 }
 
+.new-request-cta {
+  .new-request__title {
+    margin-top: 0;
+  }
+}
+
+.act_link {
+  margin-bottom: 0.5em;
+}
+
 /* TODO: Not sure this is still used – we use the request header action bar for
    tracks */
 /* #track-request .feed_link is here to override specificity in the request sidebar,
@@ -21,12 +31,6 @@
   .track-request-action {
     margin-bottom: 1em;
     display: inline-block;
-  }
-}
-
-.new-request-cta {
-  .new-request__title {
-    margin-top: 0;
   }
 }
 
@@ -45,8 +49,4 @@
   time {
     display: block;
   }
-}
-
-.act_link {
-  margin-bottom: 0.5em;
 }
--- a/app/assets/stylesheets/responsive/_sidebar_layout.scss
+++ b/app/assets/stylesheets/responsive/_sidebar_layout.scss
@@ -12,6 +12,8 @@
   }
 }
 
+/* TODO: Not sure this is still used – we use the request header action bar for
+   tracks */
 /* #track-request .feed_link is here to override specificity in the request sidebar,
    normally .feed_link should be enough */
 #track-request .feed_link,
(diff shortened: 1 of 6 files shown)


app/views/request/_sidebar.html.erb
```diff
@@ -9,32 +9,32 @@
                locals: { info_request: @info_request } %>
   <% end %>
 
-  <div class="request__special-status">
```
gbp: I would split this into two commits, one where the indentation is sorted out, the second where the empty div is corrected. Would give a better diffs.

app/assets/stylesheets/responsive/_sidebar_layout.scss
```diff
@@ -12,6 +12,8 @@
   }
 }
 
+/* TODO: Not sure this is still used – we use the request heade action bar for
```
gbp: Typo? s/heade/header/ ?

app/helpers/application_helper.rb
```diff
@@ -213,4 +213,10 @@ def site_wide_announcement
       session[:announcement_dismissals]
     ).first
   end
+
+  def show_pro_upsell?(user)
+    return false unless feature_enabled?(:alaveteli_pro)
```
gbp: As we're just dealing with booleans here are guard clauses necessary, could be written as but is it as readable?

```suggestion
    feature_enabled?(:alaveteli_pro) && (!user || user && !user.is_pro?)
```

app/views/request/_sidebar.html.erb
```diff
@@ -1,49 +1,66 @@
...
     <%= render :partial => "general/new_request" %>
   </div>
 
+  <% if show_pro_upsell?(@user) %>
+    <div class="sidebar__section pro-upsell">
+      <% pro_link = link_to account_request_index_path do %>
+        <%= AlaveteliConfiguration.pro_site_name -%>
+      <% end %>
+
+      <% pro_site_name_link = content_tag(:strong, pro_link) %>
```
gbp: Extract to helper? Avoids variable assignment in the views and so its testable too?

app/views/request/_sidebar.html.erb
```diff
@@ -1,49 +1,66 @@
...
+
+      <%= _('{{pro_site_name_link}} is a powerful, fully-featured FOI ' \
```
lizconlan: yay, this exact wording is used on the account request page so we can go ahead and lock the translations down for the release when we're ready

lizconlan: hmm, thinking in terms of translator effort and release blockers, this is the same wording as the account request page but a different variable name

https://github.com/mysociety/alaveteli/blob/develop/app/views/alaveteli_pro/account_request/index.html.erb#L4-L6

app/views/request/_sidebar.html.erb
garethrees: Will fix in merge 👍

app/views/request/_sidebar.html.erb
garethrees: Fixed in c62120410

app/helpers/application_helper.rb
garethrees: Yeah I couldn't decide between this, but happy to reduce it down. Fixed in c7860024c.

app/assets/stylesheets/responsive/_sidebar_layout.scss
garethrees: Fixed in 99507c4e8

app/views/request/_sidebar.html.erb
garethrees: I couldn't decide about this either. I was thinking the whole thing should be a helper; something like:

```ruby
<= pro_upsell if show_pro_upsell?(@user) %>
```

But, if there's success here, I'd like to also add it to one of the slots under the search bar in `/select_authority`, which would be a good time to refactor common bits? I didn't really want to delay this while I think about that.

I can't really think of a good name for a helper that does just the link at the moment. Any thoughts? It seems like it needs to be something like `strong_pro_site_name_link_to_account_request_index` to be explicit about what its doing, but that's rather unwieldy!

garethrees: I guess there could be a helper for just the text part:

```erb
<% if show_pro_upsell?(@user) %>
  <div class="sidebar__section pro-upsell">
    <%= pro_upsell %>
  </div>
<% end %>
```

garethrees: Fixed in a91111a7d.

PR #5331, opened 2019-08-06
Hotfix for 0.34.0.1

gbp:
## What does this do?

- Updates translations
- Bump alaveteli version

PR #5332, opened 2019-08-06
Add rake task reformat gettext files consistently

gbp:
## Relevant issue(s)

Used on #5331 and resulted the changes in 2c0be351c4e5af843a592c42d141e6f205a48b11

## What does this do?

Reformat gettext files consistently using regexp

## Why was this needed?

When pulling translations from Transifex it was hard to see what really had changed. This makes it much easier to see.

### DIFF
--- a/lib/tasks/gettext.rake
+++ b/lib/tasks/gettext.rake
@@ -104,6 +104,30 @@ namespace :gettext do
     end
   end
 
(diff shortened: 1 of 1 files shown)


gbp:
Well the existing clean commands actually work - I'm sure they weren't doing anything previously! This is completely not needed.

Issue #1005, opened 2013-07-22
Allow individual attachments to incoming messages to be hidden


crowbot:
As @RichardTaylor [comments](https://github.com/mysociety/alaveteli/issues/209#issuecomment-2669338) often it is a particular attachment that admins would like to hide. 

In order to enable this, I think we would need to:
- [ ] Add a prominence field to FoiAttachment
- [ ] Add a prominence notes field to FoiAttachment
- [ ] Either update get_attachments_for_display to consider the visibility of the attachments or (in order that we can show them or show the reason why they are hidden) leave it as it is and reference prominence and prominence notes when deciding how to display each attachment
- [ ] As with https://github.com/mysociety/alaveteli/issues/209, we need to make sure that cached versions of attachments are not accessible to users when the attachments are not. 
- [ ] Decide how to display a hidden attachment when it does/doesn't have notes

RichardTaylor:
Sometimes hiding isn't enough - see #2686

petterreinholdtsen:
We have seen the need for such feature in Norway too.

Gemmamysoc:
Belgium have also identified this need

equivalentideas:
We've just come across this need for Right To Know as well. We'd like to replace an attachment with a redacted version to hide a person's signature who has requested it be removed.

turukawa:
I'm more comfortable with redacting individual files and removing them entirely (rather than deleting an entire response) than I am with editing the files themselves. Editing is a problem. Redacting should be ok (and with some automated note indicating a file was removed)

nigeljonez:
We (NZ) have faced this issue occasionally.

In the interests of transparency it'd be better to be able to hide individual attachments than full responses as useful information may still be stored in non-hidden attachments.  Examples include agencies adding the requestors address(es) to cover letters but not the actual response, redaction errors, and so on.

I don't mind mucking in on this, but, taking the original task list it seems the main considerations for implementation would be...

* Do we follow the prominence options for individual messages (normal/requestor_only/hidden) or add extra?
* Do we show the filename even if the prominence is changed, if so do we make it an option to also apply the prominence to the filename?
* How do we display it to the users of the site?
* Do we need prominence for each attachment?

Answering no to the last question could be the 'neatest' solution, and would allow (from an admin view) something similar to:
![image](https://user-images.githubusercontent.com/414169/61029845-1af1d880-a410-11e9-9ab4-d5461428a99b.png)
[...]

PR #5334, opened 2019-08-08
Show incoming message attachments in admin

garethrees:
## Relevant issue(s)

https://github.com/mysociety/sysadmin/issues/1169

## What does this do?

* Minor code cleanup
* Lists `FoiAttachment`s related to an `IncomingMessage` at `AdminIncomingMessageController#edit`.

## Why was this needed?

When handling data breaches we need to find the hexdigest of the
attachment to ensure it gets purged from all caches / backups etc.

Its quite hard to do this – even from the console – and any console
commands are prone to error. Its easier to just show the associated
attachments on the admin interface, along with some less user-friendly
information that's useful for admin purposes.

It turned out it was easier to add this than explain in our [internal guide](https://wiki.mysociety.org/wiki/Handling_a_data_breach_by_an_authority)
how to find the `hexdigest` of a particular attachment through the rails
console.

This might also be a nice point at which to link to individual attachments as part of https://github.com/mysociety/alaveteli/issues/1005.

## Screenshots

Screenshot 2019-08-08 at 16 12 29

Screenshot 2019-08-08 at 16 12 43

### DIFF
--- a/app/views/admin_incoming_message/_foi_attachments.html.erb
+++ b/app/views/admin_incoming_message/_foi_attachments.html.erb
@@ -1,0 +1,30 @@
+<fieldset id="attachments" class="form-horizontal">
+  <legend>Attachments</legend>
+
+  <% if foi_attachments.any? %>
+    <table class="table">
+      <thead>
+        <tr>
+          <th>ID</th>
+          <th>Filename</th>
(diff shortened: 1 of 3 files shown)


garethrees:
Noting to myself to remember to add a changelog entry

PR #5335, opened 2019-08-09
Add some hound checks

garethrees:
Mostly around parameter indentation and spacing

### DIFF
--- a/.ruby-style.yml
+++ b/.ruby-style.yml
@@ -122,7 +122,7 @@ Layout/InitialIndentation:
 
 Layout/FirstParameterIndentation:
   Description: 'Checks the indentation of the first parameter in a method call.'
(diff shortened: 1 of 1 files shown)


garethrees:
I think its generally easier to fix only warnings on the code you change. For example, I just tried auto-correcting [`Layout/FirstParameterIndentation`](https://github.com/mysociety/alaveteli/pull/5336), but that then warns about lots of other unrelated checks. I'd then have to auto-correct lots of other checks to get a clean build, which has far more potential to introduce bugs as change happens.

I'm not against using auto-correct, but in our current situation I think its better employed:

* on files that are much closer to our target to just hoover up the last remaining issues
* on smaller files where there's less overall change
* on very well tested files

Issue #5199, opened 2019-05-23
[Rails 5] Check - Halting Active Record Callback Chains via `throw(:abort)`


lizconlan:
Part of #3969

https://edgeguides.rubyonrails.org/upgrading_ruby_on_rails.html#halting-callback-chains-via-throw-abort

New behaviour in Rails 5...

> In Rails 4.2, when a 'before' callback returns false in Active Record and Active Model, then the entire callback chain is halted.
>
> In Rails 5.0, returning false in an Active Record or Active Model callback will not have this side effect of halting the callback chain. Instead, callback chains must be explicitly halted by calling `throw(:abort)`.

I don't think we use this but we should go through all our `ActiveRecord` and `ActiveModel` files and double check. The old behaviour will be retained in Rails 5.0 with a deprecation warning until we opt in to the new behaviour via the `ActiveSupport.halt_callback_chains_on_return_false` config setting (documented in the guide, per the above link)

garethrees:
I agree, I don't think we use this behaviour to halt callback chains.

I've scanned through all `before_` callbacks and can't see anything to suggest we do.

After #5194 we should review the production logs to check that we're not seeing deprecation warnings, and then set `ActiveSupport.halt_callback_chains_on_return_false`.

Issue #3970, opened 2017-05-10
Upgrade to Rails 5.1


garethrees:
Requires https://github.com/mysociety/alaveteli/issues/3969

https://guides.rubyonrails.org/upgrading_ruby_on_rails.html#upgrading-from-rails-5-0-to-rails-5-1

garethrees:
Update ID columns to bigint. https://m.signalvnoise.com/update-on-basecamp-3-being-stuck-in-read-only-as-of-nov-8-9-22am-cst-c41df1a58352

garethrees:
Check whether https://github.com/mysociety/alaveteli/issues/5230 is a problem

lizconlan:
Check whether we can flip `ActiveSupport.halt_callback_chains_on_return_false` to false yet

https://edgeguides.rubyonrails.org/upgrading_ruby_on_rails.html#halting-callback-chains-via-throw-abort

https://github.com/mysociety/alaveteli/issues/5199

PR #5350, opened 2019-08-15
[Rails 5.1] Remove ActiveSupport.halt_callback_chains_on_return_false override

lizconlan:
## Relevant issue(s)

Required by #3970 

## What does this do?

Allows the new (since Rails 5.0) default to apply instead of overriding it. No longer halts ActiveSupport callback chains by returning false.

## Why was this needed?

This method is now deprecated Rails 5.2 - with a warning in the last version of Rails 5.1 - and causes spec errors (was supplied as a stopgap to allow missed manual callback halting via `throw(:abort)` to be caught and fixed.
https://guides.rubyonrails.org/upgrading_ruby_on_rails.html#halting-callback-chains-via-throw-abort

Removes the deprecation warning which was making specs expecting blank output fail:

```
DEPRECATION WARNING: ActiveSupport.halt_callback_chains_on_return_false= is deprecated and will be removed in Rails 5.2. 
```

## Notes to reviewer

We had intended to do this after the Rails 5.0 upgrade once we were sure it was not logging warnings in production  https://github.com/mysociety/alaveteli/issues/5199

I've had a quick look through the log files and could not see any reference to it but a double check before merging would be good!

### DIFF
--- a/config/initializers/new_framework_defaults.rb
+++ b/config/initializers/new_framework_defaults.rb
@@ -20,6 +20,3 @@ ActiveSupport.to_time_preserves_timezone = false
 
 # Require `belongs_to` associations by default. Previous versions had false.
 Rails.application.config.active_record.belongs_to_required_by_default = false
-
-# Do not halt callback chains when a callback returns false. Previous versions had true.
-ActiveSupport.halt_callback_chains_on_return_false = true


gbp:
For some callbacks we sometimes return false but not because we want to halt the callback chain so behaviour may change

Examples:
https://github.com/mysociety/alaveteli/blob/e4e7a3fbee8783567bc82a6f98254b6f7b1fe51e/app/controllers/public_body_change_requests_controller.rb#L60-L62
https://github.com/mysociety/alaveteli/blob/12763b486837c7d1fb76a2bc425b1d13671aa08e/app/controllers/request_controller.rb#L1143-L1146

gbp:
In the `set_render_recaptcha` examples above I don't think there is any impact because in both cases these are the last callbacks in the chain.

gbp:
Ah actually we're just concerned by ActiveRecord, ActiveModel callbacks [1] :)

We should expect to see ```DEPRECATION WARNING: Returning `false` in Active Record and Active Model ...``` in our logs if this is occurring.

To check I'm currently running:
`zgrep 'DEPRECATION WARNING' log/production.log*`

[1] https://blog.bigbinary.com/2016/02/13/rails-5-does-not-halt-callback-chain-when-false-is-returned.html

gbp:
Right. `zgrep` returned nothing and I've been through all relevant callbacks and the only ones which might accidentally return false due to variable assignment have already been set to return true.

Issue #5080, opened 2019-01-29
Site not respecting user locale preference from browser


lizconlan:
> Many dutch speaker users complain that Transparencia.be does not detect their browser Dutch language 

2 problems here:

Firstly, our code is not working as intended as [when we call `set_gettext_locale`](https://github.com/mysociety/alaveteli/blob/0.31.0.6/app/controllers/application_controller.rb#L71-L77) when the user visits the homepage for the first time - so the locale is not being passed in as part of the URL - we are forcibly providing the default locale rather than allowing a null value. This takes precedence so that the language doesn't constantly flip back after you've chosen one so the browser value is never considered[1].

Secondly,[ when we ask `FastGetText` to pick the best match using the browser-supplied string](https://github.com/mysociety/alaveteli/blob/0.31.0.6/lib/alaveteli_localization.rb#L33-L39), if it gets something like `"nl,en-GB;q=0.7,en;q=0.3"` (mine, from Firefox, after adding Dutch for testing purposes), it's not going to match "nl" to "nl_BE" (it handles a string with `nl-BE` just fine but we should handle both).

(When fixed, check the site is configured with `use_default_browser_language`)

---

[1] Applies when the site is configured so that the default locale is not included in the URL otherwise the browser settings will never be considered as there is, legitimately, always a locale param - that's a different, and currently theoretical, problem which is not addressed here.

garethrees:
Now that we have improved fallback handling (https://github.com/mysociety/alaveteli/commit/bdf842350d507ee1f2cc722edf17656faeb19b82) we should be able to use these to find the closest locale from the browser string – something like:

```diff
diff --git a/lib/alaveteli_localization.rb b/lib/alaveteli_localization.rb
index 55883e8d5..61b044936 100644
--- a/lib/alaveteli_localization.rb
+++ b/lib/alaveteli_localization.rb
@@ -34,6 +34,7 @@ class AlaveteliLocalization
     # rubocop:disable Naming/AccessorMethodName
     def set_session_locale(*args)
       requested = args.compact.delete_if(&:empty?).first
+      requested = Locale.parse(requested).i18n_fallbacks
       new_locale = FastGettext.best_locale_in(requested) || default_locale
       locale = Locale.parse(new_locale)
```

Might need to consider the case where the browser locale is set to a non-regionalised locale (e.g. `en`) but the site only has a regional version of that locale available:

```ruby
# User has 'en-GB' set – likely to find a good fallback
[...]

PR #5354, opened 2019-08-21
[#5080] Use browser locale

lizconlan:
## Relevant issue(s)

#5080

## What does this do?

A quick spike exploring what might be involved in accepting the user's language preferences from their browser settings (only relevant to multi language sites)

## Why was this needed?

We currently give more weight than intended to the default locale when falling back to it without a param value (when the site is configured not to have the default locale in the URL - which is the most common choice).

## Implementation notes

* No tests as I was just seeing if it was possible
* Might not work as expected when refusing cookies or running in incognito mode
* Bonus feature - we may have to patch FastGettext to get "nl" from the browser to be matched to "nl_BE"

(In other words, I may have justified nobody working on this as it's more trouble than it's worth!)

## Notes to reviewer

The unexpected finding on `collect_locales` (https://github.com/mysociety/alaveteli/commit/155799e9af64798f97ccfc0fc707be2133608d31) might be useful though - although maybe it could go in `AlaveteliLocalization.set_locales`? - if there is ever a reported problem with the language switcher working incorrectly after a locale change.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -80,6 +80,9 @@ class ApplicationController < ActionController::Base
       params_locale = params.fetch(:locale) do
         AlaveteliLocalization.default_locale
       end
+      unless session[:locale] || cookies[:locale]
+        tryfirst = browser_locale
+      end
     else
       params_locale = params[:locale]
     end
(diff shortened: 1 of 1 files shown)


garethrees:

Issue #5358, opened 2019-08-30
Handle when installations are configured to use old Geo IP database format


gbp:
When configured as `GEOIP_DATABASE: /usr/share/GeoIP/GeoIP.dat`

Running the latest release will raise:

```
A MaxMind::DB::InvalidDatabaseError occurred in services#other_country_message:

  Metadata section not found. Is this a valid MaxMind DB file?
  lib/alaveteli_geoip.rb:27:in `new'
```
We should catch this and handle gracefully.

PR #5359, opened 2019-08-30
Download geoip data on install

garethrees:
## Relevant issue(s)

Related to https://github.com/mysociety/alaveteli/issues/5358.

## What does this do?

Download geoip data on install

## Why was this needed?

c10f6c6d005cd1dc1404cda46f5031bfe1696127 introduces a rake task to
download geoip data. Once configured correctly,
42d0c1b0ec2c4a06934dc72b46835dd18aeb9d4a handles keeping it updated.

This commit ensures that the data exists on the first install of the
app.

### DIFF
--- a/script/rails-deploy-before-down
+++ b/script/rails-deploy-before-down
@@ -119,6 +119,8 @@ bundle exec rake submodules:check
 
 bundle exec rake themes:install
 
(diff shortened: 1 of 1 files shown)


garethrees:
Pretty confident about this, so merging

PR #5367, opened 2019-09-10
Add Facebook link to blog sidebar, and fix layout

zarino:
The blog sidebar had fallen out of sync with the layout used on the request page sidebar. This commit fixes that, so the two sidebars are now more similar.

Also adds a Facebook link, if a Facebook username has been supplied in the site config.

Fixes mysociety/whatdotheyknow-theme#602.

Using the `.act-link-icon` component for the RSS link meant adding an RSS icon to the `act-links-sprite`. When I opened the source file up, I noticed it already had a YouTube icon, which wasn’t in the current version of the sprite. I’ve included it in the new sprite anyway, along with an RSS icon. While I was there, I also reorganised the `.act-link-icon--…` rulesets, to match the order of the images in the sprite, left to right.

It was _very tempting_ to put a capital T in the `<%= _("Follow us on twitter") %>` link text, but I realise that would probably require a change of translations, so have refrained for now.

~Sorry, couldn’t take a screenshot, as I don’t know how to enable the blog in my local instance of Alaveteli!~

Screenshot_2019-09-10 WhatDoTheyKnow blog and tweets


garethrees:
Awesome, thanks – will get this looked at soon.

> Sorry, couldn’t take a screenshot, as I don’t know how to enable the blog in my local instance of Alaveteli!

You can set `BLOG_FEED: 'https://www.mysociety.org/category/projects/whatdotheyknow/feed/'` in `general.yml`

Issue #4636, opened 2018-04-25
Prevent User#updated_at getting updated through cron jobs


garethrees:
This made it really hard to identify intentionally updated users vs when a cron job was run when reindexing.

> We do actually update the user record in some of these tasks, so the new updated_at does reflect the last time the user record was modified.
>
> From memory, the attributes that tend to get updated in cron jobs are last_daily_track_email, daily_summary_hour and daily_summary_minute.
> 
> I guess the answer here is to extract those out to their own model & table, so that when they get updated we don't record an update for the parent user record.
> 
> – https://groups.google.com/d/msg/alaveteli-dev/-iv2fZ5ZDYI/EwALP7MtAgAJ

PR #5375, opened 2019-09-18
[#4636] Fix User#updated_at being changed by cron job

gbp:
## Relevant issue(s)

Fixes: #4636

## What does this do?

Track Mailer runs daily for every user set to receive email alerts. After running we set #last_daily_track_email to track when the mailer should next run. We shouldn't touch #updated_at while saving as it makes it hard to identify intentionally updated users.

## Why was this needed?

Increase spam accounts. With this change it will help identify accounts which have been recently updated to add spam.

### DIFF
--- a/spec/mailers/track_mailer_spec.rb
+++ b/spec/mailers/track_mailer_spec.rb
@@ -53,9 +53,20 @@ describe TrackMailer do
 
       it 'should update the time of the user\'s last daily tracking email' do
         expect(@user).to receive(:last_daily_track_email=).with(Time.zone.now)
-        expect(@user).to receive(:save!)
+        expect(@user).to receive(:save!).with(touch: false)
         TrackMailer.alert_tracks
       end
+
+      it "does not update the user's last updated timestamp" do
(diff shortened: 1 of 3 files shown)


spec/mailers/track_mailer_spec.rb
```diff
@@ -53,9 +53,20 @@
 
       it 'should update the time of the user\'s last daily tracking email' do
         expect(@user).to receive(:last_daily_track_email=).with(Time.zone.now)
-        expect(@user).to receive(:save!)
+        expect(@user).to receive(:save!).with(touch: false)
         TrackMailer.alert_tracks
       end
+
+      it 'should not update the time of the user was last updated' do
```
gbp:
```suggestion
      it 'should not update the user\'s last updated timestamp' do
```

garethrees: I'd go with:

```ruby
it "does not update the user's last updated timestamp" do
```

* http://www.betterspecs.org/#should
* https://github.com/rubocop-hq/ruby-style-guide#consistent-string-literals-double-quote

garethrees:
Oh, before merge squash a changelog entry in to this :)

PR #5400, opened 2019-10-08
[pro#597] Fix rendering of Stripe errors

gbp:
## Relevant issue(s)

Fixes: https://github.com/mysociety/alaveteli-professional/issues/597

## What does this do?

Fixes missing Stripe errors for some cases

## Why was this needed?

Errors for expired/invalid coupons and some card processing errors were
not being shown to the user due to the form submission being done via
AJAX.

When redirecting after an error the AJAX doesn't know how to handle the
response. Using the json_redirect_to method returns the URL of the
redirection to the frontend and is already handled in the JS [1].

[1] https://github.com/mysociety/alaveteli/blob/7025d65/app/assets/javascripts/alaveteli_pro/stripe.js#L146

## Screenshots

Before - no page reload (note the disabled state of the processing button):
broken

After - page reloaded and shows error message:
invalid

### DIFF
--- a/app/controllers/alaveteli_pro/subscriptions_controller.rb
+++ b/app/controllers/alaveteli_pro/subscriptions_controller.rb
@@ -79,7 +79,7 @@ class AlaveteliPro::SubscriptionsController < AlaveteliPro::BaseController
     end
 
     if flash[:error]
-      redirect_to plan_path(non_namespaced_plan_id)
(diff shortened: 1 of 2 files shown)


garethrees:
Could you include a screenshot of an error occurring before/after this change? 🙏

gbp:
> Could you include a screenshot of an error occurring before/after this change? 🙏

Added above ^

Issue #5401, opened 2019-10-08
Internal admin authority being created twice


gbp:
There have been some cases [1] where the Internal admin authority has been created twice. While at Alaveteli Con 2019 we were able to take a look at the DB of an affected installation.

It appears the Internal admin authority translated columns were created in a secondary locale. This breas the logic in `PublicBody.internal_admin_body`, where, this query: https://github.com/mysociety/alaveteli/blob/d3dc6b3c075c98d6ca119bf12516f820af087ff1/app/models/public_body.rb#L385 will return more than one result which causes the `Multiple public bodies (n) found with url_name 'internal_admin_authority'` error to be raised.

One solution was to fix the database records by running:
```ruby
PublicBody::Translation.where(url_name: 'internal_admin_authority').where.not(locale: AlaveteliLocalization.default_locale).destroy_all
```

[1] https://groups.google.com/forum/m/#!topic/alaveteli-dev/YoOrH_qVDxc

garethrees:
Just linking to https://github.com/codeforcroatia/imamopravoznati/issues/76 for future reference

PR #5402, opened 2019-10-08
[#5401] Fix multiple internal admin bodies

gbp:
# Relevant issue(s)

Fixes: #5401

## What does this do?

Scope the query to find the internal admin authority to the default
locale. This prevents multiple results for each translated locale.

## Why was this needed?

Prevent exception warning of multiple internal admin bodies and breaking whole locales.

See: https://groups.google.com/forum/m/#!topic/alaveteli-dev/YoOrH_qVDxc

## Notes to reviewer

In order to test this you'll need to run: `PublicBody::Translation.create(url_name: 'internal_admin_authority', locale: 'cy', ...)` for a non-default configured locale.

### DIFF
--- a/app/models/public_body.rb
+++ b/app/models/public_body.rb
@@ -382,7 +382,11 @@ class PublicBody < ApplicationRecord
 
   # The "internal admin" is a special body for internal use.
   def self.internal_admin_body
-    matching_pbs = PublicBody.where(:url_name => 'internal_admin_authority')
+    matching_pbs = AlaveteliLocalization.
+      with_locale(AlaveteliLocalization.default_locale) do
+      PublicBody.where(url_name: 'internal_admin_authority')
+    end
+
     case
     when matching_pbs.empty? then
       # "internal admin" exists but has the wrong default locale - fix & return


gbp:
More information on how we tested this:

1. Issue is being seen in Rails 4 only, so checkout version 0.33 of Alaveteli.
2. Assuming there are at least two `AVAILABLE_LOCALES` configured in `config/general.yml`, lets assume `en` & `cy`
3. In a Rails console: `PublicBody.find_by(url_name: 'internal_admin_authority').translations.create(url_name: 'internal_admin_authority', locale: 'cy')`
4. Now browsing to the Welsh version of the site is broken - we don't know how this extra translation is being created on live sites, possibly while upgrading
5. To replicate the issue in the console run: `AlaveteliLocalization.with_locale('cy') { PublicBody.where(:url_name => 'internal_admin_authority') }`
6. This will return more one record which causes an exception to be raised in `PublicBody.internal_admin_body`.
7. Note the query used: `SELECT "public_bodies".* FROM "public_bodies" INNER JOIN "public_body_translations" ON "public_body_translations"."public_body_id" = "public_bodies"."id" WHERE "public_body_translations"."url_name" = 'internal_admin_authority' AND "public_body_translations"."locale" IN ('cy', 'en')`
8. Now switch to Rails 5 with version 0.34 - this probably also includes an updated gloabize gem version
9. The query used is: `SELECT DISTINCT "public_bodies".* FROM "public_bodies" INNER JOIN "public_body_translations" ON "public_body_translations"."public_body_id" = "public_bodies"."id" WHERE "public_body_translations"."url_name" = 'internal_admin_authority' AND "public_body_translations"."locale" IN ('cy', 'en')`
10. Almost identical but with `DISTINCT` meaning only one row is ever returned
11. Which this change applied, the query used is:  `SELECT "public_bodies".* FROM "public_bodies" INNER JOIN "public_body_translations" ON "public_body_translations"."public_body_id" = "public_bodies"."id" WHERE "public_body_translations"."url_name" = 'internal_admin_authority' AND "public_body_translations"."locale" = 'en'`
12. Note now `"public_body_translations"."locale" = 'en'` meaning the other translations are now being ignored.

So although this isn't necessary for Rails 5 (or the newer gloablize gem) we decided this would be a good change as everywhere else the internal admin authority is loaded we're using a `with_locale` block and this makes everything more consistent and protects from future framework changes.

We're not proposing backporting as it stands as there is a fix on #5401 to tidy up the extra translation in the database.

Issue #4127, opened 2017-07-30
Being logged out quite a lot


kingqueen3065:
These last few days, I have found that I am logged out after a relatively brief period (= a couple of hours perhaps) even if selecting the "remember me" box when logging in.

crowbot:
@kingqueen3065 what browser are you seeing this in?

kingqueen3065:
Firefox on desktop Windows, and Dolphin on mobile.

crowbot:
Note from investigating so far - the recent upgrades don't seem to have affected the duration setting on the WDTK cookie when 'remember me' is selected.

crowbot:
@kingqueen3065 could you add a note to this ticket next time it happens please? Then I'll have a dig around in the logs.

kingqueen3065:
Will do! Bookmarked.

cheers

kingqueen3065:
@crowbot My difficulty in doing this is my limited memory. I had to log in just now, and think that previously I wouldn't have had to do so. But I can't remember if I last used WhatDoTheyKnow earlier today or late last night / early hours of this morning; also I don't know how long Alaveteli would by default keep me logged in. (I use the "stay logged in longer" tick box)

It's not a problem having to log in, and isn't causing me any issues, I raised it just in case it is useful information for the other login based issues being reported by users e.g. the password reset loop. But I will try and make a better note of when I next have the issue and when I was last logged in.

crowbot:
@kingqueen3065 if you're checking the 'stay logged in for longer' box and getting logged out every few days, I think that is an indication of an issue, as that option should keep you logged in for 30 days.

crowbot:
@kingqueen3065 has just experienced this

garethrees:
Going to close this as we haven't had any more reports since https://github.com/mysociety/alaveteli/pull/5407

PR #5407, opened 2019-10-14
Make session cookie lax

### LINKED ISSUE #4038: Password reset post redirect loop
> **In the last month or so**, I've seen 1-2 messages each week where a user
> complained of being stuck in an infinite password recovery loop. Below
> is a typical sample message (sent by user 136214). Essentially they
> seem to click on the "forgotten password" link, but once they receive
> the confirmation email and click on the link in there, they just get
> told that they'll be sent an email with a confirmation link again (or
> so they claim), essentially never getting to the stage where they can
> enter a new password
> 
> I haven't been able to reproduce what happens myself, but the fact
> that a number of distinct users had the same issue makes me think it
> might not just be them being incompetent, but perhaps some sort of
> actual glitch.
> 
> When I manually send them the link with the last password change token
> issued by the system for them according to the admin page, it usually
> seems to work afterwards.
> 
> Best, Michael
>
> ---------- Forwarded message ----------
>
> I've not used What Do They Know for a while but I'd like to again and
> I'm trying to access my account and change my Password but all it does
> is keep sending me the link over and over again.

### LINKED ISSUE #5046: Possibly something odd about login when following links from elsewhere
A WhatDoTheyKnow.com user has today reported:
>I tried cutting and pasting the link into a browser and that worked. Simply clicking on it seemed to be blocked on the two computers I tried.

I have previously privately noted:

>If I copy and paste links from emails then I am logged in as soon as I arrive at WhatDoTheyKnow, if I click the links from emails I'm not logged in on the first page I arrive on. 

another administrator noted:
> I follow a search result (either Google or WhatDoTheyKnow's search function) to a request and I get shown the non-logged-in version. If I hit Reload, I get the logged in / admin version.

dracos:
A samesite=strict cookie is not sent in any cross-site request, so e.g. clicking a confirmation link in webmail would try and set a cookie but it would not be sent in the redirect, and not log you in. Fixes #5046. Might well fix #4038 and #4127 also.

### DIFF
--- a/config/initializers/secure_headers.rb
+++ b/config/initializers/secure_headers.rb
@@ -29,7 +29,7 @@
     httponly: { only: ['_wdtk_cookie_session'] },
     secure: { only: ['_wdtk_cookie_session'] },
     samesite: {
-      strict: { only: ['_wdtk_cookie_session'] }
+      lax: { only: ['_wdtk_cookie_session'] }
     }
   }
 end


gbp:
From my understanding after reading https://makandracards.com/makandra/71018-samesite-cookies this seems like the change we want to make 👍

garethrees:
Looks good in principle – going to test on staging

garethrees:
> Looks good in principle – going to test on staging

Yep, seems to work 🙇

PR #5409, opened 2019-10-15
Add CHANGELOG.md symlink to doc/CHANGES.md

garethrees:
## What does this do?

Add CHANGELOG.md symlink to doc/CHANGES.md

## Why was this needed?

This is a more conventional location, and having it at the root of the
project makes it easier to find. For now its just a symlink, but in the
future we may want to make `doc/CHANGES.md` link to `CHANGELOG.md`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,1 @@
+doc/CHANGES.md


garethrees:
Argh, GitHub doesn't resolve symlinks, so defeats the point a bit :(

PR #5438, opened 2019-11-01
Add some hound checks

garethrees:
Another 10, nothing surprising really.

### DIFF
--- a/.ruby-style.yml
+++ b/.ruby-style.yml
@@ -783,21 +783,21 @@ Style/RedundantException:
 
 Style/RedundantFreeze:
   Description: "Checks usages of Object#freeze on immutable objects."
-  Enabled: false
+  Enabled: true
 
 Style/RedundantParentheses:
   Description: "Checks for parentheses that seem not to serve any purpose."
-  Enabled: false
+  Enabled: true
 
 Style/RedundantReturn:
   Description: "Don't use return where it's not required."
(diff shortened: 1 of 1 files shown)


.ruby-style.yml
```diff
@@ -807,34 +807,34 @@ Style/RegexpLiteral:
...
   Description: 'Checks for proper usage of fail and raise.'
   StyleGuide: 'https://github.com/bbatsov/ruby-style-guide#prefer-raise-over-fail'
-  Enabled: false
+  Enabled: true
 
 Style/SingleLineBlockParams:
   Description: 'Enforces the names of some block params.'
   StyleGuide: 'https://github.com/bbatsov/ruby-style-guide#reduce-blocks'
-  Enabled: false
+  Enabled: true
```
gbp: We tend to use `memo` instead of `acc`. I beelive it was you who got me into that habit, do we want to change?

See: https://rubocop.readthedocs.io/en/latest/cops_style/#stylesinglelineblockparams

garethrees: Ah, hmm yeah, I've always used `memo` because that's what Ruby Core docs tend to use. Maybe we should drop this one; I had initially thought it was more about e.g. using [underscores for unused params](https://github.com/bbatsov/ruby-style-guide#reduce-blocks).

garethrees: Removed and added https://github.com/rubocop-hq/ruby-style-guide#no-cryptic-perlisms instead

PR #5439, opened 2019-11-01
Censor rules help

garethrees:
## Relevant issue(s)

Minor improvement on https://github.com/mysociety/alaveteli/issues/2056

## What does this do?

Expands inline help around censor rules.

TL;DR is that I've added the "About" section – previously only on the global censor rules page – to the censor rule form, so that there's as much info as possible available at the point of editing the records.

## Why was this needed?

Helps inexperienced censor rule users to learn about them.

## Screenshots

Screenshot 2019-11-01 at 11 04 49

Screenshot 2019-11-01 at 11 05 23

## Notes to reviewer

Its mostly just shuffling existing content around, with a few minor tweaks.

### DIFF
--- a/app/views/admin_censor_rule/_about.html.erb
+++ b/app/views/admin_censor_rule/_about.html.erb
@@ -1,0 +1,44 @@
+<div class="row">
+  <div class="span12">
+    <div class="alert alert-info">
+      <h3>About Censor Rules</h3>
+
+      <p>Censor Rules are rules to redact information from incoming and outgoing
+         messages.</p>
+
+      <p>Redacting means removing or hiding part of a message so it cannot be
+         read: you are effectively removing part of a document from your site.
+         Typically you do this to conceal sensitive (usually, that means
+         personal) information on the public site.</p>
+
+      <p>These can be powerful, so must be used with caution.</p>
+
+      <p>Censor Rules can be associated with the following:</p>
+
+      <ul>
+        <li>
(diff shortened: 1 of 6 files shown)


RichardTaylor:
One key point for new admins that's missing here is that multi-line censor rules often won't work. 

Line by line redaction is also more likely to work when a request is quoted in a response so should be encouraged. Actually the shorter the text to be redacted the more likely it is to catch occurrences of material in responses. 

Redacting each part of a name separately is useful as if both "Richard" and "Taylor" are redacted then that would catch a mention of "Mr Taylor" whereas a rule for "Richard Taylor" wouldn't. 

The references to other places where censor rules can be imposed could be better described by saying where they can be set eg. on a request page, a user admin page, body admin page, and the global censor rule page. 

The reference to "special code" which works on PDFs is I suspect a reference to uncompressing a PDF and applying a censor rule to uncompressed PDF code, I don't think this hint is sufficient to prompt admins to read about that, and the technique isn't mentioned at https://alaveteli.org/docs/running/redaction/

garethrees:
@RichardTaylor all valid points, but beyond the scope of this change I think.

Here, I'm doing something fairly simple to provide existing content in better places. I don't disagree that the actual content could be improved, but that's a bigger task. This just lays the groundwork for having a good place to make those improvements. https://github.com/mysociety/alaveteli/issues/2056 is a good place for getting specific about what needs adding.

PR #5487, opened 2019-12-03
[#5460] Update new response email

### LINKED ISSUE #5460: New message email confusing for embargoed requests
A pro user has queried the text of the "New response to your FOI request" email;

The text in question is:

> Although all responses are **automatically published**, we depend on
you, the original requester, to evaluate them.

But for pro users who have embargoed their requests, they are not expecting to automatically publish the request. Is there a possibility to change this wording for embargoed requests?

gbp:
## Relevant issue(s)

Fixes #5460 

## What does this do?

Update new response email to take into account if the info request is embargoed or not.
Adds mailer previews [[1](http://0.0.0.0:3000/rails/mailers/request_mailer/new_response/public)] [[2](http://0.0.0.0:3000/rails/mailers/request_mailer/new_response/embargoed)] for this mail

## Why was this needed?

Original copy was incorrect and confusing when dealing with embargoed requests. 

## Implementation notes

In the original version the locale string has had a new line added which is why its wrapping differently. See #3065. 

## Screenshots

Public:
image

Embargoed:
image

### DIFF
--- a/app/views/request_mailer/new_response.text.erb
+++ b/app/views/request_mailer/new_response.text.erb
@@ -10,8 +10,13 @@
 <%= _('When you get there, please update the status to say if the response ' \
       'contains any useful information.' ) %>
 
+<% if @info_request.embargo %>
+<%= _('All responses will be published when the privacy period expires. We ' \
+      'depend on you, the original requester, to evaluate them.') %>
+<% else %>
 <%= _('Although all responses are automatically published, we depend on ' \
       'you, the original requester, to evaluate them.') %>
+<% end %>
 
 -- <%= _('the {{site_name}} team', :site_name => site_name.html_safe) %>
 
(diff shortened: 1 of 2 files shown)


app/views/request_mailer/new_response.text.erb
```diff
@@ -10,8 +10,13 @@
 <%= _('When you get there, please update the status to say if the response ' \
       'contains any useful information.' ) %>
 
+<% if @info_request.embargo %>
+<%= _('All responses will be published when the embargo expires, we depend ' \
```
garethrees: We don't actually use "embargo" in the UI – we use "privacy period". Also suggest minor tweak for readability (`s/, we/. We`)

```diff
+ <%= _('All responses will be published when the privacy period expires. We depend ' \
- <%= _('All responses will be published when the embargo expires, we depend ' \
```

PR #5501, opened 2019-12-16
[#5483] Add citations to info requests and batches

### LINKED ISSUE #5483: Link news stories and articles to requests
A subset of https://github.com/mysociety/alaveteli/issues/34.
Fixes https://github.com/mysociety/transparency-adessium/issues/26.

---

We want to build a culture of fact-based journalism.

One way of doing this is for journalists to show their sources, especially in the case of public information like FOI requests. We know it's a struggle to get them to do this at the story end, but we think there could be milage in linking the other way – from the source to the story.

Benefits for Alaveteli maintainers:

* help track and show impact for credibility and funding purposes

Benefits for Pro users:

* Give them a little free publicity once the data is published
* Encourage them to actually publish their requests (mysociety/alaveteli-professional#385), due to 1

Benefits for free users:

* Increase motivation, knowing that their seemingly small acts can have bigger impact if their requests get picked up by news publications.

## Naming

We're going to call these "Citations" – they're _similar_ to an "annotation", but with more structured data.

## Who can add citations?

For the initial launch, we're only going to allow the following users to create citations:

* Admins
* Pros
* Request owners, on their own requests

There's a little more risk of abuse from allowing request owners to add citations, but I think in general we trust our users and their mass will help this idea pick up more than if we only initially allowed Admins and Pros to add citations.

We can later think about allowing all users to add citations, possibly through a "suggestion" form similar to `PublicBodyChangeRequest`.

## The blank slate

The request page sidebar will be the kicking-off point for Citations. We think its important to give these some prime space since they're a new concept.

citations-new

We'll add a new `.sidebar__section` container with a little explainer, and a link to add a new citation. We need to work out the wording here, but I'm thinking along the lines of:

> IN THE NEWS
>
> Has this request been mentioned in a news article or academic paper?
>
> [Let us know]()

We'll disable the link to the form unless the user has permission to add citations.

The form can be pretty simple. At a minimum we'll record the source URL.

We'll aim to store the "type" of citation (whether its a news story, academic paper or other). This helps users understand the type of citations we're after. This isn't critical though.

We may also want to add a text field for a human-readable title. We could try to automatically populate this with the `<title>` tag contents from the source URL, but it depends how much time we have left.
  
On submission, we just show a flash message saying thanks.

We should reject duplicate source URLs.

## Displaying and adding more citations

citations-index

We'll show the most recent 3-5 citations in the sidebar. We _could_ prioritise "news"-type citations in the future, but probably we'll just go for most recent initially.

Its not critical for launch if we're really pushed, but the aim should be to have a page listing all citations, with further details than we can show in the sidebar.

Clicking a citation link, or "more" should take you to the citations index (`/request/:url_name/citations`).

Here we'll show a little request metadata, and expand citations, showing who added it (for accountability) and when. The citations here can then link off-site to the source URL.

## Editing and deletion

We want to avoid getting bogged down by figuring out exact permissions of who can edit/delete citations in various contexts (e.g. "Can a Pro delete a citation on their request that got added by an Admin?"), so for this first iteration we'll make editing and deletion pretty limited.

We should only allow deletion within 30 minutes of creation by the citation author. After that they'll have to get in touch with the admin team, and developers will have to remove/edit through the console.

## Batch requests

We'll aim to mirror the behaviour of citations on Batch requests – adding an "In the news" section to the sidebar.

citations-batch-show

Citations will be recorded against the `InfoRequestBatch`.

When collecting citations to show alongside a batch, we'll get both the citations directly added to the batch, _plus_ citations added to individual requests within the batch.

On a request page that belongs to a batch, we'll collect citations added to that particular request, _plus_ citations added to the parent batch.

Not essential, but when adding a citation to an individual request in a batch, we could add a tick box that allows the citation to get added to the batch record instead of the individual request: 

citations-batch-form

## The data model

We decided that tags aren't that useful for this, as they only store one value.

Instead we'll create a `Citation` model:

```ruby
# Citation
#   source_url: string    # The URL to the news story etc
#   user_id: int          # The user who added the citation
#   type: string          # News story / Academic Paper / other
#   created_at: datetime
#   updated_at: datetime
```

There's a lot of overlap with the current `Comment` model, so we'll aim to extract the common bits (visibility, reporting, admin columns) out in to an `Annotation` abstract class, with the more specific details (e.g. validations) in subclasses.

The association will have to be polymorphic, so that a `Citation` can get added to an `InfoRequest` or an `InfoRequestBatch`.

We'll start off by creating `Citation` first, if possible, so that at least we have that if we have to abandon the abstract class idea.

## Human title

When listing citations, it would be nice to show something more human friendly than just the URL. There are a couple of options for this, depending on the time available once we've got the core features done.

I think the default should be to just show the URL, less the protocol and `www`.

So for a source URL of `https://www.bbc.co.uk/news/uk-scotland-highlands-islands-50508800` we'd show: `bbc.co.uk/news/uk-scotland-highlands-islands-50508800`.

gbp:
## Relevant issue(s)

Part of #5483 

## What does this do?

Add citations to info requests and batches

## Implementation notes

* `Citation`s for a given `InfoRequest` are a collection of `Citation`s specifically added to that request, plus any `Citation`s added to the parent `InfoRequestBatch`.
* `Citation`s for a given `InfoRequestBatch` are a collection of `Citation`s specifically added to that batch, plus any `Citation`s added to _any_ `InfoRequest` belonging to the batch.

## Screenshots

No citations – authorised to create
Screenshot 2019-12-18 at 17 05 28

---

Citations – authorised to create
Screenshot 2019-12-18 at 17 05 21

---

With 1 citation – not authorised to create
Screenshot 2019-12-18 at 17 05 44

---

No citations – not authorised to create
Screenshot 2019-12-18 at 17 05 35

---

Add citation to a single request

Screenshot 2019-12-19 at 14 37 44

---

Add citation to a request in a batch

Screenshot 2019-12-19 at 14 38 06

## Notes to reviewer

TODO before merge:

- [x] Add Indexes (Rails does this [automatically](https://stackoverflow.com/a/3658980/) now 🙇)
- [x] Validation warnings
- [x] Remove "Show more" – extract to new PR (https://github.com/mysociety/alaveteli/pull/5501/commits/d7f11bbe528454ba82fa4187864a63c72b48d462)
- [x] Truncate long citations (currently creates a horizontal scrollbar)
- [x] Simple styling of citations list
- [x] Disable "add new" link if not authorised

NEW PRs:

* Uniqueness on `Citation#source_url`?
* "Show More" and citations index page
* Ability to destroy citations
* Move status update to top of pro sidebar
* Create citation from batch/show
* Add `citation_created` `InfoRequestEvent`?

### DIFF
--- a/app/controllers/citations_controller.rb
+++ b/app/controllers/citations_controller.rb
@@ -4,8 +4,10 @@
 class CitationsController < ApplicationController
   before_action :authenticate
   before_action :load_info_request_and_authorise
+  before_action :set_in_pro_area
 
   def new
+    @citation = current_user.citations.build
   end
 
   def create
@@ -32,5 +34,9 @@ class CitationsController < ApplicationController
     end
 
     authorize! :create_citation, info_request
+  end
+
+  def set_in_pro_area
+    @in_pro_area = current_user.is_pro? && info_request.user == current_user
   end
 end
--- a/app/controllers/citations_controller.rb
+++ b/app/controllers/citations_controller.rb
@@ -3,6 +3,7 @@
 #
 class CitationsController < ApplicationController
   before_action :authenticate
+  before_action :load_info_request_and_authorise
 
   def new
   end
@@ -19,5 +20,17 @@ class CitationsController < ApplicationController
       email_subject: _('Confirm your account on {{site_name}}',
                        site_name: site_name)
     )
+  end
+
+  def info_request
+    @info_request ||= InfoRequest.find_by_url_title!(params[:url_title])
+  end
+
+  def load_info_request_and_authorise
+    if cannot?(:read, info_request)
+      return render_hidden('request/hidden', response_code: 404)
+    end
+
+    authorize! :create_citation, info_request
   end
 end
--- a/app/controllers/citations_controller.rb
+++ b/app/controllers/citations_controller.rb
@@ -2,9 +2,22 @@
 # Controller to create a new Citation for an InfoRequest or an InfoRequestBatch.
 #
 class CitationsController < ApplicationController
+  before_action :authenticate
+
   def new
   end
 
   def create
+  end
+
+  private
+
+  def authenticate
+    authenticated?(
+      web: _('To add a citation'),
+      email: _('Then you can add citations'),
+      email_subject: _('Confirm your account on {{site_name}}',
+                       site_name: site_name)
+    )
   end
 end
--- a/app/controllers/citations_controller.rb
+++ b/app/controllers/citations_controller.rb
@@ -11,6 +11,16 @@ class CitationsController < ApplicationController
   end
 
   def create
+    @citation = current_user.citations.build(citation_params)
+    @citation.citable = info_request
+
+    if @citation.save
+      notice = _('Citation successfully created.')
+      redirect_to show_request_path(url_title: info_request.url_title),
+                  notice: notice
+    else
+      render :new
+    end
   end
 
   private
@@ -38,5 +48,9 @@ class CitationsController < ApplicationController
 
   def set_in_pro_area
     @in_pro_area = current_user.is_pro? && info_request.user == current_user
+  end
+
+  def citation_params
+    params.require(:citation).permit(:source_url, :type)
   end
 end
--- a/app/controllers/citations_controller.rb
+++ b/app/controllers/citations_controller.rb
@@ -12,7 +12,7 @@ class CitationsController < ApplicationController
 
   def create
     @citation = current_user.citations.build(citation_params)
-    @citation.citable = info_request
+    @citation.citable = citable
 
     if @citation.save
       notice = _('Citation successfully created.')
@@ -52,5 +52,10 @@ class CitationsController < ApplicationController
 
   def citation_params
     params.require(:citation).permit(:source_url, :type)
+  end
+
+  def citable
+    (info_request.info_request_batch if params[:applies_to_batch_request]) ||
+      info_request
   end
 end
--- a/app/controllers/citations_controller.rb
+++ b/app/controllers/citations_controller.rb
@@ -1,0 +1,10 @@
+##
+# Controller to create a new Citation for an InfoRequest or an InfoRequestBatch.
+#
+class CitationsController < ApplicationController
+  def new
+  end
+
+  def create
+  end
+end
--- a/app/controllers/request_controller.rb
+++ b/app/controllers/request_controller.rb
@@ -946,6 +946,8 @@ class RequestController < ApplicationController
     )
 
     @similar_requests, @similar_more = @info_request.similar_requests
+
+    @citations = @info_request.citations.limit(3)
   end
 
   def assign_state_transition_variables
--- a/app/models/citation.rb
+++ b/app/models/citation.rb
@@ -40,4 +40,8 @@ class Citation < ApplicationRecord
     where(citable: info_request_batch).
       or(where(citable: info_request_batch.info_requests))
   end
+
+  def applies_to_batch_request?
+    citable.is_a?(InfoRequestBatch)
+  end
 end
--- a/app/models/citation.rb
+++ b/app/models/citation.rb
@@ -22,6 +22,15 @@ class Citation < ApplicationRecord
   belongs_to :user, inverse_of: :citations
   belongs_to :citable, polymorphic: true
 
+  validates :user, :citable, presence: true
+  validates :citable_type, inclusion: { in: %w(InfoRequest InfoRequestBatch) }
+  validates :source_url, length: { maximum: 255,
+                                   message: _('Source URL is too long') },
+                         format: { with: /\Ahttps?:\/\/.*\z/,
+                                   message: _('Please enter a Source URL') }
+  validates :type, inclusion: { in: %w(news_story academic_paper other),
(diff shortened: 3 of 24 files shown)


app/models/citation.rb
```diff
@@ -0,0 +1,36 @@
...
+#
+
+##
+# A citation of a info request or info request batch in the news stories or an
+# academic paper
+#
+class Citation < ApplicationRecord
+  self.inheritance_column = nil
+
+  belongs_to :user, inverse_of: :comments
```
garethrees: `inverse_of` doesn't look right here

app/models/info_request.rb
```diff
@@ -1739,6 +1739,12 @@ def holding_pen_request?
     self == self.class.holding_pen_request
   end
 
+  def all_citations
```
garethrees: How about this to keep it cleaner?

```diff
diff --git a/app/models/citation.rb b/app/models/citation.rb
index e3ecf9f50..473d421aa 100644
--- a/app/models/citation.rb
+++ b/app/models/citation.rb
@@ -30,6 +30,11 @@ class Citation < ApplicationRecord
   validates :type, presence: true,
                    inclusion: { in: %w(news_story academic_paper other) }

+  scope :for_request, ->(info_request) do
+    where(citable: info_request).
+      or(where(citable: info_request.info_request_batch))
+  end
+
   def applies_to_batch_request?
     citable.is_a?(InfoRequestBatch)
   end
diff --git a/app/models/info_request.rb b/app/models/info_request.rb
index ffd30900e..c660ae935 100644
--- a/app/models/info_request.rb
+++ b/app/models/info_request.rb
@@ -112,7 +112,7 @@ class InfoRequest < ApplicationRecord
            :inverse_of => :info_request,
            :dependent => :destroy
   has_many :citations,
-           -> { order('created_at') },
+           -> (info_request) { unscope(:where).for_request(info_request) },
            as: :citable,
            inverse_of: :citable,
            dependent: :destroy
```

(h/t https://stackoverflow.com/questions/53427787/how-to-override-the-query-in-has-many-relationship)

garethrees: And a counterpart for collecting all citations for a batch, plus all request-specific citations in that batch:

```diff
diff --git a/app/models/citation.rb b/app/models/citation.rb
index e3ecf9f50..40e58f3d4 100644
--- a/app/models/citation.rb
+++ b/app/models/citation.rb
@@ -30,6 +30,16 @@ class Citation < ApplicationRecord
   validates :type, presence: true,
                    inclusion: { in: %w(news_story academic_paper other) }

+  scope :for_batch, ->(info_request_batch) do
+    where(citable: info_request_batch).
+      or(where(citable: info_request_batch.info_requests))
+  end
+
   def applies_to_batch_request?
     citable.is_a?(InfoRequestBatch)
   end

```

gbp: Yeah that was quick stub to see if the associations were working - happy to go with your suggestion

app/controllers/request_controller.rb
```diff
@@ -944,6 +944,8 @@ def assign_variables_for_show_template(info_request)
     @show_other_user_update_status_action = !!(
       @old_unclassified && !@render_to_file
     )
+
+    @citations = @info_request.citations.limit(3)
```
garethrees: Add test

app/models/citation.rb
```diff
@@ -0,0 +1,46 @@
...
+# A citation of a info request or info request batch in the news stories or an
+# academic paper
+#
+class Citation < ApplicationRecord
+  self.inheritance_column = nil
+
+  belongs_to :user, inverse_of: :citations
+  belongs_to :citable, polymorphic: true
+
+  validates :user, :citable, presence: true
```
garethrees: Need to use explicit messages

app/views/request/_citations.html.erb
```diff
@@ -0,0 +1,22 @@
+<h2><%= _('In the News') %></h2>
+
+<% if citations.any? %>
```
garethrees: Add test

app/models/citation.rb
garethrees: c1fdc0bbe

app/views/request/_citations.html.erb
```diff
@@ -0,0 +1,45 @@
+<%# FIXME: Slighly messy logic, but can be cleaned up when we allow any user to
+    add citations %>
```
gbp: Meh duplication, looks simple enough to improve now.

spec/models/citation_spec.rb
```diff
@@ -0,0 +1,203 @@
...
+#  citable_id   :integer
+#  source_url   :string
+#  type         :string
+#  created_at   :datetime         not null
+#  updated_at   :datetime         not null
+#
+
+require 'spec_helper'
+
+def setup_for_x_scope_data
```
gbp: Finding this setup hard to understand - any way this could be encapsulated in the factory?

app/controllers/citations_controller.rb
```diff
@@ -4,8 +4,10 @@
 class CitationsController < ApplicationController
   before_action :authenticate
   before_action :load_info_request_and_authorise
+  before_action :set_in_pro_area
```
gbp: I think this change was separate commit previously. Was it meant to be merged into this one?

PR #5503, opened 2019-12-16
Limit number of Travis CI builds

gbp:
## What does this do?

Configure Travis to use [conditional jobs](https://docs.travis-ci.com/user/conditional-builds-stages-jobs/#Conditional-Jobs) to limit the number of builds required to get a PR green.

We don't need to always test the next Ruby & Rails versions.

- Ruby 2.3.3, which comes with Debian Stretch, is now only tested on
  master and develop as this isn't used by default any more.
- Ruby-head is now only tested on master and develop as its allowed to
  fail.
- Next version of Rails is now only tested on master/develop or any
  branch beginning with `rails`.

## Why was this needed?

Travis can be slow. Implemented while waiting for builds to finish.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,18 +6,22 @@ branches:
     - master
     - develop
 rvm:
-  - 2.3.3 # Debian Stretch
   - 2.3 # Latest official 2.3.x
   - 2.4 # Latest official 2.4.x
   - 2.5 # Latest official 2.5.x
   - 2.6 # Latest official 2.6.x
   - 2.7 # Latest official 2.7.x
-  - ruby-head
 gemfile:
   - Gemfile
-  - Gemfile.rails_next
 matrix:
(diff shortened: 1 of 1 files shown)


garethrees:
I don't really know what's going on here, but for some reason a build on `master` was kicked off at midnight:

Screenshot 2019-12-31 at 05 21 36

The last commit to `master` was on [20th November](https://github.com/mysociety/alaveteli/commit/6db98282183003c1469584462143202b7926e594)! Don't know what triggered this, but seems unnecessary.

gbp:
> I don't really know what's going on here, but for some reason a build on `master` was kicked off at midnight:

Triggered by cron, which I setup [here](https://travis-ci.org/mysociety/alaveteli/settings), as I felt it was good practice to check master is green occasionally.

PR #5505, opened 2019-12-17
[#5504] Fix HTML output in text template

### LINKED ISSUE #5504: HTML <time> tag in plaintext correspondence download
When you download [a request](https://www.whatdotheyknow.com/request/cctv_camera_locations_and_models_31) as a Zip and the correspondence gets [generated as plaintext](https://github.com/mysociety/alaveteli/blob/eb8a0bf4b6bd18357b810a7570ee6019bdc1e937/app/controllers/request_controller.rb#L1017-L1020), in some cases we use a HTML `<time>` tag:

```
From:, Spelthorne Borough Council
To: REDACTED
Date: <time datetime="2019-08-22T13:35:53+01:00" title="2019-08-22 13:35:53 +0100">22 August 2019</time>

Subject Line: FOI 1920_001596 Acknowledgment
```

This is due to [calling `simple_date`](https://github.com/mysociety/alaveteli/blob/eb8a0bf4b6bd18357b810a7570ee6019bdc1e937/app/views/request/_incoming_correspondence.text.erb#L6) without specifying the [non-default](https://github.com/mysociety/alaveteli/blob/eb8a0bf4b6bd18357b810a7570ee6019bdc1e937/app/helpers/date_time_helper.rb#L7-L9) `format: :text` option.

garethrees:
## Relevant issue(s)

Fixes #5504 

## What does this do?

Fixes HTML output in text template.

## Why was this needed?

HTML isn't human-readable, which is the intention of the template.

## Implementation notes

Added some specs, though only enough to verify the bugfix; adding more would be too much effort right now.

## Notes to reviewer

Needed to stub `current_user` and `current_ability` as per https://stackoverflow.com/a/5022406/387558

### DIFF
--- a/spec/views/request/_incoming_correspondence.text.erb_spec.rb
+++ b/spec/views/request/_incoming_correspondence.text.erb_spec.rb
@@ -1,0 +1,45 @@
+# -*- encoding : utf-8 -*-
+require 'spec_helper'
+
+describe 'request/incoming_correspondence' do
+  let(:info_request) { FactoryBot.create(:info_request_with_incoming) }
+  let(:incoming_message) { info_request.incoming_messages.first }
+
+  let(:stub_locals) do
+    { incoming_message: incoming_message }
+  end
+
+  let(:ability) { Object.new.extend(CanCan::Ability) }
+
+  before do
+    allow(controller).to receive(:current_ability).and_return(ability)
+    allow(view).to receive(:current_user).and_return(info_request.user)
+    assign(:info_request, info_request)
+  end
+
+  context 'the current_user cannot read the request' do
+    before { ability.cannot :read, incoming_message }
+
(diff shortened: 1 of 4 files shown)


spec/views/request/_incoming_correspondence.text.erb_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+
+      expected = 'request/_hidden_correspondence'
+      expect(rendered).to render_template(partial: expected)
+    end
+  end
+
+  context 'the current_user can read the request' do
+    before { ability.can :read, incoming_message }
+
+    # TODO: Add better generic success case
```
gbp: Todo now or later?

garethrees: Was going for later as per daaf8eaf57d7a4e498fc47f69ebac055b091e782

PR #5517, opened 2019-12-31
Fix double-billing edge case

garethrees:
## What does this do?

* Adds the `AlaveteliPro::Subscription#past_due?` helper
* Adds the `AlaveteliPro::SubscriptionCollection#past_due` scope
* Allows `ProAccount#active?` to consider a `past_due` subscription as
  "active", in that we still consider a `past_due` sub as "OK" – we're
  assuming the payment problem is temporary, or rely on the
  auto-cancelling to clean up if the user becomes delinquent.

## Why was this needed?

When a subscription goes `past_due` (most commonly after a failed
payment) we send an email suggesting the user update their card at
`/profile/subscriptions`. We only allow access to
`AlaveteliPro::SubscriptionsController` if the user has an _active_
subscription.

Prior to this commit, we defined "active" explicitly as the
`Stripe::Subscription#status` being `"active"`. This means that when a
sub moves to `past_due` the user can no longer access the controller to
add their new card.

We use the same check for `AlaveteliPro::PlansController`. So, when a
user goes `past_due` they can now re-subscribe to Pro and end up with
two subscriptions.

## Notes to reviewer

Tested by:

* [Manually](https://github.com/mysociety/alaveteli-professional/wiki/Manual-upgrade-with-3-month-trial) creating a trial for a user with a trial end of `Time.zone.now + 5.minutes`. Make sure the customer record doesn't have any cards (payment sources) attached.
* Wait for the trial to expire; then wait an **hour** for Stripe to finalise the invoice and attempt to bill.
* After the billing attempt, the sub should move to `past_due`.

Before this change:

* When accessing `/profile/subscriptions` you'd get [redirected with the message](https://github.com/mysociety/alaveteli/blob/develop/app/controllers/alaveteli_pro/subscriptions_controller.rb#L190): "You don't currently have an active Pro subscription"
* When accessing `/plans/pro` you'd see the payment form, rather than getting [redirected to `/profile/subscriptions`](https://github.com/mysociety/alaveteli/blob/develop/app/controllers/alaveteli_pro/plans_controller.rb#L42-L45) with a message.

Both of these flows work as expected with this commit.

### DIFF
--- a/app/models/alaveteli_pro/subscription.rb
+++ b/app/models/alaveteli_pro/subscription.rb
@@ -9,6 +9,10 @@ module AlaveteliPro
       status == 'active'
     end
 
+    def past_due?
+      status == 'past_due'
+    end
+
     def incomplete?
       status == 'incomplete'
     end
--- a/app/models/alaveteli_pro/subscription_collection.rb
+++ b/app/models/alaveteli_pro/subscription_collection.rb
@@ -33,6 +33,10 @@ module AlaveteliPro
       select(&:active?)
     end
 
+    def past_due
+      select(&:past_due?)
+    end
+
     def incomplete
       select(&:incomplete?)
     end
--- a/app/models/pro_account.rb
+++ b/app/models/pro_account.rb
@@ -22,7 +22,7 @@ class ProAccount < ApplicationRecord
   validates :user, presence: true
 
   def active?
-    subscriptions.active.any?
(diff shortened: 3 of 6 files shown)


garethrees:
@gbp would be good to get your thoughts at this point to make sure you think this is the right thing to do and won't have other side effects.

PR #5526, opened 2020-01-03
Auto-correct Layout/SpaceBeforeSemicolon

garethrees:
## What does this do?

Auto-correct Layout/SpaceBeforeSemicolon

## Why was this needed?

Code cleanup

### DIFF
--- a/spec/models/outgoing_message_spec.rb
+++ b/spec/models/outgoing_message_spec.rb
@@ -68,7 +68,7 @@ describe OutgoingMessage do
 
   describe '.expected_send_errors' do
     subject { described_class.expected_send_errors }
-    class TestError < StandardError ; end
+    class TestError < StandardError; end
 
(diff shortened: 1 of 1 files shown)


spec/models/outgoing_message_spec.rb
```diff
@@ -68,7 +68,7 @@
 
   describe '.expected_send_errors' do
     subject { described_class.expected_send_errors }
-    class TestError < StandardError ; end
+    class TestError < StandardError; end
```
gbp: Still has a semi colon! When defining empty classes I prefer:

```suggestion
    TestError = Class.new(StandardError)
```

garethrees: This cop is just checking that there's no _space_ before the semicolon, rather than the semicolon not existing at all.

Not sure about `TestError = Class.new(StandardError)` – but that's a question for another cop anyway.

PR #5531, opened 2020-01-08
Rename ProAccount#active? to ProAccount#subscription?

garethrees:
## What does this do?

* Consider all Stripe subscriptions "active"
* Rename ProAccount#active? to ProAccount#subscription?
* Replace conditionals with guard clauses

## Why was this needed?

Fixes Pro account access check regression.

Avoids confusion between "a user being subscribed to Pro" and a `Stripe::Subscription` with a status of `active`.

Minor code cleanup.

### DIFF
--- a/app/models/pro_account.rb
+++ b/app/models/pro_account.rb
@@ -21,7 +21,7 @@ class ProAccount < ApplicationRecord
 
   validates :user, presence: true
 
-  def active?
+  def subscription?
     subscriptions.any?
   end
 
--- a/app/models/pro_account.rb
+++ b/app/models/pro_account.rb
@@ -38,6 +38,7 @@ class ProAccount < ApplicationRecord
   def update_stripe_customer
     return unless feature_enabled?(:pro_pricing)
 
+    @subscriptions = nil unless stripe_customer
     @stripe_customer = stripe_customer || Stripe::Customer.new
 
     update_email
--- a/app/models/pro_account.rb
+++ b/app/models/pro_account.rb
@@ -22,7 +22,7 @@ class ProAccount < ApplicationRecord
   validates :user, presence: true
 
   def active?
-    subscriptions.active.any? || subscriptions.past_due.any?
+    subscriptions.any?
   end
 
   def subscriptions
--- a/app/models/pro_account.rb
+++ b/app/models/pro_account.rb
@@ -22,7 +22,7 @@ class ProAccount < ApplicationRecord
   validates :user, presence: true
 
   def subscription?
-    subscriptions.any?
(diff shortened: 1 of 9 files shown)


spec/controllers/alaveteli_pro/subscriptions_controller_spec.rb
```diff
@@ -189,6 +189,43 @@
...
+            source: stripe_helper.generate_card_token
+          )
+        end
+
+        let(:pro_account) do
+          user.create_pro_account(stripe_customer_id: customer.id)
+        end
+
+        it 'should cancel any incomplete subscriptions' do
+          # we can't create a subscription in the incomplete status so we have
```
garethrees: Could we do something similar to https://github.com/mysociety/alaveteli/blob/develop/lib/stripe_mock_patch.rb#L6-L10 ?

app/models/pro_account.rb
```diff
@@ -22,7 +22,7 @@ class ProAccount < ApplicationRecord
   validates :user, presence: true
 
   def subscription?
-    subscriptions.any?
+    subscriptions.current.any?
```
garethrees: 👍 nice

app/models/pro_account.rb
```diff
@@ -38,6 +38,7 @@ def stripe_customer
   def update_stripe_customer
     return unless feature_enabled?(:pro_pricing)
 
+    @subscriptions = nil unless stripe_customer
```
garethrees: Not 100% sure what the problem is without this – could you post a quick code example of a before/after?

gbp: This is the key to this solution.

After cancelling the incomplete subscriptions in the controller `@subscriptions` would be set for this ProAccount instance - it would be a SubscriptionCollection object which has been passed the current Stripe::Customer object. But at this point users might not have a Stripe::Customer object as this is before we add the card and the customer is save.

Without a customer in SubscriptionCollection#build attempts to create a Stripe::Subscription without a customer which obviously fails.

So setting to `nil` here allows SubscriptionCollection to be recreated with the new Stripe::Customer when ProAccount#subscriptions is next called.

spec/controllers/alaveteli_pro/subscriptions_controller_spec.rb
gbp: Possibly yes. Want me to give that a go?

garethrees: Worth having a stab at for the remaining hour of the day. If we can't get it working by then, I think we just go with this.

PR #5545, opened 2020-01-21
Update deployments

garethrees:


garethrees:
Had to fix this by:

* Resizing images to correct size (600x400, as per the others)
* Using absolute rather than relative URLs to `/assets`

PR #5549, opened 2020-01-24
Auto-correct Style/UnlessElse

garethrees:
## What does this do?

Auto-correct Style/UnlessElse

## Why was this needed?

Code cleanup

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -50,16 +50,14 @@ module ApplicationHelper
     end
 
     error_messages = "".html_safe
-    for object in objects
+
+    objects.each do |object|
       object.errors.each do |attr, message|
         error_messages << content_tag(:li, h(message))
       end
     end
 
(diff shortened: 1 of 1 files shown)


app/helpers/application_helper.rb
```diff
@@ -37,7 +37,9 @@ def foi_error_messages_for(*params)
     options = params.last.is_a?(Hash) ? params.pop.symbolize_keys : {}
     objects = params.collect { |object_name| instance_variable_get("@#{object_name}") }.compact
     count   = objects.inject(0) { |sum, object| sum + object.errors.count }
-    unless count.zero?
+    if count.zero?
```
gbp: Returning an empty string is no different to return nil. So we could use a guard clause here instead. 

```suggestion
    return if count.zero?
```

garethrees: Fixed in Use guard clause instead of if/else

PR #5556, opened 2020-01-30
Add STRIPE_TAX_RATE

garethrees:
## Relevant issue(s)

https://github.com/mysociety/asktheeu-theme/issues/77

## What does this do?

Allows setting a correct tax rate for Pro pricing plans.

## Why was this needed?

[Spanish VAT rate](https://www.avalara.com/vatlive/en/country-guides/europe/spain/spanish-vat-rates.html) (where AsktheEU are based) is 21% – we hard-coded to 20%.

## Implementation notes

Had to replace the constant as it was getting frozen in tests.

### DIFF
--- a/app/controllers/alaveteli_pro/subscriptions_controller.rb
+++ b/app/controllers/alaveteli_pro/subscriptions_controller.rb
@@ -50,7 +50,7 @@ class AlaveteliPro::SubscriptionsController < AlaveteliPro::BaseController
       @subscription = @pro_account.subscriptions.build
       @subscription.update_attributes(
         plan: params.require(:plan_id),
-        tax_percent: 20.0,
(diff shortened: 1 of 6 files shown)


app/controllers/alaveteli_pro/subscriptions_controller.rb
```diff
@@ -50,7 +50,7 @@ def create
       @subscription = @pro_account.subscriptions.build
       @subscription.update_attributes(
         plan: params.require(:plan_id),
-        tax_percent: 20.0,
+        tax_percent: BigDecimal(AlaveteliConfiguration.stripe_tax_rate).to_f,
```
gbp: Good catch

PR #5557, opened 2020-01-30
Remove hard-coded monthly billing assumptions

garethrees:
## Relevant issue(s)

https://github.com/mysociety/asktheeu-theme/issues/77

## What does this do?

Allows billing yearly instead of monthly.

## Why was this needed?

AskTheEU are going to use an annual plan.

## Implementation notes

Could extract the "per user, per INTERVAL" in to a helper, but the whole
page needs refactoring so will leave that for now in the interests of
getting this rolled out.

The marketing pages don't have access to the `Stripe::Plan`, so just
removed the interval from the text entirely. It's a little more clunky,
but still makes sense without lots of hassle.

## Screenshots

**BEFORE**

Screenshot 2020-01-30 at 15 04 29

**AFTER**

Screenshot 2020-01-30 at 15 04 46

### DIFF
--- a/app/views/alaveteli_pro/pages/marketing_roles/campaigners/_pricing.html.erb
+++ b/app/views/alaveteli_pro/pages/marketing_roles/campaigners/_pricing.html.erb
@@ -7,7 +7,7 @@
     <div class="marketing__section__content-items">
       <ul class="marketing__section__feature-list">
         <li>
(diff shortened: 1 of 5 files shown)


gbp:
Added to hotfix 0.37.0.1 which has been merged into master & develop.

PR #5558, opened 2020-01-30
Fix issues creating InfoRequestEvent factories

gbp:
## What does this do?
Fix issues creating InfoRequestEvent factories

## Why was this needed?
Without this change we couldn't correctly set InfoRequest on nested
association for OutgoingMessage, IncomingMessage and Comment, such as:

```ruby
  e = FactoryBot.build(:sent_event, info_request: info_request)
  e.outgoing_message.info_request == info_request #=> false
```

But now this would evaluate as true.

## Implementation notes
I've corrected the one place I've found within others factories where this is an issue. There may well be specs which might need updating for this, lets see what colour CI is...

### DIFF
--- a/spec/factories/outgoing_messages.rb
+++ b/spec/factories/outgoing_messages.rb
@@ -22,6 +22,7 @@ FactoryBot.define do
   factory :outgoing_message do
     info_request
     prominence { 'normal' }
+    last_sent_at { 2.weeks.ago }
 
     factory :initial_request do
       transient do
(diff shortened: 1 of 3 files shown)


spec/factories/outgoing_messages.rb
```diff
@@ -22,6 +22,7 @@
   factory :outgoing_message do
     info_request
     prominence { 'normal' }
+    last_sent_at { 2.weeks.ago }
```
garethrees: This seems unrelated to the fact that events aren't getting created?

Might be worth separating out as its own commit, or at least describing why we also need this in this commit msg.

gbp: Yeah just noticed this and trying to remember why this is necessary. Will separate out in to its own commit.

PR #5559, opened 2020-01-31
[#eu103] Respect the locale an InfoRequestBatch was created in when sending

garethrees:
## Relevant issue(s)

Fixes https://github.com/mysociety/asktheeu-theme/issues/103

## What does this do?

* Sends batch requests in the locale they were made under (some assumptions made – see 038813c)
* Replace `User#get_locale` with `User#locale`
* Minor cleanup

## Why was this needed?

The authority name was not being interpolated in to the outgoing messages of batch requests if made in a non-default locale.

## Implementation notes

Some assumptions made – see 038813c

## Screenshots

**BEFORE**

Screenshot 2020-01-30 at 17 03 02

**AFTER**

Screenshot 2020-01-31 at 09 10 54

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -359,8 +359,14 @@ class User < ApplicationRecord
     end
   end
 
+  def locale
+    (super || AlaveteliLocalization.locale).to_s
+  end
+
   def get_locale
-    (locale || AlaveteliLocalization.locale).to_s
+    warn %q([DEPRECATION] User#get_locale will be removed in 0.38.
+            It has been replaced by User#locale).squish
+    locale
   end
 
   def name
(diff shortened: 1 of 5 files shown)


app/models/user.rb
```diff
@@ -359,8 +359,14 @@ def reindex_referencing_models
     end
   end
 
+  def locale
+    (super || AlaveteliLocalization.locale).to_s
+  end
+
   def get_locale
-    (locale || AlaveteliLocalization.locale).to_s
+    warn %q([DEPRECATION] User#get_locale will be removed in 0.37.
```
garethrees: Actually this needs to be `0.38` – `0.37` will introduce the deprecation. Will fixup.

garethrees:
> Great stuff. Well done on fixing this.
> 
> Aside: could drop [b0cc8d5](https://github.com/mysociety/alaveteli/commit/b0cc8d563b4c2bb9e6d6944909589d79348d66ed) as I don't think this method has any really changes to it. Up to you

Oh yeah, I was initially going to add the `with_locale` block at that point, but in the end opted to isolate it to the main entry-point of the background task to reduce any unintended consequences.

PR #5563, opened 2020-02-03
Set money gem rounding mode

gbp:
## What does this do?

Set money gem rounding mode

## Why was this needed?

To remove deprecation warning from spec output

## Implementation notes

Using the new default of ROUND_HALF_UP
> Indicates that digits >= 5 should be rounded up, others rounded down

Instead of ROUND_HALF_EVEN
> Round towards the even neighbor

### DIFF
--- a/config/initializers/money.rb
+++ b/config/initializers/money.rb
@@ -1,3 +1,7 @@
 # Disable i18n which requires config/locales/*.yml files. Instead this allows
 # the Money gem to format currencies
 Money.locale_backend = :currency
+
+# Use the new default round mode to get rid of deprecation warnings - this can
+# be remove when upgrading the next major release of the Money gem
+Money.rounding_mode = BigDecimal::ROUND_HALF_UP


config/initializers/money.rb
```diff
@@ -1,3 +1,7 @@
 # Disable i18n which requires config/locales/*.yml files. Instead this allows
 # the Money gem to format currencies
 Money.locale_backend = :currency
+
+# Use the new defau8lt round mode to get rid of deprecation warnings- this can
```
garethrees: `s/defau8lt/default/`

config/initializers/money.rb
```diff
@@ -1,3 +1,7 @@
...
+# be remove when upgrading the next major release of the Money gem
+Money.rounding_mode = BigDecimal::ROUND_HALF_UP
```
garethrees: Are we absolutely sure on this? AIUI our current rounding is correct as per https://makandracards.com/makandra/1505-invoices-how-to-properly-round-and-calculate-totals.

This PR doesn't link to the upstream documentation or information about the change, so its a bit hard to assess whether to give it a 👍.

gbp: There doesn't seem to much conversation upstream about this change, I've found one brief explanation [1] its also worth noting Stripe uses `ROUND_HALF_UP` so I think this change it is good.

Either way our current usage is as such that this doesn't have any effect on us.

[1] https://github.com/RubyMoney/money/pull/883#issuecomment-562847214
[2] https://support.stripe.com/questions/rounding-rules-for-stripe-fees

PR #5564, opened 2020-02-06
[pro#360] Zip download of batch requests

gbp:
## Relevant issue(s)

Depends on #5558
Fixes https://github.com/mysociety/alaveteli-professional/issues/360

## What does this do?

Adds ability for pro users to downloads an zip archive of their batch requests

## Implementation notes

Using a live streaming zip file approach we may need to run this in the background if the performance on production isn't good.

### DIFF
--- a/app/services/info_request_batch_zip.rb
+++ b/app/services/info_request_batch_zip.rb
@@ -1,0 +1,106 @@
+##
+# This service streams a ZIP file with all incoming/ outgoing messages and
+# attachments for each request for a given info request batch
+#
+class InfoRequestBatchZip
+  include DownloadHelper
+  include Enumerable
+
+  ZippableFile = Struct.new(:path, :body)
+
+  attr_reader :info_request_batch
+
+  def initialize(info_request_batch)
+    @info_request_batch = info_request_batch
+  end
+
+  def files
+    to_a
+  end
+
+  def name
+    generate_download_filename(
+      resource: 'batch',
+      id: info_request_batch.id,
+      title: info_request_batch.title,
+      type: 'export',
+      ext: 'zip'
+    )
+  end
+
+  def stream(&chunks)
+    block_writer = ZipTricks::BlockWrite.new(&chunks)
+
+    ZipTricks::Streamer.open(block_writer) do |zip|
+      each do |file|
+        zip.write_deflated_file(file.path) { |writer| writer << file.body }
+      end
+    end
+  end
+
+  private
+
+  def each(&block)
+    to_enum(:each) unless block_given?
+
+    yield prepare_dashboard_metrics
+
+    info_request_events.each do |event|
+      if event.outgoing?
+        yield prepare_outgoing_message(event.outgoing_message)
+      elsif event.response?
+        yield prepare_incoming_message(event.incoming_message)
+
+        event.incoming_message.foi_attachments.each do |attachment|
+          yield prepare_foi_attachment(attachment)
+        end
+      end
+    end
+  end
+
+  def info_request_events
+    InfoRequestEvent.where(info_request: info_request_batch.info_requests).
+      includes(:info_request, info_request: [:public_body]).
+      includes(:outgoing_message).
+      includes(:incoming_message, incoming_message: [:foi_attachments])
+  end
+
+  def prepare_dashboard_metrics
+    metrics = InfoRequestBatchMetrics.new(info_request_batch)
+    ZippableFile.new(metrics.name, metrics.to_csv)
+  end
+
+  def prepare_outgoing_message(message)
+    sent_at = message.last_sent_at.to_formatted_s(:filename)
+    name = "outgoing_#{message.id}.txt"
+    path = [base_path(message.info_request), sent_at, name].join('/')
+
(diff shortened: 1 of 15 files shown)


app/services/info_request_batch_zip.rb
```diff
@@ -0,0 +1,98 @@
+##
+# This service streams a ZIP file with all incoming/ outgoing messages and
```
garethrees: I'm not sure I'd call this a "service". To me, a service should be more like a [command method](https://www.martinfowler.com/bliki/CommandQuerySeparation.html) – i.e. it does some work and returns `nil` (or maybe `true` or `false` as an indicator of success).

This feels like a legitimate object – it _is a_ Zip. Personally I'd make this `InfoRequestBatch::Zip`.

We don't really have a good precedent on this though, so this is more of a suggestion/comment – happy to keep it as it is for now.

app/services/info_request_batch_zip.rb
```diff
@@ -0,0 +1,98 @@
...
+      includes(:outgoing_message).
+      includes(:incoming_message, incoming_message: [:foi_attachments])
+  end
+
+  def prepare_dashboard_metrics
+    metrics = InfoRequestBatchMetrics.new(info_request_batch)
+    ZippableFile.new(metrics.name, metrics.to_csv)
+  end
+
+  def perpare_outgoing_message(message)
```
garethrees: Typo `s/perpare/prepare/` – applies to all except `prepare_dashboard_metrics`

app/services/info_request_batch_zip.rb
```diff
@@ -0,0 +1,98 @@
...
+  def perpare_foi_attachment(attachment)
+    message = attachment.incoming_message
+    sent_at = message.sent_at.to_formatted_s(:filename)
+
+    path = [
+      base_path(message.info_request),
+      sent_at,
+      'attachments',
+      attachment.filename
+    ].join('/')
```
garethrees: Probably alright to do this, but should we be using [`Pathname`](https://ruby-doc.org/stdlib-2.5.3/libdoc/pathname/rdoc/Pathname.html) for this kind of thing? Don't feel super strongly.

app/views/alaveteli_pro/info_request_batches/_info_request_batch.html.erb
```diff
@@ -69,6 +69,13 @@
 
   <div class="request__download">
 
+    <%= link_to alaveteli_pro_info_request_batch_batch_download_path(
```
garethrees: Going to leave this out initially so that we can manually test

app/services/info_request_batch_zip.rb
```diff
@@ -0,0 +1,98 @@
...
+      ext: 'zip'
+    )
+  end
+
+  def stream(&chunks)
+    block_writer = ZipTricks::BlockWrite.new(&chunks)
+
+    ZipTricks::Streamer.open(block_writer) do |zip|
+      files.each do |file|
+        zip.write_deflated_file(file.path) { |writer| writer << file.body }
```
garethrees: At the moment, this isn't initiating the download until most of the work is done. `files` is doing all of the database querying and only _then_ do we start to stream each file. On a large batch this means a ~30 second wait until we get an indication that something is happening.

You can see this happening in the logs with:

```diff
 files.each do |file|
+  ::Rails.logger.info "WRITING FILE: #{file.path}"
   zip.write_deflated_file(file.path) { |writer| writer << file.body }
 end
```

Instead, we want to initiate the download straight away. Here's an example:

```ruby
  ZipTricks::Streamer.open(block_writer) do |zip|
      (1..100).each do |i|
         zip.write_deflated_file(i.to_s) do |file_writer|
           ::Rails.logger.info "WRITING FILE: #{i}"
           file_writer << (i.to_s * 1_000_000)
         end
      end
  end
```

This will illustrate the download initiating, and then writing a file at a time as the Zip continues to download.

For our app code, I think we want something more like this. **WARNING** this does start the download more immediately, but for some reason it doesn't write incoming messages or attachments to the Zip, so take care to test that it works.

```ruby
    ZipTricks::Streamer.open(block_writer) do |zip|
      file = prepare_dashboard_metrics
      zip.write_deflated_file(file.path) { |writer| writer << file.body }

      info_request_events.each do |event|
        if event.outgoing?
          file = perpare_outgoing_message(event.outgoing_message)
          ::Rails.logger.info "WRITING FILE: #{file.path}"
          zip.write_deflated_file(file.path) { |writer| writer << file.body }
        elsif event.response?
          file = perpare_incoming_message(event.incoming_message)
          ::Rails.logger.info "WRITING FILE: #{file.path}"
          zip.write_deflated_file(file.path) { |writer| writer << file.body }
          event.incoming_message.foi_attachments.each do |attachment|
            file = perpare_foi_attachment(attachment)
            ::Rails.logger.info "WRITING FILE: #{file.path}"
            zip.write_deflated_file(file.path) { |writer| writer << file.body }
          end
        end
      end
```

garethrees: Resolved by 06b418a

PR #5588, opened 2020-03-12
[pro#360] Fixes for Zip download of batch requests

### LINKED ISSUE #5587: [ERROR] ZipTricks::PathSet::Conflict occurred in batch_downloads#show
After deploying #5564 behind a feature flag we've seen some errors when testing

```
A ZipTricks::PathSet::Conflict occurred in batch_downloads#show:

  The file at "authority - request_url_title/timestamp/attachments/attachment.html" has already been included
in the archive. Adding it the second time would cause
the first file to be overwritten during unarchiving, and
could also get the archive flagged as invalid.

  app/services/info_request_batch_zip.rb:36:in `block (2 levels) in stream'
```

gbp:
## Relevant issue(s)

Connected to mysociety/alaveteli-professional#360
Fixes #5587 

## What does this do?

1. Checks messages can be read before allowing them to be exported
2. Prevents exporting message body as an attachment
3. Ensures attachments filepaths are unique

## Why was this needed?

In order to get the feature shipped

### DIFF
--- a/app/services/info_request_batch_zip.rb
+++ b/app/services/info_request_batch_zip.rb
@@ -8,10 +8,11 @@ class InfoRequestBatchZip
 
   ZippableFile = Struct.new(:path, :body)
 
-  attr_reader :info_request_batch
+  attr_reader :info_request_batch, :ability
 
-  def initialize(info_request_batch)
+  def initialize(info_request_batch, ability:)
     @info_request_batch = info_request_batch
+    @ability = ability
   end
 
   def files
--- a/app/services/info_request_batch_zip.rb
+++ b/app/services/info_request_batch_zip.rb
@@ -34,7 +34,10 @@ class InfoRequestBatchZip
   def stream(&chunks)
     block_writer = ZipTricks::BlockWrite.new(&chunks)
 
-    ZipTricks::Streamer.open(block_writer) do |zip|
+    ZipTricks::Streamer.open(
+      block_writer,
+      auto_rename_duplicate_filenames: true
+    ) do |zip|
       each do |file|
         zip.write_deflated_file(file.path) { |writer| writer << file.body }
       end
--- a/app/services/info_request_batch_zip.rb
+++ b/app/services/info_request_batch_zip.rb
@@ -10,6 +10,8 @@ class InfoRequestBatchZip
 
   attr_reader :info_request_batch, :ability
 
+  delegate :cannot?, to: :ability
+
   def initialize(info_request_batch, ability:)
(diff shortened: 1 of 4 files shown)


app/services/info_request_batch_zip.rb
```diff
@@ -106,8 +107,8 @@ def prepare_foi_attachment(attachment)
     path = [
       base_path(message.info_request),
       sent_at,
-      'attachments',
-      attachment.filename
+      "attachments-#{message.id}",
+      "#{idx}_#{attachment.filename}"
```
garethrees: I was hoping to avoid manipulating the attachment filenames. Its really confusing when they're different to what we show in the app.

How about doing something like:

```ruby
# …
begin
  attachment = prepare_foi_attachment(attachment)
rescue ZipTricks::PathSet::Conflict
  attachment = prepare_foi_attachment_with_part_namespace(attachment)
end
# …

def prepare_foi_attachment_with_namespace(attachment)
  # …
    path = [
      base_path(message.info_request),
      sent_at,
      "attachments-#{message.id}",
      attachment.url_part_number,
      attachment.filename
    ].join('/')
  # …
end
```

That way we mostly just have

```
attachments-123/
  foo.txt
  bar.txt
```

but in the odd case where there are duplicates, we get

```
attachments-123/
  1/
    foo.txt
  2/
    foo.txt
```

gbp: I think with this suggestion, we would actually get 
```
attachments-123/
  foo.txt
  2/
    foo.txt
```
as the first attachment wouldn't raise an error.

garethrees: Ah, yeah, maybe the implementation needs a little bit of thinking, but the point still stands.

If it ends up being too hard, maybe just use `url_part_number` as a prefix rather than a random ID.

gbp: Discovered an option in ZipTricks which handles automatic renaming of duplicate files:

With 54d809cabea7490b2f558d43bed5b85e74894bb0 we would get:
```
attachments-123/
  foo.txt
  foo (2).txt
```

@garethrees how is that?

garethrees: Oh nice! Yeah, that keeps things simple. Seems like a fairly uncommon case so lets just use this for now given its near zero effort.

gbp: BTW can be tested locally with an existing email fixture file `./script/mailin < spec/fixtures/files/incoming-request-attach-attachments.email`

PR #5591, opened 2020-03-17
Remove cookie storage in GA

gbp:
## Relevant issue(s)

https://github.com/mysociety/orgsites/issues/1124

## What does this do?

Remove cookie storage from analytics

## Why was this needed?

Team meeting decision to remove individual tracking 

## Notes to reviewer

Requires an update to WDTK theme privacy pages (EN & CY)

### DIFF
--- a/doc/CHANGES.md
+++ b/doc/CHANGES.md
@@ -5,6 +5,19 @@
 * Don't show profile photos of banned users (Gareth Rees)
 * Link to change request form when asking users contact us about request email
   updates (Gareth Rees)
+* Configure Google Analytics to anonymise their IP address and disable storing
+  browser cookies. (Graeme Porteous)
+
+# Upgrade Notes
+
+* The Google Analytics code has been extracted into
+  `views/application/_ga_code.html.erb`, and to avoid the GDPR requirement to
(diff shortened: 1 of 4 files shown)


doc/CHANGES.md
```diff
@@ -1,3 +1,16 @@
+# develop
+
+## Highlighted Features
+
+* Configure Google Analytics to anonymise their IP address and disable storing
+  browser cookies. (Graeme Porteous)
+
+# Upgrade Notes
+
+* Google Analytics code was extract into `views/application/_ga_code.html.erb`,
```
garethrees, line 10: s/extract/extracted/

I think we should rephrase this a little:

```
The Google Analytics code has been extracted into `views/application/_ga_code.html.erb`, and to avoid the GDPR requirement to show cookie opt-in notices, now sets cookie-less tracking. Additionally, we are instructing Google to anonymise visitors IP address using the `anonymizeIp` option. This means that visitors are no longer individually profiled. If you wish to continue individually targeting users, you'll need to override `app/views/application/_ga_code.html.erb` in your theme and set appropriate options.
```

gbp: That reads much better

PR #5592, opened 2020-03-20
Apply patch for CVE-2020-5267

garethrees:
> [CVE-2020-5267] Possible XSS vulnerability in ActionView
>
> There is a possible XSS vulnerability in ActionView's JavaScript
> literal escape helpers. Views that use the `j` or `escape_javascript`
> methods may be susceptible to XSS attacks.
>
> – https://groups.google.com/forum/#!topic/rubyonrails-security/55reWMM_Pg8

We use `j` in `app/views/general/_localised_datepicker.html.erb` for
displaying date pickers on the request search form.

Only Rails 5.2+ has received gem bumps.

The Rails 5.1 [1] and 5.2.4.1 [2] (the version before the patch)
implementations of `escape_javascript` are identical, so this patch
should be fine.

The commit message of the patch states:

> This commit escapes dollar signs and backticks to prevent JS XSS
> issues when using the `j` or `javascript_escape` helper
>
> – https://git.io/JvMcm

[1] https://git.io/JvMcI
[2] https://git.io/JvMcL


garethrees:
> I've pushed a fixup as discussed. This:
> 
> 1. wraps the patch in `unless rails_upgrade?` conditional - this will be obvious that it can be removed when we upgrade next

Thanks, sorry, I keep forgetting about this :(

Note that these commits only got pushed to GitHub, rather than our internal git servers.

> 2. inlines the JS escape map values
> 3. removes `alias` methods as we don't need to retain the old broken implementation
> 4. fixed Rubocop warnings (module documentation and line length).

I've merged these in, but in general I tend to apply patches without modification, given they're usually temporary and I don't want to risk breaking them.

PR #5593, opened 2020-03-23
[fnf#11] Add data models for projects feature

gbp:
## Relevant issue(s)

https://github.com/mysociety/transparency-fnf-whatdotheyknow-projects/issues/11

## What does this do?

Add initial data models for new projects feature.

## Implementation notes

For the project membership I originally attempted to use Rolify's `resourcify` class method and related methods but in the end it worked out easier to create a separate ProjectMembership model. We're not using currently using the Role#resource polymorphic association for an other role/resource so it seems sensible to keep separate and the Role model simpler.

### DIFF
--- a/app/models/project.rb
+++ b/app/models/project.rb
@@ -26,4 +26,19 @@ class Project < ApplicationRecord
            source_type: 'InfoRequestBatch'
 
   validates :title, :owner, presence: true
+
+  def info_requests
+    InfoRequest.
+      joins(
+        "LEFT JOIN project_resources r1 ON " \
+        "r1.resource_id = info_requests.id AND " \
+        "r1.resource_type = 'InfoRequest'"
+      ).
+      joins(
+        "LEFT JOIN project_resources r2 ON " \
+        "r2.resource_id = info_requests.info_request_batch_id AND " \
+        "r2.resource_type = 'InfoRequestBatch'"
+      ).
+      where("r1.project_id = :id OR r2.project_id = :id", id: id)
+  end
 end
--- a/app/models/project.rb
+++ b/app/models/project.rb
@@ -15,5 +15,15 @@ class Project < ApplicationRecord
   has_one  :owner, through: :owner_membership, source: :user
   has_many :contributors, through: :contributor_memberships, source: :user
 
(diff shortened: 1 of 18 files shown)


app/models/role.rb
```diff
@@ -30,12 +30,13 @@ class Role < ApplicationRecord
 
   ROLES = ['admin'].freeze
   PRO_ROLES = ['pro', 'pro_admin'].freeze
+  PROJECT_ROLES = %w[project_owner project_contributor].freeze
 
   def self.allowed_roles
```
garethrees: I don't think we want the `PROJECT_ROLES` here:

Screenshot 2020-03-25 at 14 27 36

app/models/project.rb
```diff
@@ -15,5 +15,15 @@ class Project < ApplicationRecord
   has_one  :owner, through: :owner_membership, source: :user
   has_many :contributors, through: :contributor_memberships, source: :user
 
+  has_many :resources, class_name: 'ProjectResource'
+  has_many :requests,
```
garethrees: What do you think about the deviation of naming here?

I'm wondering whether we should do:

```ruby
has_many :info_requests
# …
has_many :info_request_batches
#…
```

And then to get all requests…

```ruby
def all_info_requests
  # …
```

I'm in two minds. On minus side for the current implementation, we have to remember `requests` and `batches` which is unusual. On the plus side, we _don't_ have to remember that there's a gotcha when you just want to list all the info requests, irrespective of how they were made.

I don't think I feel strongly either way, so I guess my default is to say keep it as is, but just wanted to pose the question to see what you think.

app/models/role.rb
gbp: Oh yes, I was aware of this and have stash somewhere which sorts this - will open a PR

garethrees: Would be good to get this PR up to enable this merge

gbp: Is up here: https://github.com/mysociety/alaveteli/pull/5603

app/models/project.rb
gbp: No strong feelings on this. I think we go for what is there now and see how it works. Easy to change if we find it confusing.

PR #5601, opened 2020-04-01
[#5171] Hide banned users photo

### LINKED ISSUE #5171: Do not display profile photo for banned users
Most users we ban are spam, so lets not show the profile photo. That way we don't need need to do the extra step of removing it (time saver), and it helps us build an archive of what a spam account looks like.

For banned users that aren't spam, they'll retain their profile photo and it'll be displayed again when their ban is lifted.

garethrees:
## Relevant issue(s)

Fixes #5171

## What does this do?

* Hides profile photos of banned users
* Bit of refactoring

## Why was this needed?

* So that we don't show spam profile photos

## Screenshots

**ACTIVE SPAMMER**

Screenshot 2020-04-01 at 19 19 32

Screenshot 2020-04-01 at 19 22 26

**BANNED SPAMMER**

Screenshot 2020-04-01 at 19 21 57

Signed in, the spammer sees that they can change the profile photo, but we [prevent them actually doing that](https://github.com/mysociety/alaveteli/blob/develop/app/controllers/user_controller.rb#L268) later on. I don't think showing the link at this point is a problem.

Screenshot 2020-04-01 at 19 22 48

Non-logged in:

Screenshot 2020-04-01 at 19 22 56

## Notes to reviewer

Tried adding a profile photos factory but for some reason it kept [erroring on file format](https://github.com/mysociety/alaveteli/blob/e13195da1fbae55fefcf8157f9235b7a0809ba8e/app/models/profile_photo.rb#L79), even though I was just doing `data { load_file_fixture('parrot.png') }`. No big deal for now though IMO.

### DIFF
--- a/spec/models/user_spec.rb
+++ b/spec/models/user_spec.rb
@@ -1785,4 +1785,32 @@ describe User do
 
   end
 
+  describe '#show_profile_photo?' do
+    subject { user.show_profile_photo? }
+
+    context 'with a profile_photo' do
+      let(:user) { FactoryBot.create(:user) }
+
+      before do
+        user.create_profile_photo!(data: load_file_fixture('parrot.png'))
+      end
+
+      it { is_expected.to be_truthy }
+    end
+
+    context 'with a profile photo and banned' do
+      let(:user) { FactoryBot.create(:user, :banned) }
+
+      before do
+        user.create_profile_photo!(data: load_file_fixture('parrot.png'))
+      end
+
+      it { is_expected.to be_falsey }
+    end
+
+    context 'without a profile_photo' do
+      let(:user) { FactoryBot.build(:user) }
+      it { is_expected.to be_falsey }
+    end
+  end
 end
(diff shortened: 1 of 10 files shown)


spec/models/user_spec.rb
```diff
@@ -1785,4 +1785,27 @@ def create_user(options = {})
...
+    end
+
+    context 'without a profile_photo' do
+      let(:user) { FactoryBot.build(:user) }
+      it { is_expected.to be_falsey }
+    end
+
+    context 'when banned' do
+      let(:user) { FactoryBot.build(:user, :banned) }
+      it { is_expected.to be_falsey }
```
gbp, line 1808: This spec could be passing because the banned user doesn't have a profile photo. We should add the before block used in the 'with a profile_photo' context.

MattK1234:
Just noting thanks for implementing this, definitely helps admins when banning spammers in bulk :-)

PR #5602, opened 2020-04-02
[#5542] Share by Link - attachments

### LINKED ISSUE #5542: Share private requests by link
Fixes https://github.com/mysociety/alaveteli-professional/issues/188

Fixes https://github.com/mysociety/transparency-adessium/issues/7

We want a simple way for Pro users to share their private requests with trusted colleagues, collaborators and friends.

There's an "advanced" method of doing this (https://github.com/mysociety/alaveteli/issues/3056), but that's not what we're going for here.

For now, the goal is to be able to view a private request by visiting a "private" URL for the request, like how private [GitHub Gists](https://gist.github.com/) work. We don't have budget for fully blown team access; This is solving _an_ issue though, e.g. intern makes FOI requests; shares responses with person writing the actual story.

## The UI

We'll add a new `share-private-url` `sidebar__section` to the private request sidebar:

share-by-private-url

To minimise the noise here, it'll initially just contain "Share with private link". Clicking that will expand the section to reveal the private link, and a short explainer of the feature. We'll want to be clear that **anyone with the URL** can view the request.

## The URL

I don't have a strong preference on URL format, as long as it looks clean.

I was thinking we'd use `/r/:token` (request), like our `/c/:token` (confirm). We may be able to do it via the existing /c/ routes – maybe a post redirect that sets an access token for a particular request?

I _do_ think that the URL should mask the content in some way though – we shouldn't reveal the request title in the URL.

## Access permissions

If the private URL for a request is https://www.whatdotheyknow.com/r/773f0f28bf40068fd4582c8a58100211, visiting that URL should:

* allow signed in pros to view the request and attachments
* allow signed in admins to view the request and attachments
* allow signed in free users to view the request and attachments
* allow non-signed in visitors to view the request and attachments

Visitors shouldn't be allowed to take any actions on the request (e.g. commenting, classifying, etc). They shouldn't be allowed to browse the request owners other requests either.

We'll want to be careful that the private URL can't be calculated from the publicly viewable attributes of the request.

The pages shouldn't be indexable if someone posts a link somewhere. [Looks like](https://www.deepcrawl.com/blog/best-practice/noindex-disallow-nofollow/) we can use `rel="noindex"` or a `X-Robots-Tag: noindex` header.

gbp:
## Relevant issue(s)

Depends on #5597 
Depends on #5580 
Fixes #5542 

## What does this do?

Expands on the work in the other PRs to allow attachments to be viewable when view an info request using the public token EG `/r/123ABC`

## Why was this needed?

Without this attempting to view an attachments would return a 404.

### DIFF
--- a/app/controllers/attachments_controller.rb
+++ b/app/controllers/attachments_controller.rb
@@ -3,12 +3,13 @@
 #
 class AttachmentsController < ApplicationController
   include FragmentCachable
+  include InfoRequestHelper
 
   around_action :cache_attachments
   before_action :authenticate_attachment
   before_action :authenticate_attachment_as_html, only: :show_as_html
 
-  helper_method :info_request, :incoming_message, :attachment, :attachment_url
+  helper_method :info_request, :incoming_message, :attachment
 
   def show
     # Prevent spam to magic request address. Note that the binary
@@ -39,7 +40,7 @@ class AttachmentsController < ApplicationController
 
     html = attachment.body_as_html(
       image_dir,
-      attachment_url: Rack::Utils.escape(attachment_url),
+      attachment_url: Rack::Utils.escape(attachment_url(attachment)),
       content_for: {
         head_suffix: render_to_string(
           partial: 'request/view_html_stylesheet',
@@ -151,15 +152,6 @@ class AttachmentsController < ApplicationController
     else
       filename
     end
-  end
-
-  def attachment_url
-    get_attachment_url(
-      id: incoming_message.info_request_id,
-      incoming_message_id: incoming_message.id,
-      part: part_number,
-      file_name: original_filename
-    )
   end
 
   def content_type
--- a/app/controllers/attachments_controller.rb
+++ b/app/controllers/attachments_controller.rb
@@ -1,23 +1,22 @@
+##
+# Controller to serve info request message attachments in both raw and as HTML.
+#
 class AttachmentsController < ApplicationController
   include FragmentCachable
 
-  before_action :authenticate_attachment
   around_action :cache_attachments
+  before_action :authenticate_attachment
+  before_action :authenticate_attachment_as_html, only: :show_as_html
+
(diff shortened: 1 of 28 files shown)


gbp:
Note commits from https://github.com/mysociety/alaveteli/pull/5602/commits/12d28bf6cf6dd964296ab48692d1e0f1bf190910 are unique to this PR, the preceding ones are from the dependent PRs

app/controllers/attachments_controller.rb
```diff
@@ -60,7 +61,11 @@ def show_as_html
   private
 
   def info_request
-    @info_request ||= InfoRequest.find(params[:id])
+    if public_token
```
garethrees, line 64: Could remove the duplication in assignment here, though perhaps a little less readable?

```ruby
  @info_request ||= begin
    InfoRequest.find_by!(public_token: public_token) if public_token
    InfoRequest.find(params[:id])
  end
```

Alternatively could do:

```ruby
@info_request ||=
  if public_token
    InfoRequest.find_by!(public_token: public_token)
  else
    InfoRequest.find(params[:id])
  end
```
Don't feel strongly about any of these though – just suggestions.

garethrees: We don't have any specs for accessing attachments via a `public_token` though – I think we should.

app/helpers/info_request_helper.rb
```diff
@@ -271,7 +271,13 @@ def attachment_path(attachment, options = {})
   def attachment_url(attachment, options = {})
     attach_params = attachment_params(attachment, options)
     if options[:html]
-      get_attachment_as_html_path(attach_params)
+      if public_token?
```
garethrees, line 274: I think we should write some specs for these

garethrees:
Superseded by https://github.com/mysociety/alaveteli/pull/5608

PR #5608, opened 2020-04-08
[#5542] Share by link

gbp:
## Relevant issue(s)

Supersedes #5580 
Supersedes #5602 
Fixes #5542 

## What does this do?

Add request sidebar section with URL to allow sharing of requests to other users. This allows embargoed requests to be viewed by other people with the link.

## Why was this needed?

We want a simple way for Pro users to share their private requests with trusted colleagues, collaborators and friends.

## Implementation notes

1. To make the CanCan abilities easier to manage this overrides the `current_ability` method in the new controller. I think overall doing this saves having to duplicating/rewriting the correspondence rendering views to check different abilities for the standard request show action and the new action.
2. We might want to improve the frontend with a button to automatic copy the URL to the users clipboard

## Screenshots

image
image

PR #5612, opened 2020-04-09
Refactor embargo forms

gbp:
## Relevant issue(s)

Connected to issue with #5608

## What does this do?

Don't initialise new instances of Embargo or EmbargoExtenstion in forms as we might check for the presence of these associations later in the view rendering.

This happens a lot at the moment with CanCanCan abilities checking if an info request is public or private by checking `info_request.embargo`.

## Implementation notes

Due to `info_request.embargo` now being `nil` I had to allow for this in the `embargo_extension_options` helper - this also needed specs as it wasn't covered initially.

### DIFF
--- a/app/helpers/alaveteli_pro/info_requests_helper.rb
+++ b/app/helpers/alaveteli_pro/info_requests_helper.rb
@@ -16,7 +16,7 @@ module AlaveteliPro::InfoRequestsHelper
   end
 
   def embargo_extension_options(embargo)
-    options = embargo_options_from_date(embargo.publish_at)
+    options = embargo_options_from_date(embargo&.publish_at || Date.today)
     options.unshift([_('Choose a duration'), ''])
   end
 
--- a/app/views/alaveteli_pro/info_requests/_embargo_extension_form.html.erb
+++ b/app/views/alaveteli_pro/info_requests/_embargo_extension_form.html.erb
@@ -1,7 +1,8 @@
 <%= form_for(
-      [AlaveteliPro::EmbargoExtension.new(embargo: info_request.embargo)],
+      :alaveteli_pro_embargo_extension,
+      url: alaveteli_pro_embargo_extensions_path,
       html: {class: 'js-embargo-form'}) do |f| %>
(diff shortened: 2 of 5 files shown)


app/views/alaveteli_pro/info_requests/_embargo_extension_form.html.erb
```diff
@@ -1,7 +1,8 @@
 <%= form_for(
-      [AlaveteliPro::EmbargoExtension.new(embargo: info_request.embargo)],
+      :alaveteli_pro_embargo_extension,
+      url: alaveteli_pro_embargo_extensions_path,
       html: {class: 'js-embargo-form'}) do |f| %>
-  <%= f.hidden_field :embargo_id %>
+  <%= f.hidden_field :embargo_id, value: info_request.embargo&.id %>
```
garethrees, line 5: I think we can rely on an embargo ID existing here, since [we check that there is an embargo](https://github.com/mysociety/alaveteli/blob/develop/app/views/alaveteli_pro/info_requests/_embargo_form.html.erb#L13-L15) before rendering this partial.

The controller action [requires an `embargo_id` param](https://github.com/mysociety/alaveteli/blob/develop/app/controllers/alaveteli_pro/embargo_extensions_controller.rb#L10-L11) too, so I think this adds a little confusion.

app/helpers/alaveteli_pro/info_requests_helper.rb
```diff
@@ -16,7 +16,7 @@ def publish_at_options
   end
 
   def embargo_extension_options(embargo)
-    options = embargo_options_from_date(embargo.publish_at)
+    options = embargo_options_from_date(embargo&.publish_at || Date.today)
```
garethrees, line 19: I wouldn't do anything about this now as its fine as is, but worth noting that this would have been better as two commits.

1. Allow `InfoRequestsHelper#embargo_extension_options` to accept a `nil` embargo
2. Don't initialise new instances of `Embargo` or `EmbargoExtenstion` in forms

I find that an easier trail of thought to follow than a single "Refactor X" commit.

PR #5615, opened 2020-04-15
Further refactoring of embargo_create_form

garethrees:
## What does this do?

Further refactoring of `embargo_create_form` after https://github.com/mysociety/alaveteli/pull/5612.

## Why was this needed?

Reduce complexity.

## Implementation notes

4b43545 takes the approach of creating a new instance of `AlaveteliPro::Embargo` to build the form, but there's an alternative option and I'm sure which I prefer.

```erb
<%= form_for(AlaveteliPro::Embargo.new, html: { class: 'js-embargo-form' }) do |f| %>
```

I like this because the URL is figured out automatically.

There's a second option here, since we're not actually using the instance, of doing:

```erb
<%= form_with(url: alaveteli_pro_embargoes_path, html: { class: 'js-embargo-form' }) do |f| %>
```

I like this because it's clearer that we're not relying on a local instance of anything. On the downside, it's not as resource oriented.

### DIFF
--- a/config/initializers/inflections.rb
+++ b/config/initializers/inflections.rb
@@ -9,8 +9,7 @@ ActiveSupport::Inflector.inflections(:en) do |inflect|
 #   inflect.singular /^(ox)en/i, '\1'
 #   inflect.irregular 'person', 'people'
 #   inflect.uncountable %w( fish sheep )
-  inflect.plural /^(embargo)$/i, '\1es'
-  inflect.singular /^(embargo)es/i, '\1'
+  inflect.irregular 'embargo', 'embargoes'
 end
 
 # These inflection rules are supported but not enabled by default:
--- a/app/helpers/alaveteli_pro/info_requests_helper.rb
+++ b/app/helpers/alaveteli_pro/info_requests_helper.rb
@@ -15,7 +15,7 @@ module AlaveteliPro::InfoRequestsHelper
     options.unshift([_('Publish immediately'), ''])
   end
 
-  def embargo_extension_options(embargo)
+  def embargo_extension_options(embargo = nil)
     options = embargo_options_from_date(embargo&.publish_at || Date.today)
     options.unshift([_('Choose a duration'), ''])
   end
--- a/app/views/alaveteli_pro/info_requests/_embargo_create_form.html.erb
+++ b/app/views/alaveteli_pro/info_requests/_embargo_create_form.html.erb
@@ -1,21 +1,23 @@
 <%= form_for(
       :alaveteli_pro_embargo,
       url: alaveteli_pro_embargoes_path,
-      html: {class: 'js-embargo-form'}) do |f| %>
+      html: { class: 'js-embargo-form' }) do |f| %>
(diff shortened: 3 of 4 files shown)


config/initializers/inflections.rb
```diff
@@ -11,6 +11,8 @@
 #   inflect.uncountable %w( fish sheep )
   inflect.plural /^(embargo)$/i, '\1es'
   inflect.singular /^(embargo)es/i, '\1'
+  inflect.plural /^(alaveteli_pro_embargo)$/i, '\1es'
+  inflect.singular /^(alaveteli_pro_embargo)es/i, '\1'
```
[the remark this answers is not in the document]
garethrees: Will clean these up on merge – might be able to make them a single regexp?

garethrees:
> Nice improvements 👍

Thanks! Any thoughts on the `form_for` vs `form_with` approach?

gbp:
> Thanks! Any thoughts on the `form_for` vs `form_with` approach?

I would keep with `form_for` for consistency for now.

config/initializers/inflections.rb
```diff
@@ -9,8 +9,8 @@
 #   inflect.singular /^(ox)en/i, '\1'
 #   inflect.irregular 'person', 'people'
 #   inflect.uncountable %w( fish sheep )
-  inflect.plural /^(embargo)$/i, '\1es'
-  inflect.singular /^(embargo)es/i, '\1'
+  inflect.plural(/^(alaveteli_pro_embargo|embargo)$/i, '\1es')
+  inflect.singular(/^(alaveteli_pro_embargo|embargo)es/i, '\1')
```
gbp, line 13:
```suggestion
  inflect.plural(/(embargo)$/i, '\1es')
  inflect.singular(/(embargo)es$/i, '\1')
```

Actually I think this does everything we need.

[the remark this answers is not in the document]
gbp: Or even just:
```suggestion
  inflect.irregular 'embargo', 'embargoes'
```

gbp: Or even just:
```suggestion
  inflect.irregular 'embargo', 'embargoes'
```

garethrees: Ah yeah good call! Fixed in 56f27ac – will merge on green.

gbp: Did you see my third suggestion: https://github.com/mysociety/alaveteli/pull/5615#discussion_r409661060

garethrees: I didn't – I think I prefer `plural` and `singular` because it doesn't feel as irregular as e.g. "person" vs "people" – its just that the `e` rule is slightly different. IDK – will keep as is for now I think 🤷 .

gbp: Well... https://github.com/rails/rails/blob/b2eb1d1c55a59fee1e6c4cba7030d8ceb524267c/activesupport/lib/active_support/inflections.rb#L66

PR #5617, opened 2020-04-16
[fnf#20] Projects home

garethrees:
## Relevant issue(s)

Part of https://github.com/mysociety/transparency-fnf-whatdotheyknow-projects/issues/20

## What does this do?

* Sets up permissions for reading a `Project`
* Adds a basic `ProjectsController` so that we can view a project

## Why was this needed?

So that Project members can view a project, and gives us a place to hang other features off.

## Implementation notes

In standup I said I'd wite some notes on the direction of travel for the controllers.

Here's what I think we'll end up with at the end:

```ruby
Projects::BaseController
# Responsible for before_action access checks
Projects::ProjectsController < Projects::BaseController
# Responsible for the project "home page" – not any of the task queues
# #show – show the project to authorised users
# #create – allow eligible users to create a project
# #edit – allow owners to edit a project
# #update – allow owners to update a project
Projects::ClassificationsController < Projects::BaseController
  # Allow requests in a project to be classified
  # #show – show the project-specific view for classifying for a request in the project
Projects::ExtractsController < Projects::BaseController
# Allow requests in a project to have data extracted
# #show – show the project-specific view for extracting data for a request in the project
Projects::VerificationsController < Projects::BaseController
# Verify data extractions from contributors
# #show – show the project-specific view for verifying data extracted for a request in the project
Projects::FollowupsController < Projects::BaseController
# Verify data extractions from contributors
# #new – show the project-specific view for writing a followup for a request in the project
```

Note that the task queue controllers _don't_ have `#create` or `#update` actions. I feel that these should be handled by generic controllers.

So, taking classifications as an example, in `app/views/projects/classifications/show.html.erb` we'd show the request and also have a sidebar form that submits to `ClassificationsController#update`, and then redirects back to `Projects::ClassificationsController#show` for the next request to be classified.

## Screenshots

Not much to see at the moment – just basic info on the page:

Screenshot 2020-04-16 at 15 50 23

Not logged in:

Screenshot 2020-04-16 at 15 50 57

## Notes to reviewer

Will extract common bits of project access checks in to `Projects::BaseController` in later PRs when it's clearer what needs to get shared.

### DIFF
--- a/app/models/ability.rb
+++ b/app/models/ability.rb
@@ -150,6 +150,11 @@ class Ability
       end
     end
 
+    if feature_enabled? :projects
+      can :read, Project do |project|
+        user && (user.is_pro_admin? || project.member?(user))
+      end
+    end
   end
 
   private
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -162,6 +162,14 @@ Rails.application.routes.draw do
         :via => :post
   ####
 
+  #### Projects
+  constraints FeatureConstraint.new(:projects) do
+    scope module: :projects do
+      resources :projects, only: [:show]
+    end
+  end
+  ####
+
   resources :health_checks, :only => [:index]
 
   resources :request, :only => [] do
--- a/spec/controllers/projects/projects_controller_spec.rb
+++ b/spec/controllers/projects/projects_controller_spec.rb
@@ -1,0 +1,84 @@
+require 'spec_helper'
+
+spec_meta = {
+  type: :controller,
+  feature: :projects
+}
+
+RSpec.describe Projects::ProjectsController, spec_meta do
+  before do
+    allow(controller).to receive(:site_name).and_return('SITE')
+  end
+
+  describe 'GET show' do
+    let(:project) { FactoryBot.create(:project) }
+    let(:ability) { Object.new.extend(CanCan::Ability) }
+
+    before do
+      allow(controller).to receive(:current_ability).and_return(ability)
+    end
+
+    context 'with a logged in user who can read the project' do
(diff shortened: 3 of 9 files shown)


config/routes.rb
```diff
@@ -162,6 +162,10 @@
         :via => :post
   ####
 
+  #### Projects controller
+  resources :projects, only: [:show]
```
garethrees, line 166: Should add a feature constraint around this

app/models/ability.rb
```diff
@@ -150,6 +150,11 @@ def initialize(user)
       end
     end
 
+    if feature_enabled? :projects
+      can :read, Project do |project|
+        user && (user.is_pro_admin? || project.member?(user))
+      end
+    end
```
gbp, line 157: I think we risk overloading this ability.rb file with permisions for non-Pro, Pro, Projects and Public tokens. Happy to add this here for now but raise as a concern and a possible refactor in the future.

spec/controllers/projects/projects_controller_spec.rb
```diff
@@ -0,0 +1,76 @@
...
+    before do
+      allow(controller).to receive(:current_ability).and_return(ability)
+    end
+
+    context 'with a logged in user who can read the project' do
+      let(:user) { FactoryBot.create(:user) }
+
+      before do
+        ability.can :read, project
+        get :show, params: { id: project.id }, session: { user_id: user.id }
```
gbp, line 21: We normally in other specs we add `before { session[:user_id] = user.id }` instead of setting the session when calling the request. This can be changed when we add more specs to this context.

PR #5619, opened 2020-04-17
Use find_each to expire requests

garethrees:
`User` and `PublicBody` are likely to have many requests. Use
`find_each` to limit memory consumption while expiring many records.

Also uses Proc shorthand [1].

[1] https://rubystyle.guide/#single-action-blocks

### DIFF
--- a/app/models/public_body.rb
+++ b/app/models/public_body.rb
@@ -686,7 +686,7 @@ class PublicBody < ApplicationRecord
   end
 
   def expire_requests
-    info_requests.each { |request| request.expire }
+    info_requests.find_each(&:expire)
   end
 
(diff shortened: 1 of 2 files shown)


garethrees:
The problem here is that while the _implementation_ works, the specs fail because the implementation details are stubbed.

https://github.com/mysociety/alaveteli/blob/4b02ca7ee892688bcea18cd9a32579d6e3fb5cc1/spec/models/public_body_spec.rb#L716-L730

We stub `PublicBody#info_requests` to return a `double`, but this doesn't behave like an `ActiveRecord` collection, so fails with `NoMethodError: undefined method `find_each' for [#<Double (anonymous)>, #<Double (anonymous)>]:Array`.

PR #5625, opened 2020-04-20
[#5542] share by link prep 2

garethrees:
## Relevant issue(s)

https://github.com/mysociety/alaveteli/issues/5542

## What does this do?

Extracts some prep commits from https://github.com/mysociety/alaveteli/pull/5608

This PR is mostly about refactoring attachment paths

## Why was this needed?

Makes it easier to review the core implementation of share by link without also needing to think about unrelated refactoring.

## Notes to reviewer

### DIFF
--- a/app/helpers/info_request_helper.rb
+++ b/app/helpers/info_request_helper.rb
@@ -261,11 +261,15 @@ module InfoRequestHelper
 
     link_to image_tag(image, :class => "attachment__image",
                              :alt => "Attachment"),
-            attachment_path(incoming_message, attachment)
+            attachment_path(attachment)
   end
 
-  def attachment_path(incoming_message, attachment, options = {})
-    attach_params = attachment_params(incoming_message, attachment, options)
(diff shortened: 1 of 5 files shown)


app/helpers/info_request_helper.rb
```diff
@@ -269,7 +269,7 @@ def attachment_path(attachment, options = {})
   end
 
   def attachment_url(attachment, options = {})
-    attach_params = attachment_params(attachment, options)
+    attach_params = attachment_params(attachment, options).merge(locale: false)
```
garethrees, line 272: So one Downside of this is that we lose the translations of the view prefix:

**No locale**

Screenshot 2020-04-20 at 15 45 01


**With Locale**

Screenshot 2020-04-20 at 15 44 22

garethrees: I think what we ideally want to do is keep the locale-based views, cache all locales, but only cache the extracted attachments once, rather than only cache this view once.

We should drop this commit.

PR #5627, opened 2020-04-20
[#5542] share by link prep 3

garethrees:
## Relevant issue(s)

https://github.com/mysociety/alaveteli/issues/5542

## What does this do?

Extracts some prep commits from https://github.com/mysociety/alaveteli/pull/5608

This PR is mostly about refactoring fragment caching

## Why was this needed?

Makes it easier to review the core implementation of share by link without also needing to think about unrelated refactoring.

## Notes to reviewer


garethrees:
@gbp could you just check that this PR is what was intended with these commits?

PR #5634, opened 2020-04-22
[#5624] Ignore spam address case

### LINKED ISSUE #5624: Spam addresses list should ignore case
The spam addresses list which is used to add spam `To:` addresses which are not request emails should be case insensitive.

In this case we had an address in the spam addresses list in uppercase, which didn't prevent a spam email arriving in the holding pen with the `To:` address in lowercase.

AFAIK capitalisation doesn't matter in email addresses, so we should ignore it on the spam addresses list.

garethrees:
## Relevant issue(s)

Fixes #5624

## What does this do?

* Ignore case when checking spam in SpamAddress
* Minor code style cleanup

## Why was this needed?

Reduce admin time by not having to create a `SpamAddress` for each case that spammers use.

## Implementation notes

Note that `SpamAddress.spam?` accepts an `Array`, hence the slightly janky way of checking each address. Didn't seem to be an easy way to `OR` the addresses in SQL.

### DIFF
--- a/app/models/spam_address.rb
+++ b/app/models/spam_address.rb
@@ -14,7 +14,7 @@ class SpamAddress < ApplicationRecord
   validates_uniqueness_of :email, :message => 'This address is already marked as spam'
 
   def self.spam?(email_address)
-    exists?(:email => email_address)
+    Array(email_address).any? { |email| exists?(['email ILIKE ?', email]) }
   end
 
 end
--- a/app/models/spam_address.rb
+++ b/app/models/spam_address.rb
@@ -10,11 +10,13 @@
 #
 
 class SpamAddress < ApplicationRecord
-  validates_presence_of :email, :message => 'Please enter the email address to mark as spam'
-  validates_uniqueness_of :email, :message => 'This address is already marked as spam'
+  validates_presence_of :email,
+                        message: 'Enter the email address to mark as spam'
+
+  validates_uniqueness_of :email,
+                          message: 'This address is already marked as spam'
 
   def self.spam?(email_address)
     Array(email_address).any? { |email| exists?(['email ILIKE ?', email]) }
   end
-
 end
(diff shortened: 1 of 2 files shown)


app/models/spam_address.rb
```diff
@@ -14,7 +14,7 @@ class SpamAddress < ApplicationRecord
   validates_uniqueness_of :email, :message => 'This address is already marked as spam'
 
   def self.spam?(email_address)
-    exists?(:email => email_address)
+    Array(email_address).any? { |email| exists?(['email ILIKE ?', email]) }
```
gbp, line 20: Don't think we need this to be this complex.

For starters.. it runs a query for each email address - or at least until a match has been found.
```
pry(main)> SpamAddress.spam? ['foo@bar.com', 'a@localhost']
  SpamAddress Exists (0.4ms)  SELECT  1 AS one FROM "spam_addresses" WHERE (email ILIKE 'foo@bar.com') LIMIT $1  [["LIMIT", 1]]
  SpamAddress Exists (0.4ms)  SELECT  1 AS one FROM "spam_addresses" WHERE (email ILIKE 'a@localhost') LIMIT $1  [["LIMIT", 1]]
```

Where the existing code already handles array of addresses:
```
pry(main)> SpamAddress.exists?(email: ['foo@bar.com', 'a@localhost'])
  SpamAddress Exists (0.6ms)  SELECT  1 AS one FROM "spam_addresses" WHERE "spam_addresses"."email" IN ('foo@bar.com', 'a@localhost') LIMIT $1  [["LIMIT", 1]]
```

So we just need to do a case insensitive search, but we can't do an ILIKE IN query easily (and I think having raw SQL in the source will cause a deprecation warning - see #5568) 

Instead we could switch to the CITEXT column type: https://github.com/rails/rails/commit/0e0b41d58c0a8e5921b7eebc3a6e0e4cc012ce4d

garethrees:
Closing in favour of https://github.com/mysociety/alaveteli/pull/5709

PR #5635, opened 2020-04-23
Force Style/SignalException to use only_raise

garethrees:
This is the default in rubocop [1] and the recommended approach in the
Ruby Style Guide [2]. For some reason Hound keeps warning us to use
`fail` instead.

[1] https://docs.rubocop.org/en/stable/cops_style/#stylesignalexception
[2] https://rubystyle.guide/#prefer-raise-over-fail


garethrees:
> This should be the default already

Yeah just wanted to see if setting it explicitly resolves the hound warnings

PR #5636, opened 2020-04-23
[fnf#20] WIP Projects Extract

garethrees:


garethrees:
@gbp probably worth having a chat about this to check what we think should be extracted in to `Projects::BaseController` vs the specific action controllers, or whether some bits suit a concern rather than being inherited.

PR #5641, opened 2020-04-24
[fnf#20] Clean request/request_header

garethrees:
## Relevant issue(s)

https://github.com/mysociety/transparency-fnf-whatdotheyknow-projects/issues/20

## What does this do?

Cleans up rendering of `request/request_header`.

Extracts instance variables out to partials to make dependencies clearer.

## Why was this needed?

Makes it easier to render `request/request_header` for Projects views.

### DIFF
--- a/app/views/alaveteli_pro/info_requests/_after_actions.html.erb
+++ b/app/views/alaveteli_pro/info_requests/_after_actions.html.erb
@@ -7,24 +7,24 @@
         <li class="action-menu__menu__item">
           <ul class="action-menu__menu__submenu">
             <li>
-              <% if @last_response.nil? %>
-                <%= link_to _("Send a followup"), new_request_followup_path(:request_id => @info_request.id, :anchor => 'followup') %>
+              <% if last_response.nil? %>
+                <%= link_to _("Send a followup"), new_request_followup_path(:request_id => info_request.id, :anchor => 'followup') %>
               <% else %>
-                <%= link_to _("Write a reply"), new_request_incoming_followup_path(:request_id => @info_request.id, :incoming_message_id => @last_response.id, :anchor => 'followup') %>
+                <%= link_to _("Write a reply"), new_request_incoming_followup_path(:request_id => info_request.id, :incoming_message_id => last_response.id, :anchor => 'followup') %>
               <% end %>
             </li>
 
             <li>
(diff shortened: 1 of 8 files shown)


app/views/request/_after_actions.html.erb
```diff
@@ -4,30 +4,30 @@
       <a href="#" class="action-menu__button"><%= _('Actions') %></a>
 
       <ul class="action-menu__menu">
-        <% unless @info_request.is_external? %>
+        <% unless info_request.is_external? %>
           <li class="action-menu__menu__item">
             <span class="action-menu__menu__heading">
               <%= _('{{info_request_user_name}}',
-                    :info_request_user_name => h(@info_request.user_name)) %>
+                    :info_request_user_name => h(info_request.user_name)) %>
```
gbp, line 11:
```suggestion
                    info_request_user_name: h(info_request.user_name)) %>
```

And other old style hashes on changed lines in this file.

app/views/alaveteli_pro/info_requests/_after_actions.html.erb
```diff
@@ -7,24 +7,24 @@
         <li class="action-menu__menu__item">
           <ul class="action-menu__menu__submenu">
             <li>
-              <% if @last_response.nil? %>
-                <%= link_to _("Send a followup"), new_request_followup_path(:request_id => @info_request.id, :anchor => 'followup') %>
+              <% if last_response.nil? %>
+                <%= link_to _("Send a followup"), new_request_followup_path(:request_id => info_request.id, :anchor => 'followup') %>
```
gbp, line 11:
```suggestion
                <%= link_to _("Send a followup"), new_request_followup_path(request_id: info_request.id, anchor: 'followup') %>
```

And other old style hashes on changed lines in this file.

garethrees:
> Only thing is there are some changed lines which still use old style hashes... can't we just do a mass project wide change for these?

Maybe maybe not – not one for this PR though!

I wanted to avoid a full cleanup of _every_ template here, given there are so many affected, which is why I didn't address some of the style issues.

> We're passing local variables from request/show… which is unnecessary for the `@user` as it already has an aliased as current_user helper method. I think this could be a good refactoring to do next but probably out of scope for this PR

Yeah that sounds like a good candidate for a one-time substitution.

PR #5642, opened 2020-04-27
[fnf#13] Project classifications

gbp:
## Relevant issue(s)

Closes https://github.com/mysociety/transparency-fnf-whatdotheyknow-projects/issues/13

## What does this do?

Adds permissions and controller actions to allow project based info request classification by any member of a given project.

## Implementation notes

There is some implementation crossover with #5636 but should be anything contentious.

## Notes to reviewer

This PR doesn't add any UI to show a form which submits to the controller action.

### DIFF
--- a/app/controllers/projects/classifications_controller.rb
+++ b/app/controllers/projects/classifications_controller.rb
@@ -1,0 +1,27 @@
+##
+# Controller responsible for handling project InfoRequest classification
+#
+# Requires `url_title` to be passed in as a param
+#
+class Projects::ClassificationsController < Projects::BaseController
+  include Classifiable
+
+  def create
+    set_described_state
+
+    flash[:notice] = _('Thank you for updating this request!')
+    redirect_to project_path(@project)
+  end
+
+  private
+
+  def find_info_request
+    @info_request = @project.info_requests.find_by!(
+      url_title: params[:url_title]
+    )
+  end
+
+  def authorise_info_request
+    authorize! :update_request_state, @info_request
+  end
+end
--- a/spec/controllers/projects/classifications_controller_spec.rb
+++ b/spec/controllers/projects/classifications_controller_spec.rb
@@ -1,0 +1,214 @@
+require 'spec_helper'
+
+spec_meta = {
+  type: :controller,
+  feature: :projects
+}
+
+RSpec.describe Projects::ClassificationsController, spec_meta do
+  shared_context 'project can be found' do
+    let(:project) { FactoryBot.create(:project) }
+
+    before do
+      allow(Project).to receive(:find).with(project.id.to_s).and_return(project)
+    end
+  end
+
+  shared_context 'request can be found' do
+    include_context 'project can be found'
+
+    let(:info_request) { FactoryBot.create(:info_request, user: user) }
+
+    before do
+      info_requests = double(:info_requests_collection)
+      allow(project).to receive(:info_requests).and_return(info_requests)
+      allow(info_requests).to receive(:find_by!).
+        with(url_title: info_request.url_title).and_return(info_request)
+    end
+  end
+
+  describe '#create' do
+    let(:user) { FactoryBot.create(:pro_user) }
+    let(:ability) { Object.new.extend(CanCan::Ability) }
+
+    before do
+      session[:user_id] = user.id
+      allow(controller).to receive(:current_user).and_return(user)
+
+      allow(controller).to receive(:current_ability).and_return(ability)
+    end
+
+    context 'project to be classified can not be found' do
+      it'raises a ActiveRecord::RecordNotFound error' do
+        expect {
+          post :create, params: { project_id: 'invalid' }
+        }.to raise_error(ActiveRecord::RecordNotFound)
+      end
+    end
+
+    context 'request to be classified can not be found' do
+      include_context 'project can be found'
+
+      it'raises a ActiveRecord::RecordNotFound error' do
+        expect {
+          post :create, params: { project_id: project.id, url_title: 'invalid' }
+        }.to raise_error(ActiveRecord::RecordNotFound)
+      end
+    end
+
+    shared_context 'request to be classified can be found' do
+      include_context 'request can be found'
+
+      def post_status(status, message: nil)
+        classification = { described_state: status }
+        classification[:message] = message if message
+
+        post :create, params: {
+          classification: classification,
+          project_id: project.id,
+          url_title: info_request.url_title
+        }
+      end
+    end
+
+    context 'user is not allowed to update the request' do
+      include_context 'request to be classified can be found'
+
+      before { ability.cannot :update_request_state, info_request }
+
+      it 'raises a CanCan::AccessDenied error' do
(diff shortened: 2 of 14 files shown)


spec/controllers/projects/classifications_controller_spec.rb
```diff
@@ -0,0 +1,208 @@
...
+
+        post :create, params: {
+          classification: classification,
+          project_id: '1',
+          url_title: 'url_title'
+        }
+      end
+    end
+
+    context 'user is not allow to update the request' do
```
garethrees, line 74: s/allow/allowed/

spec/controllers/projects/classifications_controller_spec.rb
```diff
@@ -0,0 +1,208 @@
...
+
+      it 'should create status_update log' do
+        post_status('successful')
+
+        event = assigns(:status_update_event)
+        expect(event).to be_a InfoRequestEvent
+        expect(event.event_type).to eq 'status_update'
+        expect(event.params[:described_state]).to eq 'successful'
+        expect(event.params[:old_described_state]).to eq 'waiting_response'
+        expect(event.params[:user_id]).to eq info_request.user_id
```
garethrees, line 102: Ah we'll want to record the _actual_ user who classified the request, rather than assuming its the request owner.

Perhaps we want to break these up more like:

```ruby
context 'a project owner classifies their own request'
  # …

context "a project contributor classifies an owner's request"
  # …
```

Our implementation does use [`current_user.id`](https://github.com/mysociety/alaveteli/blob/4b02ca7ee892688bcea18cd9a32579d6e3fb5cc1/app/controllers/concerns/classifiable.rb#L78), but this spec doesn't illustrate that intent.

Might be worth reviewing what we do for the classification game here and whether we want to create a [`RequestClassification`](https://github.com/mysociety/alaveteli/blob/4b02ca7ee892688bcea18cd9a32579d6e3fb5cc1/app/controllers/classifications_controller.rb#L44-L64).

spec/controllers/projects/classifications_controller_spec.rb
```diff
@@ -0,0 +1,208 @@
...
+      end
+
+      it 'should call set_described_state on the request' do
```
garethrees, line 105: Should (lol) avoid using should http://www.betterspecs.org/#should

```diff
- it 'should call set_described_state on the request' do
+ it 'calls set_described_state on the request' do
```

Applies to many of these specs

spec/controllers/projects/classifications_controller_spec.rb
```diff
@@ -0,0 +1,208 @@
...
+          message_project_classification_path(
+            project_id: 1,
+            url_title: info_request.url_title,
+            described_state: 'requires_admin'
+          )
+        )
+      end
+    end
+
+    context 'user sets the request as requires_admin without a message' do
```
garethrees, line 172: s/without/with/

spec/controllers/projects/classifications_controller_spec.rb
```diff
@@ -0,0 +1,208 @@
...
+    include_context 'project can be found'
+
+    let(:info_request) { FactoryBot.create(:info_request, user: user) }
+
+    before do
+      info_requests = double(:info_requests_collection)
+      allow(project).to receive(:info_requests).and_return(info_requests)
+      allow(info_requests).to receive(:find_by!).with(url_title: 'url_title').
+        and_return(info_request)
+    end
```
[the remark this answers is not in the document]
gbp: The problem with not stubbing like:
```ruby
allow(info_requests).to receive(:find_by!).with(url_title: 'url_title').and_return(info_request)
```
is that we can't then do:
```ruby
expect(info_request).to receive(:set_described_state)
```
this leads us to needing to knowledge about and/or testing the implementation of `InfoRequest#set_described_state` in the wrong place.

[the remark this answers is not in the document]
garethrees: hmm yeah not ideal.

Would be nice if you could use a dynamic finder for like `@project.find_info_request_by_url_title(params[:url_title])` – at least then you're not having to stub an entire method chain.

Happy to leave as-is for now, but its something I'd like to find a better solution to in general.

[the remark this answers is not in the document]
garethrees: Not suggesting this for now, but might be able to actually build this ourselves:

```ruby
  def method_missing(method_id, *arguments)
    if match = /find_(?<association>[_a-zA-Z]\w+)_by_(?<attribute>[_a-zA-Z]\w+)/.match(method_id.to_s)
      return super unless respond_to?(match[:association].pluralize)
      send(match[:association].pluralize).send("find_by_#{ match[:attribute] }", *arguments)
    else
      super
    end
  end
```

```ruby
project
# => #<Project:0x00005576e62ef768 …>
project.info_requests.first.url_title
# => "skjdhbc_ohlcc"
project.find_info_request_by_url_title("skjdhbc_ohlcc").id
#  InfoRequest Load (4.0ms)  SELECT  "info_requests".* FROM "info_requests" LEFT JOIN project_resources r1 ON r1.resource_id = info_requests.id AND r1.resource_type = 'InfoRequest' LEFT JOIN project_resources r2 ON r2.resource_id = info_requests.info_request_batch_id AND r2.resource_type = 'InfoRequestBatch' WHERE (r1.project_id = 1 OR r2.project_id = 1) AND "info_requests"."url_title" = $1 LIMIT $2  [["url_title", "skjdhbc_ohlcc"], ["LIMIT", 1]]
# => 113

```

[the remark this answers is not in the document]
garethrees: In fact, this can be simplified way more to just be:

```ruby
# model
def method_missing(method_id, *arguments)
  if match = /find_(?<association>[_a-zA-Z]\w+)_by/.match(method_id.to_s)
    return super unless respond_to?(match[:association].pluralize)
    send(match[:association].pluralize).send(:find_by, *arguments)
  else
    super
  end
end

# controller
@project.find_info_request_by(url_title: params[:url_title])

# controller spec
before do
  allow(project).
    to receive(:find_info_request_by).
      with(url_title: info_request.url_title).
        and_return(info_request)
end
```

Just posting here for future reference

app/controllers/projects/classifications_controller.rb
```diff
@@ -0,0 +1,27 @@
+##
+# Controller responsible for handling project InfoRequest classification
+#
+# Requires `url_title` to be passed in as a param
+#
+class Projects::ClassificationsController < Projects::BaseController
+  include Classifiable
+
+  def create
+    set_described_state
```
garethrees, line 10: Sorry late review spot but this doesn't check that the user can read the project – we don't do that in `BaseController` right now

garethrees: Oh sorry, we check the project in `:update_request_state, so that's probably okay enough?

gbp: Yeah I think that's enough for now.

PR #5644, opened 2020-04-28
[fnf#20] Basic styling, layout and HTML stubs of project homepage

wrightmartin:
## Relevant issue(s)

Part of https://github.com/mysociety/transparency-fnf-whatdotheyknow-projects/issues/20

## What does this do?
- Adds basic layout and styles for the project homepage

## Why was this needed?
- So that projects have an index page and waymarkers to tasks

## Implementation notes
- Uses `<progress>` which I've never used before and might require some further cross-browser testing

## Screenshots

image

## Notes to reviewer

- Lots of harcoded figures and elements to be replaced

### DIFF
--- a/app/assets/stylesheets/responsive/all.scss
+++ b/app/assets/stylesheets/responsive/all.scss
@@ -104,6 +104,9 @@
 
 @import "responsive/alaveteli_pro/_stripe";
 
+@import "responsive/alaveteli_pro/_projects_layout";
+@import "responsive/alaveteli_pro/_projects_style";
+
 @import "responsive/_ms_edge";
 
 @import "responsive/custom";
--- a/app/views/projects/projects/show.html.erb
+++ b/app/views/projects/projects/show.html.erb
@@ -1,21 +1,98 @@
-<h1><%= @project.title %></h1>
+<%= render partial: 'project_nav', locals: { project: @project } %>
 
-<div>
-  <%= sanitize @project.briefing %>
-</div>
-
-<aside>
-  <h1><%= @project.owner.name %></h1>
-
-  <% if @project.owner.show_profile_photo? %>
-    <%= image_tag get_profile_photo_url(url_name: @project.owner.url_name) %>
-  <% end %>
-
-  <p><%= @project.owner.about_me %></p>
-
-  <div>
-    <%= link_to contact_user_path(url_name: @project.owner.url_name) do %>
-      <%= _('Contact {{recipient}}', recipient: @project.owner.name) %>
-    <% end %>
+<div class="inner-canvas">
+  <div class="inner-canvas-header">
+   <div class="row">
+      <h1><%= @project.title %></h1>
+    </div>
   </div>
-</aside>
+
+  <div class="inner-canvas-body">
+    <div class="row">
+      <div class="project-body">
+        <div class="project-briefing">
+          <%= sanitize @project.briefing %>
+
+          <p class="project-meta">
+            <%= _('{{contributor_count}} contributors',
+                  contributor_count: @project.contributors.count) %>
+          </p>
+
+          <div class="project-workload">
+            <h3 class="project-workload-title">
+              <%= _('Workload') %>
+            </h3>
+
+            <div class="project-workload-diagram">
+              <progress max="100" value="56" class="project-workload-indicator"></progress>
+              <span class="project-workload-value">56%</span>
+            </div>
+          </div>
+        </div>
+
+        <div class="project-task-list" id="tasks">
+          <h2>
+            <%= _('Contribute') %>
+          </h2>
+
+          <div class="project-task">
+            <h3><%= _('Extract data') %></h3>
+
+            <div class="project-task-meta-information">
+              <p class="project-meta"><%= _('Spreadsheets') %> / <%= _('Data interpretation') %></p>
+              <p class="project-meta"><%= _('60 minutes+') %> / <%= _('Laptop or desktop') %></p>
+            </div>
+
+            <p>
+              <%= _('Extract relevant information from each successful ' \
+                    'freedom of information request and enter it into the ' \
+                    'provided fields. This will populate the data file for ' \
+                    'this project.') %>
+            </p>
+
+            <div class="project-workload">
+              <h4 class="project-workload-title"><%= _('Workload') %></h4>
+
+              <div class="project-workload-diagram">
+                <progress max="100" value="10" class="project-workload-indicator"></progress>
+                <span class="project-workload-value">10%</span>
+              </div>
+            </div>
+
+            <a href="#" class="button"><%= _('Start extracting') %></a>
+          </div>
+        </div>
+      </div>
+
+      <aside class="project-aside">
+        <div class="project-owner">
+          <h2><%= _('Project owner') %></h2>
+
+          <% if @project.owner.show_profile_photo? %>
+            <%= image_tag get_profile_photo_url(url_name: @project.owner.url_name),
+                class: 'project-owner__photo' %>
+          <% end %>
+
+          <h3 class="project-owner__name"><%= @project.owner.name %></h3>
+
+          <% if @project.owner.about_me.present? %>
+            <p><%= @project.owner.about_me %></p>
+          <% end %>
+
+           <%= link_to contact_user_path(url_name: @project.owner.url_name),
+               class: 'button-secondary' do %>
+            <%= _('Contact {{recipient}}', recipient: @project.owner.name) %>
+          <% end %>
+        </div>
+
+        <div class="project-access-status">
+          <p>
+            <%= _('You are a <strong>contributor</strong> on this project') %>
+          </p>
+
+          <a href="#" class="button-secondary"><%= _('Leave project') %></a>
+        </div>
+      </aside>
+    </div>
+  </div>
+</div>
--- a/app/assets/stylesheets/responsive/alaveteli_pro/_projects_layout.scss
+++ b/app/assets/stylesheets/responsive/alaveteli_pro/_projects_layout.scss
@@ -1,0 +1,122 @@
+.projects-nav {
+  @include grid-column($columns:12);
+  display: flex;
+  flex-flow: row wrap;
+  justify-content: space-between;
+  margin: 2em 0 -1em;
+}
+
+.projects-nav__menu {
+  @include respond-min( $main_menu-mobile_menu_cutoff ){
+    position: relative;
+    top: -3px;
+  }
+}
+
+.projects-nav__title {
+ h1 {
+   font-size: 1rem;
+   margin-top: 0;
+ }
+}
+
+.project-body {
+  @include grid-column($columns:12);
+  @include respond-min( $main_menu-mobile_menu_cutoff ){
+    @include grid-column($columns:8);
+      @include ie8 {
+        padding-right: 0.9375em;
+      }
+      @include lte-ie7 {
+        width: 36.813em;
+      }
+  }
+}
+
+.project-aside {
+  @include grid-column($columns:12);
+  @include respond-min( $main_menu-mobile_menu_cutoff ){
+    @include grid-column($columns:4);
+    @include ie8 {
+      padding-left: 0.9375em;
+    }
+    @include lte-ie7 {
+      width: 17.438em;
+    }
+  }
+}
+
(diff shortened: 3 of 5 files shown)


app/views/projects/projects/show.html.erb
```diff
@@ -1,21 +1,85 @@
-<h1><%= @project.title %></h1>
+<div class="row">
+<nav class="projects-nav">
+  <div class="projects-nav__title">
+    <h1><%= @project.title %></h1>
+  </div>
+  <div class="project-nav__menu">
+    <ul class="inline-list">
+      <li>
+        <a href="#">Projects home</a>
```
garethrees, line 9: We could remove this and use `projects-nav__title` as the link back home?

Can use the [`link_to_unless_current`](https://api.rubyonrails.org/v5.1.7/classes/ActionView/Helpers/UrlHelper.html#method-i-link_to_unless_current) helper to prevent this being a link when we're on the projects homepage.

I think we'll want:

```erb
<%= link_to_unless_current @project.title, project_path(@project) %>
```

Or even `<%= link_to_unless_current @project.title, @project %>` might work.

app/views/projects/projects/show.html.erb
```diff
@@ -1,21 +1,85 @@
...
+      </li>
+      <li>
+        <a href="#">Tasks</a>
```
garethrees, line 12: Where should this link go?

app/views/projects/projects/show.html.erb
```diff
@@ -1,21 +1,85 @@
-<h1><%= @project.title %></h1>
+<div class="row">
+<nav class="projects-nav">
```
garethrees, line 2: This `nav` should be indented as it's a child of the `div` above.

app/views/projects/projects/show.html.erb
```diff
@@ -1,21 +1,85 @@
...
+        </div>
+        <div class="project-task-list">
+          <h2>Contribute</h2>
+          <div class="project-task">
+            <h3>Extract data</h3>
+            <div class="project-task-meta-information">
+              <p class="project-meta">Spreadsheets / Data interpretation</p>
+              <p class="project-meta">60 minutes+ / Laptop or desktop</p>
+            </div>
+            <p>Task description Lorem ipsum dolor sit amet, consectetur adipiscing elit. Nulla eleifend vestibulum neque at blandit. Aliquam ac rhoncus felis. Etiam scelerisque viverra condimentum. Suspendisse faucibus lectus vel magna pretium, a imperdiet lorem aliquam. Vestibulum turpis eros.</p>
```
garethrees, line 50: Can you have a stab at actually writing a real, generic description of what this task is about?

app/views/projects/projects/show.html.erb
```diff
@@ -1,21 +1,85 @@
...
+            </div>
+            
+            <a href="#" class="button">Start</a>
+          </div>
+        </div>
+      </div>
+      <aside class="project-aside">
+        <div class="project-owner">
+          <h2>Project owner</h2>
+          <img src="https://pbs.twimg.com/profile_images/1218922930018623488/C4ocKWB6_400x400.jpg" alt="" class="project-owner__photo" />
```
garethrees, line 65: This should be replaced with the actual code that was added:

```
<% if @project.owner.show_profile_photo? %>
  <%= image_tag get_profile_photo_url(url_name: @project.owner.url_name) %>
<% end %>
```

See the [`image_tag`](https://api.rubyonrails.org/v5.1.7/classes/ActionView/Helpers/AssetTagHelper.html#method-i-image_tag) API docs for adding HTML attributes.

app/views/projects/projects/show.html.erb
```diff
@@ -1,21 +1,85 @@
...
+          <% end %>
 
-  <p><%= @project.owner.about_me %></p>
+          <p><%= @project.owner.about_me %></p>
 
-  <div>
-    <%= link_to contact_user_path(url_name: @project.owner.url_name) do %>
-      <%= _('Contact {{recipient}}', recipient: @project.owner.name) %>
-    <% end %>
+           <%= link_to contact_user_path(url_name: @project.owner.url_name),  :class => 'button-secondary' do %>
```
garethrees, line 74: We use Ruby 1.9 style hashes now:

```diff
- ,  :class => 'button-secondary'
+,  class: 'button-secondary'
```

Note there's also an extra space after that comma.

app/assets/stylesheets/responsive/all.scss
```diff
@@ -104,6 +104,8 @@
 
 @import "responsive/alaveteli_pro/_stripe";
 
+@import "responsive/alaveteli_pro/projects";
```
garethrees, line 107: Shouldn't we keep the convention of a separate style and layout file?

```scss
@import "responsive/alaveteli_pro/_projects_layout";
@import "responsive/alaveteli_pro/_projects_style";
```

app/assets/stylesheets/responsive/alaveteli_pro/_projects.scss
```diff
@@ -0,0 +1,124 @@
...
+.project-task-meta-information {
+  display: flex;
+  flex-flow: row wrap;
+  justify-content: space-between;
+  :first-child.project-meta {
+    margin-right: 0.5em;
+  }
+}
+
+project-workload {
```
garethrees, line 82: Is this missing a `.` prefix, or is this an SCSS Thing?

app/views/projects/projects/show.html.erb
```diff
@@ -1,21 +1,85 @@
...
-<aside>
-  <h1><%= @project.owner.name %></h1>
+  <div class="inner-canvas-body">
+    <div class="row">
+      <div class="project-body">
+        <div class="project-briefing">
+          <%= sanitize @project.briefing %>
+          <p class="project-meta">16 contributors</p>
+          <div class="project-workload">
+            <h3 class="project-workload-title">Workload</h3>
```
garethrees, line 36: What's the reason for using "Workload" over "Progress"?

app/views/projects/projects/show.html.erb
```diff
@@ -1,21 +1,85 @@
...
+  <div class="project-nav__menu">
+    <ul class="inline-list">
+      <li>
+        <a href="#">Projects home</a>
+      </li>
+      <li>
+        <a href="#">Tasks</a>
+      </li>
+      <li>
+        <a href="#">Message project owner</a>
```
garethrees, line 15: We can link this up as it's an existing feature, as done in the `project-owner` `aside`.

app/views/projects/projects/show.html.erb
```diff
@@ -1,21 +1,85 @@
...
+  </div>
 
-<aside>
-  <h1><%= @project.owner.name %></h1>
+  <div class="inner-canvas-body">
+    <div class="row">
+      <div class="project-body">
+        <div class="project-briefing">
+          <%= sanitize @project.briefing %>
+          <p class="project-meta">16 contributors</p>
```
garethrees, line 34: Can get the real value here:

```erb
<p class="project-meta">
  <%= _('{{contributor_count}} contributors',
        contributor_count: @project.contributors.count %>
</p>
```

app/views/projects/projects/show.html.erb
```diff
@@ -1,21 +1,85 @@
...
+            </div>
+            <p>Task description Lorem ipsum dolor sit amet, consectetur adipiscing elit. Nulla eleifend vestibulum neque at blandit. Aliquam ac rhoncus felis. Etiam scelerisque viverra condimentum. Suspendisse faucibus lectus vel magna pretium, a imperdiet lorem aliquam. Vestibulum turpis eros.</p>
+            <div class="project-workload">
+              <h4 class="project-workload-title">Workload</h4>
+              <div class="project-workload-diagram">
+                <progress max="100" value="10" class="project-workload-indicator"></progress><span class="project-workload-value">100%</span>
+              </div>
+            </div>
+            
+            <a href="#" class="button">Start</a>
```
garethrees, line 58: Should we say like "Start extracting", "Start classifying", etc?

app/views/projects/projects/show.html.erb
```diff
@@ -1,21 +1,85 @@
...
 
-  <% if @project.owner.show_profile_photo? %>
-    <%= image_tag get_profile_photo_url(url_name: @project.owner.url_name) %>
-  <% end %>
+          <% if @project.owner.show_profile_photo? %>
+            <%= image_tag get_profile_photo_url(url_name: @project.owner.url_name) %>
+          <% end %>
 
-  <p><%= @project.owner.about_me %></p>
+          <p><%= @project.owner.about_me %></p>
```
garethrees, line 72: What happens if this is empty? Do we just get an empty `<p>` tag? Might want to do:

```erb
<% if @project.owner.about_me.present? %>
  <p><%= @project.owner.about_me %></p>
<% end %>
```

garethrees:
Sorry, one more comment. Applying wdtk-theme and reducing the screen width leaves no margin to the container. This is something we'll want to fix in core – don't want to have to add a theme override for every Alaveteli that enables projects.

Screenshot 2020-04-28 at 12 50 46

garethrees:
Doh, another minor find. Looks like the alignment of the project header is out:

Screenshot 2020-04-28 at 11 25 00

app/views/projects/projects/show.html.erb
wrightmartin: We decided to change this as 'progress' is too hard to measure, I can't quite remember the details but it was to do with how requests would flow through the tasks so progress might be misleading, whereas workload shows how much work is to be done and where effort is needed.

app/views/projects/projects/show.html.erb
wrightmartin: Done

app/views/projects/projects/show.html.erb
wrightmartin: Done

app/views/projects/projects/show.html.erb
wrightmartin: Done

app/views/projects/projects/show.html.erb
wrightmartin: Yup, agree

app/views/projects/projects/show.html.erb
wrightmartin: Done

app/views/projects/projects/show.html.erb
wrightmartin: It should go to #tasks (on the homepage, adding the ID for the fragment now) - quick Q how can I write the link so that if we're on the homepage already it just outputs `<a href="#tasks">Tasks</a>`, but if not it adds the full path?

app/views/projects/projects/show.html.erb
wrightmartin: `<%= link_to_unless_current @project.title, @project %>` seems to work

wrightmartin:
> Uses <progress> which I've never used before and might require some further cross-browser testing

The implications of this are mostly aesthetic, especially as we are also writing this to the page as a number. Browser support for `<progress>` is pretty wide (https://caniuse.com/#search=progress) but getting it to look exactly as we want is a bit fiddly. 

That being said I'd still prefer to use this than a CSS based solution that doesn't give the page any semantic meaning

app/views/projects/projects/show.html.erb
wrightmartin: Yep, great suggestion 👍

app/views/projects/projects/show.html.erb
garethrees: Just use the full path, as the browser figures out that it doesn't need a full page load.

```erb
<%= link_to 'Tasks', project_path(@project, anchor: 'tasks') %>
```

app/views/projects/projects/show.html.erb
garethrees, line 2: Probably makes sense to extract this out to a partial since we know we'll be using it across project pages.

app/views/projects/projects/show.html.erb
wrightmartin: 👍 Done

app/views/projects/projects/show.html.erb
wrightmartin: Done

PR #5646, opened 2020-04-29
Add Pro batch configuration variables

gbp:
## What does this do?

Adds ENABLE_PRO_BATCH_ACCESS and ENABLE_PRO_BATCH_CATEGORY_UI

Not documented in the example `general.yml` file, nor is ENABLE_PROJECTS, but that is fine until the feature ships.

## Why was this needed?

I always forget to enable in the console, with this I can just add to my `general.yml` configuration.

### DIFF
--- a/config/initializers/alaveteli_features.rb
+++ b/config/initializers/alaveteli_features.rb
@@ -12,6 +12,8 @@ features = %i[
   annotations
   alaveteli_pro
   projects
+  pro_batch_access
+  pro_batch_category_ui
   pro_pricing
   pro_self_serve
 ]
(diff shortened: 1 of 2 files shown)


config/initializers/alaveteli_features.rb
```diff
@@ -12,6 +12,8 @@
   annotations
   alaveteli_pro
   projects
+  pro_batch_access
+  pro_batch_category_ui
```
garethrees, line 16: We set these on a per-actor basis – does it make sense to include them here?

I actually think `pro_batch_access` can be done away with entirely. If you're a pro, you should be able to create batches without needing a feature flag set.

PR #5653, opened 2020-05-07
[fnf#23] Classification show controller stub

garethrees:
## Relevant issue(s)

Work on https://github.com/mysociety/transparency-fnf-whatdotheyknow-projects/issues/23

## What does this do?

Adds a basic `Projects::ClassifiesController#show`

## Why was this needed?

Gives us something to work with to start rendering the classification form.

## Implementation notes

See commit message notes.

## Screenshots

No sidebar yet – just rendering the request into the template.

Screenshot 2020-05-07 at 14 42 38

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -168,6 +168,8 @@ Rails.application.routes.draw do
     scope module: :projects do
       resources :projects, only: [:show] do
         resource :extract, only: [:show]
+        resource :classify, only: [:show]
+
         resources :classifications, only: :create, param: :described_state do
           get :message, on: :member
         end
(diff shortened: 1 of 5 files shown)


config/routes.rb
```diff
@@ -168,6 +168,8 @@
     scope module: :projects do
       resources :projects, only: [:show] do
         resource :extract, only: [:show]
+        resource :classify, only: [:show]
```
gbp, line 171: Should this not be part of the existing classifications controller, perhaps renamed to classify?

garethrees: I added a note about this in the commit message

PR #5654, opened 2020-05-07
[fnf#20] Extraction data models

gbp:
## Relevant issue(s)

mysociety/transparency-fnf-whatdotheyknow-projects#20

## What does this do?

Adds basic dataset model to store extractions

## Why was this needed?

So we can start work storing the project submission data.

## Implementation notes

Will need to be rebased/merged with the work done in https://github.com/mysociety/alaveteli/compare/develop...fnf20-projects-extract-design

## Screenshots

While there is UI implement this isn't finial and will change.

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -3,5 +3,6 @@ languages:
   PHP: true
 exclude_paths:
 - "lib/strip_attributes/test/*"
+- "db/migrate/"
 - "spec/"
 - "vendor/"
--- a/app/models/project.rb
+++ b/app/models/project.rb
@@ -40,6 +40,8 @@ class Project < ApplicationRecord
     where("r1.project_id = :id OR r2.project_id = :id", id: project.id)
   }
 
+  has_one :key_set, class_name: 'Dataset::KeySet', as: :resource
+
   validates :title, :owner, presence: true
 
   def info_request?(info_request)
--- a/app/views/projects/dataset/keys/_key.html.erb
+++ b/app/views/projects/dataset/keys/_key.html.erb
@@ -1,0 +1,5 @@
+<%= f.fields_for 'values_attributes[]', Dataset::Value.new do |nested_form| %>
+  <%= render partial: "projects/dataset/keys/#{key.format}_key",
+             locals: { f: nested_form, key: key } %>
+  <%= nested_form.hidden_field :dataset_key_id, value: key.to_param %>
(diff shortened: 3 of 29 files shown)


.codeclimate.yml
```diff
@@ -3,5 +3,6 @@ languages:
   PHP: true
 exclude_paths:
 - "lib/strip_attributes/test/*"
+- "db/migrate/"
```
gbp, line 6: Added to prevent CodeClimate complaining about duplicated code in database migrations

app/models/project.rb
```diff
@@ -40,6 +40,8 @@ class Project < ApplicationRecord
     where("r1.project_id = :id OR r2.project_id = :id", id: project.id)
   }
 
+  has_one :key_set, class_name: 'Dataset::KeySet', as: :resource
```
gbp, line 43: For now just a has one association but can be changed to has many in the future when needed.

app/views/projects/dataset/keys/_key.html.erb
```diff
@@ -0,0 +1,5 @@
+<%= f.fields_for 'values_attributes[]', Dataset::Value.new do |nested_form| %>
```
gbp, line 1: The format of the params will be compatible with `accepts_nested_attributes_for :values` on Dataset::ValueSet

app/views/projects/extracts/_sidebar.html.erb
```diff
@@ -0,0 +1,10 @@
+<div id="right_column" class="sidebar right_column" role="complementary">
+  <div class="sidebar__section">
+    <%= form_for Dataset::ValueSet.new,
+                 url: project_extract_path,
```
gbp, line 4: No POST action yet - this is added in #5656

.codeclimate.yml
garethrees: Commit messages are the place for this kind of comment ☝️ 

If in a years time I'm wondering why we excluded `"db/migrate/", first thing I'd do is git blame on the line. Takes time to then trace that back to a PR (which are more ephemeral than commits – e.g. if we decided to move to GitLab or something).

garethrees:
> Will need to be rebased/merged with the work done in https://github.com/mysociety/alaveteli/compare/develop...fnf20-projects-extract-design

Design PR can be rebased over this I think – better to get the design working with the real forms.

PR #5655, opened 2020-05-07
[fnf#28] Project queues

garethrees:
## Relevant issue(s)

* https://github.com/mysociety/transparency-fnf-whatdotheyknow-projects/issues/28
* https://github.com/mysociety/transparency-fnf-whatdotheyknow-projects/issues/20
* https://github.com/mysociety/transparency-fnf-whatdotheyknow-projects/issues/23

## What does this do?

Adds the basic query methods for finding requests for a given queue.

## Why was this needed?

Better to have these defined in the model rather than at the controller level.

## Implementation notes

We might want to add something like a `Project::TaskQueue` class that encapsulates the requests for a given queue and handle things like "checking out" a request and which have been skipped by a given user, but for now sampling the query methods is good enough.

### DIFF
--- a/app/models/project.rb
+++ b/app/models/project.rb
@@ -3,6 +3,8 @@
 # info requests.
 #
 class Project < ApplicationRecord
+  EXTRACTABLE_STATES = %w(successful partially_successful).freeze
+
   has_many :memberships, class_name: 'ProjectMembership'
   has_one  :owner_membership,
            -> { where(role: Role.project_owner_role) },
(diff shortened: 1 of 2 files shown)


app/models/project.rb
```diff
@@ -53,4 +55,11 @@ def member?(user)
   def classifiable_requests
     info_requests.where(awaiting_description: true)
   end
+
+  def extractable_requests
+    info_requests.where(
```
garethrees, line 60: We'll want to update this based on https://github.com/mysociety/alaveteli/pull/5654 – probably need extractable requests that _you_ have not submitted data to

garethrees:
Closing and extracting in to a PR per queue

PR #5656, opened 2020-05-11
[fnf#20] Project submissions

gbp:
## Relevant issue(s)

Depends on #5654
mysociety/transparency-fnf-whatdotheyknow-projects#20

## What does this do?

- Adds an extraction POST action which stores form data as Dataset::ValueSet/Value instances.
- Adds project submission model to record Project classifications and extractions

## Todo

- [x] Specs for the controller action needed
- [x] Specs elsewhere?

### DIFF
--- a/app/models/project/submission.rb
+++ b/app/models/project/submission.rb
@@ -1,0 +1,32 @@
+# == Schema Information
+# Schema version: 20200509082917
+#
+# Table name: project_submissions
+#
+#  id            :integer          not null, primary key
+#  project_id    :integer
+#  user_id       :integer
+#  resource_type :string
+#  resource_id   :integer
+#  created_at    :datetime         not null
+#  updated_at    :datetime         not null
+#
+
+##
+# A submission to a Project by one of its members. Can either be an Info Request
+# status update (Classification) or an extraction of data (Dataset::ValueSet).
+#
+class Project::Submission < ApplicationRecord
+  belongs_to :project
+  belongs_to :user
+  belongs_to :resource, polymorphic: true
+
+  RESOURCE_TYPES = %w[
+    InfoRequestEvent
+    Dataset::ValueSet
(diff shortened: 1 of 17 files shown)


app/models/project/submission.rb
```diff
@@ -0,0 +1,33 @@
...
+# Schema version: 20200509082917
+#
+# Table name: project_submissions
+#
+#  id            :integer          not null, primary key
+#  project_id    :integer
+#  user_id       :integer
+#  resource_type :string
+#  resource_id   :integer
+#  skipped       :boolean
```
garethrees, line 11: I'm thinking `skipped` should be ephemeral and stored in a session-based queue, rather than creating empty submissions. Maybe remove it for now and if we decide we should use `Project::Submission` to handle skipping we can add it in later?

app/models/classification.rb
```diff
@@ -0,0 +1,20 @@
...
+#  info_request_id :integer
+#  state           :string
+#  created_at      :datetime         not null
+#  updated_at      :datetime         not null
+#
+
+##
+# A basic Info Request classification status updates
+#
+class Classification < ApplicationRecord
```
garethrees, line 16: This is pretty close to [`RequestClassification`](https://github.com/mysociety/alaveteli/blob/develop/app/models/request_classification.rb) so I think we need to consider the additional indirection this addition will cause.

Is it intentional that this doesn't really _do_ anything at the moment? I assume the thinking was that at some point we'll extract `set_described_state` into this on a `before_save` callback?

Do we _need_ to tie classifications to `Project::Submission` at this point? What are we losing if we don't?

gbp: Not quite sure what to do here yet. I guess we could associate `Project::Submission` to an `InfoRequestEvent` with `event_type=status_update`  but that is ideal either.

PR #5659, opened 2020-05-12
[fnf#20] Projects extract design

garethrees:
## Relevant issue(s)

https://github.com/mysociety/transparency-fnf-whatdotheyknow-projects/issues/20

## What does this do?

Adds layout and basic styling to the projects 'extract' screen

## Why was this needed?

To allow us to continue the buildout of the projects UI to eventually test the whole flow

## Screenshots
image

### DIFF
--- a/app/assets/stylesheets/responsive/_global_style.scss
+++ b/app/assets/stylesheets/responsive/_global_style.scss
@@ -459,3 +459,7 @@ button,
     }
   }
 }
+
+.button--full-width {
+  width: 100%;
+}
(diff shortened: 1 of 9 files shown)


garethrees:
> @gbp force-pushed the `fnf20-projects-extract-design` branch from d0a97fb to 566fab4

This was only pushed to the `github` remote – caused a bit of confusion this morning.

gbp:
> One thing I think we should do before merge is that the labels don't seem to focus the related input. Might need to debug with @gbp.

Fixed in 5667

garethrees:
Okay so I think this is ready to merge once https://github.com/mysociety/alaveteli/pull/5667 is merged and rebasing this PR 👏

PR #5660, opened 2020-05-12
[fnf#28] Add Project#classifiable_requests

garethrees:
## Relevant issue(s)

Part of https://github.com/mysociety/transparency-fnf-whatdotheyknow-projects/issues/28.

## What does this do?

Adds a query method for requests to be classified.

## Why was this needed?

Allows us to pick out suitable requests for classification

## Implementation notes

Once we've implemented https://github.com/mysociety/transparency-fnf-whatdotheyknow-projects/issues/23 we can add progress percentage on the project homepage with something like:

```ruby
@classification_progress =
  ((@project.classifiable_requests.count / @project.info_requests.count.to_f) * 100).floor
```

### DIFF
--- a/app/models/project.rb
+++ b/app/models/project.rb
@@ -49,4 +49,8 @@ class Project < ApplicationRecord
   def member?(user)
     members.include?(user)
   end
+
+  def classifiable_requests
+    info_requests.where(awaiting_description: true)
+  end
 end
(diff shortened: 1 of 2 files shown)


garethrees:
Just going to merge this as it's not complicated so I don't think waiting for review is critical.

PR #5665, opened 2020-05-15
[fnf#23] Add classify form

garethrees:
## Relevant issue(s)

Work on https://github.com/mysociety/transparency-fnf-whatdotheyknow-projects/issues/23
Requires https://github.com/mysociety/alaveteli/pull/5663

## What does this do?

Adds Projects classification form and basic styling

## Why was this needed?

To allow contributors to classify requests in a project

## Screenshots

Screenshot 2020-05-15 at 11 40 17

PR #5669, opened 2020-05-15
[fnf#23] Classify workflow

garethrees:
## Relevant issue(s)

Work on https://github.com/mysociety/transparency-fnf-whatdotheyknow-projects/issues/23

## What does this do?

* Redirect to project homepage if there are no classifiable requests
* Redirect to more classifications after classifying a request

## Why was this needed?

Links up the actions as expected. 

## Screenshots

Message after all classifications done:

Screenshot 2020-05-15 at 16 14 38

### DIFF
--- a/app/controllers/projects/classifies_controller.rb
+++ b/app/controllers/projects/classifies_controller.rb
@@ -11,6 +11,7 @@ class Projects::ClassifiesController < Projects::BaseController
     unless @info_request
       msg = _('There are no requests to classify right now. Great job!')
       redirect_to @project, notice: msg
+      return
     end
 
     @state_transitions = @info_request.state.transitions(
--- a/app/controllers/projects/classifies_controller.rb
+++ b/app/controllers/projects/classifies_controller.rb
@@ -8,6 +8,11 @@ class Projects::ClassifiesController < Projects::BaseController
     @info_request =
       @project.info_requests.where(awaiting_description: true).sample
 
+    unless @info_request
+      msg = _('There are no requests to classify right now. Great job!')
+      redirect_to @project, notice: msg
+    end
+
     @state_transitions = @info_request.state.transitions(
       is_pro_user: false,
       is_owning_user: false,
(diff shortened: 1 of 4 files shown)


app/controllers/projects/classifies_controller.rb
```diff
@@ -7,6 +7,11 @@ def show
 
     @info_request =
       @project.info_requests.where(awaiting_description: true).sample
+
+    unless @info_request
```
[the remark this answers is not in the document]
garethrees: Hmm can't here really unless we extract finding the request to a filter. Maybe once we implement https://github.com/mysociety/transparency-fnf-whatdotheyknow-projects/issues/28.

app/controllers/projects/classifies_controller.rb
```diff
@@ -7,6 +7,11 @@ def show
...
+      msg = _('There are no requests to classify right now. Great job!')
+      redirect_to @project, notice: msg
```
garethrees, line 13: Need a `return` after this line for when we integrate https://github.com/mysociety/alaveteli/pull/5665/files#diff-3ee52fe46c4a210851d4d5beba970a93R11

garethrees: Fixed in e2300bf89900f549dff274a488e7b43c0d97214a

app/controllers/projects/classifies_controller.rb
```diff
@@ -7,6 +7,12 @@ def show
...
+      return
+    end
```
gbp, line 15: Not sure I quite follow, presumably Hound is suggesting:

```suggestion
    return if @info_request
    
    msg = _('There are no requests to classify right now. Great job!')
    redirect_to @project, notice: msg
```

This seem does seem better to me.

app/controllers/projects/classifies_controller.rb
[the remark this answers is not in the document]
gbp: Commented with suggestion: https://github.com/mysociety/alaveteli/pull/5669#discussion_r427241087

app/controllers/projects/classifies_controller.rb
garethrees: Ah sorry, really I should have waited for https://github.com/mysociety/alaveteli/pull/5665 before requesting review in hindsight. I've done so now and pushed.

garethrees: The problem is that we actually want to do stuff after this (i.e. set `@state_transitions`)

PR #5672, opened 2020-05-18
[fnf#28] Classification progress

garethrees:
## Relevant issue(s)

## What does this do?

Add classification progress to the project homepage

## Why was this needed?

Show users whether there's any work to do

## Implementation notes

Simple approach for now. Just calculate proportion of requests classified vs needing to be classified.

## Screenshots

All requests classified:

Screenshot 2020-05-18 at 16 54 15

## Notes to reviewer

I think we need to rephrase "workload" now. Doesn't really make sense that when we're all done "Workload is 100%"?

### DIFF
--- a/app/models/project.rb
+++ b/app/models/project.rb
@@ -59,4 +59,8 @@ class Project < ApplicationRecord
   def classified_requests
     info_requests.where(awaiting_description: false)
   end
+
+  def classification_progress
+    ((classified_requests.count / info_requests.count.to_f) * 100).floor
+  end
 end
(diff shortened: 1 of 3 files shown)


wrightmartin:
> I think we need to rephrase "workload" now. Doesn't really make sense that when we're all done "Workload is 100%"?

garethrees:
> Completion?

Or just "Classified"?

garethrees:
I'm actually thinking its clearer to not have a subtitle here:

Screenshot 2020-05-19 at 13 17 55

PR #5675, opened 2020-05-19
[fnf#23] Fix submission of classification message via Projects

garethrees:
## Relevant issue(s)

https://github.com/mysociety/transparency-fnf-whatdotheyknow-projects/issues/23

## What does this do?

Allows submitting a classification message via a project classification.

## Why was this needed?

Prior to this commit we weren't passing the `:url_title` param, so couldn't find the `InfoRequest` the classification was for.

## Screenshots

**BEFORE**

Screenshot 2020-05-15 at 16 52 58

Screenshot 2020-05-15 at 16 53 21

Screenshot 2020-05-15 at 16 53 25

**AFTER**

You see flash success message and get redirected back to the homepage (though soon to be redirected to another request after https://github.com/mysociety/alaveteli/pull/5669 is merged).

### DIFF
--- a/spec/controllers/projects/classifications_controller_spec.rb
+++ b/spec/controllers/projects/classifications_controller_spec.rb
@@ -39,7 +39,7 @@ RSpec.describe Projects::ClassificationsController, spec_meta do
     end
 
     context 'project to be classified can not be found' do
-      it'raises a ActiveRecord::RecordNotFound error' do
+      it 'raises a ActiveRecord::RecordNotFound error' do
         expect {
           post :create, params: { project_id: 'invalid' }
         }.to raise_error(ActiveRecord::RecordNotFound)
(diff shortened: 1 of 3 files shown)


spec/controllers/projects/classifications_controller_spec.rb
```diff
@@ -49,13 +49,23 @@
...
         }.to raise_error(ActiveRecord::RecordNotFound)
       end
     end
 
+    context 'url_title param not submitted' do
+      include_context 'project can be found'
+
+      it 'raises an ActionController::ParameterMissing error' do
+        expect {
+          post :create, params: { project_id: project.id, info_request_id: '1' }
```
garethrees, line 64: This spec is actually useless because we're stubbing `find_by!` – even though the parameter is missing we're not actually calling `url_title` because the stub happens first.

gbp, line 64:
```suggestion
          post :create, params: { project_id: project.id }
```

garethrees: Yeah guess it makes sense to just completely exclude it

PR #5676, opened 2020-05-19
Add public_network vagrant setting

garethrees:
Allow the VM to get get an IP address on the local network via DHCP.

This makes it easy to view the development instance from another device
on the network – your smartphone for example.

Only use this on networks you trust.

The public network address is _in addition to_ the private network
address.

See the Vagrant documentation for more information on this setting [1].

[1] https://www.vagrantup.com/docs/networking/public_network.html


garethrees:
@gbp you don't need to run this – it works – but just want to get your thoughts on this in principle.

PR #5679, opened 2020-05-20
[fnf#28] Queues

gbp:
## Relevant issue(s)

## What does this do?

1. Add Project#info_request.classifiable
2. Add Project#info_request.classified
3. Add Project#info_request.extractable
4. Add Project#info_request.extracted
5. Adds Project#extraction_progress

## Why was this needed?

In order to generate Project info request queues for contributors to work on.

## Implementation notes

~~Still need to add `extractable_requests` in some form like https://github.com/mysociety/transparency-fnf-whatdotheyknow-projects/issues/20#issuecomment-630893728~~

### DIFF
--- a/app/controllers/projects/extracts_controller.rb
+++ b/app/controllers/projects/extracts_controller.rb
@@ -4,6 +4,12 @@ class Projects::ExtractsController < Projects::BaseController
 
   def show
     authorize! :read, @project
+
+    unless @info_request
+      msg = _('There are no requests to extract right now. Great job!')
+      redirect_to @project, notice: msg
+      return
+    end
   end
 
   def create
--- a/app/controllers/projects/extracts_controller.rb
+++ b/app/controllers/projects/extracts_controller.rb
@@ -41,8 +41,6 @@ class Projects::ExtractsController < Projects::BaseController
   def extract_params
     params.require(:extract).permit(
       :dataset_key_set_id, values_attributes: [:dataset_key_id, :value]
-    ).merge(
-      resource: @info_request
     )
   end
 
--- a/app/controllers/projects/extracts_controller.rb
+++ b/app/controllers/projects/extracts_controller.rb
@@ -47,6 +47,10 @@ class Projects::ExtractsController < Projects::BaseController
   end
 
   def submission_params
-    { user: current_user, resource: Dataset::ValueSet.new(extract_params) }
+    {
+      user: current_user,
+      info_request: @info_request,
+      resource: Dataset::ValueSet.new(extract_params)
+    }
   end
 end
--- a/app/controllers/projects/extracts_controller.rb
+++ b/app/controllers/projects/extracts_controller.rb
@@ -29,12 +29,11 @@ class Projects::ExtractsController < Projects::BaseController
 
   def find_info_request
     if params[:url_title]
-      @info_request = @project.info_requests.find_by!(
+      @info_request = @project.info_requests.extractable.find_by!(
         url_title: params[:url_title]
       )
     else
-      # HACK: Temporarily just find a random request to render
-      @info_request = @project.info_requests.sample
+      @info_request = @project.info_requests.extractable.sample
     end
   end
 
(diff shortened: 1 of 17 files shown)


app/models/dataset/value_set.rb
```diff
@@ -29,7 +29,7 @@ class Dataset::ValueSet < ApplicationRecord
     FoiAttachment
   ].freeze
 
-  validates :resource, :key_set, presence: true
-  validates :resource_type, inclusion: { in: RESOURCE_TYPES }
+  validates :key_set, presence: true
+  validates :resource_type, inclusion: { in: RESOURCE_TYPES }, if: :resource
```
garethrees, line 33: If we're not requiring this, what's the reason for keeping the columns?

app/models/project/info_request_query.rb
```diff
@@ -10,17 +10,18 @@ def initialize(relation = InfoRequest)
     end
 
     def call(project)
+      table = ProjectResource.table_name
+
       @relation.joins(
```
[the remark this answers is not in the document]
gbp: I've changed this. Difference is this outputs returns in the Rails logs

app/models/dataset/value_set.rb
gbp: Left in because in Slack you said 
> I’d like to be able to add datasets to info requests outside of projects, so we’ll need to do it at some point

app/models/project/info_request_extension.rb
```diff
@@ -0,0 +1,25 @@
...
+  ##
+  # An ActiveRecord association extensions with extra InfoRequest scopes for
+  # projects
+  #
+  module InfoRequestExtension
+    def extracted
+      base_scope.merge(Project::Submission.extraction)
+    end
+
+    def classified
```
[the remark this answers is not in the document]
gbp: For now I've dropped these. Easy to add back if/when needed.

app/controllers/projects/extracts_controller.rb
```diff
@@ -4,6 +4,12 @@ class Projects::ExtractsController < Projects::BaseController
 
   def show
     authorize! :read, @project
+
+    unless @info_request
```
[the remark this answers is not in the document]
gbp: Ignoring as we've done for the classify controller.

app/models/project/info_request_extension.rb
```diff
@@ -11,5 +11,17 @@ def classifiable
     def classified
       where(awaiting_description: false)
     end
+
+    EXTRACTABLE_STATES = %w(successful partially_successful)
```
garethrees, line 15: Constants go at the top of the file https://rubystyle.guide/#consistent-classes

app/controllers/projects/extracts_controller.rb
```diff
@@ -29,12 +29,11 @@ def authenticate
...
   def find_info_request
     if params[:url_title]
-      @info_request = @project.info_requests.find_by!(
+      @info_request = @project.info_requests.extractable.find_by!(
         url_title: params[:url_title]
       )
     else
-      # HACK: Temporarily just find a random request to render
-      @info_request = @project.info_requests.sample
+      @info_request = @project.info_requests.extractable.first
```
garethrees, line 42: Should continue to use `sample` here because we don't want people landing on the same request all the time.

app/models/dataset/value_set.rb
garethrees: Ah, I meant architecturally lets not paint ourselves into a corner where it would be really difficult to do so. Might as well just keep the columns though as there's no harm in them being there.

garethrees:
> Conflicting files
> app/views/projects/projects/show.html.erb

Might as well rebase this on `develop` and add a commit to enable extracts again on the project homepage.

app/controllers/projects/extracts_controller.rb
gbp: Oops that was a mistake

PR #5685, opened 2020-05-21
[fnf#30] Hide correspondence footers

garethrees:
## Relevant issue(s)

https://github.com/mysociety/transparency-fnf-whatdotheyknow-projects/issues/30

## What does this do?

* Hide correspondence footers in projects
* Hide correspondence footers in followups

## Why was this needed?

* The footers don't make sense in the context of a project
* The footers are too long on mobile (https://github.com/mysociety/alaveteli/issues/5698)

### Projects Before

Screenshot 2020-06-03 at 08 50 37

### Projects After

Screenshot 2020-06-03 at 08 49 35

### Followup Before

Screenshot 2020-06-03 at 08 26 12

### Followup After

Screenshot 2020-06-03 at 08 26 21

### DIFF
--- a/app/views/comment/_single_comment.html.erb
+++ b/app/views/comment/_single_comment.html.erb
@@ -31,7 +31,7 @@
     </p>
   </div>
 
-  <% if comment.id && !@info_request.embargo %>
+  <% if comment.id && show_correspondence_footer %>
     <% link_to_this_url = comment_url(comment) %>
 
     <% report_path =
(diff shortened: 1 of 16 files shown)


garethrees:
> If I'm understanding it correctly there are 4 cases here

Yep, correct.

> Passing `show_correspondence_footer` around everywhere isn't ideal and confusing.

Yeah, on reflection I was wondering whether using ivars is clearer here – defaults as `nil`, so only have to set it where we want to display it, rather than _always_ having to set it.

> I would rather this logic to be in a decorator/helper (or evan as an ability) where it'll be more obvious what it happening and can be tested.

Yeah, this sounds like a better way forward.

Lots of the work here was just untangling things so that we can start thinking about how to do something better.

PR #5690, opened 2020-05-28
[fnf#28] classifications queue

garethrees:

### DIFF
--- a/app/controllers/projects/classifies_controller.rb
+++ b/app/controllers/projects/classifies_controller.rb
@@ -14,8 +14,6 @@ class Projects::ClassifiesController < Projects::BaseController
       return
     end
 
-    @queue.current(@info_request.id)
-
     @state_transitions = @info_request.state.transitions(
       is_pro_user: false,
       is_owning_user: false,
--- a/app/controllers/projects/classifies_controller.rb
+++ b/app/controllers/projects/classifies_controller.rb
@@ -5,13 +5,16 @@ class Projects::ClassifiesController < Projects::BaseController
   def show
     authorize! :read, @project
 
-    @info_request = @project.info_requests.classifiable.sample
+    @queue = Project::Queue::Classifiable.new(@project, current_user, session)
+    @info_request = @queue.next
 
     unless @info_request
       msg = _('There are no requests to classify right now. Great job!')
       redirect_to @project, notice: msg
       return
     end
+
+    @queue.current(@info_request.id)
 
     @state_transitions = @info_request.state.transitions(
       is_pro_user: false,
--- a/app/models/project/queue.rb
+++ b/app/models/project/queue.rb
@@ -14,10 +14,6 @@ class Project::Queue
     queue['current'] = info_request_id
   end
 
-  def clear_current
-    queue['current'] = nil
-  end
-
   def include?(info_request)
     info_requests.include?(info_request)
   end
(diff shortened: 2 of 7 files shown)


spec/models/project/queue_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+      it { is_expected.to eq(false) }
+    end
+
+    context 'with a different session' do
+      let(:other_queue) { described_class.new(project, current_user, double) }
+      it { is_expected.to eq(false) }
+    end
+  end
+
+  describe '#current' do
```
garethrees, line 44: The tests for `#current` and `#clear_current` are a bit meh.

They don't really test the _behaviour_ as such – just the return value. Feels like it's not really verifying much. Guess that's the general problem with mutating stuff

app/controllers/projects/classifies_controller.rb
```diff
@@ -5,14 +5,17 @@ class Projects::ClassifiesController < Projects::BaseController
...
+    @queue = Project::Queue::Classifiable.new(@project, current_user, session)
+    @info_request = @queue.next
 
     unless @info_request
       msg = _('There are no requests to classify right now. Great job!')
       redirect_to @project, notice: msg
       return
     end
 
+    @queue.current(@info_request.id)
```
garethrees, line 17: I think we can do this within `Project::Queue#next`, but it worked for now

app/models/project/queue.rb
```diff
@@ -0,0 +1,69 @@
...
+    @project = project
+    @current_user = current_user
+    @session = session
+  end
+
+  def next
+    find_current || sample
+  end
+
+  def current(info_request_id)
```
garethrees, line 12: I ummed and ahhed about this a little.

I started my test like:

```ruby
  describe '#current' do
    subject { queue.current = 1 }
    it { is_expected.to eq(1) }
  end
```

…and it was immediately passing when I implemented the method stub:

```ruby
def current=(info_request_id_
end
```

TIL that `=` methods always return the given value:

```
class Foo
  def bar=(x)
  end

  def bar2=(x)
    x * 20
  end

  def baz(x)
  end
end

f = Foo.new
f.bar = 1
# => 1

f.bar2 = 1
# => 1

f.baz(1)
# => nil
```

So basically my test is doing nothing.

Could dive in to the internals of `session` but that shouldn't really be exposed outside the class.

Anyway, the current test isn't doing much as mentioned, so could implement this as `current=` since the behaviour is tested through `#next`

spec/models/project/queue/classifiable_spec.rb
```diff
@@ -0,0 +1,69 @@
+require 'spec_helper'
+require 'models/project/queue_spec'
+
+RSpec.describe Project::Queue::Classifiable do
+  include_context 'Project::Queue context'
+
+  it_behaves_like 'Project::Queue'
+
+  describe '#next' do
```
garethrees, line 9: There'll probably be quite a lot of similarity when we implement `Project::Queue::Extractable`, but the results really depend on the specific requests that are in the queue. Open to ideas on extracting some of this into the shared examples.

garethrees:
Going to clean this up and create a new PR

garethrees:
Superseded by https://github.com/mysociety/alaveteli/pull/5691

PR #5691, opened 2020-05-28
[fnf#28] User-specific classifications queue

garethrees:
## Relevant issue(s)

## What does this do?

Adds a user-specific classifications queue

## Why was this needed?

As per initial feedback (https://github.com/mysociety/transparency-fnf-whatdotheyknow-projects/issues/35), if a contributor views an attachment in a request, then clicks back to go back and classify it, we actually get a different random request.

This preserves the current request the user is working on.

### DIFF
--- a/spec/controllers/projects/classifies_controller_spec.rb
+++ b/spec/controllers/projects/classifies_controller_spec.rb
@@ -75,6 +75,41 @@ RSpec.describe Projects::ClassifiesController, spec_meta do
       end
     end
 
+    context 'when there are only skipped requests to classify' do
+      let(:user) { FactoryBot.create(:user) }
+
+      before do
+        session[:user_id] = user.id
+
+        queue = Project::Queue::Classifiable.new(project, session)
+
+        project.info_requests.classifiable.each do |info_request|
+          queue.skip(info_request)
+        end
+
+        ability.can :read, project
+        get :show, params: { project_id: project.id }
+      end
+
+      it 'clears the skipped queue' do
+        skipped_requests =
+          session['projects'][project.to_param]['classifiable']['skipped']
+
+        expect(skipped_requests).to be_empty
+      end
+
+      it 'asks the user to have another go at the skipped requests' do
+        msg = 'Nice work! How about having another try at the requests you ' \
+              'skipped?'
+
+        expect(flash[:notice]).to eq(msg)
+      end
+
+      it 'redirects back to the project homepage' do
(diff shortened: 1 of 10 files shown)


app/models/project/queue.rb
```diff
@@ -0,0 +1,84 @@
+# Abstract class for Project task queues.
+# Gives the logged in User a new InfoRequest to contribute to.
+#
+# Subclasses must implement #info_requests.
+class Project::Queue
+  def initialize(project, current_user, session)
+    @project = project
+    @current_user = current_user
```
garethrees: Just noticed I don't actually use `current_user` after some refactoring, so I'll drop this.

app/models/project/queue.rb
```diff
@@ -0,0 +1,86 @@
...
+      @prime_session ||= prime_session!
+    end
+
+    def prime_session!
+      session['projects'] ||= {}
+      session['projects'][project.id.to_s] ||= {}
+      session['projects'][project.id.to_s][queue_name] ||= {}
+      session['projects'][project.id.to_s][queue_name]['current'] ||= nil
+      true
+    end
```
gbp, line 75: Actually these prime_session methods could be dropped - if we did:

```suggestion
    def queue
      session.fetch('projects', {}).
              fetch(project.to_param, {}).
              fetch(queue_name, {})
    end
```

spec/controllers/projects/classifies_controller_spec.rb
```diff
@@ -33,6 +33,23 @@
...
+      end
+
+      it 'assigns an info_request from the queue' do
+        queue = Project::Queue::Classifiable.new(project, user, session)
+        expect(queue).to include(assigns[:info_request])
+      end
+
+      it 'remembers the current request' do
+        current_request_id =
+          session['projects'][project.id.to_s]['classifiable']['current']
```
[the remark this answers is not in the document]
garethrees: Nerp :( 

```
  1) Projects::ClassifiesController GET #show with a logged in user who can read the project remembers the current request
     Failure/Error: session.dig('projects', project.to_param, 'classifiable', 'current')

     NoMethodError:
       undefined method `dig' for #<ActionController::TestSession:0x000055e84b0d6360>
```

app/models/project/queue.rb
garethrees: Doesn't pass specs and couldn't quite figure out what was wrong. Can refactor this later

PR #5692, opened 2020-05-29
[fnf#28] user queue skip

garethrees:

### DIFF
--- a/app/views/projects/classifies/_sidebar.html.erb
+++ b/app/views/projects/classifies/_sidebar.html.erb
@@ -7,6 +7,8 @@
                          project: project,
                          state_transitions: state_transitions } %>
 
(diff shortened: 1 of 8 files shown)


app/views/projects/classifies/_sidebar.html.erb
```diff
@@ -7,6 +7,8 @@
                          project: project,
                          state_transitions: state_transitions } %>
 
+    <%= link_to _('Skip'), project_classify_path(project, url_title: info_request.url_title), method: :patch, class: 'button-tertiary button--full-width form__skip-button' %>
```
garethrees, line 10: oops need to trim the line length of this and remove the JS-based button (and the skip JS while I'm here)

garethrees:
Merged into https://github.com/mysociety/alaveteli/pull/5691

PR #5704, opened 2020-06-05
Remove rack-utf8_sanitizer

garethrees:
## What does this do?

Remove `rack-utf8_sanitizer`

## Why was this needed?

Specs still pass, even without rack-utf8-sanitizer.

## Implementation notes

Should dig in to this a little before merging – has rails included it by default, etc?


garethrees:
Related https://github.com/mysociety/alaveteli/commit/e96e756c43a6b0a81d917a4f093119a51d9a4115

Issue #5069, opened 2019-01-23
Upgrade to Rails 5.2


garethrees:
https://guides.rubyonrails.org/5_2_release_notes.html

Requires https://github.com/mysociety/alaveteli/issues/3970

gbp:
Useful: http://railsdiff.org/5.1.7/5.2.0

gbp:
Failing spec since 5.2.4:

```
Finished in 0.80882 seconds (files took 9.46 seconds to load)
1 example, 1 failure

Failed examples:

rspec ./spec/controllers/admin_public_body_categories_controller_spec.rb:127 # AdminPublicBodyCategoriesController POST create on success saves the public body category's heading associations
```

Using git bisect the upstream change responsible is https://github.com/rails/rails/pull/35799

PR #5707, opened 2020-06-05
Fix category header association inverse of option

gbp:
## Relevant issue(s)

Connected to #5069 

## What does this do?

Fix category header association inverse of option

## Why was this needed?

This was causing an error [1] on Rails 5.2.4 after a fix implemented
upstream [2].

[1] https://github.com/mysociety/alaveteli/issues/5069#issuecomment-639372587
[2] https://github.com/rails/rails/pull/35799


gbp:
Have done a quick check of other association `inverse_of` option - all look correct.

PR #5709, opened 2020-06-05
[#5624] Ignore spam address case

### LINKED ISSUE #5624: Spam addresses list should ignore case

garethrees:
## Relevant issue(s)

Fixes #5624
Supersedes https://github.com/mysociety/alaveteli/pull/5634

## What does this do?

* Ignore case when checking spam in SpamAddress
* Minor code style cleanup

## Why was this needed?

Reduce admin time by not having to create a `SpamAddress` for each case that spammers use.

## Implementation notes

https://github.com/mysociety/alaveteli/pull/5634#discussion_r422912006 suggests using Postgres `citext` extension.

Unfortunately, enabling requires either the database user being a superuser, or manual intervention. Neither is a great option, so this keeps things simple from an operations point of view. I'm happy with that given this is such a small class. We can revisit if we need similar behaviour in a more complex or performance-dependent class.

### DIFF
--- a/app/models/spam_address.rb
+++ b/app/models/spam_address.rb
@@ -16,6 +16,8 @@ class SpamAddress < ApplicationRecord
   validates_uniqueness_of :email,
                           message: 'This address is already marked as spam'
 
+  strip_attributes
+
   before_save :downcase_email
 
   def self.spam?(email_address)
--- a/app/models/spam_address.rb
+++ b/app/models/spam_address.rb
@@ -13,8 +13,15 @@ class SpamAddress < ApplicationRecord
   validates_presence_of :email, :message => 'Please enter the email address to mark as spam'
   validates_uniqueness_of :email, :message => 'This address is already marked as spam'
 
+  before_save :downcase_email
+
   def self.spam?(email_address)
-    exists?(:email => email_address)
+    exists?(email: Array(email_address).compact.map(&:downcase))
   end
 
+  private
+
+  def downcase_email
+    email.downcase!
+  end
 end
--- a/app/models/spam_address.rb
+++ b/app/models/spam_address.rb
@@ -10,8 +10,11 @@
 #
 
 class SpamAddress < ApplicationRecord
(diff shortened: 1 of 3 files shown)


app/models/spam_address.rb
```diff
@@ -14,7 +14,10 @@ class SpamAddress < ApplicationRecord
   validates_uniqueness_of :email, :message => 'This address is already marked as spam'
 
   def self.spam?(email_address)
-    exists?(:email => email_address)
+    exists?(email: Array(email_address).compact.map(&:downcase))
   end
 
+  def email=(email)
+    write_attribute(:email, email.downcase)
```
gbp:
```suggestion
    super(email.downcase)
```

app/models/spam_address.rb
```diff
@@ -10,8 +10,11 @@
 #
 
 class SpamAddress < ApplicationRecord
-  validates_presence_of :email, :message => 'Please enter the email address to mark as spam'
-  validates_uniqueness_of :email, :message => 'This address is already marked as spam'
+  validates_presence_of :email,
+                        message: 'Enter the email address to mark as spam'
+
+  validates_uniqueness_of :email,
+                          message: 'This address is already marked as spam'
```
gbp, line 17: Commit message says "Quote style" but these look the same.

app/models/spam_address.rb
```diff
@@ -21,6 +21,6 @@ def self.spam?(email_address)
   end
 
   def email=(email)
-    write_attribute(:email, email.downcase)
+    write_attribute(:email, email.downcase.strip)
```
gbp, line 24: Could use the `strip_attributes` gem for this. Would need to call `valid?` in the specs tho.

PR #5750, opened 2020-06-12
[fnf#34] Project data download

gbp:
## Relevant issue(s)

Closes https://github.com/mysociety/transparency-fnf-whatdotheyknow-projects/issues/34

## What does this do?

Allows project owners to download data collected by the project classifications and extractions.

## Why was this needed?

Only other way to do this was with developer help.

## Implementation notes

Duplication around CSV generation with the Info Request Batch downloads - ultimately this should be extracted into its own service object.

### DIFF
--- a/app/models/project/export.rb
+++ b/app/models/project/export.rb
@@ -4,6 +4,8 @@
 class Project::Export
   require_dependency 'project/export/info_request'
 
+  include DownloadHelper
+
   attr_reader :project
   protected :project
 
@@ -14,6 +16,23 @@ class Project::Export
   def data
     @data ||= project.info_requests.extracted.map do |info_request|
       Project::Export::InfoRequest.new(project, info_request).data
+    end
+  end
+
+  def name
+    generate_download_filename(
+      resource: 'project',
+      id: project.id,
+      title: project.title,
+      ext: 'csv'
+    )
+  end
+
+  def to_csv
+    CSV.generate do |csv|
+      header = data.first
+      csv << header.keys.map(&:to_s) if header
+      data.each { |row| csv << row.values }
     end
   end
 end
--- a/app/models/project/export.rb
+++ b/app/models/project/export.rb
@@ -1,0 +1,19 @@
+##
+# Export a project's info request classifications and data extractions to CSV.
+#
+class Project::Export
+  require_dependency 'project/export/info_request'
+
+  attr_reader :project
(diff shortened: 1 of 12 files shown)


app/models/project/export/info_request.rb
```diff
@@ -0,0 +1,60 @@
+##
+# Export key classification and extracted data for an info request within a
+# given project.
+#
+class Project::Export::InfoRequest
+  extend Forwardable
+
+  attr_reader :project, :info_request
```
garethrees, line 8: Should make these [`private`/`protected`](https://www.garethrees.co.uk/2020/06/14/use-protected-attr-readers/#why-not-use-public-accessors).

app/models/project/export/info_request.rb
```diff
@@ -0,0 +1,60 @@
...
+
+  def initialize(project, info_request)
+    @project = project
+    @info_request = info_request
+  end
+
+  def data
+    {
+      request_url: info_request.url_title,
```
garethrees, line 17: Should be the full URL rather than just the slug

app/models/project/export.rb
```diff
@@ -0,0 +1,20 @@
+##
+# Export a project's info request classifications and data extractions to CSV.
+#
+class Project::Export
+  require_dependency 'project/export/info_request'
```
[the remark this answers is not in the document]
gbp: If it was outside the class then we get:

```
[1] pry(main)> cd Project::Export
Error: Bad object path: "Project::Export"
Failed trying to resolve: "Project::Export"
Exception: #<RuntimeError: Circular dependency detected while autoloading constant Project::Export>
```

app/models/project/export.rb
```diff
@@ -0,0 +1,20 @@
...
+
+  extend Forwardable
+
+  attr_reader :project
```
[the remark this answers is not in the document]
gbp: I've changed this and removed the spec checking `#project`

app/models/project/export/info_request.rb
gbp: As above I've changed this. Also switched the class to inherit from SimpleDelegator to remove #info_request completely

app/models/project/export/info_request.rb
gbp: Doh, of course changed!

PR #5752, opened 2020-06-16
Drop NewRelic

gbp:
## Relevant issue(s)

https://github.com/mysociety/sysadmin/issues/1327
https://github.com/mysociety/sysadmin/issues/1305
https://github.com/mysociety/sysadmin/issues/1316

Closes https://github.com/mysociety/alaveteli/pull/5729

## What does this do?

Removes NewRelic gem and integration

## Why was this needed?

Cron tasks and scripts not exiting correctly due to errors communicating with NewRelic - we don't really use NR so opting to drop for now.


garethrees:
Before I review, will also want:

- [ ] Changelog entry
- [ ] Ticket/s for removing from [privacy page](https://www.whatdotheyknow.com/help/privacy#cookies) (not sure if other alaveteli themes include this)

PR #5779, opened 2020-06-29
Fix gettext rake commands

gbp:
## Relevant issue(s)

https://github.com/mysociety/alaveteli/pull/5711#issuecomment-650079260

## What does this do?

Updates our gettext rake tasks to be compatible with gettext v3

## Why was this needed?

Running `gettext:clean` and `gettext:find` was broken so we can't do a release at the moment. 

## Implementation notes

There will be some differences in PO files because previously, in our strings we had `\\n` this apparently was wrong and fixed upstream so these are now replaced as `\n`, as a result, this means:
1. The `no-wrap` doesn't wrap new lines any more - because it is actually detecting the new lines
2. The ordering has slightly changed

### DIFF
--- a/app/views/alaveteli_pro/comment/_suggestions.html.erb
+++ b/app/views/alaveteli_pro/comment/_suggestions.html.erb
@@ -6,7 +6,7 @@
 <ul>
   <% if [ 'successful', 'partially_successful' ].include?(@info_request.described_state) %>
     <li><%= _(' <strong>Summarising</strong> the content of any information returned. ')%></li>
-    <li><%= _(" Say how you\'ve <strong>used the information</strong>, with " \
+    <li><%= _(" Say how you've <strong>used the information</strong>, with " \
                 "links if possible.")%> </li>
     <li><%= _("<strong>Thank</strong> the public authority.") %></li>
     <li> <%= _("Point to <strong>related information</strong>, campaigns or " \
--- a/lib/tasks/gettext.rake
+++ b/lib/tasks/gettext.rake
@@ -10,20 +10,17 @@ namespace :gettext do
 
   desc 'Rewrite .po files into a consistent msgmerge format'
   task :clean do
-    load_gettext
     clean_dir('locale')
   end
 
   desc 'Rewrite Alaveteli Pro .po files into a consistent msgmerge format'
   task :clean_alaveteli_pro do
-    load_gettext
     clean_dir('locale_alaveteli_pro')
   end
 
   desc "Update pot/po files for a theme."
   task :find_theme => :environment do
     theme = find_theme(ENV['THEME'])
-    load_gettext
     msgmerge = Rails.application.config.gettext_i18n_rails.msgmerge
     msgmerge ||= %w[--sort-output --no-location --no-wrap]
     GetText.update_pofiles_org(
@@ -37,7 +34,6 @@ namespace :gettext do
 
   desc "Update pot/po files for Alaveteli Pro."
   task :find_alaveteli_pro => :environment do
-    load_gettext
     msgmerge = Rails.application.config.gettext_i18n_rails.msgmerge
     msgmerge ||= %w[--sort-output --no-location --no-wrap]
     GetText.update_pofiles_org(
@@ -52,7 +48,6 @@ namespace :gettext do
   desc 'Rewrite theme .po files into a consistent msgmerge format'
   task :clean_theme do
     theme = find_theme(ENV['THEME'])
-    load_gettext
     clean_dir(theme_locale_path(theme))
   end
 
--- a/lib/tasks/gettext.rake
+++ b/lib/tasks/gettext.rake
@@ -33,8 +33,7 @@ namespace :gettext do
 
   desc 'Rewrite theme .po files into a consistent msgmerge format'
   task :clean_theme do
-    theme = find_theme(ENV['THEME'])
-    clean(root: theme_locale_path(theme))
+    clean(root: theme_locale_path)
   end
 
   def xgettext(pot_file, *files)
@@ -70,15 +69,14 @@ namespace :gettext do
     find(files: files_to_translate, root: locale_path)
   end
 
-  desc "Update pot/po files for a theme."
-  task :find_theme => :environment do
-    theme = find_theme(ENV['THEME'])
-    find(files: theme_files_to_translate(theme), root: theme_locale_path(theme))
-  end
-
   desc "Update pot/po files for Alaveteli Pro."
   task :find_alaveteli_pro => :environment do
     find(files: pro_files_to_translate, root: pro_locale_path)
+  end
+
+  desc "Update pot/po files for a theme."
+  task :find_theme => :environment do
+    find(files: theme_files_to_translate, root: theme_locale_path)
   end
 
   desc 'Remove fuzzy translations'
@@ -138,14 +136,6 @@ namespace :gettext do
     data
   end
 
-  def find_theme(theme)
-    unless theme
-      puts "Usage: Specify an Alaveteli-theme with THEME=[theme directory name]"
-      exit(1)
-    end
-    theme
-  end
-
   def find_mapping_file(file)
     unless file
       puts "Usage: Specify a csv file mapping old to new strings with MAPPING_FILE=[file name]"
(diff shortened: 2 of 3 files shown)


gbp:
After running
```
rails gettext:clean gettext:find gettext:remove_fuzzy gettext:clean
```

and
```
rails gettext:clean_alaveteli_pro gettext:find_alaveteli_pro gettext:remove_fuzzy_alaveteli_pro gettext:clean_alaveteli_pro
```

The differences I see are: https://gist.github.com/gbp/dc64bebd09d4a00401b6d33c8a4cff70

lib/tasks/gettext.rake
```diff
@@ -1,21 +1,23 @@
 # -*- encoding : utf-8 -*-
 namespace :gettext do
-  def msgmerge
-    Rails.application.config.gettext_i18n_rails.msgmerge ||
```
gbp: I attempted to set this config in the initializers but not all these rake tasks actually load the full Rails application so these options need to be defined here anyway.

lib/tasks/gettext.rake
```diff
@@ -1,21 +1,23 @@
...
-      %w[--sort-output --no-location --no-wrap]
+  def msgmerge(*files)
+    destination = files.first
+    options = %w[--sort-output --no-location --no-wrap --no-obsolete-entries]
```
gbp, line 5: Added `no-obsolete-entries` to prevent warnings in STDOUT

lib/tasks/gettext.rake
```diff
@@ -31,18 +31,26 @@ namespace :gettext do
...
-    pot_temp = Tempfile.new(pot_file)
-    pot_temp_path = pot_temp.path
+  def xgettext(pot_file, *files)
+    output = Tempfile.new(pot_file)
+    output_path = output.path
 
     # find new strings and write to temp file
-    GetText::Tools::XGetText.run('--output', pot_temp_path, *files)
+    GetText::Tools::XGetText.run(
+      '--add-comments=TRANSLATORS', '--output', output_path, *files
```
gbp, line 45: Added `add-comments` to retain the existing helpful comment for translators

lib/tasks/gettext.rake
```diff
@@ -6,9 +6,16 @@ namespace :gettext do
   end
 
   def clean_dir(dir)
-    Dir.glob("#{dir}/*/app.po") do |po_file|
-      GetText::msgmerge(po_file, po_file, 'alaveteli',
-                        :msgmerge => [:sort_output, :no_location, :no_wrap])
```
garethrees, line 7: > 1. a bug where new lines where outputted as "\\n" has now been fixed
> …
> This means all our PO files will see updates next time translations are
> imported.

Need to understand what to do about this

lib/tasks/gettext.rake
```diff
@@ -29,6 +29,18 @@ namespace :gettext do
     clean_dir('locale_alaveteli_pro')
   end
 
+  Rake::Task['find'].clear
```
garethrees: Why doesn't the `gettext_i18n_rails`-provided task work?

lib/tasks/gettext.rake
```diff
@@ -29,39 +29,31 @@ namespace :gettext do
     clean_dir('locale_alaveteli_pro')
   end
 
-  Rake::Task['find'].clear
```
garethrees: 👍

lib/tasks/gettext.rake
```diff
@@ -30,13 +30,34 @@ namespace :gettext do
   end
 
   def find(files:, root:)
```
garethrees: Commit message doesn't say what was broken or why this fixes them, so it's pretty difficult to understand what to review.

lib/tasks/gettext.rake
garethrees: 👍   – squash into commit message

lib/tasks/gettext.rake
garethrees: 👍  – squash into commit message

garethrees: might as well merge `--output` flag into these now?

lib/tasks/gettext.rake
```diff
@@ -1,21 +1,23 @@
 # -*- encoding : utf-8 -*-
 namespace :gettext do
-  def msgmerge
```
garethrees: 👍  with a minor question/comment

lib/tasks/gettext.rake
garethrees: 👍  – squash into commit message

lib/tasks/gettext.rake
```diff
@@ -31,18 +31,26 @@ namespace :gettext do
     clean_dir('locale_alaveteli_pro')
   end
 
-  def find(files:, root:)
```
garethrees: 👍

lib/tasks/gettext.rake
```diff
@@ -14,21 +14,27 @@ namespace :gettext do
     output.close!
   end
 
-  def clean_dir(dir)
```
garethrees, line 4: 👍

lib/tasks/gettext.rake
```diff
@@ -33,8 +33,7 @@ namespace :gettext do
 
   desc 'Rewrite theme .po files into a consistent msgmerge format'
   task :clean_theme do
-    theme = find_theme(ENV['THEME'])
```
garethrees: 👍

app/views/alaveteli_pro/comment/_suggestions.html.erb
```diff
@@ -6,7 +6,7 @@
 <ul>
   <% if [ 'successful', 'partially_successful' ].include?(@info_request.described_state) %>
     <li><%= _(' <strong>Summarising</strong> the content of any information returned. ')%></li>
-    <li><%= _(" Say how you\'ve <strong>used the information</strong>, with " \
+    <li><%= _(" Say how you've <strong>used the information</strong>, with " \
```
garethrees, line 9: Since we're modifying these strings, might as well remove the empty space at the start.

Might as well also do so for  `' <strong>Summarising</strong> the content of any information returned. '` too.

Applies to both files in this diff.

gbp: This change is actually to retain the same string as before. Without this change backward slash is now escaped. so the string appears different in the POT/PO files.

I beleive this could be due to same escaping bug which effected the new lines.

lib/tasks/gettext.rake
gbp: It does work but it will:
1. leave `*.edit.po` and `*.time_stamp.po` files in the locale directory
2. can't easily be re-used for the pro/theme find tasks

lib/tasks/gettext.rake
gbp: As you'll see on https://gist.github.com/gbp/dc64bebd09d4a00401b6d33c8a4cff70 strings with newlines appear as being changed.

This as far as I can see is the correct format for PO files now. What we had previously worked and may well continue to work so if we really want to `no-wrap` for the whole PO then we could regexp them out here: https://github.com/mysociety/alaveteli/pull/5779/commits/ed4ca07a861be6c22fc0d00c7a9316c4727bb2d8#diff-faeefb9d492cbafa96677b194d99f752R17

lib/tasks/gettext.rake
gbp: I'll expand on the commit message but basically as the same as https://github.com/mysociety/alaveteli/pull/5779/commits/ed4ca07a861be6c22fc0d00c7a9316c4727bb2d8 where upstream removed the helper methods we were using.

lib/tasks/gettext.rake
garethrees: Yeah seems worth doing the Right Thing, but just want to understand/check if there are any implications we'll care about, or whether transifex will handle it all transparently.

lib/tasks/gettext.rake
gbp: We could but the `--output` flag requires a value. So the command is evaluated to something like:
`msgmerge *our options* --output=$(output.path) *our files*` but without calling the command because its all done in Ruby code.

I intentional left it out so it can be next to the actual output path.

app/views/alaveteli_pro/comment/_suggestions.html.erb
garethrees: Ah okay gotcha. In that case yeah lets just leave it as it is for now

lib/tasks/gettext.rake
garethrees: Ah! Okay yeah makes more sense as is 👍

lib/tasks/gettext.rake
garethrees: 👍

garethrees:
Just assigning back to @gbp for testing pushing a locale to transifex.

gbp:
Tested with the Welsh locale...

In transifex, before pushing we would see `\n` in both the source text and translated text :
Screenshot 2020-07-13 at 17 24 30

Now we now see:
Screenshot 2020-07-13 at 17 14 20

It also warns if the string should start with a new line:
Screenshot 2020-07-13 at 17 17 54

Most importantly we can pull from Transifex without conflicts.

gbp:
Pushed a rebase over develop and updated commit messages as requested.

PR #5784, opened 2020-07-03
Remove delorean

garethrees:
Rails now has time helpers built in [1] and Delorean is
discontinued [2].

This commit adds Alaveteli::TimeHelpers to maintain Delorean’s DSL to
make it easier to transition.

The notable difference is that Rails has two methods that work slightly
differently.

`travel` only accepts an `ActiveSupport::Duration`, but allows a block.

`travel_to` accepts a `date_or_time`, but with no block.

Our pattern with Delorean has been to provide a `date_or_time` _and_ a
block, so I’ve re-implemented `Delorean#time_travel_to` [3] using Rails’
helpers.

`travel_back` does the same as `back_to_the_present` so I've just
replaced that everywhere its used.

Also removes a couple of direct calls of `Delorean`.

[1] https://api.rubyonrails.org/v5.2.4/classes/ActiveSupport/Testing/TimeHelpers.html
[2] https://github.com/bebanjo/delorean#discontinued
[3] https://github.com/bebanjo/delorean/blob/v2.0.0/lib/flux_capacitor.rb#L3-L11


gbp:
Superseded by #5813

Issue #3546, opened 2016-10-17
request#get_attachment_as_html (Encoding::CompatibilityError) "incompatible character encodings: UTF-8 and ASCII-8BIT


crowbot:
From derechoapreguntar (looks similar to https://github.com/mysociety/alaveteli/issues/2641, but seems to postdate the bugfix for that):

```
-------------------------------
Backtrace:
-------------------------------

  (erb):12:in `concat'
  (erb):12:in `render'
  /usr/lib/ruby/1.9.1/erb.rb:838:in `eval'
  /usr/lib/ruby/1.9.1/erb.rb:838:in `result'
  lib/attachment_to_html/view.rb:25:in `render'
  lib/attachment_to_html/attachment_to_html.rb:23:in `to_html'
  app/models/foi_attachment.rb:304:in `body_as_html'
  app/controllers/request_controller.rb:827:in `get_attachment_as_html'
  vendor/bundle/ruby/1.9.1/gems/actionpack-3.2.21/lib/action_controller/metal/implicit_render.rb:4:in `send_action'
  vendor/bundle/ruby/1.9.1/gems/actionpack-3.2.21/lib/abstract_controller/base.rb:167:in `process_action'
  vendor/bundle/ruby/1.9.1/gems/actionpack-3.2.21/lib/action_controller/metal/rendering.rb:10:in `process_action'
  vendor/bundle/ruby/1.9.1/gems/actionpack-3.2.21/lib/abstract_controller/callbacks.rb:18:in `block in process_action'
  vendor/bundle/ruby/1.9.1/gems/activesupport-3.2.21/lib/active_support/callbacks.rb:603:in `block (2 levels) in _run__561652700549859565__process_action__4512618588931439__callbacks'
  vendor/bundle/ruby/1.9.1/gems/activesupport-3.2.21/lib/active_support/callbacks.rb:215:in `block in _conditional_callback_around_1201'
  app/controllers/request_controller.rb:777:in `cache_attachments'
  vendor/bundle/ruby/1.9.1/gems/activesupport-3.2.21/lib/active_support/callbacks.rb:214:in `_conditional_callback_around_1201'
  vendor/bundle/ruby/1.9.1/gems/activesupport-3.2.21/lib/active_support/callbacks.rb:602:in `block in _run__561652700549859565__process_action__4512618588931439__callbacks'
[...]

PR #5821, opened 2020-07-14
[#3546] Identify broken binary censor rules

garethrees:
## Relevant issue(s)

#3546

## What does this do?

Adds a temporary rake task to check for attachments with censor rules that will be visible after fixing #3546.

## Why was this needed?

Admins couldn't check that censor rules were being correctly applied to the outputted attachments because the application was erroring. After merging #5474 these attachments will now be visible, but we don't have a way of automatically verifying that the censor rules are being applied correctly. This prints a list of URLs for admins to manually verify.

## Implementation notes

Added the monkey-patches from https://github.com/mysociety/alaveteli/issues/3546#issuecomment-658051334 to the production codebase because they're useful, and it's more stable than forgetting about them buried in a rake task.

### DIFF
--- a/lib/tasks/temp.rake
+++ b/lib/tasks/temp.rake
@@ -456,4 +456,59 @@ namespace :temp do
       Role.where(name: 'notifications_tester').destroy_all
     end
   end
+
+  desc 'Identify broken binary censor rules'
+  task identify_broken_binary_censor_rules: :environment do
+    helper = ApplicationController.helpers
+    helper.class_eval { include Rails.application.routes.url_helpers }
+
+    module ConfigHelper
+      def send_exception_notifications?
+        false
+      end
+    end
+
+    # 0.37.0.0 broken implementation of CensorRule#apply_to_binary
+    # Need to monkeypatch it here to cause the error when searching the affected
+    # attachments.
+    CensorRule.class_eval do
+      def apply_to_binary(binary_to_censor)
+        return nil if binary_to_censor.nil?
+        binary_to_censor.gsub(to_replace('ASCII-8BIT')) do |match|
+          match.gsub(single_char_regexp, 'x')
+        end
+      end
+    end
+
+    ApplicationController.allow_forgery_protection = false
+    app = ActionDispatch::Integration::Session.new(Rails.application)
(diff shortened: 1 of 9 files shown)


garethrees:
Monkeypatch old functionality into this task

lib/tasks/temp.rake
```diff
@@ -456,4 +456,54 @@ namespace :temp do
...
+
+        info_request.foi_attachments.binary.find_each do |attachment|
+          params =
+            helper.
+            send(:attachment_params, attachment, html: true, only_path: true)
+
+          path = helper.get_attachment_as_html_url(params)
+          protocol = AlaveteliConfiguration.force_ssl ? 'https' : 'http'
+          domain = AlaveteliConfiguration.domain
+          url = "#{ protocol }://#{ domain }#{ path }"
```
gbp, line 498:
```suggestion
          url = "#{protocol}://#{domain}#{path}?skip_cache=1"
```

Need to skip the cache otherwise once merged with #5474 some URLs won't be returned if successful response is already cached.

garethrees: Ah good point. Will add a randomised cache key

lib/tasks/temp.rake
```diff
@@ -495,7 +495,7 @@ namespace :temp do
           path = helper.get_attachment_as_html_url(params)
           protocol = AlaveteliConfiguration.force_ssl ? 'https' : 'http'
           domain = AlaveteliConfiguration.domain
-          url = "#{ protocol }://#{ domain }#{ path }"
```
garethrees: should be randomised

lib/tasks/temp.rake
```diff
@@ -495,7 +495,7 @@ namespace :temp do
...
+          url = "#{protocol}://#{domain}#{path}?skip_cache=1"
```
garethrees, line 499: I _think_ we should generate a random value for the `skip_cache` param.

Since we're hitting the app externally, the error page might get cached by varnish, so if you then run it again you'll always get a cached response?

gbp: I've done this a rebased.

PR #5842, opened 2020-07-30
[#4853] Fix issue where locale strings wouldn't be loaded

### LINKED ISSUE #4853: Underscore locale translations are broken by i18n v0.9.3+
i18n v0.9.3 introduced [a feature to stop storing translations which are not explicitly enabled](https://github.com/svenfuchs/i18n/commit/5077ef9f615f224afd26b00d626d78e35496f89f#diff-578120f672bb08aa3edae0031b93a21c) (later softened to skipping this if `enforce_available_locales` is false). This appears to have unintended consequences for hyphen locales as when calling `I18n.t` or `I18n.l` methods, it behaves as if it's falling back to the `:fr` locale data when passed `:fr-BE`, then failing as `:fr` is not enabled in our config, preventing the translations from being stored in memory.

(There are no specific configs for fr-BE in https://github.com/svenfuchs/rails-i18n/tree/master/rails/locale so it's going to try to use fr as a fallback, which is not loaded because we've not specified it)

```rb
I18n.available_locales
=> [:"fr-BE", :"nl-BE", :en]

I18n.locale
=> :"fr-BE"

I18n.translate('date.abbr_month_names')
=> "translation missing: fr-BE.date.abbr_month_names"
```

i18n update introduced in cafcd998

(Note - if this does get fixed in a later version of i18n, we will either need a backport or to wait to resolve it until we can upgrade Rails as `rails-i18n (< 5.0.0, ~> 4.0.0)` currently constrains us to the `< 1.0.0` branch)

Blocker for #4615

gbp:
## Relevant issue(s)

Fixes #4853

## What does this do?

When upgrading the i18n gem past 0.9.3+ [1] we re-introduced a bug [2]
where some translations wouldn't be loaded into memory.

Locales are loaded based on what `I18n.available_locales` has been set
to. This works fine for two-letter locales such as `en` or `cy` but
can break for locales with underscores or hyphens. For example both
`fr_BE` and `is_IS` are broken as they will attempt to load translations
from the rails-i18n gem locale YAML files [3], where these locales with
don't exist so the parent locale will be used, `fr` and `is`.

The problem is these parent locales haven't been loaded as they aren't
set in `I18n.available_locales` so this commit adds parent locales to be
loaded.

Specs have been added to ensure this isn't overlooked again.

[1] https://github.com/mysociety/alaveteli/commit/b836119
[2] https://github.com/mysociety/alaveteli/issues/4853
[3] https://github.com/svenfuchs/rails-i18n/tree/master/rails/locale

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -38,7 +38,7 @@ module Alaveteli
     # The default locale is :en and all translations from config/locales/*.rb,yml are auto loaded.
     # config.i18n.load_path += Dir[Rails.root.join('my', 'locales', '*.{rb,yml}').to_s]
     # config.i18n.default_locale = :de
-    I18n.config.enforce_available_locales = true
+    config.i18n.enforce_available_locales = true
 
     # Configure the default encoding used in templates for Ruby 1.9.
     config.encoding = "utf-8"
--- a/config/environments/test.rb
+++ b/config/environments/test.rb
@@ -38,6 +38,7 @@ Rails.application.configure do
   # Enable locale fallbacks for I18n (makes lookups for any locale fall back to
   # the I18n.default_locale when a translation cannot be found).
   config.i18n.fallbacks = true
+  config.i18n.enforce_available_locales = false
 
   # Print deprecation notices to the stderr.
   config.active_support.deprecation = :stderr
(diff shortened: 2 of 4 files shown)


lizconlan:
Sorry for not finishing the i18n patch (but my first draft solution gave me a headache and I never found time to finish it and … then I forgot 😞 )

https://github.com/lizconlan/i18n/pull/2

garethrees:
> Sorry for not finishing the i18n patch (but my first draft solution gave me a headache and I never found time to finish it and … then I forgot 😞 )
> 
> [lizconlan/i18n#2](https://github.com/lizconlan/i18n/pull/2)

@lizconlan no problem! Plenty more headaches from this on our end too 🍦

PR #5847, opened 2020-08-03
Ignore capistrano gem bumps

garethrees:
We're unlikely to support capistrano 3.x.x as we don't use it
internally, so ignore gem bumps until we decide what to do about it.

Supersedes https://github.com/mysociety/alaveteli/pull/5796.


garethrees:
@gbp I think this is the right thing to do now for ignoring gem bumps?

Issue #5857, opened 2020-08-10
Check that required crontab entries and daemons are generated or documented


garethrees:
- [x] Crontab
- [x] Pop Poller Daemon
- [x] Notifications Daemon

garethrees:
[The example crontab](https://github.com/mysociety/alaveteli/blob/0.38.0.0/config/crontab-example) has all the necessary scripts in place.

We already have documentation for generating the [POP poller daemon(https://github.com/mysociety/alaveteli/blob/5f242f62e4d0cba46cd35f7e4835fa2b4ff871db/docs/installing/manual_install.md#generate-mail-poller-daemon-optional).

Need to add documentation on generating [the notifications daemon](https://github.com/mysociety/alaveteli/blob/0.38.0.0/config/send-notifications-debian.example).

PR #5859, opened 2020-08-10
[#5857] Document Notifications Daemon

garethrees:
## Relevant issue(s)

Fixes https://github.com/mysociety/alaveteli/issues/5857

## What does this do?

* Extract cron & daemon setup instructions to a separate page
* Add notifications daemon documentation

## Why was this needed?

Help reusers enable Pro.

## Implementation notes

Checked for any obvious broken links and fixed the 2 that the extraction affects.


garethrees:
Slightly bigger change so might be worth @gbp taking a quick look.

PR #5891, opened 2020-09-15
[#5890] Update redirect_to back method

### LINKED ISSUE #5890: [ERROR] public_body#show (NoMethodError) "undefined method `back_url'
```
A NoMethodError occurred in public_body#show:

  undefined method `back_url' for #<PublicBodyController:0x00007f34b8a34690>
  app/controllers/public_body_controller.rb:36:in `block in show'
```

gbp:
## Relevant issue(s)

Fixes #5890

## What does this do?

Use the new Rails 5 method `redirect_back` which allows us to define a
fallback location if the browser doesn't supply HTTP referrer header.

### DIFF
--- a/app/controllers/public_body_controller.rb
+++ b/app/controllers/public_body_controller.rb
@@ -33,7 +33,7 @@ class PublicBodyController < ApplicationController
       raise ActiveRecord::RecordNotFound.new("None found") if @public_body.nil?
 
       if @public_body.url_name.nil?
-        redirect_to :back
+        redirect_back(fallback_location: root_path)
         return
       end
 


gbp:
@garethrees do you think this is the correct thing to do. We could probbaly just return a 404 as well.

garethrees (commented):
Though it doesn't address the [underlying problems](https://github.com/mysociety/alaveteli/issues/5890#issuecomment-697206769), happy with this as  `redirect_to :back` is plain wrong. I think this should introduce a spec for the behaviour too.

I think we ought to ship this as a hotfix since it's affecting Belgium and Rwanda quite a lot.

PR #5391, opened 2019-10-02
5382 specific from name alternative locale

garethrees:
## Relevant issue(s)

#5382

PR #5894, opened 2020-09-23
Add specs for IncomingMessage#specific_from_name?

garethrees:
## What does this do?

Adds some basic coverage to `IncomingMessage#specific_from_name?`

## Why was this needed?

Created in https://github.com/mysociety/alaveteli/pull/5391 – rescuing as still useful.

### DIFF
--- a/spec/models/incoming_message_spec.rb
+++ b/spec/models/incoming_message_spec.rb
@@ -231,6 +231,48 @@ describe IncomingMessage do
 
   end
 
+  describe '#specific_from_name?' do
+    subject { incoming_message.specific_from_name? }
+
+    let(:body) { FactoryBot.build(:public_body, name: 'Foo') }
+    let(:request) { FactoryBot.build(:info_request, public_body: body) }
+
+    context 'when mail_from is nil' do
+      let(:incoming_message) do
+        FactoryBot.build(:incoming_message, mail_from: nil)
+      end
+
+      it { is_expected.to eq(false) }
+    end
+
+    context 'when safe_mail_from is the same as the body name' do
+      let(:incoming_message) do
+        FactoryBot.
+          build(:incoming_message, info_request: request, mail_from: 'Foo')
+      end
+
+      it { is_expected.to eq(false) }
+    end
+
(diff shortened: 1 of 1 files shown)


spec/models/incoming_message_spec.rb
```diff
@@ -231,6 +231,52 @@
...
+      let(:incoming_message) do
+        FactoryBot.create(:global_censor_rule, text: 'Bar')
+        body = FactoryBot.build(:public_body, name: 'Foo')
+        request = FactoryBot.build(:info_request, public_body: body)
+        FactoryBot.
+          build(:incoming_message, info_request: request, mail_from: 'Bar')
+      end
+
+      it { is_expected.to eq(true) }
+    end
```
gbp, line 277:
```suggestion
    subject { incoming_message.specific_from_name? }

    let(:body) { FactoryBot.build(:public_body, name: 'Foo') }
    let(:request) { FactoryBot.build(:info_request, public_body: body) }

    context 'when mail_from is nil' do
      let(:incoming_message) do
        FactoryBot.build(:incoming_message, mail_from: nil)
      end

      it { is_expected.to eq(false) }
    end

    context 'when safe_mail_from is the same as the body name' do
      let(:incoming_message) do
        FactoryBot.
          build(:incoming_message, info_request: request, mail_from: 'Foo')
      end

      it { is_expected.to eq(false) }
    end

    context 'when safe_mail_from differs from the body name' do
      let(:incoming_message) do
        FactoryBot.
          build(:incoming_message, info_request: request, mail_from: 'Bar')
      end

      it { is_expected.to eq(true) }
    end

    context 'a censor rule masks mail_from' do
      let(:incoming_message) do
        FactoryBot.create(:global_censor_rule, text: 'Bar')
        FactoryBot.build(:incoming_message, mail_from: 'Bar')
      end

      it { is_expected.to eq(true) }
    end
```

gbp (commented):
1. The `a censor rule masks mail_from` spec doesn't need an `info_request` passed to the factory
2. Personally I would extract the `body` and `request` variables into lets.

PR #5896, opened 2020-09-24
Cleanup alaveteli localization

garethrees:
## What does this do?

Reduces complexity of `AlaveteliLocalization`, and in particular `AlaveteliLocalization.set_locales`

## Why was this needed?

Too complex to debug other localisation issues.

## Notes to reviewer

There are a few public methods where we're not yet parsing the values into `AlaveteliLocalization::Locale` classes. I think it would be nice to do this, but there's probably wider-reaching impact, so perhaps something for another day.

### DIFF
--- a/lib/alaveteli_localization.rb
+++ b/lib/alaveteli_localization.rb
@@ -8,20 +8,9 @@ class AlaveteliLocalization
     def set_locales(available_locales, default_locale)
       available, default = parse_locales(available_locales, default_locale)
 
-      FastGettext.default_available_locales =
-        available_locales.map { |locale| locale.canonicalize.to_sym }
-
-      i18n_locales = available_locales.each_with_object([]) do |locale, memo|
-        memo.concat(locale.self_and_parents)
-      end
-
-      I18n.available_locales = i18n_locales.map(&:to_s).uniq
-      I18n.locale = I18n.default_locale = default_locale.hyphenate.to_s
-
-      FastGettext.default_locale = default_locale.canonicalize.to_s
-
-      RoutingFilter::Conditionallyprependlocale.locales =
-        available_locales.map(&:to_s)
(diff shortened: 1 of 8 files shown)


lib/alaveteli_localization.rb
```diff
@@ -33,21 +23,34 @@ def set_default_locale_urls(include_default_locale_in_urls)
         include_default_locale = include_default_locale_in_urls
     end
 
-    def set_session_locale(*args)
-      requested = args.compact.delete_if { |x| x.empty? }.first
+    # rubocop:disable Naming/AccessorMethodName
+    def set_default_locale(locale)
```
gbp, line 27: There are no specs for this method but it's only ever used in `#with_default_locale` in `spec/spec_helper.rb`. Could refactor to remove this completely.

lib/alaveteli_localization.rb
```diff
@@ -33,21 +23,34 @@ def set_default_locale_urls(include_default_locale_in_urls)
...
+      locale = Locale.parse(locale)
+      I18n.default_locale = locale.hyphenate.to_s
+      FastGettext.default_locale = locale.canonicalize.to_s
+    end
+    # rubocop:enable Naming/AccessorMethodName
 
+    # rubocop:disable Naming/AccessorMethodName
+    def set_session_locale(*args)
```
gbp, line 35: I would be happy if we renamed this to `#session_locale=`

gbp (approved):
This is really great. Nice one

garethrees:
> There are no specs for this method but it's only ever used in `#with_default_locale` in `spec/spec_helper.rb`. Could refactor to remove this completely.
> …
> I would be happy if we renamed this to `#session_locale=`

Yeah both of these seem like nice improvements. Might come back to these later.

Issue #5484, opened 2019-12-03
Language switcher not working on WhatDoTheyKnow in Safari


RichardTaylor:
In Safari Version 13.0.3 the language switcher shows "Cymraeg" when the site is in English, and the switcher has no effect (clicking "English" doesn't change what's shown). 

I've noticed this myself and have seen another administrator share a screenshot while discussing another issue showing the site in English and the switcher showing "Cymraeg". 

In Firefox I get what I believe is the expected behaviour, the switcher saying "English" on opening, with the site in English, then switching to say "Cymraeg" and the site switching to Welsh if it is changed. 

On a different but connected point there doesn't appear to be a Welsh translation for two major pieces of text on the front page: "Get answers from the government and public sector" and "Make a request for information to a UK public authority: by law, they have to respond"
<img width="1124" alt="Screenshot 2019-12-03 at 00 21 33" src="https://user-images.githubusercontent.com/547695/70006721-1d430380-1566-11ea-9f47-e2340c331faa.png">

gbp:
I can't reproduce this on Safari 13.0.1. Could you confirm what URL the browser is showing, when choosing the Welsh it should redirect to https://www.whatdotheyknow.com/cy

RichardTaylor:
I too couldn't reproduce this when I first saw it. I first saw it happen a few days ago and was prompted to raise the issue when I saw another admin share a screenshot showing the problem.  I'm not seeing the issue of the word "Cymraeg" being shown while the site is in English now.  

The switcher doesn't work, it lets me switch to Welsh but not back to English.  Gif starts in English, shows a successful switch to Welsh, and then tries to show how attempts to switch back don't work. 

![language-switcher](https://user-images.githubusercontent.com/547695/70054944-7051a080-15d0-11ea-9f64-84ea45c6c5ce.gif)

MattK1234:
Just to note that I've experienced this behaviour before. If I can recreate
it I'll add to the ticket.

Matt

gbp:
Okay from the gif I reckon the /en and /cy URLs are working fine and it just seems to be an issue with the JS. If either of you are able to recreate and are able could you see if there are any errors in the web inspector.

MattK1234:
I noticed this occur today when accessing WhatDoTheyKnow via Chrome on an Android mobile device.

This first happened via an incognito tab which I couldn't screenshot, but when I opened WhatDoTheyKnow via a normal Chrome tab, the same occurred.

[...]

PR #5897, opened 2020-09-24
[eu#106] Fix the locale switcher current locale

gbp:
## Relevant issue(s)

Fixes: https://github.com/mysociety/asktheeu-theme/issues/106
Fixes: https://github.com/mysociety/alaveteli/issues/5484

## What does this do?

This changes ordering before action locale callbacks

## Why was this needed?

We need `set_gettext_locale` to run before `collect_locales` this is
because the URL locale from the params is passed into
`AlaveteliLocalization.set_session_locale` and used to set
`FastGettext.locale`.

This is used then used, via `AlaveteliLocalization.locale` in the
`collect_locales` callback to set the `@locales` varibale used when
render the locale switcher.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -14,7 +14,10 @@ class ApplicationController < ActionController::Base
   end
   class RouteNotFound < StandardError
   end
+
+  before_action :set_gettext_locale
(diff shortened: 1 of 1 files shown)


garethrees (approved):
LGTM!

PR #5898, opened 2020-09-24
Minor optimisation in Locale.parse

garethrees:
## What does this do?

* Don't parse already parsed Locales
* Extract conditional from Locale.parse

## Why was this needed?

Minor optimisations

## Implementation notes

Not 100% on bbfacdf, but I think on balance it's an improvement?

### DIFF
--- a/lib/alaveteli_localization/locale.rb
+++ b/lib/alaveteli_localization/locale.rb
@@ -5,6 +5,8 @@ class AlaveteliLocalization
     include Comparable
 
     def self.parse(identifier)
+      return identifier if identifier.is_a?(Locale)
+
       identifier = identifier.to_s
 
       klass =
--- a/lib/alaveteli_localization/locale.rb
+++ b/lib/alaveteli_localization/locale.rb
@@ -1,25 +1,26 @@
 # -*- encoding : utf-8 -*-
 class AlaveteliLocalization
+  autoload :HyphenatedLocale, 'alaveteli_localization/hyphenated_locale'
+  autoload :UnderscorredLocale, 'alaveteli_localization/underscorred_locale'
+
   # Handle transformations of a Locale identifier.
   class Locale
     include Comparable
+
+    # rubocop:disable Layout/LineLength
+    PARSERS = {
+      /\A[a-z]{2,3}-[A-Z]{2,3}\z/ => AlaveteliLocalization::HyphenatedLocale,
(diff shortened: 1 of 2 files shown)


lib/alaveteli_localization/locale.rb
```diff
@@ -4,22 +4,22 @@ class AlaveteliLocalization
   class Locale
     include Comparable
 
+    # rubocop:disable Layout/LineLength
+    PARSERS = {
+      /\A[a-z]{2,3}-[A-Z]{2,3}\z/ => 'AlaveteliLocalization::HyphenatedLocale',
+      /\A[a-z]{2,3}_[A-Z1-9]{2,3}\z/ => 'AlaveteliLocalization::UnderscorredLocale',
+      /\A[a-z]{2,3}\z/ => 'AlaveteliLocalization::Locale'
```
gbp, line 11:
```suggestion
  autoload :HyphenatedLocale, 'alaveteli_localization/hyphenated_locale'
  autoload :UnderscorredLocale, 'alaveteli_localization/underscorred_locale'

  # Handle transformations of a Locale identifier.
  class Locale
    include Comparable

    # rubocop:disable Layout/LineLength
    PARSERS = {
      /\A[a-z]{2,3}-[A-Z]{2,3}\z/ => AlaveteliLocalization::HyphenatedLocale,
      /\A[a-z]{2,3}_[A-Z1-9]{2,3}\z/ => AlaveteliLocalization::UnderscorredLocale,
      /\A[a-z]{2,3}\z/ => AlaveteliLocalization::Locale
```

lib/alaveteli_localization/locale.rb
```diff
@@ -4,22 +4,22 @@ class AlaveteliLocalization
...
-        when /\A[a-z]{2,3}_[A-Z1-9]{2,3}\z/
-          AlaveteliLocalization::UnderscorredLocale
-        when /\A[a-z]{2,3}\z/
-          AlaveteliLocalization::Locale
-        else
-          raise ArgumentError, 'invalid identifier'
-        end
-
-      klass.new(identifier)
+      klass.constantize.new(identifier)
```
gbp, line 22:
```suggestion
      klass.new(identifier)
```

gbp (commented):
It works fine. I would use actual classes for the parsers values and get rid of the call to `#constantize` - this also requires `autoload` statements under the `AlaveteliLocalization` class.

garethrees:
>  I would use actual classes for the parsers value…

Ah nice – I did initially try that but didn't know the `autoload` trick. Happy if you want to commit those suggestions.

PR #5902, opened 2020-09-29
Allow setting log level through the environment

garethrees:
Sometimes it's useful to be able to use a non-default log level.

This commit allows any log level to be set through `LOG_LEVEL` in the
environment.

If an invalid log level is set the application will fail to boot with a
`NameError`, e.g:

    $ LOG_LEVEL=foobarbaz bin/rails runner "puts Rails.application.config.log_level"
    # snip…
    uninitialized constant ActiveSupport::Logger::FOOBARBAZ (NameError)

If `LOG_LEVEL` is not set, the default for that environment will be
used.

### DIFF
--- a/config/environments/development.rb
+++ b/config/environments/development.rb
@@ -37,6 +37,10 @@ Rails.application.configure do
     'spec', 'mailers', 'previews'
   )
 
+  # Set LOG_LEVEL in the environment to a valid log level to temporarily run the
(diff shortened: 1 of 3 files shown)


gbp (commented):
No need to call `#to_sym` strings are fine.

Aside: I think staging environment is out of date and doesn't work with the our supported version of Rails.

PR #5904, opened 2020-10-01
Admin raw email controller extractions

garethrees:
## What does this do?

* Extract controller code to `PublicBody.with_domain`
* Extract controller code to `RawEmail#domain_from_email`

## Why was this needed?

Reduce controller complexity and add test coverage.


gbp (approved):
Looks good to me

PR #5906, opened 2020-10-01
[rwanda#93] i18n override improvements

gbp:
## Relevant issue(s)

Fixes https://github.com/mysociety/sobanukirwa-theme/issues/93

## What does this do?

Improves our custom override to the `n_` method to:

1. Handle any number of source strings
2. Handle counts being passed in as an string instead of integer.

## Why was this needed?

The `fr` locale works differently to the `en` & `rw` locales. In the PO file `Plural-Forms` is set to `n > 1` [1] which errors when passed a string. Where as other locales have plural forms such as `n != 1` which would return false when given a string.

This is happening on Rwanda as the count is coming from a blog JSON feed so I've improved our override to call `#to_i` on the count to ensure we compare ints with ints.

[1] https://github.com/mysociety/sobanukirwa-theme/blob/master/locale-theme/fr/app.po#L22

### DIFF
--- a/lib/i18n_fixes.rb
+++ b/lib/i18n_fixes.rb
@@ -12,13 +12,13 @@ def _(key, options = {})
 end
 
 def n_(*keys)
-  # The last parameter should be the values to do the interpolation with
-  if keys.count > 3
-    options = keys.pop
-  else
-    options = {}
-  end
-  translation = FastGettext.n_(*keys).html_safe
+  # The last parameter could be the values to do the interpolation with
+  options = keys.extract_options!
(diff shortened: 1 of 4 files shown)


gbp:
From CI it looks like there is a Ruby change between 2.4-2.5, but using RSpec bisect I've found a minimal failing command which fails even on 2.6

`rspec './spec/lib/i18n_interpolation_spec.rb[2:6,2:7]' --seed 40739`

garethrees (approved):
❇️ 👌

Issue #5815, opened 2020-07-09
[ERROR] request search_typeahead (ArgumentError) "InvalidArgumentError: Key too long…"


garethrees:
```
An ArgumentError occurred in request#search_typeahead:

InvalidArgumentError: Key too long: length was 260 bytes, maximum length of a key is 252 bytes
lib/acts_as_xapian/acts_as_xapian.rb:326:in `mset'

-------------------------------
Request:
-------------------------------

* URL : https://www.whatdotheyknow.com/request/search_ahead?q=We%20are%20a%20business%20meetings%20venue%20for%20team%20meetings,%20training%20and%20hybrid%20events.%20We%20do%20not%20fit%20into%20hospitality%20or%20Conference%20centre.%20Where%20can%20we%20get%20guidance%20on%20opening%20our%20city%20centre%20spaces.%20We%20have%20a%20good%20plan%20and%20maximum%20numbers%20could%20be%20easily%20set%20with%20social%20distancing%20and%20even%20enforcing%20the%20wedding%20of%20&requested_from=dcms&per_page=3
* HTTP Method: GET
* Parameters : {"q"=>"We are a business meetings venue for team meetings, training and hybrid events. We do not fit into hospitality or Conference centre. Where can we get guidance on opening our city centre spaces. We have a good plan and maximum numbers could be easily set with social distancing and even enforcing the wedding of ", "requested_from"=>"dcms", "per_page"=>"3", "controller"=>"request", "action"=>"search_typeahead"}
* Timestamp : 2020-07-09 11:34:01 +0100
* Rails root : /REDACTED/alaveteli-2020-07-08T12-53-53

-------------------------------
Session:
-------------------------------

* session id: [FILTERED]
* data: {"session_id"=>"REDACTED",
"locale"=>"en",
"flash"=>
{"discard"=>[],
"flashes"=>{"search_params"=>{"query"=>"Department for digital"}}}}

-------------------------------
Backtrace:
-------------------------------

lib/acts_as_xapian/acts_as_xapian.rb:326:in `mset'
lib/acts_as_xapian/acts_as_xapian.rb:326:in `block in initialize_query'
lib/acts_as_xapian/acts_as_xapian.rb:297:in `initialize_query'
lib/acts_as_xapian/acts_as_xapian.rb:492:in `initialize'
lib/typeahead_search.rb:68:in `new'
lib/typeahead_search.rb:68:in `run_query'
lib/typeahead_search.rb:22:in `xapian_search'
[...]

PR #5914, opened 2020-10-07
[#5815] Truncate search queries that are too long

garethrees:
The maximum length of a search query is 252 bytes. People often paste
long strings into the typeahead search which results in an exception
being raised.

Instead of raising an exception, just truncate the query.

We could tell the user the string is too long, but I don't think it's
worth it in this case. The typeahead search is pretty opaque to users –
it just happens as they enter content – and with a string that long I
don't think we'll lose much by truncating it.

Fixes https://github.com/mysociety/alaveteli/issues/5815.

### DIFF
--- a/doc/CHANGES.md
+++ b/doc/CHANGES.md
@@ -2,6 +2,7 @@
 
 ## Highlighted Features
 
+* Truncate search queries that are too long (Gareth Rees)
 * Remove translation markup from admin interface (Gareth Rees)
 * Improve handling of pluralised translations for locales with multiple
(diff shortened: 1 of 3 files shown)


gbp (approved):
Nice, never used `mb_chars` before

garethrees:
> Nice, never used `mb_chars` before

Yeah it's nice for this kind of thing. I think I first learned of it via https://github.com/mysociety/alaveteli/commit/5dcbded2d03505d42b27e599e94306f51ba28e64

Issue #5783, opened 2020-07-02
[ERROR] request#show (ActionView::Template::Error) "incompatible character encodings: UTF-8 and ASCII-8BIT"


garethrees:
```
An ActionView::Template::Error occurred in request#show:

incompatible character encodings: UTF-8 and ASCII-8BIT
lib/mail_handler/backends/mail_backend.rb:60:in `mail_from_outlook'

-------------------------------
Request:
-------------------------------

* URL : https://www.asktheeu.org/en/request/public_documents_related_to_disi_6
* HTTP Method: GET
* Parameters : {"controller"=>"request", "action"=>"show", "url_title"=>"public_documents_related_to_disi_6", "locale"=>"en"}
* Timestamp : 2020-07-02 11:05:38 +0200
* Rails root : /REDACTED/alaveteli-2020-06-18T09-21-33

-------------------------------
Session:
-------------------------------

* session id: [FILTERED]
* data: {"session_id"=>"REDACTED", "locale"=>"en"}

-------------------------------
Backtrace:
-------------------------------

lib/mail_handler/backends/mail_backend.rb:60:in `mail_from_outlook'
lib/mail_handler/backends/mail_backend.rb:188:in `decode_attached_part'
lib/mail_handler/backends/mail_backend.rb:241:in `expand_and_normalize_parts'
lib/mail_handler/backends/mail_backend.rb:216:in `block in expand_and_normalize_parts'
lib/mail_handler/backends/mail_backend.rb:216:in `expand_and_normalize_parts'
lib/mail_handler/backends/mail_backend.rb:284:in `get_attachment_leaves'
lib/mail_handler/backends/mail_backend.rb:355:in `get_attachment_attributes'
app/models/incoming_message.rb:536:in `extract_attachments!'
app/models/incoming_message.rb:98:in `block in parse_raw_email!'
app/models/incoming_message.rb:97:in `parse_raw_email!'
app/models/incoming_message.rb:178:in `mail_from'
app/models/incoming_message.rb:197:in `safe_mail_from'
app/models/incoming_message.rb:220:in `specific_from_name?'
app/views/request/_incoming_correspondence.html.erb:4:in `_app_views_request__incoming_correspondence_html_erb___4239059047440180258_69914119794420'
app/views/request/_correspondence.html.erb:4:in `_app_views_request__correspondence_html_erb__4133394796544516923_69914122355520'
[...]

PR #5920, opened 2020-10-13
[#5783] Improve resilience to unparsable outlook attachments

garethrees:
Sometimes we receive messages with outlook attachments that can't be
parsed due to an issue in mapi [1].

There's a potential fix [2] but it conflicts [3] with an existing patch
we apply [4].

This at least allows users to download the raw attachment, rather than
us preventing the entire request from loading because we raise an
exception.

Part of https://github.com/mysociety/alaveteli/issues/5783.

[1] https://github.com/aquasync/ruby-msg/issues/15
[2] https://github.com/aquasync/ruby-msg/pull/16
[3] https://github.com/mysociety/alaveteli/issues/5783#issuecomment-655740838
[4] https://github.com/mysociety/ruby-msg/pull/3

## Notes to reviewer

> Do not introduce global variables.

We actually rely on `$alaveteli_dir` in this case, as we use it in [`MailHandler::ReplyHandler.load_rails`](https://github.com/mysociety/alaveteli/blob/f594d74f0e1c11704c105da1c543cb0108e60159/lib/mail_handler/reply_handler.rb#L117-L120).

### DIFF
--- a/lib/mail_handler/backends/mail_backend.rb
+++ b/lib/mail_handler/backends/mail_backend.rb
@@ -35,6 +35,7 @@ end
 module MailHandler
   module Backends
     module MailBackend
+      include ConfigHelper
 
       def backend
         'Mail'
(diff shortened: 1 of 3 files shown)


gbp (approved):
Looks good to me 👍

PR #5921, opened 2020-10-14
Make RuboCop failures warnings rather than errors

garethrees:
Failures for this check are not always technical errors – they're more a
case of requiring human judgement as to whether to fix or not.

Documentation reference: https://github.com/reviewdog/action-rubocop/blob/d3161530d1ad5da4bac484302ecddcba02d186e6/README.md#level


gbp (approved):
This seems to work:

Screen Shot 2020-10-15 at 10 24 28
Screen Shot 2020-10-15 at 10 24 13

Issue #5910, opened 2020-10-05
Add support checklist to /help/unhappy to help users understand how to challenge refusals


garethrees:
On the **/help/unhappy** page we’ll introduce a new concept of a **checklist**. This will help users understand the pertinent parts of the act in an easier-to-digest format – [conceptual compression](https://m.signalvnoise.com/conceptual-compression-means-beginners-dont-need-to-know-sql-hallelujah/)!

The idea is that we help them check the refusal step by step, and progressively ask more detailed questions as we hone in on the advice we should give them.

When exemptions have been used, we’ll incorporate **specific advice** as well as the general advice for checking the request has been refused correctly.

In cases where the authority hasn’t refused correctly – say, they select “no, there wasn’t any use of exemptions” – we should highlight the **internal review link**, encouraging them to challenge.

We don’t want to prevent anyone from requesting an internal review though, so the option should always be available. This is about positively encouraging it if the authority clearly hasn’t done the right things.

As well as prompting an internal review, checklists should **filter the templates** that are shown by default on the new followup page (https://github.com/mysociety/alaveteli/issues/5911). I’m imagining that some JavaScript appends query params to the internal review link, or something like that.

We should consider how we make it easy for re-users to add checklists. They’ll be very site-specific so we probably can’t provide built-in examples, and most of our re-users aren’t particularly technical so lots of custom JavaScript won’t work.

The fallback option for re-users who want to offer better targeted guidance is to write the help page in HTML as before, where they can include links to the new followup page with hard-coded query params to filter the templates.

We’ll also want to incorporate some **case studies** on this page, to show examples of successful challenges. The case studies will be published on our blog, so this will be more about nicely presenting a snippet to provide some social proof that challenging is okay to do and can be successful.

We’ll include the **fallback options** – make a new request, write to your MP, etc – in plain text as before.

Users are also directed to **/help/unhappy** when their request is **late**, or the authority says they **don’t hold the information**. Hopefully we’ll be able to treat these in a similar way to a request being refused – provide a **checklist** and then present relevant **templates**. I’ve marked them as “bonuses” because the core goal of this project is to challenge _refusals_. We already have a reasonable workflow for late requests, and it’s difficult to do anything when an authority says they don’t have the information you’ve asked for.

![Screenshot 2020-10-05 at 12 29 58](https://user-images.githubusercontent.com/282788/95074448-80fa0800-0706-11eb-9e3a-5e53d5f1ec78.png)

[...]

Issue #5911, opened 2020-10-05
Add snippet library to new followup page


garethrees:
As per the existing workflow, once a request has been **classified** as refused we’ll take the user to **/help/unhappy** which shows them the next steps. This will either be immediately following the submission of the **classification form**, or from the **classification banner** if the request has already been classified as refused.

On the **new followup** page we’ll introduce a **template library** that allows users to better tailor their internal review based on the situation they’re faced with.

Each **template** will have some sort of title and/or short description of when to use it, and the template text itself. Clicking a template will just append the template text to the followup **message body**. This will allow the user to tweak it for their specific case, or easily delete the appended text.

**Templates** should probably be database-driven and translatable, as the specifics for each install will vary.

We’ll want some way of filtering the templates. It’s not essential that we expose filtering options to the user (that could get complicated), but they should be able to clear filters if we’ve applied them dynamically.

![Screenshot 2020-10-05 at 12 34 52](https://user-images.githubusercontent.com/282788/95074896-3036df00-0707-11eb-8cd1-163862c7d8b2.png)

garethrees:
Note that for research purposes we'll want to record that a particular template was used. A good place for this might be in the `InfoRequestEvent#params_yaml` attribute.

MyfanwyNixon:
I've started compiling snippets at https://docs.google.com/document/d/1IJnWbzoCGFiHYZUhG__AbP-6ReDvx8q3BCwvN_mglXk/edit.
At the moment these all refer to section 12 exemptions, and I've marked each instance on this doc https://docs.google.com/document/d/1oMuhEONlVmTJRDjonwsqi3VVp5xBDxhtkhc4sroRpU8/edit

garethrees:
We started a bit of a conversation in Slack around automatically pre-filling these snippets based on the Q&A.

My fears with the whole auto-filling idea are:

1. we don’t have any of the existing infrastructure around showing and editing admin-editable snippets yet; that’s not an insignificant task
2. we then get into the situation of having to be more right – people won’t necessarily read all the pre-filled stuff. Not saying they always will when they click-to-add, but more chance of it, and less chance that _every_ response we send is in some way not right because of a decision _we_ made. 
3. there's the complexity in unpicking how the mechanics of the exiting hard-coded pre-filled text and admin-editable snippets interact

I don't think we've got enough remaining time to handle these in a satisfactory way, so let's park this for now. If we figure out a nice way to do it, then we can revisit if we have time.

[...]

PR #5923, opened 2020-10-15
[#5910][WIP] Spike for rendering refusal advice

garethrees:
## Relevant issue(s)

#5910

## What does this do?

Illustrates how we can map refusal advice encoded in a data structure to present to users.

This is intended for initial feedback on approach.

## Implementation notes

Obviously this is not production-ready. Some things I'm thinking we need to figure out:

* This file is going to get big; can we have a directory of YAML files (`config/refusal_advice`?) that we glob and load?
* I've stubbed the partial rendering by using `plain` for now. Where will partials live? Should we set a conventional base path (`app/views/help/refusal_advice/`?) and expect any partials to be located relative to that?
* I don't think there'll ever be an _easy_ way for re-users to construct this advice – at the end of the day it's a lot of information and you're going to have to do quite a bit of legwork – but is YAML (or JSON) the easiest? Any other options?

## Screenshots

Screenshot 2020-10-15 at 10 58 47

The HTML generated (Not set in stone of course – just gets at the gist of what we want to do):

```html
<h2>General Advice</h2>

<div class="question">
  <p class="question__title">
    A top-level question
  </p>
  <div class="question__yes">
    <div class="question__action">
      <div class="question__render">
        You can appeal on general grounds!
      </div>
      <div class="question__template-tags" data-tags="did_not_confirm_or_deny"></div>
    </div>
  </div>
  <div class="question__no">
    <div class="question">
      <p class="question__title">
        A sub question
      </p>
      <div class="question__yes">
        <div class="question__action">
          <div class="question__render">
            You can appeal on this sub point
          </div>
          <div class="question__template-tags"></div>
        </div>
      </div>
      <div class="question__no">
        <div class="question__action">
          <div class="question__render">
            There are no general grounds to appeal :(
          </div>
          <div class="question__template-tags"></div>
        </div>
      </div>
    </div>
  </div>
</div>
<div class="question">
  <p class="question__title">
    Another top-level question
  </p>
  <div class="question__yes">
    <div class="question__action">
      <div class="question__render">
        path/to/yes-partial
      </div>
      <div class="question__template-tags"></div>
    </div>
  </div>
  <div class="question__no">
    <div class="question__action">
      <div class="question__render">
        path/to/no-partial
      </div>
      <div class="question__template-tags"></div>
    </div>
  </div>
</div>

<h2>Specific Advice</h2>

<div class="question">
  <p class="question__title">
    A specific top-level question
  </p>
  <div class="question__yes">
    <div class="question__action">
      <div class="question__render">
        You have a specific appeal!
      </div>
      <div class="question__template-tags"></div>
    </div>
  </div>
  <div class="question__no">
    <div class="question">
      <p class="question__title">
        Another specific question?
      </p>
      <div class="question__yes">
        <div class="question__action">
          <div class="question__render">
            path/to/specific_partial
          </div>
          <div class="question__template-tags"></div>
        </div>
      </div>
      <div class="question__no">
        <div class="question__action">
          <div class="question__render">
            Nothing we can do specifically soz
          </div>
          <div class="question__template-tags"></div>
        </div>
      </div>
    </div>
  </div>
</div>
```

## Notes to reviewer

You'll need a `tmp/refusals.yml`. Here's the sample I used:

```yaml
# Always shown
general_advice:
  - question: A top-level question
    yes:
      render:
        plain: 'You can appeal on general grounds!'
      followup_template_tags: did_not_confirm_or_deny
    no:
      -
        question: A sub question
        yes:
          render:
            plain: 'You can appeal on this sub point'
        no:
          render:
            plain: 'There are no general grounds to appeal :('
  - question: Another top-level question
    yes:
      render:
        plain: 'path/to/yes-partial'
    no:
      render:
        plain: 'path/to/no-partial'

# Sometimes shown
specific_advice:
  section_12:
    -
      question: A specific top-level question
      yes:
        render:
          plain: 'You have a specific appeal!'
          followup_template_tags: section_12
      no:
        -
          question: Another specific question?
          yes:
            render:
              plain: 'path/to/specific_partial'
          no:
            render:
              plain: "Nothing we can do specifically soz"
```

### DIFF
--- a/app/controllers/help_controller.rb
+++ b/app/controllers/help_controller.rb
@@ -6,6 +6,7 @@
 # Email: hello@mysociety.org; WWW: http://www.mysociety.org/
 
 class HelpController < ApplicationController
+  include RefusalsHelper
 
   # we don't even have a control subroutine for most help pages, just see their templates
 
@@ -20,6 +21,10 @@ class HelpController < ApplicationController
   def unhappy
     @country_code = AlaveteliConfiguration.iso_country_code
     @info_request = nil
+
+    @refusal_advice =
+      YAML.load(File.read('./tmp/refusals.yml')).deep_symbolize_keys
+
     if params[:url_title]
       @info_request = InfoRequest
         .not_embargoed
--- a/app/helpers/refusals_helper.rb
+++ b/app/helpers/refusals_helper.rb
@@ -1,0 +1,36 @@
+module RefusalsHelper
+  def general_advice(data)
+    build_questions(data[:general_advice])
+  end
+
+  def specific_advice(data, exemption:)
+    # TODO: Render all exemptions unless we provide one
+    build_questions(data[:specific_advice][exemption])
+  end
+
+  private
+
+  def build_questions(data)
+    data.map { |data| build_question(data) }.join.html_safe
+  end
+
+  def build_question(data)
+    return build_questions(data) if data.is_a?(Array)
+
+    tag.div(class: 'question') do
+      concat tag.p(class: 'question__title') { data[:question] }
+      concat tag.div(class: 'question__yes') { build_action(data[true]) }
+      concat tag.div(class: 'question__no') { build_action(data[false]) }
+    end
+  end
+
+  def build_action(data)
+    return build_questions(data) if data.is_a?(Array)
+
+    tag.div(class: 'question__action') do
+      concat tag.div(class: 'question__render') { render data[:render] }
+      concat tag.div(class: 'question__template-tags',
+                     data: { tags: data[:followup_template_tags] })
+    end
+  end
+end
--- a/app/views/help/unhappy.html.erb
+++ b/app/views/help/unhappy.html.erb
@@ -1,1 +1,11 @@
 Placeholder to be overriden by theme
+
+<hr />
+
+<h2>General Advice</h2>
+
+<%= general_advice(@refusal_advice) %>
+
+<h2>Specific Advice</h2>
+
+<%= specific_advice(@refusal_advice, exemption: :section_12) %>


zarino:
I went down a bit of a rabbithole on Wednesday, thinking about the structure/syntax for defining these questions. I was insipired a bit by looking at how the MOJ Form Builder saves its configuration files. [Here’s an example page from our BODS discovery work.](https://github.com/mysociety/bods-test/blob/master/metadata/page/page.ownership-by-companies.json)

The thing I wondered is whether _not_ having a tree structure to the questions would make writing the configuration simpler. eg:

```yaml
- id: "first-top-level"
  label: "This is the first top level question, do you agree?"
  options:
  - label: "Yes"
    value: "yes"
  - label: "No"
    value: "no"

- id: "conditional-example"
  show_if:
  - id: "first-top-level"
    operator: "is"
    value: "yes"
  label: "This question only appears if you selected yes to the top level question, neat huh?"
  options:
  - label: "Absolutely"
    value: "yes"
  - label: "Not really"
    value: "no"

- id: "first-recommendation"
  show_if:
  - id: "conditional-example"
    operator: "is"
    value: "yes"
  label: "You need to get out more. There is nothing we can do for you."

- id: "second-recommendation"
  show_if:
  - id: "conditional-example"
    operator: "is"
    value: "no"
  label: "Since there was ambiguity over the neatness of the message, you have grounds to appeal under Section 14(1) of the Boring Act 2014."
  next_steps:
  - type: "internal_review"
    response_template: "path/to/template/internal_review"

- id: "fallback"
  show_if:
  - id: "first-top-level"
    operator: "is"
    value: "yes"
  label: "Since you don‘t agree that the top level question was a top level question, we’d recommend you send a reply to the authority seeking further clarification."
  next_steps:
  - type: "reply"
    response_template: "path/to/template/generic_reply"
```

A things about this that struck me as useful:

1. No nesting hell.
1. Potential for quite flexible / complex conditions under the `show_if` parameter. You might have different `operator`s like `is`, `is not`, or you might have multiple conditions that all have to be true.
1. Potential for more than just yes/no binary inputs. Can have as many inputs as you want after a question, and they can be worded however you want.
1. Potential for multiple next steps based on a single condition.
1. Potential for a "step" in the flow to _both_ define a next step _and_ ask a further drill-down question.

And, as mentioned in #5910, I’m imagining that, once next steps have been identified for you, all of the associated `label`s would be saved up and presented at the end, in a summary box for the entire form, with buttons or links to begin the replies/reviews defined in the next steps.

I don’t know whether that helps you.

garethrees:
Interesting! Just noting the [GitHub Actions reference](https://docs.github.com/en/free-pro-team@latest/actions/reference/context-and-expression-syntax-for-github-actions) as a similar idea.

skenaja:
> 4. This example uses the options key to define radio buttons. But there’s potential for types of input other than options. Maybe we might want a text input so the user can fill in a number (which we then use operator: "less-than" on ;-) ) or maybe we might want multiple selection checkboxes.

I'm trying to get my head around the UX for the first few questions we'd ask of users that are manually going through the process (as opposed to arriving at a request that has already had some form of heuristics/NLP/ML applied to it to guess the small no of reasons why it was refused). 

I appreciate this is being made to work for all Alaveteli installs, but for the FOI Act alone, there are 20-30+ top-level reasons why a request might have been refused - 24 specific info-related exemptions (e.g. Commercial Interests, Legally Privileged, etc), and ~5-10 generic/request level reasons (Vexatious, Too Expensive, Info Not Held, etc), and one or more could apply.

You could gather these up front (requiring multi-select checkboxes to build up a set of future pages) but it would mean a lot of text for the user to read through and pick from which might deter some people. Or you could ask the user after they'd finished the first one if there were any other reasons to complain about, and let them go through the pick process multiple times.  Or you could have something like a typeahead box on the first page which allows users search for the reasons and to add to/remove from a list of topics that they then progress with ([a bit like this](https://select2.org/tagging#tagging-with-multi-value-select-boxes))

Thoughts?

garethrees:
> I'm trying to get my head around the UX…

Us too!

> …for the first few questions we'd ask of users that are manually going through the process

There'll be a handful of "general" questions – did they confirm or deny, did they cite an exemption – and then we'll get into the exemption specific advice based on what gets selected.

> I appreciate this is being made to work for all Alaveteli installs, but for the FOI Act alone, there are 20-30+ top-level reasons why a request might have been refused - 24 specific info-related exemptions (e.g. Commercial Interests, Legally Privileged, etc), and ~5-10 generic/request level reasons (Vexatious, Too Expensive, Info Not Held, etc), and one or more could apply.

Yeah, there's potentially a lot to cover. We're looking to do as much as we can while keeping it as simple as possible. This might mean we miss some edge cases and subtleties, but I think we should be able to provide a massive step up from where we are now, even if it's not 100% complete.

> You could gather these up front (requiring multi-select checkboxes to build up a set of future pages) but it would mean a lot of text for the user to read through and pick from which might deter some people. Or you could ask the user after they'd finished the first one if there were any other reasons to complain about, and let them go through the pick process multiple times.

This is on our minds, but I don't think we're ever going to convince _everyone_ to appeal. I think there's no way around there being a lot of content for all of this advice, but if we can at least unfold it a bit at a time when most relevant it should help a good proportion of people.

> ask the user after they'd finished the first one if there were any other reasons to complain about

This is more about helping users understand whether there are any reasons. Many users won't understand the implications of an exemption – if you do you can probably skip straight through to choose relevant templates (https://github.com/mysociety/alaveteli/issues/5911).

> Or you could have something like a typeahead box on the first page which allows users search for the reasons and to add to/remove from a list of topics that they then progress with (a bit like this)

Interesting one to explore. I'm not sure how this would work with the unfolding, but might be able to do something interesting.

Issue #1353, opened 2014-02-19
Remove translation from strings in admin interface


crowbot:
The admin interface for the site mostly appears in English, but some strings have been marked up for translation, resulting in a mix of English and the locale language appearing. Clean up the translated strings to reduce the amount of translation that needs to be done and clarify the interface.

PR #5926, opened 2020-10-16
[#5925] Remove model attribute translations strings

### LINKED ISSUE #5925: Remove model attribute translations
Since upgrading `gettext_i18n_rails` [1] when we generate translation strings we get a lot of changes where the old strings prefixed with `ApplicationRecord|` from the `local/model_attributes.rb` get correctly renamed to be prefixed with the correct model class (although it still seems broken for nested models such as our `AlaveteliPro` ones).

This is great... but the diffs in the PO are really confusing beacuse the strings are reordered into new positions in the file. So far I've yet to commit these changes as we don't actually recommend reusers to translation these strings [2].

> they do not appear on the site anywhere at the moment, and when they do, they will only be used in the admin interface. If you do translate them, only translate the text that comes after the |.

We could:

1. remove the model attribute translations completely as we don't have plans of translating the admin interface anymore
2. alternatively move these string out into a thrid locale resource and disable the `Your translators can translate resource strings` setting in Transifex.

[1] https://github.com/mysociety/alaveteli/commit/745b48a2533
[2] https://alaveteli.org/docs/customising/translation/

gbp:
## Relevant issue(s)

Fixes #5925

## What does this do?

Remove model attribute translations strings

## Why was this needed?

1. As far as I can see these strings aren't used
2. These strings will been renamed/replaced next time `rake gettext:store_model_attributes` causing a huge diffs

## Post release

- [ ] push to Transifex to remove strings
- [ ] update GitHub wiki pages to remove references to `rake gettext:store_model_attributes`
- [ ] update https://alaveteli.org/docs/customising/translation/


garethrees:
@crowbot is there any reason not to do this?

We explicitly don't translate the admin interface (https://github.com/mysociety/alaveteli/issues/1353), and the [documentation](https://alaveteli.org/docs/customising/translation/) suggests that's the only place they're intended for use:

> Do not prioritise these types of text when translating – they do not appear on the site anywhere at the moment, and when they do, they will only be used in the admin interface. If you do translate them, only translate the text that comes _after_ the `|`

garethrees (approved):
> @crowbot is there any reason not to do this?

Caught up with Louise on this and nothing came to mind. I think we can merge when we have time to do so and complete the rollout steps.

Issue #5909, opened 2020-10-05
Identify exemptions used in correspondence


garethrees:
> Automatically identifying what exemptions have been used in a refusal​, makes it quicker and easier for users to understand the grounds for an appeal. Once a user knows which exemption has been used, they are in a better position to examine whether it has been correctly applied and to rebut it effectively.

The goals for this feature are to:

1. Pull out exemptions that have been used in correspondence
2. Help the user understand the basic implications
3. Encourage them to challenge them

On the **request page** we’ll add a panel to the **incoming message** that shows any **exemptions** we’ve been able to identify.

![Screenshot 2020-10-05 at 12 26 56](https://user-images.githubusercontent.com/282788/95074222-134ddc00-0706-11eb-871c-6c4f4a79fd5f.png)

Our initial version of the identification will be quite basic, so we’ll have to be clear that these are _possible_ exemptions. The exemptions should be _separate_ from the incoming message body. It will be too complicated to manipulate the HTML generated for the message body in the time we have available. We won’t add any sort of learning or ways to remove the identified exemptions.

We should list each exemption that’s identified. Each exemption should have a name, a short description explaining what the exemption is about and its implications. We’ll possibly add a link that will take the user to a **help page** about that exemption.

We won’t trigger any automated decision making at this point. A user will still have to manually classify their request as refused to be pushed into the flow to challenge a refusal. Identifying exemptions helps give the user a leg up in understanding that “refused” is the option they need to pick, and that’s what will take them into the flow for challenging the refusal.

zarino:
So, the most obvious treatment here is to reuse/expand the existing `.correspondence_delivery` styling, and either present the message at the top or the bottom of the authority’s reply:

![Screenshot_2020-11-13 The cost of boring - a Freedom of Information request to Department for Humpadinking(1)](https://user-images.githubusercontent.com/739624/99091815-49fdf880-25c8-11eb-90e7-e62a88c2b56f.png)

This is quick and easy for us to do, but I’m a bit worried it’s adding an action in a place that actions don’t currently exist (ie: in the middle of the otherwise read-only correspondence elements) and it feels a bit contradictory to the _two absoluely gigantic_ "describe state" boxes above and below the correspondence thread.

Indeed, those "describe state" boxes already contain a "my request has been refused" option. They feel like the most logical place, to me, to put anything related to taking the next step after a refusal. Maybe something like this, where the checkboxes are pre-ticked, based on which terms were detected in the correspondence?

[...]

PR #5931, opened 2020-10-22
[#5909] Legislation detection

gbp:
## Relevant issue(s)

Connected to #5909

## What does this do?

Detects references to legislation in incominng messages

## Why was this needed?

So we can offer refusal advice

### DIFF
--- a/app/models/incoming_message.rb
+++ b/app/models/incoming_message.rb
@@ -79,6 +79,7 @@ class IncomingMessage < ApplicationRecord
   delegate :message_id, to: :raw_email
   delegate :multipart?, to: :raw_email
   delegate :parts, to: :raw_email
+  delegate :legislation, to: :info_request
 
   # Given that there are in theory many info request events, a convenience method for
   # getting the response event
@@ -741,5 +742,15 @@ class IncomingMessage < ApplicationRecord
   # Return space separated list of all file extensions known
   def self.get_all_file_extensions
     return AlaveteliFileTypes.all_extensions.join(" ")
+  end
+
+  def refusals
+    legislation_references.select(&:refusal?)
+  end
+
+  private
+
+  def legislation_references
+    legislation.find_references(get_main_body_text_folded)
   end
 end
--- a/app/models/legislation.rb
+++ b/app/models/legislation.rb
@@ -55,6 +55,7 @@ class Legislation
 
   def initialize(key:, **variants)
     @key = key
+    @refusals = variants.fetch(:refusals, [])
     @variants = variants
   end
 
@@ -64,5 +65,15 @@ class Legislation
     raise UnknownLegislationVariant.new(
       "Unknown variant #{variant} in legislation #{key}."
     )
+  end
+
+  def find_references(text)
+    Legislation::ReferenceCollection.new(legislation: self).match(text)
+  end
+
+  def refusals
+    @refusals.map do |reference|
+      Legislation::Reference.new(legislation: self, reference: reference)
+    end
   end
 end
--- a/app/models/legislation/constants.rb
+++ b/app/models/legislation/constants.rb
@@ -1,0 +1,45 @@
+class Legislation
+  module Constants
+    TYPE = /([a-z]+)/i
+    ELEMENT = /([a-z0-9]+)/i
+    NUM_ELEMENT = /([0-9]+)/i
+    BRACKETED_ELEMENT = /\(#{ELEMENT}\)/
+    DECIMAL_ELEMENT = /\.#{ELEMENT}/
+    SUB_ELEMENTS = /#{BRACKETED_ELEMENT}|#{DECIMAL_ELEMENT}/
+    ALL_ELEMENTS = /#{ELEMENT}|#{SUB_ELEMENTS}/
+
+    SPACE_OR_DOT = /\s*|\.\s*/
+    AND_SEPARATOR = /&(?:amp;)?/
+    OR_SEPARATOR = /,\s*|\s*or\s*/i
+
+    # Match whole references strings
+    REGEXP = /
+      #{TYPE}
+      #{SPACE_OR_DOT}?
+      (
+        (?:
+          #{NUM_ELEMENT}
+          (?:
+            \s*
+            (?:#{SUB_ELEMENTS})+
+            (?:#{AND_SEPARATOR}#{ALL_ELEMENTS})*
+          )?
+          #{OR_SEPARATOR}?
+        )+
+      )
+    /x
+
+    # Match references with AND separator, with named capture groups for easier
+    # replacement
+    AND_REGEXP = /
+      \b
+      (?<sup>.*?)
+      \.
+      (?<sub_1>#{ELEMENT})
+      #{AND_SEPARATOR}
+      \.?
+      (?<sub_2>#{ELEMENT})
+      \b
+    /x
+  end
+end
--- a/app/models/legislation/reference.rb
+++ b/app/models/legislation/reference.rb
@@ -1,0 +1,60 @@
+class Legislation
+  InvalidReferenceType = Class.new(StandardError)
+
+  ##
+  # Reference representing a section of a legislation
+  #
+  # See https://en.wikipedia.org/wiki/Citation_of_United_Kingdom_legislation#Primary_legislation
+  #
+  # Example:
+  #   Legislation::Reference.new(legislation: foi, reference: 's 12(1)') =>
+  #     #<Legislation::Reference
+  #       @legislation=foi,
+  #       @type="Section",
+  #       @elements=["12", "1"]>
+  #
+  class Reference
+    attr_reader :legislation, :type, :elements
+
+    def initialize(legislation:, reference:)
+      @legislation = legislation
+
+      type, elements = *reference.split(' ', 2)
+      @type = parse_type(type)
+      @elements = elements.gsub(
+        /\.?#{Constants::BRACKETED_ELEMENT}/,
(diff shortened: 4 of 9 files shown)


skenaja:
Probably a good idea to also look at the public_body tags for the request, as the exemption numbers vary by jurisdiction

 * if they contain `foi_no` or `foi_refuse`, then they're not going to be covered by FOI. `foi_voluntary` *might* by OK.
 * `scotland` - covered by FOI(S)A or EIR(S)
 * `eir_only` will be EIR or EIR(S) (the latter in conjunction with `scotland`)
 * `nsbody` will be covered by the North/South Bodies Freedom of Information Code of Practice applicable to Northern Ireland / Ireland joint bodies.

gbp:
> Probably a good idea to also look at the public_body tags for the request...

Yep, I agree on all of this. we also need to do this in away its re-useable for other installations

mdeuk:
> Probably a good idea to also look at the public_body tags for the request, as the exemption numbers vary by jurisdiction
> 
> * if they contain `foi_no` or `foi_refuse`, then they're not going to be covered by FOI. `foi_voluntary` _might_ by OK.

+1

> * `scotland` - covered by FOI(S)A or EIR(S)
> * `eir_only` will be EIR or EIR(S) (the latter in conjunction with `scotland`)

I was just coming to mention this one - thanks!

The exemptions in many cases are quite similar but there is some subtle numbering differences, and of course, the regime is different in terms of when an applicant can make a request to OSIC for a decision (e.g. the internal review is generally a mandatory step).

skenaja:
Regarding re-use, IIRC there were some similar design considerations discussed in #2819 for deciding which ombudsman to direct an appeal to.

I had been thinking of how you might classify exemptions against requests a while back (eg as part of a categorisation game), you could have generic code to set request-level tags (customised at theme level) for the exemptions, for example something like `foi:14_1`, `eir:12_4_b` which could then get picked up by the advisor for rendering to the user. This would also allow you to run the process asynchronously, especially if you end up going down the ML route and/or reading the contents of attachments. It would also allow all the advice to be held at the theme level as config, for example:

```yaml
---
key: vexatious
title: "Vexatious / manifestly unreasonable request"
description: "blah"
applicable_law: #array of request-level tags that an automated exemption classifier might have set, which could then be used to draw upon this advice fragment.
  - "foi:14_1"
priority: 100 #sort order for presenting multiple fragments
search_tokens: #array of strings that a user might type in to find detailed help about a refusal
  - section 14(1)
  - vexatious
  - manifestly unreasonable
  - 14(1)
  - "141"
  - "14.1"
text: |
  lorem ipsum *markdown text*
  blah
```

garethrees (changes requested):
I think this is going in the right general direction, and we have lots of the pieces we need, but I'm not sure they're all in the right place.

I think we should refer to "exemptions" as "refusals" in the codebase, as it matches the `described_state` better (and "exemption" is probably quite a UK-specific term, and not very human-friendly).

I hope this feedback isn't too confusing. It took me a while to get my head around (especially with the feedback about law variations) and figure out what the crux of the issues are for this initial piece of the puzzle.

## `LegislationReference`

I think the extraction of the various forms of legislation references in this class is :ok_hand:, but I find the class a bit confusing as it took me quite a while to grok what parts of the extracted strings mean what. It's mostly class methods, and not really encapsulating acting, but acting only as a value object when initialised. It feels quite "top heavy".

I feel like there are two concerns that we're munging together at the class-level here.

1. Extracting parts of a string that match the "Legislation Reference" format
2. Parsing out the _attributes_ of the Legislation Reference

I think the attribute parsing would be more appropriately located as private instance-level methods.

```ruby
str = 'foo bar baz Section 15(1)(b) foo bar baz Section 3(1)'

LegislationReference.match(str)
# => ['Section 15(1)(b)', 'Section 3(1)']
```

```ruby
s = LegislationReference.new('Section 5(1)(b)')
# => <LegislationReference::Section …>

s.section_number
# => 5

s.subsection_number
# => 1

s.paragraph
# => 'b'

s.to_sym
# => :section_5_1_b
```

However, these are kind of minor details. The parser works, so we can at least pull out these strings.

## `IncomingMessage`

I think my main concern with the implementation as it stands is that there's no way to swap out the `LegislationReference` parser, and no interface to define how to detect a _refusal_ from the parsed references.

While we think that `LegislationReference` covers a widely applicable format, it's not translated, and I think we can be fairly confident that we'll need to make tweaks for other installs (we can and should post to the mailing lists to ask for examples of how requests are refused internationally).

More importantly, there's no interface for actually saying which of the patterns are refusals. There's a stub _implementation_, but that only covers a single case in UK law.

So, how do we actually define these? I think we should keep it simple; something like:

```ruby
# Theme
LegislationReference.refusals = [
  :section_12,
  :section_21,
  # …
  :section_44
]
```

```ruby
class LegislationReference
  def refusal?
    self.class.refusals.include?(to_sym)
  end
end
```

## Different Laws

@skenaja and @mdeuk make good points about there being nuances between legislation, even within a single install.

I agree that we'll want to be mindful of this, but equally we want to model the world as we think makes it easier to understand, not necessarily how it _is_ in practice.

However, it does seem like a good idea to have a stab at this to see how it feels. With this in mind, it does seem like `IncomingMessage` is the wrong place to define the "exemption parser".

There are two places that seem like a better fit – `InfoRequest` and `PublicBody`.

`InfoRequest` has a `#law_used` attribute, which we might want to take into account at some point.

An `InfoRequest` is made to a `PublicBody`, which has a `#tag_string` that may include tags that result in special behaviour. It seems like `PublicBody` should have some knowledge of which refusal reasons it can apply.

I haven't had enough time to think deeply about this. I think for now, the main thing to focus on to get this merged is getting the configuration points in place for customising both the parsing out the format of a legislation reference, and defining which of them are a refusal.

app/views/request/_incoming_correspondence.html.erb
```diff
@@ -23,6 +23,25 @@
                             :body => incoming_message.get_body_for_html_display(@collapse_quotes),
                             :attachments => incoming_message.get_attachments_for_display }  %>
 
+    <% if incoming_message.exemptions.any? %>
```
garethrees, line 26: Let's drop this from the PR as it's not intended to get merged

garethrees:
> I haven't had enough time to think deeply about this. I think for now, the main thing to focus on to get this merged is getting the configuration points in place for customising both the parsing out the format of a legislation reference, and defining which of them are a refusal.

One other thing. We _can_ launch this under a limited set of conditions:

```ruby
class PublicBody
  def has_refusal_advice?
    !tagged_with?('scotland') && !tagged_with?('eir_only') # …
  end
end

# somewhere appropriate…
info_request.law_used == 'foi' && public_body.has_refusal_advice?
```

I'd rather do this than try to cover every single case and end up running out of time :)

app/views/request/_incoming_correspondence.html.erb
gbp: Done

app/models/incoming_message.rb
```diff
@@ -80,6 +80,14 @@ class IncomingMessage < ApplicationRecord
   delegate :multipart?, to: :raw_email
   delegate :parts, to: :raw_email
 
+  def legislation_references
+    LegislationReference.match(get_main_body_text_folded)
```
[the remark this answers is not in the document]
gbp: This doesn't work, I've left as `get_main_body_text_folded` for now

gbp:
I think this is ready for a review, its been rebased over the legislation work in #5945. This configure refusals you'll need to add:

```diff
diff --git a/app/models/legislation.rb b/app/models/legislation.rb
index f62d14eaa9..c71d146fbc 100644
--- a/app/models/legislation.rb
+++ b/app/models/legislation.rb
@@ -12,7 +12,8 @@ def self.all
         short: _('FOI'),
         full: _('Freedom of Information'),
         with_a: _('A Freedom of Information request'),
-        act: _('Freedom of Information Act')
+        act: _('Freedom of Information Act'),
+        refusals: ['s 12', 's 14', 's 40', 's 21', 's 31']
       ),
       new(
         key: 'eir',
```

gbp:
> ```ruby
> s = LegislationReference.new('Section 5(1)(b)')
> # => <LegislationReference::Section …>
> 
> s.section_number
> # => 5
> 
> s.subsection_number
> # => 1
> 
> s.paragraph
> # => 'b'
> 
> s.to_sym
> # => :section_5_1_b
> ```

I've opted not to go with adding these accessors as in the end these weren't needed as I added a `#cover?` method which deals with situations such as: `'s 12'.covers?('s 12(a)')` and offers more flexibility when comparing to the predefined list of `Legislation#refusals`.

app/models/incoming_message.rb
```diff
@@ -79,6 +79,15 @@ class IncomingMessage < ApplicationRecord
...
   delegate :multipart?, to: :raw_email
   delegate :parts, to: :raw_email
+  delegate :legislation, to: :info_request
+
+  def legislation_references
+    legislation.find_references(get_main_body_text_folded)
+  end
+
+  def refusals
+    legislation_references.select(&:refusal?)
```
garethrees, line 89: Nether of these have specs.

I think we can make `legislation_references` private for now at least since we're only using it internally.

Really, we just want a simple integration spec along the lines of:

Given a message with content X, we we should identify N number of refusals.

garethrees (approved):
All looks good on whole!

I'm not totally satisfied with overriding `Legislation.all` for configuration, but that's fine, we can improve that later.

I agree with the rubocop comments.

PR #5945, opened 2020-10-30
[#5909] Legislations

gbp:
## Relevant issue(s)

Connected to #5909

## What does this do?

Adds a legislation class to represent the law used to make an information request

## Why was this needed?

To pass the legislation through to the legislation reference detection - see #5931 - inorder to correctly detect exemptions.

## Implementation notes

By default this contains hardcoded FOI and EIR legislations but the intention is to allow themes to customise this and add other legislations for example the FOISA etc...

## Notes to reviewer

I've done more work (not included here) to remove `#eir_only?` conditionals through out the codebase. This has raised some isses with existings translations where we need to interpolate the translated law into other strings. It would be fine for English but other languages might use different wording based on the context. In order to do this properly we will need change from doing:
```
_(Your {{law_used}} request', law_used: 'Environmental Information Regulation') => 'Your Environmental Information Regulation request'
```
We should make use to gettext context feature:
```
_('foi|Your information request') => 'Your Freedom of Infomation request'`
_('eir|Your information request') => 'Your Environmental Information Regulation request'
```

It is made more complex due to themes being able to define their own legislations so we would need to do `_("#{law_used}|Your information request")`. This currently is blocked as these wouldn't be picked up by the `gettext:find` rake task.


locale/af_ZA/app.po
```diff
@@ -1895,7 +1895,7 @@ msgstr ""
 msgid "Invalid token"
 msgstr ""
 
-msgid "Is {{email_address}} the wrong address for {{type_of_request}} requests to {{public_body_name}}? If so, please contact us using this form:"
```
garethrees, line 1898: Preserving the existing translations is a good thing to do, but we need to be a bit careful about how we push these to Transifex without destroying translations that we haven't pulled into the repo yet. What was your rollout plan here?

spec/models/info_request_spec.rb
```diff
@@ -1770,16 +1694,18 @@
 
   describe 'when generating a user name slug' do
 
-    before do
```
garethrees, line 1740: This doesn't seem related to the commit

app/models/info_request.rb
```diff
@@ -888,11 +888,7 @@ def legislation
   end
 
   def law_used_human(key = :full)
-    begin
-      applicable_law.fetch(key)
-    rescue KeyError
-      raise "Unknown key '#{key}' for '#{law_used}'"
```
garethrees, line 893: We've lost this behaviour:

```ruby
r = InfoRequest.new(law_used: 'unknown')
r.law_used_human(:full)
# => "Freedom of Information"
```

I think an error is the correct thing to do when an invalid `law_used` key is provided.

app/models/info_request.rb
```diff
@@ -888,6 +888,8 @@ def legislation
   end
 
   def law_used_human(key = :full)
+    warn %q([DEPRECATION] law_used_human will be replaced with
```
garethrees, line 891: To make this clearer, I'd do:

```diff
- warn %q([DEPRECATION] law_used_human will be replaced with
+ warn %q([DEPRECATION] InfoRequest#law_used_human will be replaced with

File: app/models/info_request.rb [hunk 5]
```diff
@@ -1256,7 +1258,7 @@ def recipient_name_and_email
       # TRANSLATORS: Please don't use double quotes (") in this translation
       # or it will break the site's ability to send emails to authorities!
       _("{{law_used_short}} requests at {{public_body}}",
-        law_used_short: law_used_human(:short),
+        law_used_short: legislation.to_s,
```
garethrees, line 1261: Do we need to actually call `#to_s` for these? Might just call it by default?

doc/CHANGES.md
```diff
@@ -4,6 +4,12 @@
 
 ## Upgrade Notes
 
+* The `InfoRequest` method `law_used_human` has been deprecated and will be
+  removed in a future release. The new method `legislation` has been supplied
+  instead which is an Legislation instance with an `#to_s` method equivalent to
```
garethrees, line 9: Minor:

* s/an Legislation instance/a Legislation instance`
* s/an `#to_s` method/a `#to_s` method/

garethrees (commented):
Couple of minor comments but looks great in general! 🎠

spec/models/info_request_spec.rb
gbp: Actually this errors with these changes. I've updated the commit message in e670c9cc7

app/models/info_request.rb
gbp: Changed in 09e860c

doc/CHANGES.md
gbp: Changed in 40c0bfcc

app/models/info_request.rb
gbp: Of course - changed in 92367a7705

locale/af_ZA/app.po
gbp: Will need to do a manual process of:
1. pull from Transifex - this replace these changes
2. re-apply this commit patch
3. push back up to Transifex

app/models/track_thing.rb
```diff
@@ -229,11 +229,11 @@ def public_body_updates_params
                       public_body_name: public_body.name),
       verb_on_page_already: _("Following"),
       # Email
-      title_in_email: _("{{foi_law}} requests to '{{public_body_name}}'",
-                        foi_law: public_body.law_only_short,
+      title_in_email: _("{{law_used_short}} requests to '{{public_body_name}}'",
+                        law_used_short: public_body.law_used_human(:short),
```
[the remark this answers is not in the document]
gbp: Not sure how that happend - Reverted in 071e5fd633cf6

app/models/info_request.rb
gbp: Restored in 3673650e49

garethrees (approved):
LGTM. I haven't run any of this or looked through in a _huge_ amount of detail, but looks sensible, so happy if you're happy 🚀

Issue #5939, opened 2020-10-27
Build refusals advice Q&A interface


garethrees:
Requires https://github.com/mysociety/alaveteli/issues/5937.

We've decided to merge the best bits of the [wizard](https://github.com/mysociety/alaveteli/issues/5910#issuecomment-712778655) and [chatbot](https://github.com/mysociety/alaveteli/issues/5910#issuecomment-713685509) interfaces.

The idea is that:

* We present a question one-by-one
* Answering unfolds a new question until we run out of questions or the user gets bored
* We keep answered questions so that the user can understand the "history" of what they've done, in case they want to change their minds on answers
* The next steps get built up underneath the Q&A as before
* We number each "ground" for the next actions

garethrees:
Fixed by https://github.com/mysociety/alaveteli/pull/5948

Issue #5937, opened 2020-10-27
Parse refusal advice configuration and generate HTML


garethrees:
The [data format](https://github.com/mysociety/foi-refusals-prototype/blob/master/_data/section_12.yml) @zarino has created seems like a good starting point.

We'll need to parse a file like this, and [generate HTML](https://github.com/mysociety/foi-refusals-prototype/blob/master/_layouts/wizard.html) for each question.

We'll want to support translations of the data in some way. What are our options here?

* A separate file per locale (`refusal_advice.LOCALE.yml`)? Lots of duplication in defining the Q&A multiple times.
* An `erb.yml` file that uses gettext to add theme translations? Probably the simplest option, but limited in our use of Rails helpers?
* Instead of adding the `label` text directly in the config, use the name of a partial to render? This would be a bit annoying for the "Yes" or "No" labels, but more flexible for the main text labels, since we'll be able to use Rails helpers and the like.

If a re-user wants to include the refusal advice Q&A on a help page, they should be able to do so like so:

```erb
<%= refusal_advice(refusal_reasons: params[:refusal_reasons]) %>
```

We'll need a way of pre-selecting any identified refusals, so we could do that through the params.

garethrees:
I've mostly broken the back of the work to parse our data files in https://github.com/mysociety/alaveteli/pull/5951.

Now, the question comes to how we _localise_ them.

I've added the option to the `label` to be able to use Rails' built in `render` to use either plain text, html, or a partial:

```yaml
      label: {
        plain: 'Refusing a request on cost grounds…'
      },

      label: {
        partial: 'path/to/_partial'
      },
```

In the latter case, we can use the normal gettext mechanism.

Rails supports erb config templates, and it's easy to embed gettext strings in YAML files:

```diff
diff --git a/lib/tasks/gettext.rake b/lib/tasks/gettext.rake
index 779bbfb84..64468e86e 100644
--- a/lib/tasks/gettext.rake
+++ b/lib/tasks/gettext.rake
@@ -177,7 +177,7 @@ def find_theme(theme)

   def theme_files_to_translate
     theme = find_theme(ENV['THEME'])
-    Dir.glob("{lib/themes/#{theme}/lib}/**/*.{rb,erb}")
+    FileList.new("lib/themes/#{theme}/{lib,config}/**/*.{rb,erb}")
   end

   def theme_locale_path
```

```erb
# THEME_ROOT/config/refusal_advice/all.yml.erb
      label: {
[...]

PR #5948, opened 2020-11-03
[#5937] JRCT refusals wizard styling, and interactivity

zarino:
## Relevant issue(s)

#5910 #5939

## What does this do?

Static example of markup and styling for the refusals checklist.

Markup is all hard-coded for now. Aim is that eventually questions will be generated by the server.

The JavaScript plugin handles progressing through the questions and updating the status feedback elements. There’s no special logic yet for branching based on input values.

## Screenshots

foi-checklist

### DIFF
--- a/app/assets/javascripts/wizard.js
+++ b/app/assets/javascripts/wizard.js
@@ -1,0 +1,283 @@
+(function($) {
+  // TODO: Needed:
+  // Questions remaining text
+  // x Suggestion markup
+  // x Combining suggestions - group by suggestion.action
+  // TODO: Nice to have:
+  // x Initial state on reload
+  // Local storage ?
+  // x Other actions always visible - dynamically remove if suggested
+
+  var RefusalWizard = function(target, options) {
+    this.$el = $(target);
+
+    var defaults = {
+      debug: false,
+
+      questionClass: "wizard__question",
+      questionAnswerableClass: "wizard__question--answerable",
+      questionAnsweredClass: "wizard__question--answered",
+
+      suggestionClass: "wizard__suggestion",
+      suggestionActiveClass: "wizard__suggestion--active",
+
+      actionClass: "wizard__action",
+      actionActiveClass: "wizard__action--active",
+
+      nextStepClass: "wizard__next-step",
+      nextStepActiveClass: "wizard__next-step--active"
+    };
+
+    this.options = $.extend(true, defaults, options);
+
+    this._init(target);
+
+    return this;
+  };
+
+  RefusalWizard.prototype._init = function(target) {
+    var wizard = this;
+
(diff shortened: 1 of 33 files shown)


garethrees (commented):
Looks good in principle, really like how the interaction works.

This doesn't include examples of [all possible fields/attributes](https://github.com/mysociety/alaveteli/pull/5951/commits/d5e012eaae41faa6790a1f25155cba4273e42728#diff-4f95384fc60227db65d815a4c18dc265d60d187a5f4a872c6bf27246bdf05048R4-R26), so we'll need those to build the HTML.

Minor nitpick, but Alaveteli should use 2-space indentation in all files.

app/assets/javascripts/wizard.js
```diff
@@ -0,0 +1,113 @@
...
+  };
+
+  // Declare public methods here. They should return the
+  // plugin target (this.$el) for jQuery chaining, eg:
+  //
+  // Plugin.prototype.examplePublicMethod = function () {
+  //   return this.$el;
+  // };
+
+  $.fn[ WIZARD_PLUGIN_NS ] = function( methodOrOptions ) {
```
gbp, line 75: The code from this point took me a while to understand what it was doing. Am I correct in assuming this is boilerplate code for custom jQuery plugins to get the `.refusalWizard` method on jQuery elements?

I'm curious why you chose a plugin as I don't think we've currently used this approach anywhere else in Alaveteli?

I've tired stayed away from using jQuery inorder to make life easier when upgrading to webpack once we get the Rails upgrades out of the way.

It a shame thats not ready yet now that we're adding more complex UI elements. For now I'm happy to continue with this jQuery approach.

app/assets/javascripts/wizard.js
```diff
@@ -0,0 +1,113 @@
+(function ($) {
+
+  var WIZARD_PLUGIN_NS = 'refusalWizard';
+
+  var Plugin = function (target, options) {
```
gbp, line 5:
```suggestion
  var RefusalWizard = function (target, options) {
```

gbp (commented):
Good start. A couple of comments/questions inline.

app/assets/javascripts/wizard.js
zarino: I started writing it as just a collection of lines of javascript, but once I broke different parts of the process out into separate functions, I started spending a lot of time passing references to DOM elements around between those functions, and setting global variables to hold information about the wizard’s state.

Wrapping it all up as a jQuery plugin solved a lot of these issues for me, and helped me keep the actual nuts and bolts of the interaction stuff quite simple. But I agree, that comes at the cost of quite a lot of boilerplate.

Happy to pair on transforming it to a different format, if you’d prefer?

PR #5951, opened 2020-11-04
[#5937] JRCT – Parse refusal advice YAML files

garethrees:
## Relevant issue(s)

Part of https://github.com/mysociety/alaveteli/issues/5937.

## What does this do?

Parses refusal advice from YAML into usable Ruby objects.

## Implementation notes

Allows flexibility in keeping all advice in a single file, or spreading
over several files.

Handles name-spacing questions under a legislation key.

## Notes to reviewer

Refusal advice can be defined in a single file…

```yml
# tmp/refusal_advice_example.yml
foi:
  - id: have-confirmation
    label:
      plain: Has the authority confirmed or denied?
    options:
    - label: "Yes"
      value: "yes"
    - label: "No"
      value: "no"

eir:
  - id: is-confirmation-too-costly
    show_if:
    - id: some-other-question
      operator: is
      value: "no"
    label:
      partial: path/to/partial
    options:
    - label: "Yes"
      value: "yes"
    - label: "No"
      value: "no"
```

…or a file per legislation…

```yml
# tmp/foi.yml
foi:
  - id: have-confirmation
    label:
      plain: Has the authority confirmed or denied?
    options:
    - label: "Yes"
      value: "yes"
    - label: "No"
      value: "no"

# tmp/eir.yml
eir:
  - id: is-confirmation-too-costly
    show_if:
    - id: some-other-question
      operator: is
      value: "no"
    label:
      partial: path/to/partial
    options:
    - label: "Yes"
      value: "yes"
    - label: "No"
      value: "no"
```

…or any number of files that can be globbed from a directory. It all gets munged into a data structure of `legislation_key => [collection of questions]`.

```ruby
r = RefusalAdvice.load(Rails.root + 'tmp/refusal_advice_example.yml')
r.questions(:foi).first.id
# => 'have-confirmation'
```

Refusal can be added to the theme simply by creating the YAML files in `THEME_ROOT/config/refusal_advice/*.yml`.

### DIFF
--- a/app/models/refusal_advice.rb
+++ b/app/models/refusal_advice.rb
@@ -1,0 +1,35 @@
+##
+# A collection of Questions that help users challenge refusals.
+#
+class RefusalAdvice
+  def self.default
+    files = Rails.configuration.paths['config/refusal_advice'].existent
+    new(Store.from_yaml(files))
+  end
+
+  def initialize(data)
+    @data = data
+  end
+
+  def legislation
+    Legislation.default
+  end
+
+  def questions
+    data[legislation.key.to_sym][:questions].
+      map { |question| Question.new(question) }
+  end
+
+  def actions
+    data[legislation.key.to_sym][:actions].
+      map { |action| Action.new(action) }
+  end
+
+  def ==(other)
+    data == other.data
+  end
+
+  protected
+
+  attr_reader :data
+end
--- a/app/models/refusal_advice/question.rb
+++ b/app/models/refusal_advice/question.rb
@@ -1,0 +1,12 @@
+##
+# A single question that we present to users to help them challenge refusals.
+#
+class RefusalAdvice::Question < RefusalAdvice::Block
+  def hint
+    renderable_object(data[:hint])
+  end
+
+  def options
+    collection(data[:options])
+  end
+end
--- a/config/initializers/theme_loader.rb
+++ b/config/initializers/theme_loader.rb
@@ -3,13 +3,20 @@
 # It is used by our config/routes.rb to decide which route extension files to load.
 $alaveteli_route_extensions = []
 
+def theme_root(theme_name)
+  Rails.root.join('lib/themes', theme_name)
+end
+
 def require_theme(theme_name)
-  theme_lib = Rails.root.join 'lib', 'themes', theme_name, 'lib'
+  root = theme_root(theme_name)
+  theme_lib = root.join('lib')
   $LOAD_PATH.unshift theme_lib.to_s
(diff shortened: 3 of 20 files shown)


config/initializers/theme_loader.rb
```diff
@@ -3,15 +3,34 @@
...
+  require theme_main_include
+
+  Rails.configuration.paths.add(
+    'config/refusal_advice',
+     with: root.join('config/refusal_advice'),
+     glob: '*.yml'
+  )
 end
 
+Rails.configuration.paths.add(
```
garethrees, line 28: Did I mean to commit this? Hmm.

app/models/refusal_advice.rb
```diff
@@ -0,0 +1,27 @@
+# A collection of Questions that help users challenge refusals.
+class RefusalAdvice
+  def self.default
+    files = Rails.configuration.paths['config/refusal_advice'].existent
+    new(RefusalAdvice::Data.from_yaml(files))
```
garethrees, line 5: Zarino is getting an error along the lines of "Undefined method `from_yaml` on class `Data`". His VM is using Ruby 2.3. Works for me on 2.5 and doesn't fail in CI.

garethrees:
Next steps:

- [ ] Resolve Ruby 2.3 namespace issue calling `.from_yaml` on top-level `Data` constant.
- [ ] Resolve CI failure (ordering of file glob vs hard-coded ruby example data)
- [ ] Parse [updated data file](https://github.com/mysociety/alaveteli/pull/5954#issuecomment-724032497) which makes clear distinction between question and suggestion items
  * Add a `RefusalAdvice::Suggestion` class
  * Suggestion gets a label now
  * Hint is now an "object" rather than a "value"
- [ ] Return HTML-safe strings for `Question`/`Suggestion` methods that can be handed to `render`.

config/initializers/theme_loader.rb
garethrees: Probably not – it's unlikely we'll provide any default refusal advice configuration.

garethrees: We can drop this; no need to worry about none theme as it'll only be us using it.

app/models/refusal_advice/question.rb
```diff
@@ -0,0 +1,22 @@
...
+class RefusalAdvice::Question < RefusalAdvice::Block
+  def hint
+    renderable_object(question[:hint])
+  end
+
+  def options
+    collection(question[:options])
+  end
+
+  def to_partial_path
```
garethrees, line 13: Probs move this to "render" branch

garethrees:
Removed `awaiting-deploy` because there's a genuine build failure here.

PR #5965, opened 2020-11-10
Gh pages add wobknopargentina

JenMysoc:
## Relevant issue(s)
Closes https://github.com/mysociety/alaveteli-project/issues/118 

## What does this do?
Adds Argentina and Wob-Knop to the deployments page

## Why was this needed?
Showcasing up to date list of sites


zarino (changes requested):
Thanks for doing this @JenMysoc! But we should definitely optimise the filesize of those images. They should probably be 600x400 pixels too, to match the aspect ratio of all the others.

zarino (approved):
I‘ve created new images at the right size, and I‘ve also copied the changes onto the spanish version of the page too. This looks good to go from my perspective.

gbp (approved):

Issue #5967, opened 2020-11-12
[ERROR] script/update-xapian-index warning: regular expression has ']' without escape


garethrees:
```
[cron-alaveteli] Cron <foi@redacted> /usr/bin/env XAPIAN_MAX_CHANGESETS=10 run-with-lockfile -n /APP_ROOT/change-xapian-database.lock "/APP_ROOT/alaveteli/script/update-xapian-index verbose=true" >> /APP_ROOT/alaveteli/log/update-xapian-index.log || echo "stalled?"

/APP_ROOT/alaveteli-2020-10-15T14-44-42/app/models/censor_rule.rb:119: warning: regular expression has ']' without escape
/APP_ROOT/alaveteli-2020-10-15T14-44-42/app/models/censor_rule.rb:119: warning: regular expression has ']' without escape
/APP_ROOT/alaveteli-2020-10-15T14-44-42/app/models/censor_rule.rb:119: warning: regular expression has ']' without escape
/APP_ROOT/alaveteli-2020-10-15T14-44-42/app/models/censor_rule.rb:119: warning: regular expression has ']' without escape
/APP_ROOT/alaveteli-2020-10-15T14-44-42/app/models/censor_rule.rb:119: warning: regular expression has ']' without escape
/APP_ROOT/alaveteli-2020-10-15T14-44-42/app/models/censor_rule.rb:119: warning: regular expression has ']' without escape
/APP_ROOT/alaveteli-2020-10-15T14-44-42/app/models/censor_rule.rb:119: warning: regular expression has ']' without escape
/APP_ROOT/alaveteli-2020-10-15T14-44-42/app/models/censor_rule.rb:119: warning: regular expression has ']' without escape
/APP_ROOT/alaveteli-2020-10-15T14-44-42/app/models/censor_rule.rb:119: warning: regular expression has ']' without escape
/APP_ROOT/alaveteli-2020-10-15T14-44-42/app/models/censor_rule.rb:119: warning: regular expression has ']' without escape
/APP_ROOT/alaveteli-2020-10-15T14-44-42/app/models/censor_rule.rb:119: warning: regular expression has ']' without escape
/APP_ROOT/alaveteli-2020-10-15T14-44-42/app/models/censor_rule.rb:119: warning: regular expression has ']' without escape
/APP_ROOT/alaveteli-2020-10-15T14-44-42/app/models/censor_rule.rb:49: warning: regular expression has ']' without escape
[...]

PR #5969, opened 2020-11-12
[#5967] Prevent CensorRules being saved when Regexp is bad

garethrees:
We allow the regexp to be created from a user inputted string.

In some cases, we see rules that don’t have correct escaping, e.g:

    Regexp.new('foo]')
    # (irb):19: warning: regular expression has ']' without escape: /foo]/
    # => /foo]/

This warning behaviour is not configurable, so instead I’ve added
Warning.with_raised_warnings. This simply makes any call to Warning.warn
raise the given message instead of outputting it to stderr.

We then use the raised exception to massage the message into a
RegexpError, which we already have handling that presents this as an
ActiveModel error to the user.

Fixes https://github.com/mysociety/alaveteli/issues/5967

Screenshot 2020-11-12 at 11 16 13

```ruby
r.text
# => "foo\\]"

r.replacement
# => "bar"

r.apply_to_text('foo] bar')
# => "bar bar"
```

### DIFF
--- a/app/models/censor_rule.rb
+++ b/app/models/censor_rule.rb
@@ -117,7 +117,10 @@ class CensorRule < ApplicationRecord
   end
 
   def make_regexp(encoding)
-    Regexp.new(encoded_text(encoding), Regexp::MULTILINE)
+    ::Warning.with_raised_warnings do
+      Regexp.new(encoded_text(encoding), Regexp::MULTILINE)
+    end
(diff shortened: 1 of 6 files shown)


garethrees:
Should be able to clean all this up and resolve remaining build failures after https://github.com/mysociety/alaveteli/pull/6052

gbp (approved):
This works well

PR #5984, opened 2020-11-25
[#5269] Limit incoming message guesses

### LINKED ISSUE #5269: IncomingMessage.guess_by_incoming_subject can return too many things
There's an incoming message currently sitting in the holding pen which - as the subject line is short and generic - has 4376 guesses for the subject line alone. And is listing all of them.

The admin page for the raw email itself is slow loading and the results are not particularly helpful.

IncomingMessage: 1386337

gbp:
## Relevant issue(s)

Fixes #5269

## What does this do?

Limit number of info requests guessed by subject 

## Why was this needed?

Receiving a request with a subject like "Freedom of Information Request"
which ends up in the holding pen will cause issues for administrations.
The number of info request guesses would be far too large and this
potentially breaks the admin UI due to the page timing out when
rendering each info request guess.

This change limits to just 25 info requests. While it would be good to
allow these requests to be paginated. It's more likely the admin would
inspect the raw email instead of going through pages of guesses.

### DIFF
--- a/app/models/info_request.rb
+++ b/app/models/info_request.rb
@@ -325,7 +325,8 @@ class InfoRequest < ApplicationRecord
 
     requests = requests_by_title.or(requests_by_subject).
       distinct.
-      where.not(url_title: 'holding_pen')
+      where.not(url_title: 'holding_pen').
+      limit(25)
 
     guesses = requests.each.reduce([]) do |memo, request|
       memo << Guess.new(request, subject_line, :subject)
--- a/app/models/info_request.rb
+++ b/app/models/info_request.rb
@@ -314,16 +314,19 @@ class InfoRequest < ApplicationRecord
 
     # try to find a match on InfoRequest#title
     reply_format = InfoRequest.new(title: '').email_subject_followup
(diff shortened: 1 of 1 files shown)


garethrees:
Should this have a spec before review?

gbp:
> Should this have a spec before review?

In was on the fence but in the end I decided not to add one. Seems poinless (and slow) to create 25+ records in the test database to test this when its a core Rails method which is tested upstream.

garethrees (approved):
> > Should this have a spec before review?
>
> In was on the fence but in the end I decided not to add one. Seems poinless (and slow) to create 25+ records in the test database to test this when its a core Rails method which is tested upstream.

Yeah, ISWYM.

I don't see the test so much as testing `limit` but on how `guess_by_incoming_subject` works overall, more as documentation in this case.

That said, I don't feel strongly about this case, and there is a tradeoff in spec performance, so yeah, let's go with this. Nice one! 👍 🍰

gbp:
This has been pushed directly to the WDTK branch so it can be released today

PR #5985, opened 2020-11-25
[gh-pages] Highlight pro sites on deployment page

zarino:
## What does this do?

Updates the `/deployments` and `/es/deployments` pages to…

* include an "Alaveteli Pro" page on any Pro deployments, which links to that deployment’s `/pro` page
* feature a more up-to-date list of deployments (with all the Pro sites featured at the top)
* be generated automatically from the same `deployments.yml` file, preventing them from getting out of sync
* use a proper column layout, rather than `display: inline`, preventing the need for HTML comments between list items

## Why was this needed?

The yaml and column refactoring wasn’t strictly necessary for the requested change (displaying the "Alaveteli Pro" badges) but it’s long been a source of annoyance that the English and Spanish pages were frequently out of sync with each other, and the weird list layout (which required HTML comments between list items to prevent them breaking onto new lines) was just odd and made editing the pages even tricker.

Putting all the deployments data into a single yaml file should make it easier to add / remove deployments from the page, and will also prevent the deployments differing between the English and Spanish pages.

## Screenshots

Screenshot_2020-11-25 Deployments Alaveteli mySociety(1)

## Notes to reviewer

Still to do:

1. Spanish translations for the new "featured" sites – @JenMysoc is working on these

Questions:

1. Do we want to give the lists an explicit sort order? Like, sorting by site name, A-Z? (Personally, I find it a bit odd that the sites are aparently randomly sorted.)
2. Do we need to explain the concept of Alaveteli Pro anywhere on this page?
3. While we’re here, should we update the header to say "29 jurisdictions" (by my count) rather than "25"? Are there any other numbers on this page that are out of date?


garethrees (approved):
Looks great to me!

garethrees:
> 1. Do we want to give the lists an explicit sort order? Like, sorting by site name, A-Z? (Personally, I find it a bit odd that the sites are aparently randomly sorted.)

Nope, I think it's fine as is.

> 2. Do we need to explain the concept of Alaveteli Pro anywhere on this page?

I don't think we _need_ to in order to merge this PR – clicking through describes what it's all about.

> 3. While we’re here, should we update the header to say "29 jurisdictions" (by my count) rather than "25"? Are there any other numbers on this page that are out of date?

How about:

```diff
- 20+ languages, 25 jurisdictions
+ 20+ languages, 25+ jurisdictions
- 800,000 requests for information
+ 800,000+ requests for information
```

garethrees:
@JenMysoc was this what you had in mind?

PR #5987, opened 2020-11-26
Fix HasTagString#add_tag_if_not_already_present

garethrees:
The original implementation would add the tag irrespective of whether it
was already present, as it simply concatenated the existing and given
strings together.

This makes the method do what it says, and also fixes minor style
issues.

* Fixes quote style
* Fixes redundant self

### DIFF
--- a/lib/has_tag_string/has_tag_string.rb
+++ b/lib/has_tag_string/has_tag_string.rb
@@ -135,7 +135,8 @@ module HasTagString
 
     # Adds a new tag to the model, if it isn't already there
     def add_tag_if_not_already_present(tag_as_string)
-      self.tag_string = self.tag_string + " " + tag_as_string
+      return tag_string if has_tag?(tag_as_string)
+      self.tag_string = tag_string + ' ' + tag_as_string
     end
   end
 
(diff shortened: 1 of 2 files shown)


lib/has_tag_string/has_tag_string.rb
```diff
@@ -135,7 +135,8 @@ def get_tag_values(tag_as_string)
 
     # Adds a new tag to the model, if it isn't already there
     def add_tag_if_not_already_present(tag_as_string)
-      self.tag_string = self.tag_string + " " + tag_as_string
+      return tag_string if tag_string.include?(tag_as_string)
```
gbp, line 138: This would prevent `foo` being added if `foobar` is already present. Real world example would be adding `NHS` if `NHSChoicesID` was present

gbp (changes requested):

lib/has_tag_string/has_tag_string.rb
garethrees: Ah good catch! Fixed in 5c265d33d.

gbp (approved):

Issue #5982, opened 2020-11-24
Default vagrant setup doesn't start up because of missing virtualbox guest additions


mlandauer:
With a default vagrant (without any plugins installed) doing a fresh checkout of the latest release of the alavateli code:
```
$ git checkout 0.38.3.0
$ git submodule update --init
$ vagrant up
Bringing machine 'default' up with 'virtualbox' provider...
==> default: Box 'debian/stretch64' could not be found. Attempting to find and install...
    default: Box Provider: virtualbox
    default: Box Version: >= 0
...
...
==> default: Checking for guest additions in VM...
    default: No guest additions were detected on the base box for this VM! Guest
    default: additions are required for forwarded ports, shared folders, host only
    default: networking, and more. If SSH fails on this machine, please install
    default: the guest additions and repackage the box to continue.
    default: 
    default: This is not an error message; everything may continue to work properly,
    default: in which case you may ignore this message.
==> default: Setting hostname...
==> default: Configuring and enabling network interfaces...
==> default: Mounting shared folders...
    default: /home/vagrant/alaveteli => /home/matthew/git/oaf/alaveteli
Vagrant was unable to mount VirtualBox shared folders. This is usually
because the filesystem "vboxsf" is not available. This filesystem is
made available via the VirtualBox Guest Additions and kernel module.
Please verify that these guest additions are properly installed in the
guest. This is not a bug in Vagrant and is usually caused by a faulty
Vagrant box. For context, the command attempted was:

mount -t vboxsf -o uid=1000,gid=1000 home_vagrant_alaveteli /home/vagrant/alaveteli

The error output from the command was:
```

Following advice in https://github.com/scotch-io/scotch-box/issues/296#issuecomment-288759311 I added a vagrant plugin to automatically install the guest additions:
```
vagrant plugin install vagrant-vbguest
vagrant vbguest
```

Then doing `vagrant up` worked perfectly.

If you can reproduce this error I'd be happy to put in a PR to suggest adding the vagrant plugin in the documentation in `Vagrantfile` and at https://alaveteli.org/docs/installing/vagrant/

garethrees:
> I added a vagrant plugin to automatically install the guest additions

Ah, yes, I always have this installed!

> If you can reproduce this error I'd be happy to put in a PR to suggest adding the vagrant plugin in the documentation in `Vagrantfile` and at https://alaveteli.org/docs/installing/vagrant/

Yeah, that sounds like a great idea! 🙇

[...]

PR #5992, opened 2020-11-30
Add vagrant plugin docs to main documention

mlandauer:
This is the second part to the fix for #5982


garethrees:
Thanks! Made a minor tweak (you don't need to manually run `vbguest`, and at the point of telling people to run it they don't have the repo cloned yet) and merged in 752c934b2.

PR #5997, opened 2020-12-01
Revert removal of temp:migrate_admins_and_pros_to_roles rake task

mlandauer:
This is a partial revert of f12018c46b4539d85d8487a27eb55698fa13d2fe

Other rake tasks in temp used for migrating old versions are kept around in case 
people are upgrading from very old versions. This was one is missing because it was
removed (I'm assuming accidentally).

### DIFF
--- a/lib/tasks/temp.rake
+++ b/lib/tasks/temp.rake
@@ -117,6 +117,20 @@ namespace :temp do
     end
   end
 
+  desc 'Migrate admins and pro users to role-based backend'
+  task :migrate_admins_and_pros_to_roles => :environment do
+    User.where(:admin_level => 'super').each do |admin|
+      admin.add_role :admin
+    end
+    pro_users = User.
+      includes(:pro_account).
+        where("pro_accounts.id IS NOT NULL").
+          references(:pro_accounts)
+    pro_users.each do |pro_user|
+      pro_user.add_role :pro
+    end
+  end
+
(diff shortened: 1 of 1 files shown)


mlandauer:
Oops. I pasted the wrong commit hash. It's supposed to be 9b9202d7237ac739ed00d1e883883b1d9d39500c

garethrees:
> Other rake tasks in temp used for migrating old versions are kept around in case
people are upgrading from very old versions. This was one is missing because it was
removed (I'm assuming accidentally).

The removal of `temp:migrate_admins_and_pros_to_roles` would have been intentional, as we always recommend upgrading to each major release individually so that the necessary tasks exist at the right points in time. In general the tasks in `temp` namespace are indeed temporary and, when we remember, get removed after they've served their purpose.

Thanks for the contribution – all are very much appreciated – but I think we'll decline this one since it's not something we need going forward. 🙏

mlandauer:
Okay that makes perfect sense. In that case please remove all the other old migration scripts in temp for consistency. We're upgrading from a 0.25 release and *all* the other temp migration scripts required for upgrading are still there in the current release *except* for the temp:migrate_admins_and_pros_to_roles.

garethrees:
> In that case please remove all the other old migration scripts in temp for consistency.

Done in https://github.com/mysociety/alaveteli/pull/6000 🙏

Issue #6008, opened 2020-12-09
ES6 syntax breaks asset compilation


garethrees:
```
rake aborted!
Uglifier::Error: Unexpected token name «of», expected punc «;». To use ES6 syntax, harmony mode must be enabled with Uglifier.new(:harmony => true).
--
 19971         return false;
 19972       }
 19973
 19974       if (!showIfArray) {
 19975         wizard.log("MATCH:", this.block, "no show-if");
 19976         return true;
 19977       }
 19978
    =>       for (var showIf of showIfArray) {
 19980         // can't be a dependent if showIf is for a different question ID
 19981         // check showIf operator
 19982         if (
 19983           showIf.id === question.block &&
 19984           showIf.operator === "is" &&
 19985           question.values().indexOf(showIf.value) > -1
 19986         ) {
 19987           wizard.log("MATCH:", this.block, showIf, question.values());
==
/data/vhost/www.whatdotheyknow.com/alaveteli-2020-12-09T12-27-54/app/assets/config/manifest.js:2
/data/vhost/www.whatdotheyknow.com/shared/vendor/bundle/ruby/2.6.0/gems/uglifier-4.2.0/lib/uglifier.rb:291:in `parse_result'
/data/vhost/www.whatdotheyknow.com/shared/vendor/bundle/ruby/2.6.0/gems/uglifier-4.2.0/lib/uglifier.rb:221:in `run_uglifyjs'
/data/vhost/www.whatdotheyknow.com/shared/vendor/bundle/ruby/2.6.0/gems/uglifier-4.2.0/lib/uglifier.rb:166:in `compile'
/data/vhost/www.whatdotheyknow.com/shared/vendor/bundle/ruby/2.6.0/gems/sprockets-3.7.2/lib/sprockets/uglifier_compressor.rb:53:in `call'
/data/vhost/www.whatdotheyknow.com/shared/vendor/bundle/ruby/2.6.0/gems/sprockets-3.7.2/lib/sprockets/uglifier_compressor.rb:28:in `call'
/data/vhost/www.whatdotheyknow.com/shared/vendor/bundle/ruby/2.6.0/gems/sprockets-3.7.2/lib/sprockets/processor_utils.rb:75:in `call_processor'
/data/vhost/www.whatdotheyknow.com/shared/vendor/bundle/ruby/2.6.0/gems/sprockets-3.7.2/lib/sprockets/processor_utils.rb:57:in `block in call_processors'
[...]

PR #6009, opened 2020-12-09
[#6008] Use Uglifier's experimental Harmony mode for ES6

garethrees:
4d2ab5a941 introduces ES6 syntax, which Uglifier doesn't support by
default [1].

This commit enables the experimental "Harmony mode", along with
disabling `mangle` and `compress` as per the usage recommendations [2].

This doesn't seem like a good long-term solution, since development of
Harmony mode has stopped and may be removed. This at least allows us to
deploy 4d2ab5a941 for now.

Fixes https://github.com/mysociety/alaveteli/issues/6008.

[1] https://github.com/lautis/uglifier#es6--es2015--harmony-mode
[2] https://github.com/lautis/uglifier#usage
[3] https://github.com/lautis/uglifier/issues/142#issuecomment-404872931

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -20,7 +20,8 @@ Rails.application.configure do
   config.public_file_server.enabled = ENV['RAILS_SERVE_STATIC_FILES'].present?
 
   # Compress JavaScripts and CSS.
-  config.assets.js_compressor = :uglifier
+  config.assets.js_compressor =
+    Uglifier.new(harmony: true, mangle: false, compress: false)
   # config.assets.css_compressor = :sass
 
   # Do not fallback to assets pipeline if a precompiled asset is missed.


garethrees:
The alternative is to remove ES6 syntax from 4d2ab5a,  but I don't know off hand how tricky that would be.

gbp:
Superseded by #6010

gbp:
Looks like we might need this after all as ES6 syntax is also used in clipboard.js introduced in 366d87c54f7af47a5ee5c660ba443f80a8ebb89e

gbp (approved):
uglifier HEAD has dropped the harmony option and are recommending https://github.com/ahorek/terser-ruby as a better option. 

I've tested both and exactly the same JS is outputted. 

For now lets stick with uglifier until:
1. the next version is relaesed and switch to terser then 
2. we migrate to webpacker

Issue #6002, opened 2020-12-07
Configure whatdotheyknow-theme with Legislation References


garethrees:
Need to add refusals for WDTK.

https://github.com/mysociety/alaveteli/pull/5931

PR #6015, opened 2020-12-10
[#6002] Add refusals configuration

garethrees:
## Relevant issue(s)

Part of #6002.

## What does this do?

Allows refusals to be configured without theme overrides.

## Why was this needed?

Theme overrides are gross and more difficult to maintain.

## Implementation notes

Configure in theme like:

```ruby
# THEME/lib/model_patches.rb
Legislation.refusals = {
  foi: ['s 12', 's 14', 's 31'],
  eir: []
}
```

### DIFF
--- a/app/models/legislation.rb
+++ b/app/models/legislation.rb
@@ -5,21 +5,37 @@ class Legislation
   UnknownLegislation = Class.new(StandardError)
   UnknownLegislationVariant = Class.new(StandardError)
 
+  def self.refusals=(hash)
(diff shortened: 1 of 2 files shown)


gbp (commented):
Comment left theme PR here: https://github.com/mysociety/whatdotheyknow-theme/pull/744#pullrequestreview-569291444

Making that change would remove the need for `with_indifferent_access` too which I would prefer.

garethrees:
Requires https://github.com/mysociety/alaveteli/pull/6072

Issue #6003, opened 2020-12-07
Allow refusal advice to be translated


garethrees:
Part of https://github.com/mysociety/alaveteli/issues/5937.

See https://github.com/mysociety/alaveteli/issues/5937#issuecomment-722271911 and https://github.com/mysociety/alaveteli/issues/5937#issuecomment-722284979 for basic idea of how we'll do this.

Issue #6011, opened 2020-12-10
Upgrade to Rails 6.1


gbp:
Now out: https://weblog.rubyonrails.org/2020/12/9/Rails-6-1-0-release/

- [x] Requires #5159

* [Release notes](https://guides.rubyonrails.org/v6.1/6_1_release_notes.html)
* [Upgrade notes](https://guides.rubyonrails.org/v6.1/upgrading_ruby_on_rails.html#upgrading-from-rails-6-0-to-rails-6-1)

garethrees:
May require https://github.com/mysociety/sysadmin/issues/1407.

garethrees:
From the upgrade notes it doesn't look like much, if anything, will have a major impact on us.

gbp:
Looks like some changes might be required in our `conditionallyprependlocale` route filter.

See: https://github.com/svenfuchs/routing-filter/commit/8b85058acf5503877e3830158afbbbc415eae71f#diff-ff2e2792509c994a000cb388544e2b4452a96b6eaf8c065382f09f6e124ece9c

gbp:
Need to run `bundle exec rails active_storage:update` and commit new migrations

PR #6016, opened 2020-12-10
[#6003] Allow refusal advice data to be translated

garethrees:
## Relevant issue(s)

Fixes https://github.com/mysociety/alaveteli/issues/6003.

## What does this do?

Allow refusal advice data to be translated

## Why was this needed?

So that refusal advice can be presented in all locales that the site supports.

## Implementation notes

Uses existing translation mechanism for translations so that we can benefit from crowdsourcing of the translations and handling of falling back to default locale.

Uses Rails' `ActiveSupport::ConfigurationFile` [1] to parse ERB embedded in the YAML configuration files. Once we upgrade to Rails 6.1 (https://github.com/mysociety/alaveteli/issues/6011) we'll get a deprecation warning telling us to remove the patch.

[1] https://github.com/rails/rails/blob/v6.1.0/activesupport/lib/active_support/configuration_file.rb

## Notes to reviewer

```sh
# Add some refusal advice…
$ cat config/refusal_advice/data.yml.erb
foi:
  group:
  - id: <%= _('Refusal Advice') %>

# Find new translation strings
$ THEME=whatdotheyknow-theme bin/rake gettext:find_theme

# Strings are added
$ rg "Refusal Advice" locale-theme/
locale-theme/en/app.po
268:msgid "Refusal Advice"

locale-theme/app.pot
268:msgid "Refusal Advice"
```

```ruby
# Check that we actually load the translated string in the app context
AlaveteliLocalization.with_locale(:cy) { RefusalAdvice::Store.from_yaml(Dir.glob('./spec/fixtures/refusal_advice/data/*.yml.erb')) }
# => #<RefusalAdvice::Store:0x000056211506d088 @data=[{"foi"=>{"group"=>[{"id"=>"Ceisiadau Rhyddid Gwybodaeth"}]}}]>

```

### DIFF
--- a/app/models/refusal_advice/store.rb
+++ b/app/models/refusal_advice/store.rb
@@ -7,7 +7,7 @@ require 'yaml'
 class RefusalAdvice::Store
   def self.from_yaml(files)
     yamls = files.sort.inject([]) do |memo, file|
-      yaml = YAML.load(File.read(file))
+      yaml = ActiveSupport::ConfigurationFile.parse(file)
       memo << yaml if yaml
       memo
     end
--- a/config/initializers/alaveteli.rb
+++ b/config/initializers/alaveteli.rb
@@ -66,6 +66,7 @@ require 'alaveteli_mail_poller'
 require 'safe_redirect'
 require 'alaveteli_pro/metrics_report'
 require 'alaveteli_pro/webhook_endpoints'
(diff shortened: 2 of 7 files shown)


gbp (approved):
Nice. This works great. LGTM

PR #5621, opened 2020-04-18
[#1431] Fix transient spec failure

gbp:
## Relevant issue(s)

Fixes #1431 
Fixes #4346

## What does this do?

Fixes: 

```
1) When errors occur when considering all requests local (by default all in development) should show a full trace for general errors
   Failure/Error: expect(response.body).to match('<div id="traces"')
     expected "" to match "<div id=\"traces\""
   # ./spec/integration/errors_spec.rb:39:in `block (3 levels) in <top (required)>'
```

## Why was this needed?

This seems to be one of the most common spec failures.

## Notes to reviewer

All the seeds I've tried, including the ones on the linked issues, now work. I've also ran this in a loop for a while without seeing failures.
```
rspec spec/integration/errors_spec.rb --seed=16271
rspec spec/integration/errors_spec.rb --seed=46014
rspec spec/integration/errors_spec.rb --seed=58016
```

PR #6019, opened 2020-12-11
[#6018] fix hidden exceptions in integration specs

### LINKED ISSUE #6018: Rendering exceptions while running integration specs can disguise failures
Due to having `Rails.application.config.action_dispatch.show_exceptions = true` in a before suite callback [1] means Rails' development in-browser exception rendering is turned on. If we have integration specs where with an expectation like:
```
expect(page).to have_content("foo")
```
the `have_content` line could appear in the stacktrace meaning this expectation will pass.

This hid the errors seen in #6006 when `spec/integration/classify_request_spec.rb` was tested before some (all?) controller spec. Somehow if controller spec was run first it resulted in the exceptions not being shown.

We should consider not setting `show_exceptions = true`.

 [1] https://github.com/mysociety/alaveteli/blob/8959f3f/spec/spec_helper.rb#L116-L122

gbp:
## Relevant issue(s)

Fixes #6018

## What does this do?

Remove before test suite callback 

## Why was this needed?

Don't set show exceptions for integration tests. This can hide failures
by returning the in-browser development exception viewer and a 200
response instead of 500.

## Implementation notes

The lines removed were added as part of https://github.com/mysociety/alaveteli/commit/071b5aab0fc1b05b90ac60bc1f259c6b409f7120 the rest of the "consider all requests local" work added at the same time was rewritten recently in #5621 so I believe this callback can now be removed.

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -113,14 +113,6 @@ RSpec.configure do |config|
     ActionMailer::Base.deliveries = []
   end
 
-  # For integration tests, make sure the app renders exceptions rather
-  # than passing them to the test itself.
-  config.before(:each) do |example|
-    if [:request].include? example.metadata[:type]
-      Rails.application.config.action_dispatch.show_exceptions = true
-    end
-  end
-
   config.before(:suite) do
     if ENV['ALAVETELI_USE_OINK']
       oink_log = Rails.root + 'log/oink.log'


gbp:
This requires #6021 in order for specs to pass.

garethrees (approved):

PR #6021, opened 2020-12-11
[#6006] fix broken help/unhappy page when not using a theme

### LINKED ISSUE #6006: Transient spec failure in RefusalAdvice.default
Seed 36145

```
Failures:

  1) classifying a request marking request as rejected displays a thank you message post redirect
     Failure/Error: files = Rails.configuration.paths['config/refusal_advice'].existent

     NoMethodError:
       undefined method `existent' for nil:NilClass
     # ./app/models/refusal_advice.rb:6:in `default'
     # ./app/controllers/help_controller.rb:29:in `unhappy'
     # ./app/controllers/application_controller.rb:118:in `record_memory'
     # ./lib/strip_empty_sessions.rb:15:in `call'
     # ./spec/integration/alaveteli_dsl.rb:93:in `classify_request'
     # ./spec/integration/classify_request_spec.rb:176:in `block (4 levels) in <top (required)>'
     # ./spec/integration/classify_request_spec.rb:175:in `block (3 levels) in <top (required)>'
```

gbp:
## Relevant issue(s)

Fixes #6006

## What does this do?

Fix broken help/unhappy page when not using a theme

## Why was this needed?

Before attempting to retrieve refusal advice we need to check its been
configured. If for example we're not using theme then the paths aren't
set up and will return `nil`.

This is the case when running CI by default.

Also allows `RefusalAdvice.default` to be called with no argument
instead of requiring `nil` if there is no info request.

### DIFF
--- a/app/models/refusal_advice.rb
+++ b/app/models/refusal_advice.rb
@@ -2,7 +2,7 @@
 # A collection of Questions that help users challenge refusals.
 #
 class RefusalAdvice
-  def self.default(info_request)
+  def self.default(info_request = nil)
     files = Rails.configuration.paths['config/refusal_advice'].existent
     new(Store.from_yaml(files), info_request: info_request)
   end
--- a/config/initializers/theme_loader.rb
+++ b/config/initializers/theme_loader.rb
@@ -18,12 +18,12 @@ def require_theme(theme_name)
 
   require theme_main_include
 
-  Rails.configuration.paths.add(
-    'config/refusal_advice',
-     with: root.join('config/refusal_advice'),
-     glob: '*.yml'
+  Rails.configuration.paths['config/refusal_advice'].push(
+    root.join('config/refusal_advice')
   )
 end
+
+Rails.configuration.paths.add('config/refusal_advice', glob: '*.yml')
 
 if Rails.env == "test"
   # By setting this ALAVETELI_TEST_THEME to a theme name, theme tests can run in the Rails
--- a/spec/models/refusal_advice_spec.rb
+++ b/spec/models/refusal_advice_spec.rb
@@ -9,12 +9,14 @@ RSpec.describe RefusalAdvice do
   describe '.default' do
     subject { described_class.default(info_request) }
 
+    let(:path) { Rails.root.join('spec/fixtures/refusal_advice') }
+
     before do
-      Rails.configuration.paths.add(
-        'config/refusal_advice',
-         with: Rails.root.join('spec/fixtures/refusal_advice'),
-         glob: '*.yml'
-      )
+      Rails.configuration.paths['config/refusal_advice'].push(path)
+    end
+
+    after do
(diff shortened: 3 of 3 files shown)


app/models/refusal_advice.rb
```diff
@@ -2,8 +2,11 @@
 # A collection of Questions that help users challenge refusals.
 #
 class RefusalAdvice
-  def self.default(info_request)
-    files = Rails.configuration.paths['config/refusal_advice'].existent
+  def self.default(info_request = nil)
+    paths = Rails.configuration.paths['config/refusal_advice']
+    return [] unless paths
+
+    files = paths.existent
```
garethrees, line 9: Probably don't need local files variable in this case?

```ruby
new(Store.from_yaml(paths.existent), info_request: info_request)
```

gbp:
This is still and issue with this, EG: `rspec --seed 46141 spec/models/refusal_advice_spec.rb`

Looks liek we need to revert the `Rails.configuration.paths.add` call after the relevent specs have run.

config/initializers/theme_loader.rb
```diff
@@ -18,13 +18,13 @@ def require_theme(theme_name)
 
   require theme_main_include
 
-  Rails.configuration.paths.add(
-    'config/refusal_advice',
-     with: root.join('config/refusal_advice'),
-     glob: '*.yml'
+  Rails.configuration.paths['config/refusal_advice'].push(
+    root.join('config/refusal_advice')
```
garethrees, line 22: I'm assuming the glob added on L26 also applies to the theme path because they're the same "key" (`'config/refusal_advice'`)?

garethrees (commented):
> This is still and issue with this, EG: `rspec --seed 46141 spec/models/refusal_advice_spec.rb`

Couldn't reproduce locally on `develop` @ c70a0d186.

Haven't run, but makes sense in principle.

Looks liek we need to revert the `Rails.configuration.paths.add` call after the relevent specs have run.

gbp:
> Couldn't reproduce locally on `develop` @ [c70a0d1](https://github.com/mysociety/alaveteli/commit/c70a0d186601b0f54b362dec9573c270bafcdd2b).

Error on  d088a77 which was the original commit attempting to fix this, but has now been force pushed over.

config/initializers/theme_loader.rb
gbp: Yeah as I understand it, line 26 is setting up a configuration path which can exist in core but doesn't need to - and we don't have this path. This path can also have alternative paths which is what we add when loading the theme or in specs.

Rails doesn't have a way of doing `Rails.configuration.paths.remove` but it can add/remove these alternative paths which is what is done in the spec after callback hook using `unshift`.

garethrees: Cool makes sense 👍

PR #5971, opened 2020-11-13
[#5909] JRCT – Show exemptions from incoming_message.refusals

gbp:
## Relevant issue(s)

https://github.com/mysociety/alaveteli/issues/5909

## What does this do?

If an `incoming_message` has any `refusals`, they’re shown in a panel at the top of the correspondence body, along with a link to the `help_unhappy_path`.

## Screenshots

Here it is running in the WDTK theme.

Screenshot 2020-12-07 at 11 31 05

## Notes to reviewer

@garethrees @gbp note this link _doesn’t_ currently pass any information about the request, or the detected exemptions. So I’m guessing we might want to add that?

The message has a pale grey background by default in the core theme. We should probably override that to something a bit more colourful in the WDTK theme – I guess that should be done in a pull request on whatdotheyknow-theme?

PR #6022, opened 2020-12-11
[#6020] render unique references for refusals

### LINKED ISSUE #6020: IncomingMessage#refusals should only return unique refusals
Screenshot 2020-12-11 at 09 26 46

gbp:
## Relevant issue(s)

Fixes #6020

## What does this do?

- Rename legislation reference main element
- Add LegislationReference equality method
- Add LegislationReference#parent method
- Update IncomingMessage#refusals

## Why was this needed?

Request refusal partial from #5971 was rendering the same reference more than
once these changes ensures we show unique parent reference.

A parent reference being a reference with all sub-sections removed. In this case
we can ignore sub-sections as the refusal wizard isn't using these.

### DIFF
--- a/app/models/legislation/reference.rb
+++ b/app/models/legislation/reference.rb
@@ -32,6 +32,10 @@ class Legislation
       parent_reference + "(#{sub_elements.join(')(')})"
     end
 
+    def parent
+      self.class.new(legislation: legislation, reference: parent_reference)
+    end
+
     def cover?(other)
       legislation == other.legislation && type == other.type &&
         elements == other.elements[0...elements.count]
--- a/app/models/legislation/reference.rb
+++ b/app/models/legislation/reference.rb
@@ -41,6 +41,11 @@ class Legislation
       legislation.refusals.any? { |reference| reference.cover?(self) }
     end
 
+    def ==(other)
+      legislation == other.legislation && type == other.type &&
+        elements == other.elements
+    end
+
     private
 
     def parent_reference
--- a/app/models/legislation/reference.rb
+++ b/app/models/legislation/reference.rb
@@ -28,9 +28,8 @@ class Legislation
     end
 
     def to_s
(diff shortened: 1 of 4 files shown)


spec/models/legislation/reference_spec.rb
```diff
@@ -104,4 +104,41 @@ def reference_with_type(type)
       it { is_expected.to eq false }
     end
   end
+
+  describe '#==' do
+    let(:reference) do
+      Legislation::Reference.new(legislation: foi, reference: 'Section 12(1)')
+    end
+
+    subject { reference.==(other) }
```
garethrees, line 139: Why `.==` here?

spec/models/legislation/reference_spec.rb
```diff
@@ -104,4 +104,41 @@ def reference_with_type(type)
...
+    context 'other reference has different elements' do
+      let(:other) do
+        Legislation::Reference.new(legislation: foi, reference: 'Section 12(2)')
+      end
+      it { is_expected.to eq false }
+    end
+
+    context 'other reference has different type' do
+      let(:other) do
+        Legislation::Reference.new(legislation: foi, reference: 'Article 12(2)')
```
garethrees, line 158: This has a different "elements" as well if I understand correctly. I _think_ this should be:

```diff
- Article 12(2)
+ Article 12(1)
```

app/models/legislation/reference.rb
```diff
@@ -32,6 +32,10 @@ def to_s
       parent_reference + "(#{sub_elements.join(')(')})"
     end
 
+    def parent
+      Reference.new(legislation: legislation, reference: parent_reference)
```
garethrees, line 36: `self.class.new` to be more resilient to renames? No big deal either way though

garethrees (approved):
LGTM – couple of minor questions/suggestions but no need to re-review 👍

PR #6035, opened 2020-12-21
Replace rubyracer

gbp:
## Relevant issue(s)

See https://github.com/rails/rails/pull/29285

## What does this do?

Use mini_racer which uses a up to date version of libv8 which has
security fixes and recommended upstream in Rails for new applications
which don't have pre-existing JavaScript runtimes in production.

## Why was this needed?

Can't get older libv8 gems installed in GitHub action workflows. I think this might help.


gbp:
Running `rails assets:precompile RAILS_ENV=production` locally also works fine

garethrees (approved):

PR #6037, opened 2020-12-22
Extract converage setup

gbp:
Extract coverage setup. Did this while moving to GitHub actions where we need to change this setup a bit to get it working.

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -5,7 +5,7 @@ require 'stripe_mock_patch'
 require "alaveteli_features/spec_helpers"
 require 'strip_attributes/matchers'
 
-require_relative 'spec_coverage'
+require_relative 'spec_coverage' if ENV['COVERAGE'] == 'local' || ENV['CI']
 
 # This file is copied to spec/ when you run 'rails generate rspec:install'
 ENV["RAILS_ENV"] ||= 'test'
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -1,29 +1,11 @@
 # -*- encoding : utf-8 -*-
 require 'rubygems'
-require 'simplecov'
-require 'coveralls'
 require 'webmock/rspec'
 require 'stripe_mock_patch'
 require "alaveteli_features/spec_helpers"
 require 'strip_attributes/matchers'
 
-cov_formats = [Coveralls::SimpleCov::Formatter]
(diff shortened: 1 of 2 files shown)


spec/spec_helper.rb
```diff
@@ -5,7 +5,7 @@
 require "alaveteli_features/spec_helpers"
 require 'strip_attributes/matchers'
 
-require_relative 'spec_coverage'
+require_relative 'spec_coverage' if ENV['COVERAGE'] == 'local' || ENV['CI']
```
garethrees, line 8: I think:

```diff
- Don't load simplecov if when not checking coverage
+ Don't load simplecov when not checking coverage
```

I also think the commit message should explain _why_ this change was necessary, as I don't really understand the problem based on the change alone. Something to do with GH actions I know; I thought it was that we didn't want to run coverage when running specs (needed separate step), but then `|| ENV['CI']` makes it look like we _do_ want to load this file when running specs on CI?

spec/spec_helper.rb
```diff
@@ -1,29 +1,11 @@
...
-SimpleCov.start('rails') do
-  add_filter  'commonlib'
-  add_filter  'vendor/plugins'
-  add_filter  'lib/attachment_to_html'
-  add_filter  'lib/has_tag_string'
-  add_filter  'lib/acts_as_xapian'
-  add_filter  'lib/themes'
-  add_filter  '.bundle'
-end
+require_relative 'spec_coverage'
```
garethrees: Would probably make this `spec/support/spec_coverage`?

garethrees (commented):
Changes look fine in principle but I think need further explanation.

spec/spec_helper.rb
gbp: I did initially the problem with that is we auto require all files in `spec/support`

spec/spec_helper.rb
gbp: Update the commit message

gbp:
Neither of these commits are nessecary for switching to Github actions. The actual coverage changes are currently at https://github.com/mysociety/alaveteli/pull/6038/commits/b97b0dc807a20e2a9d69f6b66954960fa0ef4b93.

This PR does reduces complexity in the spec helper file which I've alwasy felt needed to be improved.

garethrees (approved):
> Neither of these commits are nessecary for switching to Github actions. The actual coverage changes are currently at b97b0dc.
>
> This PR does reduces complexity in the spec helper file which I've alwasy felt needed to be improved.

Ah gotcha! Yep, lgtm!

PR #6036, opened 2020-12-22
Update README shield badges

gbp:
This makes the shield badges look consitent. Noticed this when moving to GitHub actions

Screenshot 2020-12-22 at 17 10 24

Issue #5841, opened 2020-07-29
ImportError: No module named yaml


gbp:
Some of our scripts rely on `read_conf` shell function which, I think, doesn't run out of the box on macOS Catalina, EG:

```shell
$ bash -c 'source commonlib/shlib/deployfns; read_conf config/general; echo $OPTION_SITE_NAME'
Traceback (most recent call last):
  File "/Users/gbp/Documents/mySociety/transparency/alaveteli/commonlib/bin/yaml2sh", line 9, in <module>
    import optparse, sys, re, yaml
ImportError: No module named yaml
: shlib/deployfns read_conf: error loading config file config/general.yml
```

System python (installed at `/usr/bin/python` which is what is expected https://github.com/mysociety/commonlib/blob/master/bin/yaml2sh#L1) is still 2.7.16, but doesn't come with `pyyaml` installed by default.

To get this working first I had to switch python versions - I use [pyenv](https://github.com/pyenv/pyenv) to manage my python versions, so I ran:
1. `pyenv local system`
2. `sudo /usr/bin/easy_install pip`
3. `pip install pyyaml==3.13`

The [latest version of `pyyaml`](https://pypi.org/project/PyYAML/5.3.1/) will output an deprecation warning:

```shell
$ bash -c 'source commonlib/shlib/deployfns; read_conf config/general; echo $OPTION_SITE_NAME'
/Users/gbp/Documents/mySociety/transparency/alaveteli/commonlib/bin/yaml2sh:29: YAMLLoadWarning: calling yaml.load() without Loader=... is deprecated, as the default Loader is unsafe. Please read https://msg.pyyaml.org/load for full details.
  data = yaml.load(fh)
Alaveteli
```

This is all far from ideal.

We should:
- [x] change shebang in `yaml2sh` to `#!/usr/bin/env python2` so we're not reliant on system Python

We should consider, and test on supported distributions, if we can:
- [x] support the newer version of `pyyaml`
- [x] allowing Python 3 - it will error currently

gbp:
Fixed due to https://github.com/mysociety/commonlib/commit/e6d18158940242bc682dc765602a969b7d14af88

PR #6038, opened 2020-12-22
Migrate to GitHub actions

gbp:
## Relevant issue(s)

Depends on #6035 
Depends on #6036
Depends on #6037 
Depends on https://github.com/mysociety/commonlib/pull/71 

## What does this do?

Migrates from Travis to GitHub actions CI

## Why was this needed?

Travis doesn't like open source anymore.

## Implementation notes

1. This uses the Ubuntu 20.04 base image due to needing to install `pdftk` I couldn't workout how to add the focial package repo - to install like we currenlty do on Travis.
2. This users the FMS postgres image which has the `en_GB` locale already installed
3. ~~Needs to install Python 2.7 too :-1: with an old version of `pyyaml` so the `read_conf` command can run. See #5841~~
4. Opted to run RSpec with the [Fivemat](https://github.com/tpope/fivemat) formatter, which is a happy medium or the progress and documentation formatters, due to the GitHub action web UI only updating when a linefeed to outputted. This allows us to monitor the spec run in real time.


gbp:
One downside of GitHub action is it doesn't yet support re-running individual builds - its all or nothing. I'm happy with this as I think it'll force us to once and for all resolve our transient spec errors

garethrees:
Build is failed because of Ruby 2.7 – does GH actions support an equivalent of `allow_failures`?

gbp:
> Build is failed because of Ruby 2.7 – does GH actions support an equivalent of `allow_failures`?

No they don't. See: https://github.com/actions/toolkit/issues/399, I've just setup marking specific checks as required in the repo settings but that still results in the red cross.

I wonder if we could add a step like:
```yaml
    - name: Ignore experimental faliures
      if: faliure() && ${{ matrix.experimental || false }}
      run: |
        exit 0
```

I will test...

gbp:
Oh we should also make a Ruby 2.7 gemfile which includes a newer version (a yet to be built version) of `xapian-full-alaveteli` so we actually get pass `bundle install` step.

gbp:
No - doesn't seem to work and looking around it doesn't seem like its possible. @garethrees if you agree I'll drop 2.7 from CI until we can add a gemfile for it and a newer version of `xapian-full-alaveteli`?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,92 @@
...
+
+    - name: Setup database
+      run: |
+        psql postgres://postgres:postgres@localhost:5432 <<-EOSQL
+          CREATE DATABASE template_utf8 TEMPLATE template0 ENCODING "UTF-8";
+          UPDATE pg_database SET datistemplate=true WHERE datname='template_utf8';
+          CREATE DATABASE alaveteli_test TEMPLATE template_utf8;
+        EOSQL
+
+    - name: Install theme
```
garethrees, line 81: Should this be handled by `RAILS_ENV=test ./script/rails-post-deploy` as per travis? Assuming that causes some issues?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,92 @@
...
+        bundle config path vendor/bundle
+        bundle config build.libv8 --with-system-v8
+        bundle install --jobs $(nproc) --retry 3
+
+    - name: Setup database
+      run: |
+        psql postgres://postgres:postgres@localhost:5432 <<-EOSQL
+          CREATE DATABASE template_utf8 TEMPLATE template0 ENCODING "UTF-8";
+          UPDATE pg_database SET datistemplate=true WHERE datname='template_utf8';
+          CREATE DATABASE alaveteli_test TEMPLATE template_utf8;
```
garethrees, line 78: Does the FMS image also include the `en_GB` collation?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,92 @@
...
+        EOSQL
+
+    - name: Install theme
+      run: |
+        cp config/general.yml-example config/general.yml
+        bundle exec rake themes:install
+
+    - name: Migrate database
+      run: |
+        bundle exec rails db:migrate
```
garethrees, line 88: OOI, can we `db schema:load` here?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,111 @@
...
+
+    - name: Cache RubyGems
+      uses: actions/cache@v2
+      with:
+        path: vendor/bundle
+        key: ${{ runner.os }}-${{ matrix.ruby }}-gems-unit-${{ hashFiles(format('{0}.lock', matrix.gemfile || 'Gemfile')) }}
+        restore-keys: |
+          ${{ runner.os }}-${{ matrix.ruby }}-gems-unit-
+
+    - name: Install RubyGems
```
garethrees, line 67: Should this be handled by `RAILS_ENV=test ./script/rails-post-deploy` as per travis? Assuming that causes some issues?

garethrees (commented):
LGTM in general.

Left a couple of comments but not sure if any action needs taking.

> @garethrees if you agree I'll drop 2.7 from CI until we can add a gemfile for it and a newer version of `xapian-full-alaveteli`?

👍

.github/workflows/ci.yml
gbp: Initially I had the `ruby/setup-ruby` action install the gems which it can do along with saving and restoring the bundle cache - all as one step instead of three. I had to split this out to run `bundle config build.libv8 --with-system-v8`.

I guess we could use the `rails-post-deploy` script but I think I prefer having these as seporate workflow steps.

.github/workflows/ci.yml
gbp: Yes. See: https://github.com/mysociety/public-builds/blob/master/docker/postgres/Dockerfile

.github/workflows/ci.yml
gbp: As above

.github/workflows/ci.yml
gbp: No - we still don't `db/schema.rb` file

garethrees: Oh, ofc 🤦

.github/workflows/ci.yml
garethrees: I guess there's more chance of us missing things we add to `rails-post-deploy` in future by not doing so – e.g. `db:seed` is called in `rails-deploy-while-down`. I don't have a really strong preference right now though.

gbp: Well we don't need the test DB seeded but I see your point if new actions are added in the future. Personally I think these deployment script are confusing and the GH action steps as they are simplier and easier to follow.

Lets give it a go as it is for now.

Issue #2906, opened 2015-11-19
Remove commonlib submodule


crowbot:
The submodule adds another level of complexity - is this worth it?

crowbot:
As @davewhiteland points out, we could retain the shared code in the form of a gem.

garethrees:
We add in some rake tasks in the test environment https://github.com/mysociety/alaveteli/blob/e0fc3d2d6348e16b34471cdb7d8fad11e1b8b3cc/Rakefile#L8

garethrees:
https://github.com/mysociety/alaveteli/blob/e0fc3d2d6348e16b34471cdb7d8fad11e1b8b3cc/lib/configuration.rb#L2

garethrees:
> These files must be executable Ruby files. Files that use bash or other interpreters will not work.
>
> – http://guides.rubygems.org/specification-reference/#executables

garethrees:
Probably don't need to publish to rubygems – just use the `git` option:

```ruby
gem 'mysociety-commonlib', :git => 'https://github.com/mysociety/commonlib.git', :branch => 'master'
```

Need to check when this gets updated; may only be on `gem update`.

garethrees:
## `Rakefile`

Loads ruby specs

https://github.com/mysociety/alaveteli/blob/develop/Rakefile#L7-L9

## `Vagrantfile`

`install-site.sh` downloaded directly

## `config/crontab-example`

Uses `run-with-lockfile.sh` several times

https://github.com/mysociety/alaveteli/blob/develop/config/crontab-example#L15

## `config/httpd.conf-example`

Makes `jslib` available. Don't know if we use it?

https://github.com/mysociety/alaveteli/blob/develop/config/httpd.conf-example#L88-L93

## `config/initializers/alaveteli.rb`

Makes rblib available to the app

https://github.com/mysociety/alaveteli/blob/develop/config/initializers/alaveteli.rb#L2-L5

## `lib/configuration.rb`

Loads `MySociety::Config` – a config parser that shells out to PHP?

- https://github.com/mysociety/alaveteli/blob/develop/lib/configuration.rb#L2-L10
- https://github.com/mysociety/commonlib/blob/master/rblib/config.rb

## `lib/tasks/submodules.rake`

Not a direct use of commonlib but checks whether it needs initializing/updating

https://github.com/mysociety/alaveteli/blob/develop/lib/tasks/submodules.rake

## `lib/tasks/themes.rake

[...]

PR #6053, opened 2021-01-06
Remove dependency on Python

gbp:
Instead of using yaml2sh from commonlib this commits adds a new Ruby
script at bin/config which can retrieve individual configuration
variables from config/general.yml.

To get a configuration variable run:
  bin/config SITE_NAME

This bin command doesn't initialise the Rails stack so is fast

It also *doesn't* use lib/configuration.rb Ruby module, although it
could if some configuration variables had defaults assigned. These are:
1. BUNDLE_PATH
2. SHARED_DIRECTORIES
3. SHARED_FILES
4. SHARED_FILES_PATH
5. STAGING_SITE

As I understand it these haven't be given a default because they are
only used from the installation shell scripts. Which seems reasonable to
me.

### DIFF
--- a/bin/config
+++ b/bin/config
@@ -1,0 +1,9 @@
+#!/usr/bin/env ruby
+
+require 'yaml'
+
+settings = YAML.load_file(
+  File.join(File.dirname(__FILE__), '..', 'config', 'general.yml')
+)
+
+puts settings.fetch(ARGV[0].upcase)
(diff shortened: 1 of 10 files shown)


gbp:
This is not strictly needed once https://github.com/mysociety/commonlib/pull/71 is merged

gbp:
This would be a step towards #2906

gbp:
Have double checked to confirm (through building vagrant boxes for #6615) that the install scripts are still working.

PR #6054, opened 2021-01-06
Fix 'view as HTML' for non-cacheable attachments

gbp:
## Relevant issue(s)

This reverts part of #6049

## What does this do?

Specs broken and view as HTML for "backpaged" attachments were returning 404s

## Why was this needed?

We only need to check if a message is public to allow an attachment to
be viewed in browser.

Attachments are viewed in browser using the Google Docs viewer and these
need to be publicly accessible. We raise a 404 when the messages are
private.

We don't need to raise a 404 when the messages are not cacheable. Not
caching attachments is to ensure X-Robots-Tag header is present and
shouldn't have an impact on if an attachment can be viewed in browser.


garethrees (approved):
LGTM

PR #6072, opened 2021-01-19
 Legislation tweaks

garethrees:
## Relevant issue(s)

Fixes https://github.com/mysociety/alaveteli/issues/6020

## What does this do?

Memoise Legislation.all
Add Legislation#==

## Why was this needed?

Addresses review feedback from configuring refusals [1] and fixes an equality error [2].

[1] https://github.com/mysociety/alaveteli/pull/6015#pullrequestreview-569293283
[2] https://github.com/mysociety/alaveteli/issues/6020

## Implementation notes

Would ideally like a more real life spec for the failure in https://github.com/mysociety/alaveteli/issues/6020, but not sure we have a good place for it, and it's pretty edge case so not a huge concern?

### DIFF
--- a/app/models/legislation.rb
+++ b/app/models/legislation.rb
@@ -71,6 +71,10 @@ class Legislation
     )
   end
 
+  def ==(other)
+    other&.to_sym == to_sym
+  end
+
   def find_references(text)
     Legislation::ReferenceCollection.new(legislation: self).match(text)
   end
--- a/app/models/legislation.rb
+++ b/app/models/legislation.rb
@@ -6,7 +6,7 @@ class Legislation
   UnknownLegislationVariant = Class.new(StandardError)
 
   def self.all
-    [
+    @all ||= [
       new(
         key: 'foi',
         short: _('FOI'),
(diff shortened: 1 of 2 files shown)


app/models/legislation.rb
```diff
@@ -71,6 +71,10 @@ def to_s(variant = :short)
     )
   end
 
+  def ==(other)
+    other.to_sym == to_sym
```
garethrees, line 75: Currently causes failures with other specs – all trying to call `#to_sym` on a `RSpec::Matchers::BuiltIn::HaveAttributes`. Dunno if this isn rspec bug?!

```
  14) Legislation.for_public_body public body tagged as eir_only does not contains FOI legislation
      Failure/Error: other.to_sym == to_sym

      NoMethodError:
        undefined method `to_sym' for #<RSpec::Matchers::BuiltIn::HaveAttributes:0x000055f97c029da8>
        Did you mean?  to_s
      # ./app/models/legislation.rb:75:in `=='
      # ./spec/models/legislation_spec.rb:93:in `block (4 levels) in <top (required)>'
```

gbp (approved):
Nice, this looks good.

I think the specs to change the `have_attributes` matchers could be move the the first commit and still pass?

garethrees:
> Nice, this looks good.
>
> I think the specs to change the `have_attributes` matchers could be move the the first commit and still pass?

Cheers! It's the addition of `Legislation#==` that causes the failure, so makes more sense for the spec changes to align with that?

PR #6074, opened 2021-01-22
[#5911][jrct] Snippet Library WIP

garethrees:
https://github.com/mysociety/alaveteli/issues/5911

Creating a draft PR for easier visibility and WIP discussion.

Will create a new one when this is ready for merge


app/controllers/refusal_advice_controller.rb
```diff
@@ -1,10 +1,47 @@
 class RefusalAdviceController < ApplicationController
   def create
-    @params = parsed_refusal_advice_params # TODO: move into log event method
```
garethrees: The approach of logging a separate event and passing through a separate controller feels like it's adding quite a bit of complexity. I don't have fully-formed thoughts on this yet though.

There are two things rattling around in my head.

1. We already log `followup_sent`, which is the main "event" that's happening. I can appreciate the technical reasons for why a `refusal_advice_wizard` has been added, but it doesn't feel like we're logging something that's particularly actionable or driven by user intent. IDK, I could be convinced otherwise I guess. My thinking though, was that we'd temporarily store some data from the refusal advice form and then record that in the `params_yaml` of the subsequent `followup_sent` event. 
2. Maybe more in my mind as feeling weird is that some installs won't configure refusal advice, so they'll need to be concerned about where to send the user from their implementation of `/help/unhappy`. Again, I'm not totally against what we have here – this is just thinking out loud. It feels like this couples followups with refusal advice more than it should. We do want to be able to show snippets based on refusal advice though, and this makes it easier for reusers to configure in that they don't have to worry about matching up the `response_template` key with tags in the database?

I'm wondering whether there's a middle ground where we can reduce the complexity of this action, but still with the gains of reduced configuration.

app/views/followups/_snippets.html.erb
```diff
@@ -7,20 +7,19 @@
   <p><strong>Challenging a refusal?</strong> Here are some snippets that may help:</p>
 <% end %>
 
-<div class="snippet-library">
+<div class="snippet-library js-snippet-library">
```
garethrees, line 10: Not quite sure where the best place for this comment is, but looks like we'll need to handle "no snippets for this section" (or, just don't show the section at all)

Screenshot 2021-01-22 at 09 39 16

db/seeds.rb
```diff
@@ -37,3 +37,77 @@
     AlaveteliPro::RequestSummaryCategory.create(slug: category_slug)
   end
 end
+
+OutgoingMessageSnippet.create(
```
garethrees, line 41: Be good to add ~50+ snippets – even if they're lorem ipsum – just so we can see how it feels with several snippets appearing under a given section.

garethrees (commented):
Looking great so far! I appreciate this is very WIP, but added some thoughts as I was giving it a prod.

Attaching screenshot as it is so far:

Screenshot 2021-01-22 at 09 39 51

app/controllers/followups_controller.rb
```diff
@@ -59,6 +60,12 @@ def set_internal_review
     end
   end
 
+  def set_refusal_advice_snippets
+    action = @internal_review ? 'internal_review' : 'clarification'
```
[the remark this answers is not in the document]
gbp: Yeah I get where you are coming from although for the refusal advice actions are grouped under `clarification`. Now this isn't ideal due to this being configurable in theme to be something else completely so I agree wo don;t want to hard code this here.

So with this change, suppose the requester completes the wizard and clicks a `clarification` action we would show both the `clarification` and `internal_review` snippets to them. Is that an issue?

We should also change the case statement in `RefusalAdviceController#create` then too.

app/controllers/refusal_advice_controller.rb
gbp: I was kinda viewing this as an event which happend to the info request. We talked about recording usage/results of the wizard which this does and allows us to anaylise usage at a later date.  

I don't think its too complex, all `parsed_refusal_advice_params` is doing is turning `"true"`/`"false"` into booleans to be stored in the database. Would creating a seporate refusal advice answers model be better - It could do this parsing for us then?

gbp: For now I've continued with creating `InfoRequestEvent` but am open to changing now or in the future.

app/controllers/followups_controller.rb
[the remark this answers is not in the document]
garethrees: > So with this change, suppose the requester completes the wizard and clicks a `clarification` action…

I'd say in that case we show any snippets tagged `refusal_advice` that _aren't_ tagged `internal_review`?

app/controllers/refusal_advice_controller.rb
garethrees: > I was kinda viewing this as an event which happend to the info request. We talked about recording usage/results of the wizard which this does and allows us to anaylise usage at a later date.

Yeah, I'm not against per-say, just made me pause for thought.

> I don't think its too complex, all `parsed_refusal_advice_params` is doing is turning `"true"`/`"false"` into booleans to be stored in the database. Would creating a seporate refusal advice answers model be better - It could do this parsing for us then?

A PORO like `RefusalAdvice::WizardAnswers.new(params).to_hash` might be nice, but not essential.

garethrees: > 2\. Maybe more in my mind as feeling weird is that some installs won't configure refusal advice, so they'll need to be concerned about where to send the user from their implementation of `/help/unhappy`. Again, I'm not totally against what we have here – this is just thinking out loud. It feels like this couples followups with refusal advice more than it should. We do want to be able to show snippets based on refusal advice though, and this makes it easier for reusers to configure in that they don't have to worry about matching up the `response_template` key with tags in the database?

This is maybe my main concern here. How will this work?

spec/controllers/refusal_advice_controller_spec.rb
```diff
@@ -2,9 +2,65 @@
...
+      }
+    end
+
+    context 'valid params' do
+      it 'returns a success' do
+        post :create, params: params
+        expect(response).to be_successful
+      end
+
+      it 'logs event with correctly parsed params' do
```
gbp: `it 'parses params correctly'`

app/views/followups/_snippets.html.erb
```diff
@@ -0,0 +1,24 @@
...
+
+<div class="snippet-library js-snippet-library">
+  <div class="snippet-library__filters">
+    <!-- TODO: Non-JavaScript fallback? -->
+    <label for="snippet-relation">Snippets related to</label>
+    <select id="snippet-relation">
+      <% if @refusal_advice.answers %>
+        <option value="<%= @refusal_advice.answers.join(' ') %>">my previous answers</option>
+      <% end %>
+      <option value="all">any section</option>
```
garethrees, line 17: `s/any section/any reason/` ?

PR #6082, opened 2021-01-28
[#5909][JRCT] Show identified exemptions

garethrees:
## Relevant issue(s)

Fixes https://github.com/mysociety/alaveteli/issues/5909.

Requires https://github.com/mysociety/alaveteli/pull/6080 to fix the duplicates.

Ideally I'd like to have a way of getting a bit more explanation of these refusal reasons – certainly on the public requests – rather than just their names.

## What does this do?

Show identified refusals on incoming messages.

## Why was this needed?

Help speed up classification and provide a way of linking through to refusal advice.

## Implementation notes

Feels like these take up quite a lot of the space with the heading, so maybe make it a little more concise without losing too much (197b3bc)?

Screenshot 2021-01-28 at 16 58 49

### Public Request

When the request needs classification link the user down to the classification form.

Screenshot 2021-01-28 at 17 04 46

When the request has been classified in a state where refusals are identified, point the user to `/help/unhappy` to challenge it.

Screenshot 2021-01-28 at 17 04 00

Just show that there might be refusals if the request isn't in an "unhappy" state.

Screenshot 2021-01-28 at 17 03 25

### Private Request

When a request needs classification don't show any CTA since there's no page element to take them to.

Screenshot 2021-01-28 at 16 56 31

When the request has been classified in a state where refusals are identified, point the user to `/help/unhappy` to challenge it.

Screenshot 2021-01-28 at 16 54 39

Just show that there might be refusals if the request isn't in an "unhappy" state.

Screenshot 2021-01-28 at 16 54 17

## Notes to reviewer

I kept going back and forth between showing the CTA when the request isn't in an "unhappy" state. I _think_ this is best, because it feels weird to be pushing the user to make a challenge when they've told us it's successful (or "responded by post" or whatever). Feels like we should only push it when they've said its refused in some way.

### DIFF
--- a/app/views/request/_incoming_refusals.html.erb
+++ b/app/views/request/_incoming_refusals.html.erb
@@ -1,0 +1,20 @@
+<% if can?(:update_request_state, @info_request) && incoming_message.refusals? %>
+  <div class="correspondence__suggestion">
+    <p>
+      <%= _('{{authority_name}} may have refused all or part of your request ' \
+            'under <strong>{{exemptions}}</strong>.',
+            authority_name: @info_request.public_body.name,
+            exemptions: incoming_message.refusals.to_sentence) %>
+
+      <% if @info_request.awaiting_description? && @show_bottom_describe_state_form %>
+        <%= link_to '#describe_state_form_2' do %>
+          <%= _('Let us know and we’ll help you challenge it.') %>
+        <% end %>
+      <% elsif @info_request.reason_to_be_unhappy? %>
+        <%= link_to help_unhappy_path(@info_request.url_title) do %>
+          <%= _('Get help to challenge it.') %>
+        <% end %>
+      <% end %>
+    </p>
+  </div>
+<% end %>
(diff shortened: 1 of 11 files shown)


app/views/request/_incoming_refusals.html.erb
```diff
@@ -0,0 +1,20 @@
...
+            'under <strong>{{exemptions}}</strong>.',
+            authority_name: @info_request.public_body.name,
+            exemptions: incoming_message.refusals.to_sentence) %>
+
+      <% if @info_request.awaiting_description? && @show_bottom_describe_state_form %>
+        <%= link_to '#describe_state_form_2' do %>
+          <%= _('Let us know and we’ll help you challenge it.') %>
+        <% end %>
+      <% elsif @info_request.reason_to_be_unhappy? %>
+        <%= link_to help_unhappy_path do %>
```
gbp, line 14: This need to have the request `url_title` param

gbp (approved):

PR #6086, opened 2021-01-29
[#5911][JRCT] Store wizard answers

### LINKED ISSUE #6103: Linking wizard actions to followup form with snippets
- [x] Allow next steps actions to be shown on click by user
- [x] Allow actions to be rendered even if no suggestions defined in yaml
- [x] Give `RefusalAdvice::Action` a "body" method (`question.label`, `action.body`, `suggestion.advice`)
- [ ] Give `RefusalAdvice::Action` a "target" method 
- [x] `RefusalAdvice::Action#header` should return  `#title` by default if not configured (https://github.com/mysociety/alaveteli/issues/6105)
- [x] Allow an empty set of questions (`Array(data[legislation.to_sym][:questions])`) (https://github.com/mysociety/alaveteli/pull/6104)
- [x] Render actions within next steps?
- [ ] Update controller to handle external target and new request internal actions
- [ ] Rename clarification to followup in `RefusalAdviceController` case

```ruby
RefusalAdvice::Action.find(id).target
target: { internal: 'followup' }
target: { help_page: 'ico' }
target: { external: 'http://...' }
```

gbp:
## Relevant issue(s)

Connected to #5911 
Fixes #6103

## What does this do?

Stores the answers, and suggested actions from the refusal advice wizard. This gets stored as an `InfoRequestEvent` record.

## Why was this needed?

So we can filter the snippets on the new follow up form.

### DIFF
--- a/app/controllers/refusal_advice_controller.rb
+++ b/app/controllers/refusal_advice_controller.rb
@@ -1,0 +1,8 @@
+##
+# Controller to store advice, from the refusal advice wizard, given to users
+# after their Info Requests are marked as refused/rejected.
+#
+class RefusalAdviceController < ApplicationController
+  def create
+  end
+end
--- a/app/controllers/refusal_advice_controller.rb
+++ b/app/controllers/refusal_advice_controller.rb
@@ -4,5 +4,21 @@
 #
 class RefusalAdviceController < ApplicationController
   def create
+    @params = parsed_refusal_advice_params # TODO: move into log event method
+  end
+
+  private
+
+  def refusal_advice_params
+    params.require('refusal_advice').permit(:id, questions: {}, actions: {})
+  end
+
+  def parsed_refusal_advice_params
+    refusal_advice_params.merge(
+      actions: refusal_advice_params.fetch(:actions).
+        each_pair do |_, suggestions|
+          suggestions.transform_values! { |v| v == 'true' }
+        end
+    )
   end
 end
--- a/app/controllers/refusal_advice_controller.rb
+++ b/app/controllers/refusal_advice_controller.rb
@@ -7,6 +7,14 @@ class RefusalAdviceController < ApplicationController
 
   def create
     log_event
+
+    internal_redirect_to ||
+      help_page_redirect ||
+      external_redirect  ||
+      raise(
+        RefusalAdvice::Action::RedirectionError,
+        "Can't redirect to #{action.target}"
+      )
   end
 
   private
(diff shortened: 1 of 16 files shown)


app/controllers/refusal_advice_controller.rb
```diff
@@ -3,12 +3,45 @@
...
   end
 
   private
 
+  def authenticate
+    authenticated?({})
+  end
+
+  def info_request
+    @info_request ||= current_user.info_requests.
```
garethrees, line 34: should check what happens with "sign in as" here.

app/controllers/refusal_advice_controller.rb
```diff
@@ -0,0 +1,57 @@
...
+
+    when 'internal_review'
+      redirect_to new_request_followup_path(
+        request_id: info_request.id, internal_review: '1', anchor: 'followup'
+      )
+
+      # TODO: there are other actions - define what to do in the refusal advice
+      # YAML?
+    end
+  end
```
garethrees, line 25: Why not just do?

```ruby
    case parsed_refusal_advice_params.fetch(:button)
    when 'internal_review'
      redirect_to new_request_followup_path(
        request_id: info_request.id, internal_review: '1', anchor: 'followup'
      )
    else
      redirect_to new_request_followup_path(
        request_id: info_request.id, anchor: 'followup'
      )
    end
```

garethrees: Because all actions from the YAML get redirected through this controller.

app/controllers/refusal_advice_controller.rb
```diff
@@ -0,0 +1,57 @@
+##
+# Controller to store advice, from the refusal advice wizard, given to users
+# after their Info Requests are marked as refused/rejected.
+#
+class RefusalAdviceController < ApplicationController
+  before_action :authenticate
```
garethrees, line 6: Seems to fail with:

```

ActionView::Template::Error (undefined method `empty?' for nil:NilClass):
    35:   <div id="sign_together" class="sign_together">
    36:     <% if !@post_redirect.nil? %>
    37:       <h1 id="sign_in_reason" class="sign_in_reason">
    38:         <% if @post_redirect.reason_params[:web].empty? %>
    39:           <%= _('Please create an account or sign in') %>
    40:         <% else %>
    41:           <%= _('{{reason}}, create an account or sign in',

app/views/user/sign.html.erb:38:in `_app_views_user_sign_html_erb___1886059217492605600_47049850305220'
app/controllers/users/sessions_controller.rb:24:in `new'
app/controllers/application_controller.rb:118:in `record_memory'
lib/strip_empty_sessions.rb:15:in `call'
Started GET "/favicon.ico" for 10.10.10.1 at 2021-02-03 12:33:25 +0000
Processing by GeneralController#not_found as
  Parameters: {"path"=>"favicon"}
Completed 500 Internal Server Error in 6ms (ActiveRecord: 0.0ms)

```

app/controllers/refusal_advice_controller.rb
garethrees: Currently, if a user submits an action which is not defined here, we get a `no_content` response:

```
No template found for RefusalAdviceController#create, rendering head :no_content
```

app/controllers/refusal_advice_controller.rb
gbp: Fixed in #6098

app/controllers/refusal_advice_controller.rb
gbp: Confirmed this isn't an issue

app/controllers/refusal_advice_controller.rb
gbp: YAML format has been updated to include the `target` as discussed in #6103

garethrees (approved):
Looks great! ✨

PR #6107, opened 2021-02-09
[#6103][JRCT] Refusal advice updates

gbp:
## Relevant issue(s)

#6103

## What does this do?

Reworks the refusal wizard:
- Rename methods/classes to be more logical
- Some YAML options are now optional
- In-lines the actions within the list of next steps
- Allows expanding of the actions so they are visible regardless of being suggested by the wizard

### DIFF
--- a/app/views/help/refusal_advice/_action.html.erb
+++ b/app/views/help/refusal_advice/_action.html.erb
@@ -1,5 +1,6 @@
 <div class="wizard__action" data-block="<%= action.id %>">
   <h2><%= action.header %></h2>
+  <div><%= render action.body %></div>
 
   <%= render action.suggestions %>
 
--- a/spec/models/refusal_advice/action_spec.rb
+++ b/spec/models/refusal_advice/action_spec.rb
@@ -41,6 +41,12 @@ RSpec.describe RefusalAdvice::Action do
         RefusalAdvice::Suggestion.new(id: 'confirmation-not-too-costly')
       )
     end
+
+    context 'when no suggestions are defined' do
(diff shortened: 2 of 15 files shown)


spec/models/refusal_advice/action_spec.rb
```diff
@@ -59,6 +59,14 @@
         )
       end
     end
+
+    context 'without render options' do
+      before { data.delete(:body) }
+
+      it 'will renders an empty string' do
```
garethrees, line 67: `s/will renders an empty string/renders an empty string/`

app/views/help/refusal_advice/_action.html.erb
```diff
@@ -1,6 +1,7 @@
 <div class="wizard__action" data-block="<%= action.id %>">
   <h2><%= action.header %></h2>
 
+  <%= render action.body %>
```
garethrees, line 4: Wrap in a `<div>` so that we're sure it's enclosed in a block element to prevent the button appearing on the same line:

Screenshot 2021-02-09 at 12 15 21

garethrees: Might as well also wrap `<%= render action.suggestions %>` in a `<div>` too.

garethrees (approved):
Fantastic!

app/views/help/refusal_advice/_action.html.erb
gbp: Don't think `<%= render action.suggestions %>` needs this as the partial already wraps

PR #6111, opened 2021-02-11
[JRCT] Translated record form

garethrees:
## Relevant issue(s)

In service of #5911.

## What does this do?

Add `TranslatedRecordForm` form builder

## Why was this needed?

Make it easier to add translated record admin interfaces.

## Notes to reviewer

Extracted from https://github.com/mysociety/alaveteli/tree/5911-snippets-admin. I'd like to rebase https://github.com/mysociety/alaveteli/tree/5911-snippets-admin on top of this to make sure it all actually works, as a couple of tweaks were made while writing tests.

### DIFF
--- a/spec/helpers/admin/translated_record_helper_spec.rb
+++ b/spec/helpers/admin/translated_record_helper_spec.rb
@@ -1,0 +1,13 @@
+require 'spec_helper'
+
+describe Admin::TranslatedRecordHelper, type: :helper do
+  describe '#translated_form_for' do
+    let(:record) { FactoryBot.build(:announcement) }
+
+    it 'uses a custom form builder' do
+      helper.translated_form_for(record, url: '#') do |f|
+        expect(f).to be_an(Admin::TranslatedRecordForm)
+      end
+    end
+  end
+end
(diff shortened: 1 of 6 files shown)


garethrees:
Transient spec error – `bin/rspec spec/helpers/` sometimes passes locally.

gbp:
> Transient spec error – `bin/rspec spec/helpers/` sometimes passes locally.

Example failing spec command:
`rspec './spec/helpers/admin/translated_record_form_spec.rb[1:1:2:1]' './spec/helpers/public_body_helper_spec.rb[1:2:5:1]' --seed 33252`

gbp:
Transient spec error is due to rows not being cleared from the database between runs. This case is fixed in #6112

spec/helpers/admin/translated_record_helper_spec.rb
```diff
@@ -0,0 +1,17 @@
...
+      helper.translated_form_for(record) { |f| '' }
+    end
+
+    let(:record) { FactoryBot.create(:announcement) }
+
+    let(:expected) do
+      helper.form_for(record, builder: Admin::TranslatedRecordForm) { |f| '' }
+    end
+
+    it { is_expected.to eq(expected) }
```
gbp, line 15: This spec isn't proving anything without the new form builder methods being used. We don't want to expect those methods are outputting the HTML here so I would consider using a `to receive` or `to have_received` matcher to check the `builder` argument is used.

gbp (changes requested):
This is almost there... comments inline

PR #6120, opened 2021-02-15
[#6089] Fix typeahead searches

### LINKED ISSUE #6089: Authority search very poor
The search box at:

https://www.whatdotheyknow.com/select_authority

is particularly poor, it produces worse results than the general site search. 

A search for cabinet office on the select authority page on WhatDoTheyKnow.com currently gives no results. 

I suspect a number of the cases of reported problems logged at #1179 are not actually issues with the general site search, but with the select authority search. 

See also:  #4426

gbp:
## Relevant issue(s)

Fixes #6089

## What does this do?

Update expected exception message when wildcard searches error allowing
fallback to an non-wildcard search to continue.

## Why was this needed?

Poor search performance

## Implementation notes

Locally in dev and test env all was working fine. I could only replicate in production, this is due to the size of the index. A large index can't always do a wildcard search, for example, when searching for `Cabinet Office` it will raise an exception with the message like: 
`WildcardError: Wildcard office* expands to more than 1000 terms`

This message seems to have changed at some point when upgrading Xapian in the past.

### DIFF
--- a/lib/acts_as_xapian/acts_as_xapian.rb
+++ b/lib/acts_as_xapian/acts_as_xapian.rb
@@ -88,6 +88,12 @@ module ActsAsXapian
   def self.config
     @@config
   end
+  def self.max_wildcard_expansion=(max_wildcard_expansion)
+    @@max_wildcard_expansion = max_wildcard_expansion
+  end
+  def self.max_wildcard_expansion
+    @@max_wildcard_expansion
+  end
 
   ######################################################################
   # Initialisation
@@ -121,6 +127,8 @@ module ActsAsXapian
     Dir.mkdir(db_parent_path) unless File.exist?(db_parent_path)
 
     @@db_path = File.join(db_parent_path, environment)
+
+    @@max_wildcard_expansion = config.fetch('max_wildcard_expansion', 1000)
 
     # make some things that don't depend on the db
     # TODO: this gets made once for each acts_as_xapian. Oh well.
@@ -170,7 +178,9 @@ module ActsAsXapian
     # Large installations of Alaveteli should consider
     # upgrading, because uncontrolled wildcard expansion
     # can crash the whole server: see http://trac.xapian.org/ticket/350
-    @@query_parser.set_max_wildcard_expansion(1000) if @@query_parser.respond_to? :set_max_wildcard_expansion
+    if @@query_parser.respond_to? :set_max_wildcard_expansion
+      @@query_parser.set_max_wildcard_expansion(@@max_wildcard_expansion)
+    end
 
     @@stopper = Xapian::SimpleStopper.new
     @@stopper.add("and")
--- a/lib/acts_as_xapian/acts_as_xapian.rb
+++ b/lib/acts_as_xapian/acts_as_xapian.rb
@@ -42,8 +42,8 @@ module ActsAsXapian
   def self.bindings_available
     $acts_as_xapian_bindings_available
   end
-  class NoXapianRubyBindingsError < StandardError
-  end
+  NoXapianRubyBindingsError = Class.new(StandardError)
+  UnhandledRuntimeError = Class.new(StandardError)
 
   @@db = nil
   @@db_path = nil
@@ -351,6 +351,8 @@ module ActsAsXapian
           else
             raise
           end
+        rescue RuntimeError => ex
+          raise UnhandledRuntimeError, ex.message
         end
         self.cached_results = nil
       }
--- a/lib/typeahead_search.rb
+++ b/lib/typeahead_search.rb
@@ -20,13 +20,11 @@ class TypeaheadSearch
     ActsAsXapian.query_parser.default_op = Xapian::Query::OP_OR
     begin
       xapian_search = run_query
-    rescue RuntimeError => e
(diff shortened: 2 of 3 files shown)


gbp:
It appears`QueryParserError` still exists in the Xapian so maybe we need to catch both `QueryParserError` and `WildcardError`

gbp:
Exception message matching was first added in https://github.com/mysociety/alaveteli/commit/22b2db48c005be1d1b87dd6b58f830a2a74d2108

garethrees (approved):
Makes sense in principle!

Would be worth expanding the details into the commit message, and ideally link to the upstream code where these errors exist so that we have a better future reference.

> It appears `QueryParserError` still exists in the Xapian so maybe we need to catch both `QueryParserError` and `WildcardError`

Yeah maybe they split these into two distinct concepts. [We reference `QueryParserError` elsewhere](https://github.com/mysociety/alaveteli/blob/ad8a2db3e30e28ffea1b9692fa4ab6e8c16b7141/spec/controllers/general_controller_spec.rb#L422) so I assume both still exist, but whether we need to rescue from both here IDK.

> I could only replicate in production

I think it's worth trying to test this fix against the prod DB somehow. Worst case just temporarily patch the live file and run the query in the console.

gbp:
@garethrees I've made some changes:

1. managed to replicate this error in a spec by adding a configuration option to `ActsAsXapian`. So this is now tested.
2. moved the `RuntimeError` rescue into `ActsAsXapian` and now re-raising as exception within the `ActsAsXapian` model to allow us to catch know exception classes in the core Alaveteli code.
3. added exception notifications for us

gbp:
In trying to write a spec to show the original `QueryParserError` exception message I've worked out that this only is raised when using the `Xapian::QueryParser::FLAG_BOOLEAN` flag which we don't use in typeahead search.

```
[1] pry(main)> ActsAsXapian.readable_init
=> snip
[2] pry(main)> ActsAsXapian.query_parser.parse_query('dog AND', Xapian::QueryParser::FLAG_BOOLEAN)
RuntimeError: QueryParserError: Syntax: <expression> AND <expression>
from (pry):2:in `parse_query'
[3] pry(main)> ActsAsXapian.query_parser.parse_query('dog AND', Xapian::QueryParser::FLAG_LOVEHATE | Xapian::QueryParser::FLAG_SPELLING_CORRECTION | Xapian::QueryParser::FLAG_WILDCARD)
=> #<Xapian::Query:0x00007fd9082652d0 @__swigtype__="_p_Xapian__Query">
```

gbp:
In production we are hitting the wildcard expansion limit, which is currently set to 1000. We should consider increasing this limit. With can do this with 9d57051 by adding a `config/xapian.yml` file, something like:

```
test:
  max_wildcard_expansion: 10

development:
  max_wildcard_expansion: 500

production:
  max_wildcard_expansion: 1000
```

From my tests in the production Rails console, increasing to 1075 allows the wildcard query for 'Cabinet Office' to succeed. We would also need to consider server memory usage /cc @sagepe

lib/acts_as_xapian/acts_as_xapian.rb
```diff
@@ -42,8 +42,8 @@ module ActsAsXapian
   def self.bindings_available
     $acts_as_xapian_bindings_available
   end
-  class NoXapianRubyBindingsError < StandardError
-  end
+  NoXapianRubyBindingsError = Class.new(StandardError)
+  UnhandleRuntimeError = Class.new(StandardError)
```
garethrees, line 46:
```diff
- UnhandleRuntimeError
+ UnhandledRuntimeError
```

lib/typeahead_search.rb
```diff
@@ -20,13 +20,11 @@ def xapian_search
...
-    rescue RuntimeError => e
-      if e.message =~ /^QueryParserError: Wildcard/
-        # Wildcard expands to too many terms
-        Rails.logger.info "Wildcard query '#{query}' caused: #{e.message.force_encoding('UTF-8')}"
-        @wildcard = false
-        xapian_search = run_query
-      end
+    rescue ActsAsXapian::UnhandleRuntimeError => e
+      # Wildcard expands to too many terms
+      Rails.logger.info "Wildcard query '#{query}' caused: #{e.message.force_encoding('UTF-8')}"
```
garethrees: Wondering whether this should be `warn` rather than `info`? Not that it would make much difference to our visibility at the moment, but maybe it might be a way to keep more of an eye on this type of problem that only shows itself over time.

lib/typeahead_search.rb
```diff
@@ -21,8 +23,10 @@ def xapian_search
     begin
       xapian_search = run_query
     rescue ActsAsXapian::UnhandleRuntimeError => e
-      # Wildcard expands to too many terms
-      Rails.logger.info "Wildcard query '#{query}' caused: #{e.message.force_encoding('UTF-8')}"
```
garethrees: I think we should keep the log output as well as the optional exception notification.

garethrees (approved):
LGTM! Amazing work 🦸‍♂️ 

Couple of tiny changes suggested but no need for re-review.

PR #6122, opened 2021-02-16
Add example Xapian configuration

gbp:
## What does this do?

Add example Xapian configuration

## Why was this needed?

This will be used when the application is next deployed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,6 +19,7 @@
 /config/newrelic.yml
 /config/rails_env.rb
 /config/user_spam_scorer.yml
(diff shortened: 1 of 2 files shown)


gbp:
To test in the production console, run:

```
ActsAsXapian.readable_init
ActsAsXapian.max_wildcard_expansion = 1500
search = TypeaheadSearch.new('Cabinet Office', { model: PublicBody })
search.xapian_search
search.wildcard
```

This should return `true` if search was successfull - ie. expansion limit hasn't been reached

PR #6127, opened 2021-02-18
[#5911][JRCT] Add admin UI for snippets

### LINKED ISSUE #5911: Add snippet library to new followup page

gbp:
## Relevant issue(s)

Fixes #5911
Depends on #6126 
Depends on #6128

## What does this do?

Add admin UI for snippets

## Why was this needed?

So we can add resfual advice which will be offered as outgoing message snippets when composing followup replies and requests for internal reviews.

## Implementation notes

Do we want to put this behind a feature flag?

### DIFF
--- a/app/views/admin/outgoing_messages/snippets/edit.html.erb
+++ b/app/views/admin/outgoing_messages/snippets/edit.html.erb
@@ -1,0 +1,16 @@
+<h1><%= @title %></h1>
+
+<div class="row">
+  <div class="span8">
+    <div id="admin_snippet_form">
+      <%= translated_form_for @snippet, url: admin_snippet_path(@snippet), method: :put, html: { class: 'form form-horizontal' } do |f| %>
+        <%= render partial: 'form', locals: { f: f } %>
+
+        <div class="form-actions">
+          <%= f.submit 'Save', accesskey: 's', class: 'btn btn-success' %>
+          <%= link_to 'List all', admin_snippets_path, class: 'btn' %>
+        </div>
+      <% end %>
+    </div>
+  </div>
+</div>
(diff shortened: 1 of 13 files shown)


gbp:
Probably could refactor 0b91d10 into a seporate PR

app/views/admin/outgoing_messages/snippets/edit.html.erb
```diff
@@ -12,5 +12,10 @@
         </div>
       <% end %>
     </div>
+
+    <%= form_tag admin_snippet_path(@snippet), class: 'form form-inline', method: 'delete' do %>
+      <%= submit_tag "Destroy #{@snippet.name}", class: 'btn btn-danger' %>
```
garethrees, line 17: `name` is likely going to be far too long.

Screenshot 2021-02-23 at 09 25 22

Should either truncate or just use `'Destroy snippet'`.

garethrees (approved):
LGTM!

One minor comment and pushed some minor style fixes that can be squashed on merge.

> Do we want to put this behind a feature flag?

Maybe pop the nav link behind a flag.

PR #6130, opened 2021-02-18
[#6129] Fix broken headers when signing in

### LINKED ISSUE #6129: Bug in heading text on sign in page
At eg. 

www.whatdotheyknow.com/profile/sign_in

The top heading is:
>**, create an account or sign in**

it used to be, and probably still should be:

>**Please create an account or sign in**

Apologies if this is a https://github.com/mysociety/whatdotheyknow-theme issue.

gbp:
## Relevant issue(s)

Fixes #6129

## What does this do?

This commit fixes broken headers when signing in by checking if the web reason param is either nil or empty by using the `#blank?` method.

## Why was this needed?

In 2866a97 we made a minor change to the controller authentication which introduced an issue where the wrong string was being used on the sign view.

### DIFF
--- a/app/views/user/sign.html.erb
+++ b/app/views/user/sign.html.erb
@@ -3,7 +3,7 @@
   <div id="sign_alone" class="sign_alone">
 
     <p id="sign_in_reason" class="sign_in_reason">
-      <% if @post_redirect.reason_params[:web].nil? %>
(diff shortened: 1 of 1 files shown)


garethrees (approved):
Haven't run it but looks sensible so happy to merge.

PR #6137, opened 2021-02-24
Updates expected refusal advice tag

gbp:
## What does this do?

Updates expected refusal advice tag

## Why was this needed?

This matches what is expected in the `RefusalAdvice#snippets` and will allow snippets to have a single key+value tag, EG:
  `refusal_advice:section-12`
Rather than needing two tags:
  `refusal_advice refusal:section-12`


garethrees (approved):
Nice, much nicer!

Spec failing but fix is obvious. No need for rereview.

Issue #6026, opened 2020-12-14
Pre-check automatically identified refusals


garethrees:
If we land on `/help/unhappy` with a request that has refusals, we should pre-select the applicable refusal reasons.

PR #6139, opened 2021-02-26
Create RefusalAdviceQuestionForm

gbp:
## Relevant issue(s)

Initially thought this might be useful for #6026 - not too sure now but still a good refactoring I think

## What does this do?

Create RefusalAdviceQuestionForm

## Why was this needed?

Extract and rename helper methods into a FormBuilder which allows us to
simplify method attributes and the view.


garethrees (approved):
Nice improvement!

PR #6140, opened 2021-02-26
[#6026][JRCT] Automatically check identified refusals

### LINKED ISSUE #6026: Pre-check automatically identified refusals

gbp:
## Relevant issue(s)

Depends on #6139 - although it doesn't need to if we don't want that PR
Fixes #6026

## What does this do?

Makes improvments to the wizard JS so it will:
1. automatically check identified refusals
2. use a new wizard question option class

## Why was this needed?

The app can identified refusals references so it makes sense to automattically check these checkboxes.

### DIFF
--- a/app/assets/javascripts/wizard.js
+++ b/app/assets/javascripts/wizard.js
@@ -32,6 +32,8 @@
   RefusalWizard.prototype._init = function(target) {
     var wizard = this;
 
+    wizard.$form = wizard.$el.find('form');
+
     wizard.$blocks = wizard.$el.find(
       "." +
         wizard.options.questionClass +
@@ -91,6 +93,8 @@
     wizard.$questions.on("change", function() {
       wizard._update($(this));
     });
+
+    wizard._checkIdentifiedRefusals();
   };
 
   RefusalWizard.prototype._setupNextSteps = function() {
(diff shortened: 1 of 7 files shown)


app/helpers/refusal_advice_helper.rb
```diff
@@ -4,4 +4,20 @@ def refusal_advice_actionable?(action, info_request:)
...
+  end
+
+  private
+
+  def latest_refusals_as_json(info_request)
+    latest_refusals(info_request).map(&:to_param)
+  end
+
+  def latest_refusals(info_request)
+    info_request.incoming_messages&.map(&:refusals).reject(&:empty?).last || []
```
garethrees, line 21: Feels like at least some of this should be at the model layer. Probably `InfoRequest#refusals`?

app/helpers/refusal_advice_helper.rb
```diff
@@ -4,4 +4,20 @@ def refusal_advice_actionable?(action, info_request:)
...
+
+  def refusal_advice_form_data(info_request)
+    return {} unless info_request
+
+    { refusals: latest_refusals_as_json(info_request) }
+  end
+
+  private
+
+  def latest_refusals_as_json(info_request)
```
garethrees, line 16: We're not actually converting to JSON here. Given the other comment below, I _think_ we can condense the `private` methods here down to:

```ruby
def latest_refusals(info_request)
  Array(info_request.refusals.reject(&:empty?).last).map(&:to_param)
end
```

app/assets/javascripts/wizard.js
```diff
@@ -253,6 +257,18 @@
     $options.prop("checked", false);
   };
 
+  RefusalWizard.prototype._checkInitialQuestions = function() {
```
garethrees, line 260: Think a better name for this would be `_checkIdentifiedRefusals`?

garethrees (approved):
Bit of minor extracting that would be worth doing but looks great overall! 🕺

PR #6147, opened 2021-03-01
Add a reference to IPv6 and AAAA records to production server page

sagepe:
This simply suggests ensuring this is set-up should the hosting company support it.
Screenshot 2021-03-01 at 14 39 49


garethrees (commented):
👍  Thanks!

PR #6160, opened 2021-03-15
Simpler Wording for Personal Circumstnaces

benrfairless:
Refer https://github.com/openaustralia/righttoknow/issues/757#issuecomment-773366578

## What does this do?
- Changes the wording on the "Personal Circumstances" hide reason in the Request Admin view to be easier to read.

## Why was this needed?
- The previous version isn't clear why we are hiding requests for personal information. 
- In some jurisdictions, requests for personal information are covered by FOI
- The real reason Alaveteli doesn't allow requests for Personal info is because it could make sensitive info public inadvertently.

## Implementation notes
- Anyone using the view in their theme should consider if they need to consider doing so.


garethrees:
Thanks! Merged as bb2d2d3b5b71a031bcc809bd05841deb52fc05ce (I fixed up the newline with empty whitespace).

PR #6165, opened 2021-03-18
Use traits in info requests factory

mlandauer:
This is a straightforward refactor of the info_requests factory to extract [traits](https://github.com/thoughtbot/factory_bot/blob/master/GETTING_STARTED.md#traits)

## What does this do?

By extracting traits you can more easily create requests in tests that have multiple things set - e.g. a request that is embargoed, has an incoming message on it and is awaiting a description to be set.

## Why was this needed?

It's needed for something I'm working on at the moment to automatically create screenshots for the Pro marketing pages (for example the four screenshots on https://www.whatdotheyknow.com/pro). The idea is to make it easy to recreate the screenshots automatically with the styling of your particular theme and also make it easier to update the content of the screenshots to be more appropriate to your location.

I will likely open a PR for a proof-of-concept of this as a separate thing to discuss but I figured this change was useful enough on its own that it was worth putting in first.

### DIFF
--- a/spec/factories/info_requests.rb
+++ b/spec/factories/info_requests.rb
@@ -48,146 +48,35 @@ FactoryBot.define do
       initial_request.save
     end
 
-    factory :info_request_with_incoming do
+    trait :with_incoming do
       after(:create) do |info_request, evaluator|
         incoming_message = create(:incoming_message, :info_request => info_request)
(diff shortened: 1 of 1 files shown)


gbp:
@mlandauer thanks for this. I merged #6189 with some minor changes made to the last commit where I think you ran the auto-correct rubocop command.

PR #6185, opened 2021-04-06
Embargo extensions cleanup

garethrees:
## What does this do?

Minor cleanup of `AlaveteliPro::EmbargoExtensionsController`

## Why was this needed?

Was reading this in relation to a Pro support query and found it a bit hard work so cleaned it up as I read.

## Implementation notes

Not intended to be a full refactoring; just making it a little better than it was before.


garethrees:
Merging since there's a known issue with 2.3 builds.

PR #6208, opened 2021-04-16
Add support for Glass Xapian database format

gbp:
## Relevant issue(s)

https://github.com/mysociety/sysadmin/issues/1409#issuecomment-815632882

## What does this do?

Add support for Glass Xapian database format 

## Why was this needed?

Framework modernisation

### DIFF
--- a/doc/CHANGES.md
+++ b/doc/CHANGES.md
@@ -20,6 +20,15 @@
 
 ## Upgrade Notes
 
+* We now support "Glass" formatted Xapian databases. We will continue to support
+  existing "Chert" formatted databases for as long as possible. You can convert
+  your database by either:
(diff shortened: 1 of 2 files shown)


garethrees:
Need to check what happens on reindexing here. Do we create a copy and then reindex offline, or does it create a "new" database based on the old one?

gbp:
> Need to check what happens on reindexing here. Do we create a copy and then reindex offline, or does it create a "new" database based on the old one?

There was indeed an issue when reindexing. Updating an existing Chert DB resulted in Glass files in the Xapian DB directory. I have made a change to check for existing `iamchert` file when opening an existing DB for writing.

gbp:
Tested with this script: https://gist.github.com/gbp/b850b1f1cf80a036909534075ce315e8

garethrees (approved):
LGTM!

PR #6209, opened 2021-04-16
Fix snippets error when there are no suggestions

gbp:
Fixes:
  NoMethodError: undefined method `inject' for nil:NilClass

### DIFF
--- a/app/models/refusal_advice.rb
+++ b/app/models/refusal_advice.rb
@@ -49,7 +49,7 @@ class RefusalAdvice
     action = wizard_answer.params[:id]
     suggestions = wizard_answer.params[:actions][action.to_sym]
 
-    suggestions.inject([]) do |memo, (k, v)|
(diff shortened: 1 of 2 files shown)


app/models/refusal_advice.rb
```diff
@@ -50,6 +50,7 @@ def suggested_actions
 
     action = wizard_answer.params[:id]
     suggestions = wizard_answer.params[:actions][action.to_sym]
+    return unless suggestions
 
     suggestions.inject([]) do |memo, (k, v)|
```
garethrees, line 55: How about this instead to reduce conditionals and method body length?

```diff
- suggestions.inject([])
+ Array(suggestions).inject([])
```

garethrees:
I've pushed some fixups to get this closer to being over the line.

@gbp could you just tweak that spec?

Issue #5283, opened 2019-07-08
Error in wording of admin flash notice on marking a response as requesting clarification


RichardTaylor:
The admin flash notice:

>Old response marked as having been a clarification

should state

>Old response marked as having been a request for clarification

Relatively unimportant but could confuse or worry new administrators. 

I'm not sure this is important enough to warrant the bright red "bug" label.

PR #6213, opened 2021-04-18
[#5283] Correct flash notice when marking response as a clarification request

MattK1234:
## Relevant issue(s)
#5283

## What does this do?

This changes the wording of the flash notice described when marking a response as a request for clarification in the admin interface. Instead of "Old response marked as having been a clarification" we now say "Old response marked as having been a request for clarification".

## Why was this needed?

Could confuse or worry newer admins that an error has been made.

## Implementation notes

Have changed this in the spec file as well, as part of this PR.

## Screenshots
Old:
image

New:
image

## Notes to reviewer
First time making a PR on Alavateli, please be gentle :-)

### DIFF
--- a/app/controllers/admin_info_request_event_controller.rb
+++ b/app/controllers/admin_info_request_event_controller.rb
@@ -21,7 +21,7 @@ class AdminInfoRequestEventController < AdminController
     # Reset the due dates for the request if necessary
     @info_request_event.recheck_due_dates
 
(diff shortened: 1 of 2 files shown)


garethrees:
Thanks! I've updated this against the most recent development branch and squashed the commits together in https://github.com/mysociety/alaveteli/pull/6241

PR #6216, opened 2021-04-20
Convert templates from Ugly to ERB

gbp:
## Relevant issue(s)

Connected to #6052

## What does this do?

Convert templates from Ugly to ERB

## Why was this needed?

Our Ugly implementation isn't fully complete compared to the mySociety
deployment scripts where Ugly templates are parsed using Perl.

Switching to ERB allows us to use conditionals. This will allow us to
selectively use rbenv for daemons and crontab.

### DIFF
--- a/config/alert-tracks-debian.example
+++ b/config/alert-tracks-debian.example
@@ -1,33 +1,33 @@
 #!/bin/bash
 #
 ### BEGIN INIT INFO
-# Provides: !!(*= $daemon_name *)!!
+# Provides: <%= daemon_name %>
 # Required-Start: $local_fs $syslog
 # Required-Stop: $local_fs $syslog
 # Default-Start: 2 3 4 5
 # Default-Stop: 0 1 6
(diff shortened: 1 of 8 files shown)


gbp:
@sagepe as discussed, it would be good to get your thoughts on this. I don't think any of these example files would be used in a mySociety deployment as we've got overrides in the server repo.

sagepe:
I don't see a problem with this approach, as you say none of the files will be used internally (with the exception of `contab-example`, and in this case that file's presence will only trigger the deployment script to build a crontab using the template in the internal configuration repository). And as discussed, it makes sense to me to use the most appropriate templating system for the language and framework in use in a given project.

garethrees (commented):
Haven't run it but happy in principle. Worth adding a changelog entry about this.

Should also update https://alaveteli.org/docs/installing/cron_and_daemons which is I think the only place where we mention `.ugly`.

PR #6224, opened 2021-05-01
Add development Docker environment

gbp:
## What does this do?

Add development Docker environment

## Why was this needed?

As Virtualbox/Vagrant isn't compatible with M1 Macs we need to be able to easily setup a valid development envionment. Will be required by @wrightmartin when he does some design changes on the recent refusal advice work.

## Implementation notes

Install Docker desktop and run:
`./docker/bootstrap` and then `./docker/server`

This will install `alavetelitheme` as default, for other themes, check them out inside a `alaveteli-themes` directory which should sit along side your `alaveteli` repo dir.

`./script/switch-theme.rb` has been updated to generate relative symbolic links which should work in containers and other more standard environments.


sagepe (changes requested):
Some of the Rails errors may well be to do with config/setup issues at my end - might be good to pair on getting it working then we can talk about some of the other bits and pieces?

docker/script/setup
```diff
@@ -0,0 +1,25 @@
...
+SQL
+
+createdb -h db -U postgres -T template_utf8 alaveteli_development
+createdb -h db -U postgres -T template_utf8 alaveteli_test
+
+bin/rails db:migrate db:seed
+bin/rails db:migrate RAILS_ENV=test
+
+bundle exec script/load-sample-data
+bundle exec script/update-xapian-index
```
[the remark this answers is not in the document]
sagepe: Due to older checkout of the default theme, resolved when that was updated.

docker/script/setup
```diff
@@ -0,0 +1,25 @@
+if [[ -z "$DOCKER" ]]; then
+  docker compose -f docker/development.yml -p alaveteli run app ./docker/script/setup
+  exit
+fi
+
+bundle install
+
+git clone https://github.com/mysociety/alavetelitheme.git lib/themes/alavetelitheme
```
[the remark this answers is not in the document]
sagepe: Also noting that the older version of this theme that was already checked out here caused the rails errors, so that should probably be checked in some way?

docker/script/setup
```diff
@@ -0,0 +1,25 @@
...
+cp config/database.yml-docker config/database.yml
+cp config/general.yml-example config/general.yml
+
+createdb -h db -U postgres -T template0 -E UTF-8 template_utf8
+psql -h db -U postgres -q <<SQL
+UPDATE pg_database SET datistemplate=true, datallowconn=false WHERE datname='template_utf8';
+SQL
+
+createdb -h db -U postgres -T template_utf8 alaveteli_development
+createdb -h db -U postgres -T template_utf8 alaveteli_test
```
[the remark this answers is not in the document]
sagepe: This would also mean that we could remove the `POSTGRES_HOST_AUTH_METHOD` variable from the compose file.

docker-compose.yml
```diff
@@ -0,0 +1,29 @@
+version: '3.8'
+
+services:
+  app:
+    build:
+      dockerfile: docker/Dockerfile
```
sagepe, line 6: There needs to be a`context` key here:
```
  build:
    context: .
    dockerfile: docker/Dockerfile
```

We probably could build the images with a context of `./docker`, avoiding the need for the `.dockerignore`, the paths in the `dockerfile` keys and the `ADD` line in the `Dockerfile-postges`, but it doesn't make any practical difference :shrug:

docker-compose.yml
```diff
@@ -0,0 +1,29 @@
...
+    volumes:
+      - ./:/alaveteli
+      - ../alaveteli-themes:/alaveteli-themes
+      - bundle:/bundle
+    depends_on:
+      - db
+
+  db:
+    build:
+      dockerfile: docker/Dockerfile-postgres
```
sagepe, line 21: As per previous comment.

docker/bootstrap
```diff
@@ -0,0 +1,47 @@
...
+set -e
+
+cd "$(dirname "$0")/.."
+
+error_msg() { printf "\033[31m%s\033[0m\n" "$*"; }
+notice_msg() { printf "\033[33m%s\033[0m " "$*"; }
+success_msg() { printf "\033[32m%s\033[0m\n" "$*"; }
+
+notice_msg "Checking dependencies..."
+if ! which docker &>/dev/null; then
```
sagepe, line 12: I'd use `if ! command -v docker >/dev/null 2>&1; then` here - a bit more verbose, but more portable given we're calling this script with `/bin/sh`.

docker/bootstrap
```diff
@@ -0,0 +1,47 @@
...
+git submodule update --init
+success_msg 'done'
+
+notice_msg "Getting latest alavetelitheme commits..."
+alavetelitheme="$themes_dir/alavetelitheme"
+[ ! -d $alavetelitheme ] && git clone https://github.com/mysociety/alavetelitheme.git $alavetelitheme >/dev/null
+pushd $alavetelitheme >/dev/null && git pull >/dev/null && popd >/dev/null
+success_msg 'done'
+
+./docker/setup
```
sagepe, line 47: Just a suggestion, but I think it would be closer to the scripts-to-rule-them-all pattern to be able to run `docker/bootstrap` without also calling `docker/setup`. Personally speaking I expect the bootstrap script simply to deal with pre-requisites for running things, and for that script to then be called elsewhere where necessary.

I suggest replacing this line will a call to `docker/bootstrap` inside the `if [ -z "$DOCKER ]; then` block in `docker/setup`.

sagepe (commented):
Looking good! I have a working environment from a fresh clone 🎉

I've just a couple of really small tweaks.

docker/bootstrap
```diff
@@ -0,0 +1,47 @@
...
+success_msg 'done'
+
+notice_msg "Syncing git submodules..."
+git submodule update --init
+success_msg 'done'
+
+notice_msg "Getting latest alavetelitheme commits..."
+alavetelitheme="$themes_dir/alavetelitheme"
+[ ! -d $alavetelitheme ] && git clone https://github.com/mysociety/alavetelitheme.git $alavetelitheme >/dev/null
+pushd $alavetelitheme >/dev/null && git pull >/dev/null && popd >/dev/null
```
sagepe, line 44: Also `pushd` and `popd` should be avoided with `/bin/sh` as they aren't POSIX compliant.

docker-compose.yml
gbp: Not sure why I took that out, guess I assumed the working directory would be the default context

Issue #5405, opened 2019-10-10
Overridden scope PublicBodyChangeRequest.open


garethrees:
Seeing lots of these in production logs:

```
W, [2019-10-10T14:17:40.179390 #30917]  WARN -- : Creating scope :open. Overwriting existing method PublicBodyChangeRequest.open
```

garethrees:
> All Ruby objects inherit from Object, which includes the Kernel module, which in turn, has an open method. In my situation, I do not use the Kernel module’s open method and I need to keep the naming consistent with the existing project. To silence these messages we can make the following change to order_status.rb:
>
> ```
> class OrderStatus < ApplicationRecord 
>   class << self
>     undef open
>   end 
>   enum status: {'open':1,'hold':2,'closed':3}
>   OPEN='open'
>   HOLD='hold'
>   CLOSED='closed'
> end
> ```
>
> _– https://medium.com/quark-works/ruby-on-rails-active-record-enum-6a08df7f3685_

Slightly nicer fix in https://github.com/rails/rails/issues/31234#issuecomment-824144388 though:

```ruby
  singleton_class.undef_method :open # hide Kernel.open, avoiding a warning when defining scope :open
  scope :open, -> { where(status: [:pending, :requested]) }
```

Issue #5222, opened 2019-05-30
Drop support for Ruby 2.3


garethrees:
* Debian Stretch: [`ruby2.3`](https://packages.debian.org/stretch/ruby)
* Debian Buster: [`ruby2.5`](https://packages.debian.org/buster/ruby)
* Ubuntu Bionic (18.04LTS): [`ruby2.5`](https://packages.ubuntu.com/bionic/ruby)
* Ubuntu Focal (20.04LTS): [`ruby2.7`](https://packages.ubuntu.com/focal/ruby)
- [x] Remove Unicode gem (https://github.com/mysociety/alaveteli/pull/5907)

garethrees:
Some [notes on rbenv](https://github.com/mysociety/alaveteli/wiki/Migrating-an-existing-Alaveteli-site-from-ruby-1.8.7) from when we [dropped 1.8.7](https://github.com/mysociety/alaveteli/issues/2048)

garethrees:
> To retain compatibility with Stretch we could do something like: https://gist.github.com/gbp/6aa433ef88daef780b63a5d9d0add96d

This needs `sudo apt-get install -y libssl-dev`

PR #6228, opened 2021-05-07
[#5404] Undef Kernel.open before creating open scope

garethrees:
All Ruby objects inherit from Object, which includes the Kernel module,
which in turn, has an open method. Redefining `.open` was creating lots
of WARN level messages in the logs:

    W, [2019-10-10T14:17:40.179390 #30917]  WARN -- : Creating scope
    :open. Overwriting existing method PublicBodyChangeRequest.open

Since we're not using `Kernel.open`, we can explicitly undefine it prior
to defining our own `.open` in the form of a `scope`.

Fixes https://github.com/mysociety/alaveteli/issues/5405.

### DIFF
--- a/app/models/public_body_change_request.rb
+++ b/app/models/public_body_change_request.rb
@@ -42,6 +42,8 @@ class PublicBodyChangeRequest < ApplicationRecord
   scope :body_update_requests, -> {
     where("public_body_id IS NOT NULL").order("created_at")
   }
+
+  singleton_class.undef_method :open # Undefine Kernel.open to avoid warning
   scope :open, -> { where(is_open: true) }
 
   def self.from_params(params, user)


garethrees:
Blocked by #5222/#5223, will be mergeable after https://github.com/mysociety/alaveteli/pull/6052.

gbp (approved):
Nice, I always wondered why these were appearing.

I don't really understand why we need to call `undef_method` on the `singleton_class` instead of calling as a class method but this certainly works.

garethrees:
> I don't really understand why we need to call `undef_method` on the `singleton_class` instead of calling as a class method but this certainly works.

Ah, I left some earlier investigation in https://github.com/mysociety/alaveteli/issues/5405#issuecomment-834237671. I don't fully understand it either tbh. I assume it's something to do with being explicit about removing the `open` method of `PublicBodyChangeRequest` rather than the `open` of `Kernel`.

Issue #6121, opened 2021-02-16
Tighter scoping of wizard anchor CSS


garethrees:
Need to be able to render links in the suggestion text.

![Screenshot 2021-02-16 at 10 09 27](https://user-images.githubusercontent.com/282788/108048605-2622d680-703f-11eb-99a1-63057d234b31.png)

garethrees:
Fixed by bce6ba58d

PR #6245, opened 2021-05-14
Issues/6121 tighter css scope

wrightmartin:
## Relevant issue(s)

Fixes https://github.com/mysociety/alaveteli/issues/6121

## What does this do?
Changes the specificity to be tighter so it's targeting `.wizard__next-step__title` instead of all anchor tags within `.wizard__next-step`

## Why was this needed?
It was causing all anchor tags to have block level display modes resulting in broken lines of text if they had links

## Implementation notes
None

## Screenshots
image

## Notes to reviewer
I've tested this with as many combinations of question answers to check there's no other reason the `display: block` was applied to anchors here, and while I can't test everything it looks good. 

This was done on the docker branch so I imagine when merging we'll want to cherry pick this commit or something?


garethrees:
Thanks! Merged as bce6ba58d.

Issue #6246, opened 2021-05-14
[ERROR] help#unhappy (ActionView::Template::Error) "undefined method `[]' for nil:NilClass"


garethrees:
```
An ActionView::Template::Error occurred in help#unhappy:

undefined method `[]' for nil:NilClass
app/models/refusal_advice.rb:23:in `questions'


-------------------------------
Request:
-------------------------------

* URL : https://www.whatdotheyknow.com/help/unhappy/asbestos_question_3
* HTTP Method: GET
* Parameters : {"controller"=>"help", "action"=>"unhappy", "url_title"=>"asbestos_question_3"}
* Timestamp : 2021-05-14 11:04:55 +0100
* Rails root : /data/vhost/www.whatdotheyknow.com/alaveteli-2021-05-07T10-14-32

-------------------------------
Session:
-------------------------------

* session id: [FILTERED]
* data: {"session_id"=>"REDACTED",
"locale"=>"en",
"user_id"=>181819,
"ttl"=>"2021-05-14T11:04:55.444+01:00",
"remember_me"=>false,
"post_redirect_token"=>"REDACTED",
"_csrf_token"=>"REDACTED",
"flash"=>
{"discard"=>[],
"flashes"=>
{"notice"=>
{"partial"=>"request/describe_notices/waiting_response_very_overdue",
"locals"=>{"info_request_id"=>706831, "annotations_enabled"=>true}}}}}

-------------------------------
Backtrace:
-------------------------------

app/models/refusal_advice.rb:23:in `questions'
app/views/help/_refusal_advice.html.erb:4:in `block in _app_views_help__refusal_advice_html_erb__782931655989036458_69977395820940'
app/views/help/_refusal_advice.html.erb:2:in `_app_views_help__refusal_advice_html_erb__782931655989036458_69977395820940'
lib/themes/whatdotheyknow-theme/lib/views/help/unhappy.html.erb:3:in `_lib_themes_whatdotheyknow_theme_lib_views_help_unhappy_html_erb___1122025999156432795_69977269275080'
app/controllers/application_controller.rb:113:in `record_memory'
lib/strip_empty_sessions.rb:15:in `call'

garethrees:
Might be similar to https://github.com/mysociety/alaveteli/pull/6209.

garethrees:
Both requests are EIR:

```ruby
InfoRequest.find_by(url_title: 'asbestos_question_3').law_used
# => "eir"
[...]

PR #6247, opened 2021-05-14
[#6246] Handle unconfigured legislation questions

garethrees:
Fixes https://github.com/mysociety/alaveteli/issues/6246.

`RefusalAdvice#questions` and `RefusalAdvice#actions` expects a `Legislation` to be configured. 

It allows empty values, but currently we have not even configured EIR.

This leads to failures when we try to render the refusal advice for an EIR request.

## Screenshots

```ruby
InfoRequest.find_by(url_title: 'example_title_40').legislation.to_s
# => "EIR"
```

BEFORE

Screenshot 2021-05-14 at 16 14 02

AFTER

Screenshot 2021-05-14 at 16 11 44

### DIFF
--- a/app/models/refusal_advice.rb
+++ b/app/models/refusal_advice.rb
@@ -20,7 +20,7 @@ class RefusalAdvice
   end
 
   def questions
-    Array(data[legislation.to_sym][:questions]).
+    Array(data.dig(legislation.to_sym, :questions)).
       map { |question| Question.new(question) }
   end
 
--- a/app/models/refusal_advice.rb
+++ b/app/models/refusal_advice.rb
@@ -20,13 +20,11 @@ class RefusalAdvice
   end
 
   def questions
-    Array(data.dig(legislation.to_sym, :questions)).
-      map { |question| Question.new(question) }
+    build_from_data(:questions)
   end
 
   def actions
-    Array(data.dig(legislation.to_sym, :actions)).
-      map { |action| Action.new(action) }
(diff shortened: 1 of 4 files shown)


app/models/refusal_advice.rb
```diff
@@ -81,6 +79,12 @@ def ==(other)
 
   private
 
+  def build_from_data(key)
+    Array(data.dig(legislation.to_sym, key)).map do |value|
+      "#{ self.class }::#{ key.to_s.classify }".constantize.new(value)
```
gbp, line 84: Not sure why Rubocop didn't catch this:
```suggestion
      "#{self.class}::#{key.to_s.classify}".constantize.new(value)
```

gbp (approved):

app/models/refusal_advice.rb
garethrees: > Not sure why Rubocop didn't catch this

This [check is disabled](https://github.com/mysociety/alaveteli/blob/0f831a00b134c5a244859391ff2031d11de339a0/.ruby-style.yml#L324-L325) and has been [since it was added](https://github.com/mysociety/alaveteli/commit/a45df67b30dbc48accd6391d7067280be3c4a994).

I can't quickly find anything where this decision has been documented, but Alaveteli has always implicitly used [`EnforcedStyle: space`](https://docs.rubocop.org/rubocop/1.0/cops_layout.html#enforcedstyle-space-4).

PR #6056, opened 2021-01-07
[#5226] Add buster support

gbp:
## Relevant issue(s)

Fixes https://github.com/mysociety/alaveteli/issues/5226
See https://github.com/mysociety/sysadmin/issues/867
Requires https://github.com/mysociety/commonlib/pull/70

## What does this do?

1. Switches the wkhtmltox package
2. Adds vagrant buster box

## Why was this needed?

To support the Debian version

## Implementation notes

These are commits from commonlib package repo changes

PR #6257, opened 2021-05-18
Fix PDF request summary

gbp:
## Relevant issue(s)

Connected to #6056 

## What does this do?

Add `-q` option to calls to `wkhtmltopdf`

This makes the `wkhtmltopdf` command less verbose which is important as we check for output to determine if the command was successful or not.

Really should be checking the exit code but this would mean a change in the `AlaveteliExternalCommand` module which is also used else where.

## Why was this needed?

This is broken in production and only `txt` correspondence is being downloaded. Not sure when this broken possiblly related to https://github.com/mysociety/sysadmin/issues/867

### DIFF
--- a/app/controllers/request_controller.rb
+++ b/app/controllers/request_controller.rb
@@ -679,7 +679,7 @@ class RequestController < ApplicationController
       tmp_input.close
       tmp_output = Tempfile.new('foihtml2pdf-output')
       output = AlaveteliExternalCommand.run(
-        convert_command, tmp_input.path, tmp_output.path
+        convert_command, '-q', tmp_input.path, tmp_output.path
       )
       if !output.nil?
         file_info = { :filename => 'correspondence.pdf',
--- a/app/controllers/request_controller.rb
+++ b/app/controllers/request_controller.rb
@@ -678,13 +678,18 @@ class RequestController < ApplicationController
       tmp_input.write(html_output)
       tmp_input.close
       tmp_output = Tempfile.new('foihtml2pdf-output')
-      output = AlaveteliExternalCommand.run(convert_command, tmp_input.path, tmp_output.path)
+      output = AlaveteliExternalCommand.run(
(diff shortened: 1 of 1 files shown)


app/controllers/request_controller.rb
```diff
@@ -678,13 +678,18 @@ def make_request_summary_file(info_request)
       tmp_input.write(html_output)
       tmp_input.close
       tmp_output = Tempfile.new('foihtml2pdf-output')
-      output = AlaveteliExternalCommand.run(convert_command, tmp_input.path, tmp_output.path)
```
garethrees, line 681: Think we could make this work by changing the `convert_command` to preserve the interface here.

```yaml
# config/general.yml
HTML_TO_PDF_COMMAND: "/usr/local/bin/wkhtmltopdf -q"
```

If this works then we'll want to update `config/general.yml-example` so that it gets set like this for everyone, and also note in the changelog that this is needed.

gbp: This won't work without additional changes as this would eval to false:

```ruby
convert_command = AlaveteliConfiguration.html_to_pdf_command
File.exist?(convert_command)
```

garethrees: How about something like:

```yaml
# Default
HTML_TO_PDF_COMMAND: "/usr/local/bin/wkhtmltopdf"

# Optional
HTML_TO_PDF_COMMAND: ["/usr/local/bin/wkhtmltopdf", "-q"]
```

```ruby
convert_command = Array(AlaveteliConfiguration.html_to_pdf_command)
cmd = convert_command.shift
File.exist?(cmd)
output = AlaveteliExternalCommand.run(cmd, convert_command, tmp_input.path, tmp_output.path)
```

garethrees:
Superseded by https://github.com/mysociety/alaveteli/pull/6488?

PR #6261, opened 2021-05-19
Remove Admin URL from requires admin notifications

garethrees:
I frequently wonder whether to remove this when I reply to user support
requests. Most of the time I don't bother, since one needs authorisation
to view the page, but it still doesn't _look_ good to reveal to
outsiders.

We also tend to redact these URLs from SAR responses, which creates
significant extra work.

This commit removes the URL. It adds an extra click for admins who need
to act on the request, but personally speaking I almost always use the
public URL first to review the content before visiting the admin page to
act.

### DIFF
--- a/app/mailers/request_mailer.rb
+++ b/app/mailers/request_mailer.rb
@@ -67,7 +67,6 @@ class RequestMailer < ApplicationMailer
     user = set_by || info_request.user
     @reported_by = user
     @url = request_url(info_request)
-    @admin_url = admin_request_url(info_request)
     @info_request = info_request
     @message = message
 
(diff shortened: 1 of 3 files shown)


RichardTaylor:
I agree with this change. 

Ideally we'd remove these admin links from correspondence before replying to users too, we often don't. This change will tidy up correspondence with users as well as have the benefit mentioned in respect of SAR response preparation.

mdeuk:
I also agree with this change - there’s no substantial downside from an administrative perspective for us, and the benefits outweigh any slight increase in admin time.

Issue #1881, opened 2014-10-09
Censor Rules do not get applied to Info Request titles


garethrees:
Added a censor rule to a user to redact `Localhost`.

Created the following request:

```
Date: Thu, 09 Oct 2014 14:16:12 +0000
From: Example User <foi+request-125-51208a77@localhost>
To: FOI requests at TGQ <geraldine-requests@localhost>
Message-ID: <ogm-27+543698ac2be57-aa98@localhost>
Subject: Freedom of Information request - How many schools are rated "Very
 Good" in Localhost?
Mime-Version: 1.0
Content-Type: text/plain;
 charset=UTF-8
Content-Transfer-Encoding: 7bit

Dear Geraldine Quango,

How many schools are rated "Very Good" in [REDACTED] between 2013 and 2014?

Yours faithfully,

Example User

-------------------------------------------------------------------

Please use this email address for all replies to this request:
foi+request-125-51208a77@localhost

Requestor details:
===================================================================
# Name: Example User
# Identity Card Number: 123-456789-1234Z
# Domicile: Localhost
# Occupation: Example
# Marital status: Single
# Date of birth: 1980-10-09
===================================================================

Is geraldine-requests@localhost the wrong address for Freedom of Information requests to Geraldine Quango? If so, please contact us using this form:
http://10.10.10.30/change_request/new?body=tgq

Disclaimer: This message and any reply that you make will be published on the internet. Our privacy and copyright policies:
http://10.10.10.30/help/officers

If you find this service useful as an FOI officer, please ask your web manager to link to us from your organisation's FOI page.


-------------------------------------------------------------------
```

`Localhost` is not redacted in the request title.

![screen shot 2014-10-09 at 15 19 10](https://cloud.githubusercontent.com/assets/282788/4577789/97bba508-4fbf-11e4-8a91-897911eae6bc.png)

Issue #48, opened 2011-06-08
Correcting a typo in a URL ought not to break links to the uncorrected one


robinhouston:
Sometimes a word is mistyped in the title of a request, and if we correct it the URL changes, potentially breaking any links to the mistyped version.

RichardTaylor:
Any implementation of this would need to leave open the option of removing defamatory content (or other content we don't want to publish) from a request title and URL.

HelenWDTK:
This issue has been automatically closed due to a lack of discussion or resolution for over 12 months.
Should we decide to revisit this issue in the future, it can be reopened.

PR #6264, opened 2021-05-21
[#1881] Apply censor rules to InfoRequest#title

garethrees:
Seeing if this passes…

Spike on https://github.com/mysociety/alaveteli/issues/1881.

## Implementation notes / Screenshots

The problem with this as-is is that the redaction gets applied when we render the edit form. On save, this will re-write the `title` attribute and generate a new `url_title`. I haven't checked yet, but I imagine this would happen on any `save` call too.

I think this needs some version of https://github.com/mysociety/alaveteli/issues/48 where we partially detach `title` from `url_title`.

Screenshot 2021-05-24 at 09 31 30

### DIFF
--- a/app/models/censor_rule.rb
+++ b/app/models/censor_rule.rb
@@ -46,6 +46,7 @@ class CensorRule < ApplicationRecord
 
   def apply_to_text(text_to_censor)
     return nil if text_to_censor.nil?
+    return text_to_censor if replacement.nil?
     text_to_censor.gsub(to_replace('UTF-8'), replacement)
   end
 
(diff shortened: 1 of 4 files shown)


garethrees:
Closing this because the approach needs some work.

I think we need to [detach the redaction process](https://github.com/mysociety/alaveteli/pull/2787#issue-44814412) from the censorable model.

How would that look? Probably something like:

```ruby
# Very much a sketch just to get the gist across;
# Lots more thinking needed on the specific design
class InfoRequest::PublicDisplay
  def title
    with_redactions(@info_request.title)
  end

  def with_redactions(attr)
    @info_request.censor_rules.apply(attr)
  end
end

class SomeController
  def show
    @info_request = InfoRequest::PublicDisplay.new(InfoRequest.find(params[:id]))
  end
end
```

Issue #6017, opened 2020-12-10
Collapse large option sets in refusal advice Q&A


garethrees:
There are lots of exemptions!

![Screenshot 2020-12-10 at 15 03 23](https://user-images.githubusercontent.com/282788/101789205-e8fb0f00-3af8-11eb-9555-d624c2a89350.png)

I think we should:

* Order by most common in the YAML file
* Only show the top 6 by default
* Have a "Show more" link to reveal the remainder
* Always show all of them if one outside the top 6 is pre-selected

wrightmartin:
> Always show all of them if one outside the top 6 is pre-selected

When one 'below the line' is selected by the user should we allow them to hide again and keep it pre-selected?

Also, if we're doing this we should definitely consider moving away from the three columns to a single column of options, there's plenty of evidence that a single scannable list will be easier to comprehend than the grid approach

zarino:
+1 for no longer being in columns – that was just a temporary hack from me during the first prototypes. Worked ok when there were three or four options, but deffo not any more.

garethrees:
> When one 'below the line' is selected by the user should we allow them to hide again and keep it pre-selected?

Yeah, it doesn't need to be that clever. I just think we want a default of not bombarding users with ~30 checkboxes. The [blue "Learning and assessment" section here](https://www.cardiff.ac.uk/study/postgraduate/taught/courses/course/advanced-chemistry-msc) is basically what I was thinking.

> +1 for no longer being in columns

I don't disagree, but there's some logic in the [form builder](https://github.com/mysociety/alaveteli/blob/develop/app/helpers/refusal_advice_question_form.rb) that would need to change, and we don't have spare programmer capacity for all the things we're doing right now.

I don't want us to get held up on this, so let's create a separate issue for this suggestion. The "too many options" problem needs solving irrespective of whether we use a column or grid, so let's solve that first.

garethrees:
> changes in whatdotheyknow-theme or alaveteli?
>
> – @wrightmartin (slack)


Alaveteli plz! It should “Just work” for any theme that renders the refusal advice.

garethrees:
> Order by most common in the YAML file

Here's a sorted version of the top exemptions identified on the 10,000 most recent requests (Oct 2020) https://github.com/mysociety/alaveteli/issues/5912#issuecomment-717186502

```ruby
 {"Section 12"=>371,
 "Section 1"=>297,
 "Section 14"=>288,
 "Section 17"=>279,
[...]

PR #6272, opened 2021-05-28
[#6017] collapse large option sets

wrightmartin:
## Relevant issue(s)

Part of https://github.com/mysociety/alaveteli/issues/6017

## What does this do?
Collapses what can be a large number of options down so you will only initially see the most relevant 

## Why was this needed?
Too many options displayed at once was making the UI too overwhelming

## Implementation notes
Can be tweak to show more or less options by changing the `max-height` property

## Screenshots

Screenshot 2021-05-28 at 13 40 20
image

### DIFF
--- a/app/assets/javascripts/wizard.js
+++ b/app/assets/javascripts/wizard.js
@@ -312,4 +312,49 @@
   };
 
   $(".js-wizard").refusalWizard({ debug: false });
+
+
+  // Hide/reveal on long list of exemption checkboxes
+
+  function shouldHideCheckboxes () {
+    var checkboxCount = $('[data-block="exemption"] input[type="checkbox"]').length;
+
+    // 4 seems about right for small and large screens
+    if(checkboxCount >= 4 ) {
+      return true;
+    }
+  }
+
+  function expandFieldset() {
+    // fetch translated strings from the HTML
+    var textExpand = $('.wizard').data('expand-button-text-expand');
+    var textShrink = $('.wizard').data('expand-button-text-shrink');
+    
+    // Hide the long list now we're sure we've got JS
+    $('[data-block="exemption"]').toggleClass('expanded');
+   
+    if ($('.maximise-questions').text() == textExpand ) {
+      $('.maximise-questions').html(textShrink);
+    } else {
+      $('.maximise-questions').html(textExpand);
+    }
+  }
+
+
+  if(shouldHideCheckboxes()) {
+
+    $('[data-block="exemption"] fieldset').after("<button class='button-secondary maximise-questions'>" + $('.wizard').data('expand-button-text-expand') + "</button>");
+    
+    //if a checkbox is pre-selected, expand the list
+    if ($('[data-block="exemption"] input[type="checkbox"]').is(":checked")) {
+      expandFieldset();
+    }
+  }
+
+  $('.maximise-questions').click(function(e){
+    e.preventDefault();
+    expandFieldset();
+  });
+  
+
 })(window.jQuery);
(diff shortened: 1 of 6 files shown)


app/assets/stylesheets/responsive/_wizard_layout.scss
```diff
@@ -63,3 +63,32 @@
...
+  [data-block="exemption"].expanded fieldset {
+    /*
+    * We're toggling the .expanded class with JS
+    */
+    max-height: 50em;
+    transition: max-height 0.2s ease-out;
+  }
+
+  .maximise-questions {
+    position: relative;
```
garethrees, line 89: Feels a little disconnected from the fieldset. Any way we can bring it vertically a bit closer? "Not easily" is a perfectly fine answer.

app/assets/stylesheets/responsive/_wizard_layout.scss
```diff
@@ -63,3 +63,32 @@
...
+    left: 50%;
+    transform: translateX(-50%);
+    border: 0;
+  }
+}
```
garethrees, line 94: Needs a newline at the end of the file. https://thoughtbot.com/blog/no-newline-at-end-of-file#following-the-rules-in-your-editor might be useful.

app/assets/javascripts/wizard.js
```diff
@@ -312,4 +312,19 @@
   };
 
   $(".js-wizard").refusalWizard({ debug: false });
+
+
+  $('[data-block="exemption"] fieldset').after("<button class='button-secondary maximise-questions'>Show more</button>");
```
garethrees, line 317: Any user-visible text data-attributes on the HTML so that we can translate them. Should be existing patterns to follow for this.

garethrees (changes requested):
Looks good in principle, but misses a critical piece:

> Always show all of them if one outside the top 6 is pre-selected

At the moment this doesn't happen. Let's say someone get's a "Section 43" refusal:

Screenshot 2021-05-28 at 16 18 44

When they click through to get help (which takes them to `/help/unhappy`), because we've identified the exemption, we automatically tick it in the refusal advice wizard. Currently they can't see that though:

Screenshot 2021-05-28 at 16 21 19

In this situation the collapse should have been _open_ so that they could see the pre-checked field.

app/assets/javascripts/wizard.js
wrightmartin: Ah, yep, makes sense. I've had a browse around the codebase but can't find a pattern to follow, would you mind pointing me towards one? Cheers

app/assets/stylesheets/responsive/_wizard_layout.scss
wrightmartin: Thanks, will add this

app/assets/javascripts/wizard.js
garethrees: Ah, sorry, this was no where as obvious as I thought it was. Here are a couple:

* https://github.com/mysociety/alaveteli/blob/3add3721d0dc2d2011d63127c2f51fc637c5e34d/app/views/request/_after_actions.html.erb#L89
* https://github.com/mysociety/alaveteli/blob/3add3721d0dc2d2011d63127c2f51fc637c5e34d/app/views/alaveteli_pro/batch_request_authority_searches/index.html.erb#L51-L54
* https://github.com/mysociety/alaveteli/blob/3add3721d0dc2d2011d63127c2f51fc637c5e34d/app/views/alaveteli_pro/info_requests/_select_authority_form.html.erb#L13

wrightmartin:
I've pushed a new version of this with all the requested changes

garethrees:
Thanks! Merged as 4195c69a6.

PR #6300, opened 2021-06-10
 Add a back link to request internal review

### LINKED ISSUE #6138: [JRCT] Include a way to go back and change your Refusal Advice wizard answers
I think we should include a way to go back and change your answers along the lines of:

Screenshot 2021-02-23 at 10 17 22

_Originally posted by @garethrees in https://github.com/mysociety/alaveteli/pull/6099#discussion_r580918297_

wrightmartin:
## Relevant issue(s)

Fixes #6138

## What does this do?

Adds a back link to the request internal review page if the user has come from the refusal advice wizard

## Why was this needed?

To allow the user to go back and amend their answers (and to show them that they can)

## Screenshots

image

## Notes to reviewer

Used a `<br />` tag - possibly bad but seemed like the simplest way to add space with little complexity

### DIFF
--- a/app/views/followups/_form_title.html.erb
+++ b/app/views/followups/_form_title.html.erb
@@ -1,3 +1,9 @@
+<% if @refusal_advice.suggested_actions %>
+  <br />
+  <%= link_to _('Go back to refusal advice questions'),
+              help_unhappy_path(url_title: @info_request.url_title) %>
+<% end %>
+
 <% if @internal_review %>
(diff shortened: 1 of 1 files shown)


gbp:
Is there a reason why this is within the `if @internal_review` conditional - should show for follow ups too

wrightmartin:
> Is there a reason why this is within the `if @internal_review` conditional - should show for follow ups too

ah, no, I think I thought that was the only condition we should show it

Issue #5938, opened 2020-10-27
Decide what the refusal advice should present when we don't have JavaScript enabled


garethrees:

garethrees:
We're going to assume this will require JS for now.

We're going to render the HTML through rails, with the JS just dealing with interaction.

This means we'll have to do something to hide the advice when JS is disabled

garethrees:
Maybe just render the actions without suggestions?

wrightmartin:
Ok viewing the page without Javascript gives you this...

![image](https://user-images.githubusercontent.com/2292925/119097273-7037a180-ba0c-11eb-9219-44713100d812.png)

There's two approaches

**Build a JS-free version**

This would be a multistage form that works just like the JS driven one but uses traditional GET/POST to receive answers. We could then use JS to enhance this and make it work similar to how it works now. This is a big undertaking and probably a rewrite, however the bulk of the work of this feature, the logic and the content can be reused so not unachievable. We should put this on the roadmap. 

**Remove the questionnaire and adjust the next steps display**

This pretty much what we have now. I'm thinking not showing the advice based on your questions is fine, as arguably an accessible alternative is to read through them and determine what applies based on the advice given. We should fix the `wizard__next-steps` controls to show by default and only hide when JavaScript is available.

wrightmartin:
Actually on reflection this won't be enough as we've got the snippets to deal with too....

I think for now the non-JS version should work as everything did previously for now. We hide the questionnaire and the `wizard__next-steps` links completely from non-JS clients

garethrees:
In https://github.com/mysociety/whatdotheyknow-theme/pull/817 I've added some text which says "use our wizard". We _could_ render that with JS, or, maybe we should just render a box in place of the wizard saying "JavaScript must be enabled, sorry".

wrightmartin:
> "JavaScript must be enabled, sorry".

Never sure about these 'look at what you could have won' messages as it's rare JS is disabled as a choice they can easily revert

garethrees:
Fixed by https://github.com/mysociety/alaveteli/pull/6306 and https://github.com/mysociety/whatdotheyknow-theme/pull/827.

PR #6306, opened 2021-06-10
Hide wizard content from non js users

wrightmartin:
## Relevant issue(s)

#5938 

## What does this do?

Hides the wizard and wizard content from clients without javascript

## Why was this needed?

The wizard is (currently) JS dependant so won't work with JS disabled or where the JS files haven't loaded

### DIFF
--- a/app/assets/stylesheets/responsive/_wizard_layout.scss
+++ b/app/assets/stylesheets/responsive/_wizard_layout.scss
@@ -1,5 +1,13 @@
 .wizard {
   margin: 2rem 0;
+  display: block;
+}
+
+.no-js {
+  .wizard {
+    display: none;
+    visibility: hidden;
+  }
 }
 
 // Hide questions by default.


garethrees:
@wrightmartin got any screenshots for the PR so that I know what I'm expecting to see?

garethrees (changes requested):
I've had a quick look at this. I don't know if I'm missing something vs what you intended, but I'm not sure this does anything differently to the behaviour prior to the change?

I can see two issues at the moment that this change doesn't cover.

1. We say "Hey, use our refusals wizard" below, which won't be there.
2. We say "Get help with refusals" and render nothing. That doesn't feel great.

Screenshot 2021-06-10 at 15 27 36

wrightmartin:
ah sorry for the confusion, there's a WDTK PR that comes along with it that hides that content, I've just found the PR sat in an open tab not submitted

gbp (approved):

PR #6328, opened 2021-06-28
Remove ignored dependencies

gbp:
## What does this do?

Remove ignored dependencies

## Why was this needed?

Now we've dropped support for Ruby 2.3 and 2.4 these dependencies can be
bumped.

## Implementation notes

This seems a good time to do this now 0.39 has been released.


garethrees (approved):

gbp:
Merged 91fa962fd079f7695cff1aba6ab908d4b2b8b763

PR #6332, opened 2021-06-30
New response reminders: Add request title to subject

### LINKED ISSUE #6329: Change subject of email titled "Please update the status of your request"
In my email client all emails with this title are bundled up into the same message thread, even though they concern different requests. Other email clients may behave differently.

Perhaps it would be better to introduce the request title into the email subject? This way it will be a separate thread, and also perhaps make the title better clarify what subject the email refers to. For example: "Please update the status of your request: 'Mortgage Rescue Scheme'" is more clear than "Please update the status of your request"

itsaphel:
Resolves #6329


itsaphel:
Uhh, will split those two commits into separate branches...

PR #6334, opened 2021-06-30
Add Mailcatcher to Docker

itsaphel:
Install mailcatcher's gem, and run it when starting Docker app. Expose mailcatcher's web server port. Was helpful for a development install, possibly should be done by default (unless the Docker image has non-development uses as well?)


gbp (approved):
Thanks for this, it seems to work well.

gbp:
Merged 32424169cf290837eff8ce9b09af0ae0b88cad89

PR #6337, opened 2021-06-30
Extract Statistics::General from GeneralController

garethrees:
## What does this do?

Reduce complexity of controller (and its tests) by extracting logic to a
more appropriate model.

Allows us to remove a non-controller method from ApplicationController.

Also allows easier access from the command line:

    bin/rails runner "pp Statistics::General.new.to_h"

## Why was this needed?

Just general cleanup that looked quick and easy while working on https://github.com/mysociety/alaveteli-project/issues/135.

## Implementation notes

It's a heavily coupled class, but I'm not sure there's a great way around that.

## Screenshots

Still appears to work 😅 

Screenshot 2021-06-30 at 18 59 48

Screenshot 2021-06-30 at 19 01 26

## Notes to reviewer

At some point we should migrate the current `Statistics` class to e.g. `Statistics::PublicBodies` and `Statistics::Users` or something like that.

### DIFF
--- a/app/controllers/admin_general_controller.rb
+++ b/app/controllers/admin_general_controller.rb
@@ -156,7 +156,7 @@ class AdminGeneralController < AdminController
 
   def debug
     @admin_current_user = admin_current_user
-    @current_commit = alaveteli_git_commit
+    @current_commit = Statistics::General.new.to_h[:alaveteli_git_commit]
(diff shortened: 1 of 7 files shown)


gbp (approved):
We could resolve the new TODO comments, `s/class Statistics/module Statistics/` should be enough, shouldn't it?

Issue #6339, opened 2021-07-01
Remove legacy batch UI 


garethrees:
Reduce complexity and maintenance.

List of things to remove:

- [ ] `User#can_make_batch_requests`
- [ ] `RequestController#new_batch`
- [ ] Lots of conditionals from `app/views/request/new.html.erb`
- [ ] `ALLOW_BATCH_REQUESTS` config

PR #4084, opened 2017-07-04
Remove unconventional PublicBody constraints and defaults

garethrees:
- [x] Changelog notes
- [ ] Upgrade notes

PR #6355, opened 2021-07-07
[#6322] Fix info request law used

### LINKED ISSUE #6322: 'eir_only' tag does not appear to change legislation on a request
WDTK Body 92627 - Electricity North West Limited

This body has the tag `eir_only` yet all but the first request (704413) show
`law used: foi`

This means the public request page mentions FOI and not EIR. Requests with this issue are 716316, 716317, 716318, 738434, 749218. This may also affect advice given on requests re timings / exemptions / next actions etc.

We cannot see from the admin page when the `eir_only` tag was added but believe it to be from when the body was first created (03/03/2020).

This seems to be across all bodies tagged `eir_only` not just ENWL, I've also seen this at request 757169 to body 110962 (Abellio Scotrail Limited).

Labelling as a bug as I believe something has stopped the legislation change working.

gbp:
## Relevant issue(s)

Fixes #6322

## What does this do?

Reworks how the `InfoRequest#law_used` attribute is set. 

## Why was this needed?

Requests to authorities which are EIR only had `law_used` set to `foi` this was coming from the database column default value.

### DIFF
--- a/app/models/info_request.rb
+++ b/app/models/info_request.rb
@@ -1,6 +1,6 @@
 # -*- encoding : utf-8 -*-
 # == Schema Information
-# Schema version: 20210114161442
+# Schema version: 20210706195253
 #
 # Table name: info_requests
 #
@@ -14,7 +14,7 @@
 #  awaiting_description                  :boolean          default("false"), not null
 #  prominence                            :string           default("normal"), not null
 #  url_title                             :text             not null
-#  law_used                              :string           default("foi"), not null
+#  law_used                              :string           not null
 #  allow_new_responses_from              :string           default("anybody"), not null
 #  handle_rejected_responses             :string           default("bounce"), not null
 #  idhash                                :string           not null
--- a/app/models/info_request.rb
+++ b/app/models/info_request.rb
@@ -165,7 +165,7 @@ class InfoRequest < ApplicationRecord
   validate :must_be_valid_state
   validates_inclusion_of :prominence, :in => Prominence::VALUES
 
-  validates_inclusion_of :law_used, in: Legislation.keys
+  validates_inclusion_of :law_used, in: Legislation.keys, allow_nil: true
 
   # who can send new responses
   validates_inclusion_of :allow_new_responses_from, :in => [
--- a/app/models/info_request.rb
+++ b/app/models/info_request.rb
@@ -185,7 +185,7 @@ class InfoRequest < ApplicationRecord
   after_destroy :update_counter_cache
   after_update :reindex_some_request_events
   before_destroy :expire
-  before_validation :compute_idhash
+  before_validation :compute_idhash, :set_law_used
   before_create :set_use_notifications
 
   # Return info request corresponding to an incoming email address, or nil if
@@ -1872,8 +1872,10 @@ class InfoRequest < ApplicationRecord
       # this should only happen on Model.exists? call. It can be safely ignored.
       # See http://www.tatvartha.com/2011/03/activerecordmissingattributeerror-missing-attribute-a-bug-or-a-features/
     end
+  end
 
-    self.law_used ||= legislation.key if new_record?
+  def set_law_used
+    self.law_used ||= public_body.legislation.key if public_body
   end
 
   def set_use_notifications
--- a/db/migrate/20210706195253_remove_info_request_law_used_default.rb
+++ b/db/migrate/20210706195253_remove_info_request_law_used_default.rb
@@ -1,0 +1,5 @@
+class RemoveInfoRequestLawUsedDefault < ActiveRecord::Migration[5.2]
+  def change
+    change_column_default :info_requests, :law_used, from: 'foi', to: nil
+  end
+end
--- a/spec/factories/info_requests.rb
+++ b/spec/factories/info_requests.rb
@@ -1,6 +1,6 @@
 # -*- encoding : utf-8 -*-
 # == Schema Information
-# Schema version: 20210114161442
+# Schema version: 20210706195253
 #
 # Table name: info_requests
 #
@@ -14,7 +14,7 @@
 #  awaiting_description                  :boolean          default("false"), not null
 #  prominence                            :string           default("normal"), not null
 #  url_title                             :text             not null
-#  law_used                              :string           default("foi"), not null
+#  law_used                              :string           not null
 #  allow_new_responses_from              :string           default("anybody"), not null
 #  handle_rejected_responses             :string           default("bounce"), not null
 #  idhash                                :string           not null
--- a/spec/fixtures/info_requests.yml
+++ b/spec/fixtures/info_requests.yml
@@ -1,5 +1,5 @@
 # == Schema Information
-# Schema version: 20210114161442
+# Schema version: 20210706195253
 #
 # Table name: info_requests
 #
@@ -13,7 +13,7 @@
 #  awaiting_description                  :boolean          default("false"), not null
 #  prominence                            :string           default("normal"), not null
 #  url_title                             :text             not null
-#  law_used                              :string           default("foi"), not null
+#  law_used                              :string           not null
 #  allow_new_responses_from              :string           default("anybody"), not null
 #  handle_rejected_responses             :string           default("bounce"), not null
 #  idhash                                :string           not null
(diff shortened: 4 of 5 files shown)


app/models/info_request.rb
```diff
@@ -14,7 +14,7 @@
 #  awaiting_description                  :boolean          default("false"), not null
 #  prominence                            :string           default("normal"), not null
 #  url_title                             :text             not null
-#  law_used                              :string           default("foi"), not null
+#  law_used                              :string           not null
```
garethrees, line 17: This commit ought to explain _why_ the column default "prevents the correct `law_used` from being assigned when the model is initialised." I have some suggestions for later commits, but they _may_ be based on an incorrect understanding of why this was necessary. It's also good to include for future understanding.

app/models/info_request.rb
```diff
@@ -165,7 +165,7 @@ class << self
   validate :must_be_valid_state
   validates_inclusion_of :prominence, :in => Prominence::VALUES
 
-  validates_inclusion_of :law_used, in: Legislation.keys
+  validates_inclusion_of :law_used, in: Legislation.keys, allow_nil: true
```
garethrees, line 168: I don't think we want to allow a `nil` or `blank` value to be saved.

We have the database constraint of course, which the model validations should mirror. If we don't do this, there's a possibility of the app raising a `PG::NotNullViolation` instead of gracefully handling the issue.

app/models/info_request.rb
```diff
@@ -185,7 +185,7 @@ class << self
   after_destroy :update_counter_cache
   after_update :reindex_some_request_events
   before_destroy :expire
-  before_validation :compute_idhash
+  before_validation :compute_idhash, :set_law_used
```
garethrees, line 188: It would make more sense conceptually to use an [association callback](https://guides.rubyonrails.org/v5.2/association_basics.html#association-callbacks), since this is closer to what we _mean_ to happen – when we update the associated `PublicBody`, make sure the `InfoRequest` is being made under the correct `Legislation`.

I think this would resolve the comments on the previous two commits. We'd be able to set a default (since we know it will get updated correctly when the body is assigned) and we can retain the existing validations that mirror the database constraints, since we no longer need to support a `nil` `law_used` attribute.

The thing to consider here is what should happen if a request is moved from, say, an FOI-able body to an EIR-only body. Something like the following might work?

```ruby
before_add: [:check_law_used_supported, :update_law_used]

def check_law_used_supported(public_body)
  return true if new_record?
  raise SomeError unless public_body.covered_by?(legislation)
end

def update_law_used(public_body)
  self.law_used = public_body.legislation
end
```

I don't think I'd worry too much about handling the exception in the admin UI at this point – if it happens a lot then we can ticket up and do something.

app/models/info_request.rb
gbp: Once #6339 this can be reverted as this is to allow specs to pass

app/models/info_request.rb
gbp: Rails doesn't support association callbacks on `belongs_to` associations (see https://github.com/rails/rails/issues/586). We could override the `#public_body=` if you prefer but a before validation callback seemed simpler.

Currently when moving a request to a different public body we update the `#law_used` https://github.com/mysociety/alaveteli/blob/b1d999103889f6754d78d49e92cfad41e3302017/app/models/info_request.rb#L1620-L1622

This PR doesn't change this behaviour as its not really within scope but I don't think we should be updating `#law_used` as once the request has been created the application wouldn't know if the user wrote a FOI or EIR request. Its entirely possible a user would attempt to send a FOI to an EIR-only body.

We could even lose the correct law used when moving, suppose a EIR request was written to a FOI&EIR public body, the request moved to another FOI&EIR body - I think the law used would now be set to `foi`. If you agree I'll open a ticket for this -  the solution could when look a bit like your `#check_law_used_supported`.

app/models/info_request.rb
gbp: The specs this fixes are:

```
Failures:

  1) RequestController#new_batch when batch requests is enabled when the current user can make batch requests should render 'preview' when given a good title and body
     Failure/Error: expect(response).to render_template('preview')
       expecting <"preview"> but rendering with <["request/new", "layouts/default"]>
     # ./spec/controllers/request_controller_spec.rb:2008:in `block (4 levels) in <top (required)>'
     # ./.direnv/gems/2.6.7p197/gems/webmock-3.13.0/lib/webmock/rspec.rb:37:in `block (2 levels) in <top (required)>'

  2) RequestController#new_batch when batch requests is enabled when the current user can make batch requests on success sets the batch_sent flash to true
     Failure/Error: expect(flash[:batch_sent]).to be true

       expected true
            got nil
     # ./spec/controllers/request_controller_spec.rb:2059:in `block (5 levels) in <top (required)>'
     # ./.direnv/gems/2.6.7p197/gems/webmock-3.13.0/lib/webmock/rspec.rb:37:in `block (2 levels) in <top (required)>'

  3) RequestController#new_batch when batch requests is enabled when the current user can make batch requests on success should create an info request batch and redirect to the new batch on success
     Failure/Error: expect(new_info_request_batch).not_to be_nil

       expected: not nil
            got: nil
     # ./spec/controllers/request_controller_spec.rb:2045:in `block (5 levels) in <top (required)>'
     # ./.direnv/gems/2.6.7p197/gems/webmock-3.13.0/lib/webmock/rspec.rb:37:in `block (2 levels) in <top (required)>'

  4) RequestController#new_batch when batch requests is enabled when the current user can make batch requests on success should prevent double submission of a batch request
     Failure/Error: expect(assigns[:existing_batch]).not_to be_nil

       expected: not nil
            got: nil
     # ./spec/controllers/request_controller_spec.rb:2054:in `block (5 levels) in <top (required)>'
     # ./.direnv/gems/2.6.7p197/gems/webmock-3.13.0/lib/webmock/rspec.rb:37:in `block (2 levels) in <top (required)>'
```

What is happening is this conditional is true since teh `@info_request` has no law used since the default value was removed in the previous commit in this PR
https://github.com/mysociety/alaveteli/blob/b1d999103889f6754d78d49e92cfad41e3302017/app/controllers/request_controller.rb#L245-L252

app/models/info_request.rb
gbp: Will update to include:
> The `#law_used` is assigned using the conditional assignment operator so with this default the value can never be set to the public body legislation key.

garethrees (commented):
We had a bit of a chat about this today. Main points:

* It's not the end of the world if we merge it as-is.
* Sorry, I didn't realise `belongs_to` doesn't have a `before_add` – that's annoying! Using `before_validation` is fine, since that's also close to what we mean to happen ("before we validate this record, check to see that it's as valid as we can make it").
* I think we should introduce a spec which fails on `develop`. The actual problem here is that in practice, we initialise a new `InfoRequest` and at some later point add the `PublicBody` to it, rather than at the point of initialisation. At that point `law_used` already has the default value, so is never updated.
* Extracting the setting of `law_used` to `set_law_used` is nice – that gives us a bit more flexibility.
* I agree on not changing behaviour on moving requests to a different body – let's ticket up anything which needs looking at after this gets merged.
* It would be preferable not to introduce a situation where validations don't match database constraints. We've put a lot of work in over the years (https://github.com/mysociety/alaveteli/pull/1834, https://github.com/mysociety/alaveteli/pull/4041, https://github.com/mysociety/alaveteli/issues/4032, https://github.com/mysociety/alaveteli/pull/4084, etc) to clean up these types of issues, and it might be (literally) years before we make it to https://github.com/mysociety/alaveteli/issues/6339.
* I think the default value of `"foi"` for `InfoRequest#law_used` is correct – we are _intentionally_ communicating that we think most requests should be made under "FOI" laws in any Alaveteli install, with other legislations being the edge case. I guess we're handling that via `InfoRequest#legislation` as well, which is a little easier to change for us and re-users. I'm not _too_ hung up on this.

I think we can make this change simpler with something like:

```ruby
before_validation :update_law_used, on: :create

def update_law_used
  self.law_used = legislation.key
end
```

AIUI, this _should_ keep setting `"foi"`, unless we then add an `eir_only?` body, at which point we'll set `law_used = "eir"`. This won't update old requests if, say, the body becomes `eir` only later.

gbp:
Superseded by #6385

PR #6356, opened 2021-07-07
Unfold the details tag if they've come to the page with a hash link

wrightmartin:
## Relevant issue(s)
https://github.com/mysociety/whatdotheyknow-theme/issues/832

## What does this do?
Adds functionality whereby if you're visiting the /help/unhappy page with a known url hash eg `#internal_review` it opens the associated `details` element and highlights it with a css fading style

## Why was this needed?
It wasn't clear that the page has the relevant information when it's hidden inside a closed `details` element

## Implementation notes
The JavaScript is hard coded to the known ID tags of the sections - this means if we ever add new ones it will need updating

## Screenshots

image

### DIFF
--- a/app/assets/javascripts/wizard.js
+++ b/app/assets/javascripts/wizard.js
@@ -356,5 +356,17 @@
     expandFieldset();
   });
 
+  // work out what hashes we're looking for
+  var IDs = [];
+  $("#help_unhappy h2[ID]").each(function(){ 
+    IDs.push(this.id); 
+  });
+
+  // show unrolled details tag if they've come to the page with a known location hash
+  var hash = window.location.hash;
+
+  if(IDs.includes(hash.slice(1))) {
+    $(hash + ' + details' ).attr('open', '').addClass('flash');
+  }
 
 })(window.jQuery);
--- a/app/assets/stylesheets/responsive/_wizard_style.scss
+++ b/app/assets/stylesheets/responsive/_wizard_style.scss
@@ -51,3 +51,12 @@
     margin: 0.5em 0 0 0;
   }
 }
+
+.flash {
+    animation: target-fade 3s 1;
+}
+
+@keyframes target-fade {
+    0% { background-color: #F3BD2A; }
+    100% { background-color: transparent; }
+}


app/assets/javascripts/wizard.js
```diff
@@ -355,6 +355,11 @@
     e.preventDefault();
     expandFieldset();
   });
-  
+
+  // show unrolled details tag if they've come to the page with a known location hash
+  var hash = window.location.hash;
+  if(hash == '#other_means' || hash == '#refer-to-ico' || hash == '#internal_review') {
```
garethrees, line 361: This won't work for international reuses where there are different next actions.

Should be possible to look up whether there's a target that matches the `window.location.hash` before adding the flash so that we don't have to hard code them?

wrightmartin: ah, of course. I was going super simple and went too super simple. I'll have a think

wrightmartin:
@garethrees have pushed an update - it now finds all the possible IDs in that div and puts them in an array, then checks the URL against the url instead

garethrees (commented):
I don't know if it's just my ageing machine / being slow in development mode, but here's what I see.

It looks like the fade might be being blocked by the wizard rendering, and when it does happen it's a suuuuper short flash of yellow. Is that intentional? If so, I feel like it could tail out a bit slower.

https://user-images.githubusercontent.com/282788/124793862-531c4b00-df46-11eb-89e3-01ba3f093f7b.mov

wrightmartin:
ooh that looks quite a lot different to what happens on mine - the animation should last 3 seconds `animation: target-fade 3s 1;`

There is a chance it's being blocked by the wizard rendering but I figured that was the correct approach to stop the content from jumping around after the fade

garethrees:
Could you upload a video of what it does for you? Might be worth just merging and seeing what it's like on production when all the assets have been compiled etc.

wrightmartin:
@garethrees I haven't forgotten about this I've just been tied up on other stuff, will look at it before the end of Friday

wrightmartin:
Here it is happening on my local machine
https://user-images.githubusercontent.com/2292925/126763798-3117ddfe-7923-4e1d-be2d-fdb40a33c64e.mov

garethrees:
Cool, I guess let's just merge this as is (squashing the commits into one) then and if it's still slow for some users then reconsider.

garethrees:
Deployed and much quicker in production.

PR #6357, opened 2021-07-07
[#5649] Fix internal email forwarding

### LINKED ISSUE #5649: Mail sent To team@whatdotheyknow.com with Cc pro_team@whatdotheyknow.com does not get delivered to team@
When an email is received that was sent both to a regular support address and a Pro support address (as per the example in the title), the mail handling code will override the regular support address with the pro support address defined in `PRO_CONTACT_EMAIL`.

See [these lines](https://github.com/mysociety/alaveteli/blob/4b02ca7ee892688bcea18cd9a32579d6e3fb5cc1/lib/mail_handler/reply_handler.rb#L108-L111) in `reply_handler.rb`:
```ruby
def self.get_forward_to_address(message)
  forward_to = AlaveteliConfiguration.forward_nonbounce_responses_to
  if AlaveteliConfiguration.enable_alaveteli_pro
    pro_contact_email = AlaveteliConfiguration.pro_contact_email
    original_to = message ? MailHandler.get_all_addresses(message) : []
    if original_to.include?(pro_contact_email)
      forward_to = AlaveteliConfiguration.forward_pro_nonbounce_responses_to
    end
  end
  forward_to
end
```
 The call to `MailHandler.get_all_addresses(message)` in `mail_backend.rb`[returns an array that includes things in CC](https://github.com/mysociety/alaveteli/blob/4b02ca7ee892688bcea18cd9a32579d6e3fb5cc1/lib/mail_handler/backends/mail_backend.rb#L135-L143):
```ruby
def get_all_addresses(mail, include_invalid: false)
  addrs = []
  addrs << mail.to
  addrs << mail[:to].try(:value) if mail.to.nil? && include_invalid
  addrs << mail.cc
  addrs << mail[:cc].try(:value) if mail.cc.nil? && include_invalid
  addrs << (mail['envelope-to'] ? mail['envelope-to'].value.to_s : nil)
  addrs.flatten.compact.uniq
end
```
The first snippet then resets `forward_to` to `AlaveteliConfiguration.forward_pro_nonbounce_responses_to` if `AlaveteliConfiguration.pro_contact_email` is found in the `original_to` array - which it will be in this case.

More details available in [this internal ticket](https://github.com/mysociety/sysadmin/issues/1276).

garethrees:
## Relevant issue(s)

Fixes #5649

## What does this do?

Fix mail forwarding on to both addresses

## Why was this needed?

Mail sent `To: team@` and `Cc: pro_team@` was not getting sent to `team@`, because we were overwriting it.

## Implementation notes

This works by setting multiple recipients when we shell out to `sendmail`.

Prior to this PR, and for most messages, we do `/usr/sbin/sendmail -i team_delivery@example.com`.

As of this PR, when we have a case of `To: team@` and `Cc: pro_team@`, we'll now do `/usr/sbin/sendmail -i "team_delivery@example.com,pro_team_delivery@example.com"`

I literally started 5 branches to try to tidy this up in various ways.

I tried a basic attempt at making `MailHandler::ReplyHandler` a class – there's no reason it should be a bag of functions – but that got complicated because of the interplay between the unparsed email and the parsed `Mail::Message`. It's certainly resolvable, but the lack of existing tests means its a huge amount of work to be confident in the refactoring.

`MailHandler::ReplyHandler.get_forward_to_address` is complicated enough to be a class in its own right. I tried extracting it to `MailHandler::ForwardOnRecipients`, but because we're trying to not use rails we can't use  the RSpec integration of `feature_enabled?` to test whether Pro is enabled. Using `AlaveteliConfiguration` made the specs way more complicated.

I did at least manage to shuffle around _some_ of the complexity of `script/handle-mail-replies.rb` itself by extracting some methods to better communicate each return clause, but it still feels like all this is uber procedural and, as a result, overcomplicated.

## Notes to reviewer

Need to actually check that this works IRL.

### DIFF
--- a/lib/mail_handler/reply_handler.rb
+++ b/lib/mail_handler/reply_handler.rb
@@ -103,13 +103,16 @@ module MailHandler
 
     def self.get_forward_to_address(message)
       forward_to = AlaveteliConfiguration.forward_nonbounce_responses_to
+
       if AlaveteliConfiguration.enable_alaveteli_pro
         pro_contact_email = AlaveteliConfiguration.pro_contact_email
         original_to = message ? MailHandler.get_all_addresses(message) : []
+
         if original_to.include?(pro_contact_email)
           forward_to = AlaveteliConfiguration.forward_pro_nonbounce_responses_to
         end
       end
+
       forward_to
     end
 
--- a/lib/mail_handler/reply_handler.rb
+++ b/lib/mail_handler/reply_handler.rb
@@ -95,25 +95,49 @@ module MailHandler
 
     def self.forward_on(raw_message, message = nil)
       forward_to = self.get_forward_to_address(message)
-      IO.popen("/usr/sbin/sendmail -i #{forward_to}", "wb") do |f|
+      IO.popen(%Q(/usr/sbin/sendmail -i "#{forward_to}"), 'wb') do |f|
         f.write(raw_message);
         f.close;
       end
     end
 
     def self.get_forward_to_address(message)
-      forward_to = AlaveteliConfiguration.forward_nonbounce_responses_to
+      non_default_recipient(message) || default_recipient
+    end
 
+    def self.non_default_recipient(message)
       if AlaveteliConfiguration.enable_alaveteli_pro
-        pro_contact_email = AlaveteliConfiguration.pro_contact_email
-        original_to = message ? MailHandler.get_all_addresses(message) : []
-
-        if original_to.include?(pro_contact_email)
-          forward_to = AlaveteliConfiguration.forward_pro_nonbounce_responses_to
+        if addressed_to_both_contacts?(message)
+          [default_recipient, pro_recipient].join(',')
+        elsif addressed_to_pro_contact?(message)
+          pro_recipient
         end
       end
+    end
 
-      forward_to
+    def self.addressed_to_pro_contact?(message)
+      pro_contact_email = AlaveteliConfiguration.pro_contact_email
+      original_recipients(message).include?(pro_contact_email)
+    end
+
+    def self.addressed_to_both_contacts?(message)
+      contact_email = AlaveteliConfiguration.contact_email
+      pro_contact_email = AlaveteliConfiguration.pro_contact_email
+
+      original_recipients(message).include?(contact_email) &&
+        original_recipients(message).include?(pro_contact_email)
+    end
+
+    def self.default_recipient
+      AlaveteliConfiguration.forward_nonbounce_responses_to
+    end
+
+    def self.pro_recipient
+      AlaveteliConfiguration.forward_pro_nonbounce_responses_to
+    end
+
+    def self.original_recipients(message)
+      message ? MailHandler.get_all_addresses(message) : []
     end
 
     def self.load_rails
--- a/script/handle-mail-replies.rb
+++ b/script/handle-mail-replies.rb
@@ -34,9 +34,40 @@ require 'reply_handler'
 # the default encoding for IO is utf-8, and we use utf-8 internally
 Encoding.default_external = Encoding.default_internal = Encoding::UTF_8
 
+def content_type(message)
+  MailHandler.get_content_type(message)
+end
+
+def temporary_failure?(message)
+  content_type(message) == 'multipart/report'
+end
+
+def delivery_delay?(message)
+  subject = MailHandler.get_header_string('Subject', message)
+
+  content_type(message) == 'multipart/mixed' &&
+    subject == 'Delivery Status Notification (Delay)'
+end
+
+def record_permanently_failed(message, raw_message, in_test_mode)
+  permanently_failed_addresses =
+    MailHandler::ReplyHandler.permanently_failed_addresses(message)
+
+  permanently_failed_addresses.each do |pfa|
+    if in_test_mode
+      puts pfa
+    else
+      MailHandler::ReplyHandler.record_bounce(pfa, raw_message)
+    end
+  end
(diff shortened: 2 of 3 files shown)


garethrees:
> This works by setting multiple recipients when we shell out to `sendmail`.
> 
> Prior to this PR, and for most messages, we do `/usr/sbin/sendmail -i team_delivery@example.com`.
> 
> As of this PR, when we have a case of `To: team@` and `Cc: pro_team@`, we'll now do `/usr/sbin/sendmail -i "team_delivery@example.com,pro_team_delivery@example.com"`

@sagepe could you just sense check this?

* It [seems like](https://stackoverflow.com/a/13390926) this is a valid `sendmail` argument?
* Does this mean that instead of `To:` and `Cc:`, we'll end up with both addresses in the `To:`?
* Can you think of an easy way of testing that this does what we want?

lib/mail_handler/reply_handler.rb
```diff
@@ -96,24 +96,46 @@ def self.is_oof?(message)
...
+
       if AlaveteliConfiguration.enable_alaveteli_pro
-        pro_contact_email = AlaveteliConfiguration.pro_contact_email
-        original_to = message ? MailHandler.get_all_addresses(message) : []
-        if original_to.include?(pro_contact_email)
+        if addressed_to_both_contacts?(message)
+          forward_to = [
+            forward_to,
+            AlaveteliConfiguration.forward_pro_nonbounce_responses_to
+          ].join(',')
```
gbp, line 113: Would extract `AlaveteliConfiguration.forward_pro_nonbounce_responses_to` into a variable and just `return` within this conditional

```suggestion
          return [forward_to, forward_pro_to].join(',')
```

garethrees: Cleaned up in 4426991

sagepe:
> > This works by setting multiple recipients when we shell out to `sendmail`.
> > Prior to this PR, and for most messages, we do `/usr/sbin/sendmail -i team_delivery@example.com`.
> > As of this PR, when we have a case of `To: team@` and `Cc: pro_team@`, we'll now do `/usr/sbin/sendmail -i "team_delivery@example.com,pro_team_delivery@example.com"`
> 
> @sagepe could you just sense check this?
> 
>  * It [seems like](https://stackoverflow.com/a/13390926) this is a valid `sendmail` argument?

Yes. As per the `man` page: "...if there is at least one command line argument, `-bm` (accept a local message on the standard input, with  the  arguments specifying  the  recipients) is assumed... The recipients are given as the command arguments... Each argument can be a comma-separated list of RFC 2822 addresses."

>  * Does this mean that instead of `To:` and `Cc:`, we'll end up with both addresses in the `To:`?

Yes, that is my understanding. One way to change this behaviour would be to include them as headers body of the response rather than passing as arguments and invoke with `-t`.

>  * Can you think of an easy way of testing that this does what we want?

There's actually an example of doing this interactively on the `man` page; to paraphrase:

```
sendmail -i address1@domain.com,address2@domain.com
<message content, including all the header lines>
CTRL-D
```

sagepe:
Just to confirm then that any headers included in a message sent using the above example will be respected, so:

```
sendmail -i address1@domain.com,address2@domain.com
Subject: Test message
To: address1@domain.com
Cc: address2@domain.com

This is a test To: address1@domain.com and Cc: address2@domain.com
^D
```

Does what you might expect it would do. Note that omitting the To: and Cc: headers results in message delivery without these included at all, which may not be what you want.

garethrees:
Also tested this like…

```
This is a test message. Sent with "sendmail -i team_delivery@whatdotheyknow.com,pro_team_delivery@whatdotheyknow.com" on the command line, the body of the message included a To: header with team@whatdotheyknow.com and a Cc: header with pro_team@whatdotheyknow.com.
```

…and we received the message as expected:

Screenshot 2021-07-20 at 12 50 18

lib/mail_handler/reply_handler.rb
```diff
@@ -96,22 +96,50 @@ def self.is_oof?(message)
...
+          elsif addressed_to_pro_contact?(message)
+            pro_recipient
+          end
+
+        return non_default_recipient if non_default_recipient
       end
-      forward_to
+
+      default_recipient
+    end
```
garethrees, line 118: I suppose we could make the top-level method (`get_forward_to_address`) a little simpler, at the expense of more class methods passing the `message` around. This should really be a class that encapsulates `message` rather than a module.

```ruby
def self.get_forward_to_address(message)
  non_default_recipient(message) || default_recipient
end

def self.non_default_recipient(message)
  if AlaveteliConfiguration.enable_alaveteli_pro
    if addressed_to_both_contacts?(message)
      [default_recipient, pro_recipient].join(',')
    elsif addressed_to_pro_contact?(message)
      pro_recipient
    end
  end
end

def self.default_recipient
  AlaveteliConfiguration.forward_nonbounce_responses_to
end
```

sagepe (approved):
Based on the fact that the sendmail integration has been tested and behaves as desired, I'm approving this PR.

garethrees:
> Based on the fact that the sendmail integration has been tested and behaves as desired, I'm approving this PR.

Thanks!

I've squashed this together and added my [suggestion](https://github.com/mysociety/alaveteli/pull/6357#pullrequestreview-710489979) to make `get_forward_to_address` even clearer in intent. I think just a quick review from @gbp now that we've confirmed the principle on the infra side would be great and then we can get this merged.

gbp (approved):
LGTM

PR #6373, opened 2021-07-14
Add local bin directory to test ENV utility path

gbp:
This is needed so I can run tests locally without failures.

Instead of using system packages or homebrew I provision utilities using
nix-shell. Each utility is installed in its own directory and these
paths might change over time:
```
  /nix/store/9vv5556b9w8kjl71m1iyzg25jfpwgllx-pdftk-3.2.1/bin/pdftk
  /nix/store/15xkf5bdyw2cw39ymqz121dh9xskk697-poppler-utils-21.06.1/bin/pdftohtml
  ...
```

This means they can't be found in the default utility search paths.
Using nix-shell's shellHook option I can symlink utilities into the
local bin directory:
```
  ln -sf ${pdftk}/bin/pdftk bin/pdftk
  ln -sf ${poppler_utils}/bin/pdftohtml bin/pdftohtml
  ...
```

And these symlinks are ignored by adding them to `.git/info/exclude`.

This change makes it possible to provision and find utilities into the
local bin directory so test pass. It does requires a change to how
`pdftohtml` and `unrtf` are called. Previously we `cd` into the `tmp`
directory. This means the utility still wouldn't be found as its would
looking in `tmp/bin` for them.

### DIFF
--- a/config/test.yml
+++ b/config/test.yml
@@ -122,3 +122,8 @@ MAX_REQUESTS_PER_USER_PER_DAY: 2
 
 
 SKIP_ADMIN_AUTH: true
+
+UTILITY_SEARCH_PATH:
+  - ./bin
+  - /usr/bin
+  - /usr/local/bin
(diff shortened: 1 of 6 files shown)


garethrees (approved):
Seems to all work okay!

PR #6393, opened 2021-07-27
[#1292] Absolute Xapian path

### LINKED ISSUE #1292: acts_as_xapian: make xapianbs location configurable, and move on our servers
At the moment, the location of the xapian databases is hard-coded to be `lib/acts_as_xapian/xapiandbs`, which creates various problems. For example, it means that:
- The xapian databases have to be kept under the repository tree, which is awkward for deployments - for example, extra care (or manipulation of symlinks) would have to be done if deploying to one directory which is swapped into place by a symlink - I consider that this issue is blocking #1289
- That path is git-ignored, which means that in certain situations git may treat it as disposable, since the ignore mechanism is really intended for build products. This is potentially dangerous with something so important as the xapian database.

For example, on our systems, it would be better to have the databases at:

```
/var/www/some.host/xapiandbs/
```

... instead of:

```
/var/www/some.host/alaveteli/lib/acts_as_xapian/xapiandbs
```

(I think @crowbot said that this was something @dracos had suggested in the past.) To change this creates some issues, though:
- What's the right way of making the location of the xapian databases configurable, ideally still maintaining some separation of acts_as_xapian from Alaveteli's configuration system?
- How do we safely move the xapiandbs directory when the location is changed in the configuration? We automatically moved the xapian database with the last change of location, but that was only possible because the paths were completely predictable. I think if it's configurable, the xapiandbs directory would have to moved manually.

garethrees:
## Relevant issue(s)

Fixes #1292

## What does this do?

Allow configuration of absolute Xapian DB path and use that to move the test xapiandbs outside the VirtualBox share on Vagrant dev installs

## Why was this needed?

Prevent the corruption issues described in 47bfdea.

## Implementation notes

Sigh, no existing tests so the cost of adding them too great.

## Notes to reviewer

**Testing error on both keys being set:**

```yaml
# config/xapian.yml
development:
  absolute_base_db_path: "/tmp/xapiandbs"
  base_db_path: "/var/xapiandbs"  
```

```
bin/rails runner "puts ActsAsXapian.prepare_environment"
Traceback (most recent call last):
        11: from bin/rails:4:in `<main>'
        10: from bin/rails:4:in `require'
         9: from /home/vagrant/bundle/ruby/2.7.0/gems/railties-5.2.6/lib/rails/commands.rb:18:in `<top (required)>'
         8: from /home/vagrant/bundle/ruby/2.7.0/gems/railties-5.2.6/lib/rails/command.rb:46:in `invoke'
         7: from /home/vagrant/bundle/ruby/2.7.0/gems/railties-5.2.6/lib/rails/command/base.rb:69:in `perform'
         6: from /home/vagrant/bundle/ruby/2.7.0/gems/thor-1.1.0/lib/thor.rb:392:in `dispatch'
         5: from /home/vagrant/bundle/ruby/2.7.0/gems/thor-1.1.0/lib/thor/invocation.rb:127:in `invoke_command'
         4: from /home/vagrant/bundle/ruby/2.7.0/gems/thor-1.1.0/lib/thor/command.rb:27:in `run'
         3: from /home/vagrant/bundle/ruby/2.7.0/gems/railties-5.2.6/lib/rails/commands/runner/runner_command.rb:41:in `perform'
         2: from /home/vagrant/bundle/ruby/2.7.0/gems/railties-5.2.6/lib/rails/commands/runner/runner_command.rb:41:in `eval'
         1: from /home/vagrant/bundle/ruby/2.7.0/gems/railties-5.2.6/lib/rails/commands/runner/runner_command.rb:41:in `<main>'
/home/vagrant/alaveteli/lib/acts_as_xapian/acts_as_xapian.rb:122:in `prepare_environment': Only set one of base_db_path OR absolute_base_db_path (ArgumentError)
```

**Testing that we don't overwrite an existing config file:**

```sh
vagrant provision

echo "# existing copy" >> config/xapian.yml

vagrant provision

grep "existing copy" config/xapian.yml
# existing copy
```

### DIFF
--- a/lib/acts_as_xapian/acts_as_xapian.rb
+++ b/lib/acts_as_xapian/acts_as_xapian.rb
@@ -123,13 +123,14 @@ module ActsAsXapian
     end
 
     # figure out where the DBs should go
-    if config['base_db_path']
-      db_parent_path = Rails.root.join(config['base_db_path'])
-    elsif config['absolute_base_db_path']
-      db_parent_path = config['absolute_base_db_path']
-    else
-      db_parent_path = File.join(File.dirname(__FILE__), 'xapiandbs')
-    end
+    db_parent_path =
+      if config['base_db_path']
+        Rails.root.join(config['base_db_path'])
+      elsif config['absolute_base_db_path']
+        config['absolute_base_db_path']
+      else
+        File.join(File.dirname(__FILE__), 'xapiandbs')
+      end
 
     # make the directory for the xapian databases to go in
     Dir.mkdir(db_parent_path) unless File.exist?(db_parent_path)
--- a/lib/acts_as_xapian/acts_as_xapian.rb
+++ b/lib/acts_as_xapian/acts_as_xapian.rb
@@ -117,9 +117,16 @@ module ActsAsXapian
     config_file = Rails.root.join("config","xapian.yml")
     @@config = File.exist?(config_file) ? YAML.load_file(config_file)[environment] : {}
 
+    if config['base_db_path'] && config['absolute_base_db_path']
+      msg = 'Only set one of base_db_path OR absolute_base_db_path'
+      raise ArgumentError, msg
+    end
+
     # figure out where the DBs should go
     if config['base_db_path']
       db_parent_path = Rails.root.join(config['base_db_path'])
+    elsif config['absolute_base_db_path']
+      db_parent_path = config['absolute_base_db_path']
     else
       db_parent_path = File.join(File.dirname(__FILE__), 'xapiandbs')
     end
--- a/script/site-specific-install.sh
+++ b/script/site-specific-install.sh
@@ -277,6 +277,20 @@ if [ "$DEVELOPMENT_INSTALL" = true ]; then
   gem install mailcatcher
 fi
 
+if [ "$DEVELOPMENT_INSTALL" = true ] && [ x"$SUDO_USER" = x"vagrant" ]
+then
+  VAGRANT_DEV_INSTALL=true
+fi
+
+if [ "$VAGRANT_DEV_INSTALL" = true ] && [ ! -e "$REPOSITORY/config/xapian.yml" ]
+  then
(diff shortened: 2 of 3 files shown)


lib/acts_as_xapian/acts_as_xapian.rb
```diff
@@ -117,13 +117,21 @@ def self.prepare_environment
...
+      raise ArgumentError, msg
     end
 
+    # figure out where the DBs should go
+    db_parent_path =
+      if config['base_db_path']
+        Rails.root.join(config['base_db_path'])
+      elsif config['absolute_base_db_path']
+        config['absolute_base_db_path']
+      else
```
gbp, line 131:
```suggestion
    # figure out where the DBs should go
    db_parent_path =
      if config['base_db_path']
        config['base_db_path']
      else
```

gbp (commented):
Wonder if we could combine `base_db_path` and `absolute_base_db_path` into a single config variable? 

Did you try this?

lib/acts_as_xapian/acts_as_xapian.rb
garethrees: > Wonder if we could combine `base_db_path` and `absolute_base_db_path` into a single config variable?
> 
> Did you try this?

It certainly occurred to me, but the current expectation is that `base_db_path` is relative to the application root, so didn't want to just change that behaviour without managed deprecation. I opted not to do that now given the lack of existing testing. It would reduce some complexity though.

garethrees: Oh, actually, it looks like when given a path starting `/`, `Rails.root.join` doesn't actually join it relative to `Rails.root`?!

```ruby
Rails.root.join('/foo').to_s
# => "/foo"

Rails.root.join('foo').to_s
# => "/home/vagrant/alaveteli/foo"
```

So basically this Just Works without this hunk of the commit. 🤦

gbp: Oh nice - TIL

Issue #6295, opened 2021-06-09
Add minimal CI test run for quicker feedback


garethrees:
We know the full spec suite takes a long time to run https://github.com/mysociety/alaveteli/issues/626.

One problem this causes is that just getting a sense check of some feature development takes quite a bit of time.

We've made a couple of attempts at adding [parallel specs](https://github.com/mysociety/alaveteli/tree/parallel-specs) over the years, but Xapian makes life difficult there.

Instead of parallelising the whole suite, what we could do is add a `ci-minimal.yml` which runs individual spec directories as a poor-man's parallelisation. FMS do this for their [coverage specs](https://github.com/mysociety/fixmystreet/blob/2ec40474cf6badbc446bbc17833eafbec06e62a0/.github/workflows/coverage.yml#L46-L56). Could split up like:

```yml
- name: Test models
  if: matrix.part == 1
  run: bundle exec rspec spec/models

- name: Test views
  if: matrix.part == 2
  run: bundle exec rspec spec/models

- name: Test controllers
  if: matrix.part == 3
  run: bundle exec rspec spec/models

- name: Test integration
  if: matrix.part == 4
  run: bundle exec rspec spec/integration

# This is a bit of a gotcha in that we might easily miss a new directoy
# (e.g. spec/jobs). Maybe we can figure out a file glob that excludes the above
# directories
- name: Test everything else
  if: matrix.part == 5
  run: bundle exec rspec spec/lib spec/mailers spec/script spec/services spec/validators
```

This might at least make life a bit easier for one of the problems caused by slow specs.

garethrees:
Oops, like I just missed `spec/helpers` out of the above list. Definitely want a glob.

HelenWDTK:
This issue has been automatically closed due to a lack of discussion or resolution for over 12 months.
Should we decide to revisit this issue in the future, it can be reopened.

PR #6394, opened 2021-07-27
WIP Track supported OS ruby versions in CI

garethrees:
The normal practice for installing Alaveteli is to use the Ruby provided
by the operating system distribution. We support a variety of operating
systems for re-installers, so we build against multiple Ruby versions
to ensure that Alaveteli is going to run correctly when installed on any
of the given systems.

Up until 0ab1c1a we explicitly targeted the Ruby versions provided by
the supported operating systems at the time. This commit restores this
synchronisation.

Using the latest release of any given Ruby is an edge case for more
experienced users who manage the Ruby version themselves (like us at
mySociety). We tend to use only the very latest Ruby, so I've included
that here.

For faster CI feedback, it would probably be worth extracting a separate
GitHub Workflow for this "Operating System Support" matrix, and use only
the latest supported Ruby for the first-pass CI build.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,11 +12,16 @@ jobs:
 
     strategy:
       fail-fast: false
+      # This matrix should track the Ruby versions available in supported
+      # operating systems, as well as versions installed using rbenv.
       matrix:
         include:
-        - { ruby: 2.5 }
+        # - { ruby: 2.3.3 } # Debian 9 (Stretch) – Must use rbenv
+        - { ruby: 2.5.1 } # Ubuntu 18.04 LTS (Bionic)
+        - { ruby: 2.5.5 } # Debian 10 (Buster)
         - { ruby: 2.6 }
-        - { ruby: 2.7 }
+        - { ruby: 2.7.0 } # Ubuntu 20.04 LTS (Focal)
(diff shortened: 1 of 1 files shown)


gbp:
Big thumbs down for this 👎

We shouldn't track Ruby versions based on the patch number. I want to keep CI fast and this is unnecessary slows it down

garethrees:
> I want to keep CI fast

I agree!

> We shouldn't track Ruby versions based on the patch number

Why not? We don't need every patch number for every elderly version; only those provided by supported operating systems.

I wouldn't object to doing something like this for the most ageing supported versions:

```yaml
# - { ruby: 2.5.1 } # Ubuntu 18.04 LTS (Bionic) – dropping support soon so relying on Debian 10 build
- { ruby: 2.5.5 } # Debian 10 (Buster)
```

At the moment we have a situation where we're not actually testing against our deployment environments. It doesn't matter if the app runs on `2.7.4` if it doesn't work on `2.7.0` that's provided by Focal – that's what others will be using and it will break for them. Installing rubies using rbenv is an edge case; not the default.

Of course, I'm open to alternative suggestions for tackling the issue.

> this is unnecessary slows it down

I have made a proposal in the commit message (very similar to / combine with https://github.com/mysociety/alaveteli/issues/6295) for mitigating the impact. Again, open to suggestion.

PR #6396, opened 2021-07-27
Add /docker-compose.override.yml to .gitignore

itsaphel:
The file is used to extend the normal docker-compose.yml (https://docs.docker.com/compose/extends/) and is usually ignored I believe. (eg I'm using the override file to expose the Postgres port to my host).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -42,3 +42,4 @@
 /tmp/
 /vendor/bundle/
 /vendor/data/
+/docker-compose.override.yml


gbp:
Thanks but I don't think we will merge this change. It [doesn't seem like a used convention](https://github.com/search?q=docker-compose.override.yml+filename%3A.gitignore&type=Code&l=&l=).

If you have a local file which you want to ignore then you should add to it `.git/info/exclude`. I do this for my `shell.nix` and `.envrc` files.

itsaphel:
@gbp That works too, I didn't know of the `.git/info/exclude` approach.

I think something is broken with the GitHub search though. It's not an unusual convention AFAIK, e.g. https://github.com/jitsi/docker-jitsi-meet/blob/master/.gitignore#L4. I think GitHub's search can't handle the dot; [this broader search](https://github.com/search?l=&p=2&q=docker-compose+filename%3A.gitignore&ref=advsearch&type=Code) has a lot of examples scattered within the results.

garethrees:
You can also set a [global `.gitignore`](https://thoughtbot.com/blog/global-gitignore) which might be what you're after?

Issue #1384, opened 2014-03-11
install script with --dev creates an unnecessary symlink


mhl:
After `vagrant up` with the new Vagrant file, the install script has created an `alaveteli` -> `/home/vagrant/alaveteli` symlink in the repository  which isn't useful either inside the guest or on the host. I think this is being created by https://github.com/mysociety/alaveteli/blob/rails-3-develop/script/install-as-user#L44 in the wrong directory; it shouldn't be doing this when invoked with `--dev`

PR #6399, opened 2021-07-28
[#1384] Only symlink repository different from destination

garethrees:
After `vagrant up`  the install script creates an
`alaveteli` -> `/home/vagrant/alaveteli` symlink in the repository which
isn't useful either inside the guest or on the host. It also appears as
a new file in git, which one has to remember to remove.

This commit fixes the issue by only creating the symlink if the
`REPOSITORY` and `LINK_DESTINATION` differ.

Fixes https://github.com/mysociety/alaveteli/issues/1384

### DIFF
--- a/script/install-as-user
+++ b/script/install-as-user
@@ -43,7 +43,11 @@ fi
 REPOSITORY="$DIRECTORY/alaveteli"
 LINK_DESTINATION="$HOME/alaveteli"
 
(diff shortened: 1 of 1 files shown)


garethrees:
Ugh will fix the commit msg:

```diff
- Only symlink repository different from destination
+ Only symlink repository when different from destination
```

gbp (approved):

Issue #6423, opened 2021-08-03
Admin page issues


RichardTaylor:
Issues with /admin/

* Classify responses...  number always states 20 even though [we have 94,732 on WhatDoTheyKnow at the time of writing. ](https://www.whatdotheyknow.com/categorise/play)
* Find missing FOI email ... no way of viewing more than the first 20 

Could point to the classification game rather than have an entry for "classify responses" at all? 

Could request a developer to list all bodies missing a contact address for review as a one-off.

garethrees:
> Classify responses... number always states 20 even though we have 94,732 on WhatDoTheyKnow at the time of writing.

Fixed by https://github.com/mysociety/alaveteli/pull/6425.

> Could point to the classification game rather than have an entry for "classify responses" at all?

I think this would be solved by https://github.com/mysociety/alaveteli/issues/1513.

> Find missing FOI email ... no way of viewing more than the first 20

Extracted to https://github.com/mysociety/alaveteli/issues/6424.

> Could request a developer to list all bodies missing a contact address for review as a one-off.

Ticketed in https://github.com/mysociety/whatdotheyknow-theme/issues/859.

RichardTaylor:
Also I'd get rid of the "(try phoning!)" ... that's certainly not part of the process at WhatDoTheyKnow.com - we try to keep records in writing of everything we do, and why - it's surprising how often that comes in useful.

PR #6425, opened 2021-08-04
[#6423] Show true count of old unclassified requests

garethrees:
We show the total count of bodies without email, but don't for old
unclassified requests. This commit fixes that.

Noted in https://github.com/mysociety/alaveteli/issues/6423.

Now shows a count (25) greater than the limit of the list (20):

Screenshot 2021-08-04 at 09 28 28


gbp (approved):
This seems to work and implenment is the same way as `@blank_contact_count`. LGTM

PR #6430, opened 2021-08-04
Update network error spec expectation

gbp:
On some systems (namely macOS) when `Errno::ETIMEDOUT` is raised the
exception message is 'Operation timed out' and causes this spec to fail.

### DIFF
--- a/spec/support/shared_examples_for_network_errors_during_send.rb
+++ b/spec/support/shared_examples_for_network_errors_during_send.rb
@@ -26,7 +26,7 @@ shared_examples_for 'NetworkSendErrors' do
 
     it 'stores the reason for the failure' do
       event = request.reload.info_request_events.last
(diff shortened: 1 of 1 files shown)


spec/support/shared_examples_for_network_errors_during_send.rb
```diff
@@ -26,7 +26,7 @@
 
     it 'stores the reason for the failure' do
       event = request.reload.info_request_events.last
-      expect(event.params[:reason]).to eq 'Connection timed out'
+      expect(event.params[:reason]).to match /timed out/
```
garethrees, line 29: +1 to the rubocop suggestion:

```diff
- expect(event.params[:reason]).to match /timed out/
+ expect(event.params[:reason]).to match(/timed out/)
```

garethrees (approved):
👍

Issue #4612, opened 2018-04-13
Require those reporting/flagging requests to administrators to give more information


RichardTaylor:
Consider changes to the "report this request" form, and the form presented to those classifying a request has having received an error message, to at least much more strongly encourage, if not require, those making reports to explain their reasons. 

Currently on the report this request form there is a drop-down showing various categories of reason and a "Please tell us more:" box. 

The "Request for Personal Information" drop down option is the only one which can reasonably be used without providing further details. 

One idea is pre-populating the "Please tell us more" box with text which the person making the report needs to replace eg. 

eg. for defamatory material:

[Please precisely identify the material you consider defamatory which you are drawing to our attention, ideally by quoting it, and explain to us why you consider our publication of the material to be problematic.]

or a not a valid request:

[Please explain why you think this isn't a valid request]

The form presented to those marking a request as having received an error message could be changed to say something like; to help the volunteer administrators it would be useful if you could briefly state what's happened and what you think needs to be done. The classification option description could be changed from "error message" to "requires administrator attention", which is how it's described on the admin side I think and is a broader catch-all. 

The aim of this ticket is easier administration; it would I think help administrators if users could be encouraged to clarify what the issue they're reporting is at a early stage. This would save us correspondence seeking details. Often administrators get requests reported to them with no details as to why and administrators have to read through the correspondence sometimes to find the likely problem.

RichardTaylor:
Just adding a +1 to a ticket I raised. 

I think blank reports to administrators should certainly be prevented. 

Before users report a request for admin attention ideally they would be prompted to consider if their issue would be best addressed by a follow-up message to the public body.

RichardTaylor:
Just to note that sometimes requesters flag overdue requests for administrator attention; this is a case where they should be prompted to send a follow-up to the public body.

RichardTaylor:
Some users confuse the "Request for Personal Information" option as a a route for making a request for personal information.

RichardTaylor:
+1 As an administrator receiving an email note that a thread has been reported for admin attention, but with no explanation as to why the report has been made is frustrating as it causes unnecessary work in reviewing the correspondence to try and identify possible issues and seeking clarification.

RichardTaylor:
If users were required to give reasons in the form of selecting options, they could be redirected to the right place if they, for example, as is reasonably common, want to report a request to the admin team because they're unhappy with a response, this should be prevented and the user directed to /help/unhappy and/or prompted to classify their request.

[...]

PR #6439, opened 2021-08-06
[#4612] Require report reasons

garethrees:
Adds HTML validation to require users provide more reasons for reporting
content.

Almost all reporting reasons need further detail, so this is a better
balance for the majority of cases.

Could go further and add validations at the model layer, but this should
prevent the majority of incomplete reports.

Fixes https://github.com/mysociety/alaveteli/issues/4612.

Screenshot 2021-08-06 at 16 01 50

Screenshot 2021-08-06 at 16 02 20


gbp (approved):
Lets see how this works out. Suspect people might just enter a few characters to get around this

PR #6449, opened 2021-08-09
Remove conditionals checking for String#encode

gbp:
This method is included in all Ruby versions we currently support.


garethrees (approved):
🧹

gbp:
Went to merge and found more conditionals for `String#encoding` `String#force_encoding` & `String#valid_encoding`.

Will merge if still green.

gbp:
Some of the `value.respond_to?(:force_encoding)` conditionals are required as these values can be integers and other types as they are coming from YAML data.

Issue #6444, opened 2021-08-08
Improve clarity of documentation relating to CSV upload


RichardTaylor:
At  `/admin/bodies/import_csv` the statement "The fields name and request_email are required" is made, but it would be useful to add:

> request_email can be left blank, doing so will result in request_email not being changed. 

There is a link to the Alaveteli Docs and there is is a statement in the table at https://alaveteli.org/docs/running/admin_manual/#creating-changing-and-uploading-public-authority-data which in respect of request_email states "This column must be present, but can be left empty".

Under the table there is a statement "a blank entry leaves the existing value for that column unchanged" so technically it is all documented but not as obviously as it could be.

garethrees:
> The fields `name` and `request_email` are required

This is incorrect if the suggested addition is true.

PR #6450, opened 2021-08-09
[#6444] Clarify CSV import documentation

garethrees:
`name` and `request_email` _columns_ must be present, but
`request_email` _values_ may be blank.

Fixes https://github.com/mysociety/alaveteli/issues/6444.

Screenshot 2021-08-09 at 11 09 44

### DIFF
--- a/app/views/admin_public_body/import_csv.html.erb
+++ b/app/views/admin_public_body/import_csv.html.erb
@@ -50,10 +50,12 @@
   <div class="span8">
     <div class="alert alert-info">
       <p><strong>CSV file format:</strong> The first row should be a list
-        of fields, starting with <code>#</code>. The fields <code>name</code> and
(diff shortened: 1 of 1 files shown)


RichardTaylor:
"to preserve the persisted value" isn't a phrase which makes much sense to me. 

"to leave the existing request_address unchanged" would be clearer.

RichardTaylor (commented):
"to preserve the persisted value" isn't a phrase which makes much sense to me.

"to leave the existing request_address unchanged" would be clearer.

mdeuk:
> "to leave the existing request_address unchanged" would be clearer.

+1 - I can work out what the original is saying, but this is definitely clearer.

garethrees:
Thanks both – agree on better suggestion. Fixed and merged.

PR #6452, opened 2021-08-09
Fire flash style when the URL hashes are clicked after page load

wrightmartin:
## Relevant issue(s)
Fixes https://github.com/mysociety/whatdotheyknow-theme/issues/832

## What does this do?
Improved the functionality introduced by #6356 by adding the ability to highlight the `<details>` elements being opened after page load

## Why was this needed?
Functionality improvement requested after #6356 went live

### DIFF
--- a/app/assets/javascripts/wizard.js
+++ b/app/assets/javascripts/wizard.js
@@ -356,17 +356,26 @@
     expandFieldset();
   });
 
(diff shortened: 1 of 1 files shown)


garethrees (approved):
Nice! 📸

PR #6455, opened 2021-08-10
Extract 'write a reponse/reply' into it's own button 

### LINKED ISSUE #6363: Extract followup option from actions menu to its own button
Fixes https://github.com/mysociety/alaveteli/issues/5441

The most commonly used option in the Actions menu – write a reply – is in quite an obscure place and leads to lots of user support.

Screenshot 2021-07-12 at 15 27 54

This ticket is to extract that option to its own button, and rename the remaining items "more actions":

Screenshot 2021-07-12 at 15 27 55

wrightmartin:
## Relevant issue(s)
Fixes #6363 

## What does this do?
Moves the oft-used 'write a reply/response' button to it's own prominent button at the bottom of request threads

## Why was this needed?
It's reduces the number of clicks needed to access a well-used function. It will also make this function more prominent in the UI and help show what kinds of things are possible on the site

## Implementation notes
I've left the action in the drop down menu as it's used at the top of the request thread and in other places. As a result I haven't renamed it 'more actions' as the original ticket suggests

## Screenshots
image

## Notes to reviewer
I extracted the reply button to a partial, you may or may not agree with that decision - I'll leave that to the reviewer to approve

### DIFF
--- a/app/views/request/show.html.erb
+++ b/app/views/request/show.html.erb
@@ -58,13 +58,16 @@
                      locals: { info_request: @info_request,
                                last_response: @last_response } %>
         <% else %>
+          <%= render partial: 'request/reply',
+                     locals: { info_request: @info_request,
+                               last_response: @last_response } %>
+                               
           <%= render partial: 'request/after_actions',
                      locals: { info_request: @info_request,
                                track_thing: @track_thing,
                                show_owner_update_status_action: @show_owner_update_status_action,
                                show_other_user_update_status_action: @show_other_user_update_status_action,
                                last_response: @last_response } %>
-
           <%= render partial: 'track/tracking_links_simple',
                      locals: { track_thing: @track_thing,
                                user: @user,
--- a/app/assets/stylesheets/responsive/_global_layout.scss
+++ b/app/assets/stylesheets/responsive/_global_layout.scss
@@ -199,7 +199,8 @@ textarea{
 .action-bar__make-request,
 .action-bar__follow,
 .action-bar__follow-button,
-.action-bar__follower-count {
+.action-bar__follower-count,
+.action-bar__reply {
  vertical-align: middle;
  font-size: 16px;
  margin: 1em 0;
--- a/app/assets/stylesheets/responsive/_request_layout.scss
+++ b/app/assets/stylesheets/responsive/_request_layout.scss
@@ -75,8 +75,9 @@
 
   .after-actions {
     display: inline-block;
+    margin-bottom: 1em; 
     @include respond-min( $main_menu-mobile_menu_cutoff) {
-
+      margin: 0;
     }
     .action-menu__button {
       width: 100%;
@@ -95,6 +96,7 @@
     margin: 0;
     @include respond-min( 30em ){
       margin-left: 1em;
+      
     }
     @include respond-min( $main_menu-mobile_menu_cutoff ){
       text-align: right;
@@ -108,6 +110,17 @@
     margin: 0;
   }
 }
+
+  .action-bar__reply {
+    margin-bottom: -1em; //seems easier to do this here than completely override the whole action styles
+    @media (min-width: 30em){
+      display: inline-block;
+      margin: 0 1em 0 0;
+    }
+    .button {
+      line-height: 1.5em;
+    }
+  }
 
 .request-header__action-bar__actions--narrow {
   //we use --narrow when a profile photo is available
--- a/app/views/alaveteli_pro/info_requests/_after_actions.html.erb
+++ b/app/views/alaveteli_pro/info_requests/_after_actions.html.erb
@@ -1,4 +1,9 @@
+<%= render partial: 'request/reply',
+  locals: { info_request: @info_request,
+          last_response: @last_response } %>
+          
 <div id="after_actions" class="after-actions">
+
   <ul class="action-menu after-actions__action-menu">
     <li>
       <a href="#" class="action-menu__button">Actions</a>
--- a/app/views/request/_reply.html.erb
+++ b/app/views/request/_reply.html.erb
@@ -1,0 +1,9 @@
+<div class="action-bar__reply">
+  <% if last_response.nil? %>
+    <%= link_to _("Send a followup"), new_request_followup_path(:request_id => info_request.id, :anchor => 'followup'),
+    :class => "action-bar__reply-button" %>
+  <% else %>
+    <%= link_to _("Write a reply"), new_request_incoming_followup_path(:request_id => info_request.id, :incoming_message_id => last_response.id, :anchor => 'followup'),
+    :class => "action-bar__reply-button" %>
+  <% end %>
+</div>


app/views/request/show.html.erb
```diff
@@ -64,7 +64,14 @@
...
                                show_other_user_update_status_action: @show_other_user_update_status_action,
                                last_response: @last_response } %>
-
+                               
+          <%= render partial: 'request/reply',
+          locals: { info_request: @info_request,
+                    track_thing: @track_thing,
+                    show_owner_update_status_action: @show_owner_update_status_action,
+                    show_other_user_update_status_action: @show_other_user_update_status_action,
+                    last_response: @last_response } %>
```
garethrees, line 73: The `request/reply` partial only requires the `info_request` and `last_response` local variables.

Also want to check the source code alignment of the `locals` hash.

app/views/request/show.html.erb
```diff
@@ -64,7 +64,14 @@
                                show_owner_update_status_action: @show_owner_update_status_action,
                                show_other_user_update_status_action: @show_other_user_update_status_action,
                                last_response: @last_response } %>
-
+
```
garethrees, line 67: This line is full of spaces – should be empty.

app/views/request/show.html.erb
```diff
@@ -64,7 +64,14 @@
...
                                last_response: @last_response } %>
-
+                               
+          <%= render partial: 'request/reply',
+          locals: { info_request: @info_request,
+                    track_thing: @track_thing,
+                    show_owner_update_status_action: @show_owner_update_status_action,
+                    show_other_user_update_status_action: @show_other_user_update_status_action,
+                    last_response: @last_response } %>
+
```
garethrees, line 74: This line is full of spaces – should be empty.

garethrees (changes requested):
Agree with using a partial here.

Looks good on WDTK, but looks a bit borked elsewhere. Wouldn't expect re-users to need to make theme updates for this – I'm not sure why WDTK handles the change without modification?

Screenshot 2021-08-10 at 12 23 26

Screenshot 2021-08-10 at 12 11 46

There's a separate set of actions for pros, so we'll need to render this there too.

Screenshot 2021-08-10 at 12 36 41

garethrees:
Also, I think "Reply" should come before "Actions", since it's the most common action to take.

Screenshot 2021-08-10 at 12 35 03

That does make the button colours a bit weird. Really the "follow" button shouldn't be a primary action, but I'll ticket that up separately.

wrightmartin:
I've pushed those improvements but it looks like removing the buttony style removes the styling for WDTK theme

I'm looking at the pro display now

wrightmartin:
All feedback addressed and ready to look at again

app/views/request/show.html.erb
```diff
@@ -54,17 +54,22 @@
     <div class="request-footer__action-bar-container">
       <div class="request-footer__action-bar__actions">
         <% if @in_pro_area %>
+          <%= render partial: 'request/reply',
+                     locals: { info_request: @info_request,
+                               last_response: @last_response } %>
```
garethrees, line 59: I think it would be nicer to render this within `alaveteli_pro/info_requests/after_actions`?

```diff
diff --git a/app/views/alaveteli_pro/info_requests/_after_actions.html.erb b/app/views/alaveteli_pro/info_requests/_after_actions.html.erb
index d9b830634..9230df490 100644
--- a/app/views/alaveteli_pro/info_requests/_after_actions.html.erb
+++ b/app/views/alaveteli_pro/info_requests/_after_actions.html.erb
@@ -1,3 +1,7 @@
+<%= render partial: 'request/reply',
+           locals: { info_request: info_request,
+                     last_response: last_response } %>
+
 <div id="after_actions" class="after-actions">
   <ul class="action-menu after-actions__action-menu">
     <li>
diff --git a/app/views/request/show.html.erb b/app/views/request/show.html.erb
index b184ec6b3..1e7b1fc47 100644
--- a/app/views/request/show.html.erb
+++ b/app/views/request/show.html.erb
@@ -54,9 +54,6 @@
     <div class="request-footer__action-bar-container">
       <div class="request-footer__action-bar__actions">
         <% if @in_pro_area %>
-          <%= render partial: 'request/reply',
-                     locals: { info_request: @info_request,
-                               last_response: @last_response } %>
           <%= render partial: 'alaveteli_pro/info_requests/after_actions',
                      locals: { info_request: @info_request,
                                last_response: @last_response } %>
```

app/views/request/show.html.erb
```diff
@@ -54,17 +54,22 @@
...
           <%= render partial: 'request/after_actions',
                      locals: { info_request: @info_request,
                                track_thing: @track_thing,
                                show_owner_update_status_action: @show_owner_update_status_action,
                                show_other_user_update_status_action: @show_other_user_update_status_action,
                                last_response: @last_response } %>
-
```
garethrees, line 67: Keep this newline

app/views/request/show.html.erb
```diff
@@ -54,17 +54,22 @@
...
+          <%= render partial: 'request/reply',
+                     locals: { info_request: @info_request,
+                               last_response: @last_response } %>
           <%= render partial: 'alaveteli_pro/info_requests/after_actions',
                      locals: { info_request: @info_request,
                                last_response: @last_response } %>
         <% else %>
+          <%= render partial: 'request/reply',
+                     locals: { info_request: @info_request,
+                               last_response: @last_response } %>
```
garethrees, line 66: Add a newline after this for clearer separation of big code chunks.

garethrees (commented):
Couple of super minor comments inline, but I think the main thing that would be great to address is making this zero-config for re-users. I can see we've made an update to WDTK (https://github.com/mysociety/whatdotheyknow-theme/pull/871), but we're going to have to remember to do this for _all_ themes.

Is there a better way? If the answer is "Not without a _load_ of effort", then I guess we'll have to settle for this. Have you had a look at whether we can include the styles in Alaveteli core at all and have existing theme styles "Just Work"?

wrightmartin:
> Couple of super minor comments inline, but I think the main thing that would be great to address is making this zero-config for re-users. I can see we've made an update to WDTK ([mysociety/whatdotheyknow-theme#871](https://github.com/mysociety/whatdotheyknow-theme/pull/871)), but we're going to have to remember to do this for _all_ themes.
> 
> Is there a better way? If the answer is "Not without a _load_ of effort", then I guess we'll have to settle for this. Have you had a look at whether we can include the styles in Alaveteli core at all and have existing theme styles "Just Work"?


Yeah this is a tough one - for _reasons_ we didn't include button styles in the core but this means, as you say, additions like this aren't zero-config, without knowing more about how the various cobrands have implemented their own button styles any solution is going to be a guess if it will work, eg if we add a button style we think is a sensible default we still don't know how it will interact with their own styles they've implemented. We could add styles to alaveletitheme with the idea they could cherry-pick them but that still results in them having to do *something*

Also if we add a button style for this one, it's then an exception as no other buttons have styles so it ends up a bit weird.

In a perfect world we'd revisit the theming as I think the white-label ended up a bit too spartan in retrospect. Sorry for the long reply, I'll try and get to the smaller feedback stuff before the end of the sprint

garethrees:
> without knowing more about how the various cobrands have implemented their own button styles any solution is going to be a guess if it will work … Also if we add a button style for this one, it's then an exception as no other buttons have styles so it ends up a bit weird.

Gotcha, let's go with current approach.

> We could add styles to alaveletitheme with the idea they could cherry-pick them but that still results in them having to do something

Yeah, do this so that it makes life easier for everyone. We'll also want a line in the "Upgrade notes" section of the changelog to remind reusers to add this.

> Sorry for the long reply

No problems here – I'd rather us document this stuff for future reference.

> I'll try and get to the smaller feedback stuff before the end of the sprint

Cheers!

wrightmartin:
Pushed those changes now 👍

garethrees:

PR #6457, opened 2021-08-11
[#6436] Improve test ENV attachment caching

### LINKED ISSUE #6436: Running attachment controller specs twice results in error
Running:

```
rspec ./spec/controllers/attachments_controller_spec.rb:212
rspec ./spec/controllers/attachments_controller_spec.rb:212
```

Results in an error like:

```
      Ability.new(#<User>, {:project=>#<Project>})
          expected: 1 time with arguments: (#<User>, {:project=>#<Project>})
          received: 0 times
```

It appears the attactments being cached during the spec run aren't been cleanered up.

While looking at this it has made me wounder why the cache callback is happening before the authenication callbacks... should these be the other way around? 
https://github.com/mysociety/alaveteli/blob/f44ab48dad80c0911dbeb5452c7c09a778e88964/app/controllers/attachments_controller.rb#L13-L16

gbp:
## Relevant issue(s)

Fixes #6436

## What does this do?

Prevent attachment controller cached files from being written to disk
expect for one test where we need to check the file permissions.

## Why was this needed?

Without this re-running the same spec more than once could result in an
error due to the normal execution being stopped early.

For example running this spec twice would error due to the standard
ability check to be skipped:

```
  rspec ./spec/controllers/attachments_controller_spec.rb:212

  Ability.new(#<User>, {:project=>#<Project>})
      expected: 1 time with arguments: (#<User>, {:project=>#<Project>})
      received: 0 times
```

## Notes to reviewer

Still need to consider:

> While looking at this it has made me wounder why the cache callback is happening before the authenication callbacks... should these be the other way around? 
https://github.com/mysociety/alaveteli/blob/f44ab48dad80c0911dbeb5452c7c09a778e88964/app/controllers/attachments_controller.rb#L13-L16

### DIFF
--- a/spec/controllers/attachments_controller_spec.rb
+++ b/spec/controllers/attachments_controller_spec.rb
@@ -2,6 +2,10 @@ require 'spec_helper'
 
 RSpec.describe AttachmentsController, type: :controller do
 
+  before do
+    allow(@controller).to receive(:foi_fragment_cache_write)
+  end
+
   describe 'GET show' do
 
     let(:info_request) do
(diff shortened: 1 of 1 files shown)


garethrees (approved):
💥 

> While looking at this it has made me wounder why the cache callback is happening before the authenication callbacks... should these be the other way around?

Let's ticket that up

PR #6461, opened 2021-08-11
[#6408] Update controllers default response to be HTML

### LINKED ISSUE #4228: [ERROR] request#new (ActionView::MissingTemplate) "Missing template request/new, application/ne...
```
An ActionView::MissingTemplate occurred in request#new:

  Missing template request/new, application/new with {:locale=>[:en], :formats=>[:json], :variants=>[], :handlers=>[:erb, :builder, :raw, :ruby, :coffee]}. Searched in:
  * "[FILTERED]/www.whatdotheyknow.com/alaveteli-2017-09-08T13-15-28/lib/themes/whatdotheyknow-theme/lib/views"
  * "[FILTERED]/www.whatdotheyknow.com/alaveteli-2017-09-08T13-15-28/app/views"
  * "[FILTERED]/www.whatdotheyknow.com/shared/vendor/bundle/ruby/2.3.0/gems/foundation-rails-5.5.3.2/app/views"

  app/controllers/request_controller.rb:1088:in `render_new_compose'


Request:

  * URL        : https://www.whatdotheyknow.com/new/liverpool_city_council.json
  * HTTP Method: GET
  * IP address : [FILTERED]
  * Parameters : {"controller"=>"request", "action"=>"new", "url_name"=>"liverpool_city_council", "format"=>"json"}
  * Timestamp  : 2017-09-11 11:36:32 +0100
  * Server : [FILTERED]
  * Rails root : [FILTERED]/www.whatdotheyknow.com/alaveteli-2017-09-08T13-15-28
  * Process: 8612

Backtrace:

  app/controllers/request_controller.rb:1088:in `render_new_compose'
  app/controllers/request_controller.rb:323:in `new'
```

### LINKED ISSUE #6408: [ERROR] sessions#new (ActionView::MissingTemplate) "Missing template user/sign…"
```
An ActionView::MissingTemplate occurred in sessions#new:

Missing template user/sign with {:locale=>[:"en-UG", :en], :formats=>[:json], :variants=>[], :handlers=>[:raw, :erb, :html, :builder, :ruby]}. Searched in:
* "/data/vhost/uganda.alaveteli.org/alaveteli-2021-07-22T10-22-20/lib/themes/askyourgovug-theme/lib/views"
* "/data/vhost/uganda.alaveteli.org/alaveteli-2021-07-22T10-22-20/app/views"

app/controllers/users/sessions_controller.rb:23:in `new'

Request:

* URL : https://askyourgov.ug/profile/sign_in?token=REDACTED
* HTTP Method: GET
* Parameters : {"token"=>"REDACTED", "controller"=>"users/sessions", "action"=>"new"}
* Timestamp : 2021-07-28 09:08:34 +0300
* Rails root : /data/vhost/uganda.alaveteli.org/alaveteli-2021-07-22T10-22-20

Session:

* session id: [FILTERED]
* data: {"session_id"=>"REDACTED", "locale"=>"en_UG"}

Backtrace:

app/controllers/users/sessions_controller.rb:23:in `new'
lib/strip_empty_sessions.rb:14:in `call'
```

gbp:
## Relevant issue(s)

Fixes #6408
Fixes #4228

## What does this do?

Adds a before action to `ApplicationController` to set the default response to HTML for all actions.

We obviously have actions which response to JSON, CSV, ZIP, PNG, etc so for these actions, in their own controller, we call `skip_before_action` to bypass the default.

## Why was this needed?

This prevents `ActionView::MissingTemplate` exceptions when JSON is requested for a route which we have explicitly said responds to it.

## Implementation notes

With this change `ActionController::UnknownFormat` errors will be raised which we already catch within `ApplicationController` and return a 406 response to the user.

This means we are now responding with 406 bad request, where we used to respond with either a 403 forbidden or a 404 not found.

It means we can't handle invalid format, for `GeneralController#search` we used to gracefully degrade to HTML for any unspecified formats. With this change we return a 406 which seems like an acceptable behaviour.

We can also simplify controller `respond_to` blocks if `format.html` is the only format specified.

### DIFF
--- a/app/controllers/admin_raw_email_controller.rb
+++ b/app/controllers/admin_raw_email_controller.rb
@@ -5,6 +5,8 @@
 # Email: hello@mysociety.org; WWW: http://www.mysociety.org/
 
 class AdminRawEmailController < AdminController
+  skip_before_action :html_response
+
   before_action :set_raw_email, only: [:show]
 
   def show
--- a/app/controllers/alaveteli_pro/batch_downloads_controller.rb
+++ b/app/controllers/alaveteli_pro/batch_downloads_controller.rb
@@ -5,11 +5,12 @@
 class AlaveteliPro::BatchDownloadsController < AlaveteliPro::BaseController
   include ActionController::Live
 
+  skip_before_action :html_response
+
   def show
     authorize! :download, info_request_batch
 
     respond_to do |format|
-      format.html { head :bad_request }
       format.zip { download_zip }
       format.csv do
         metrics = InfoRequestBatchMetrics.new(info_request_batch)
--- a/app/controllers/alaveteli_pro/public_bodies_controller.rb
+++ b/app/controllers/alaveteli_pro/public_bodies_controller.rb
@@ -1,6 +1,8 @@
 class AlaveteliPro::PublicBodiesController < AlaveteliPro::BaseController
   include AlaveteliPro::PublicBodiesHelper
 
+  skip_before_action :html_response
+
   def index
     query = params[:query] || ""
     xapian_results = typeahead_search(query, :model => PublicBody,
--- a/app/controllers/alaveteli_pro/stripe_webhooks_controller.rb
+++ b/app/controllers/alaveteli_pro/stripe_webhooks_controller.rb
@@ -3,6 +3,8 @@ class AlaveteliPro::StripeWebhooksController < ApplicationController
   class MissingTypeStripeWebhookError < StandardError; end
   class UnknownPlanStripeWebhookError < StandardError; end
 
+  skip_before_action :html_response
+
   rescue_from JSON::ParserError, MissingTypeStripeWebhookError do |exception|
     # Invalid payload, reject the webhook
     notify_exception(exception)
--- a/app/controllers/alaveteli_pro/subscriptions_controller.rb
+++ b/app/controllers/alaveteli_pro/subscriptions_controller.rb
@@ -1,6 +1,7 @@
 class AlaveteliPro::SubscriptionsController < AlaveteliPro::BaseController
   include AlaveteliPro::StripeNamespace
 
+  skip_before_action :html_response, only: [:create, :authorise]
   skip_before_action :pro_user_authenticated?, only: [:create, :authorise]
   before_action :authenticate, only: [:create, :authorise]
   before_action :prevent_duplicate_submission, only: [:create]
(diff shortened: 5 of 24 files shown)


garethrees (commented):
I think the approach is broadly correct here – render HTML by default and then do something special when that's not the case – but I'm not sure on the actual implementation.

It [looks like](https://www.justinweiss.com/articles/respond-to-without-all-the-pain/) you can set `respond_to` at the controller level. Rather than using filters, can we not set that in `ApplicationController` for others to inherit from, and then override that default in actions which have a custom response format?

```ruby
class ApplicationController < ActionController::Base
  respond_to :html
end

class AnotherController < ApplicationController
  # Responds to HTML from inherited `respond_to`
  def index
    # ...
  end

  def show
    # …
    respond_to do |format|
      format.html
      format.json do
        # some custom JSON thing for this action
      end
    end
  end

  # This action _only_ responds to JSON
  def foo
    # …
    respond_to :json
  end
end

class ApiController < ApplicationController
  # This controller only responds to JSON, so override the inherited value
  respond_to :json

  # …
end
```

I might be missing something with this suggestion, and if that's the case then I think we can go with your PR, but I can't help thinking that something doesn't smell quite right.

gbp:
That article is old. In Rails 4.2 the class level `respond_to` was [removed](https://edgeguides.rubyonrails.org/4_2_release_notes.html#respond-with-class-level-respond-to). It was extracted out into the [responders](https://github.com/heartcombo/responders) gem. While we could add to our Gemfile and make some changes to this PR but I would prefer not adding another dependency.

gbp:
> Just briefly summarising the next step on this, which is to create a new rails app and see what happens by default when you access an unsupported format. It feels like by default this should be a 400-level response code without us having to jump through hoops.

Rails, by default, has renderers for JSON, JS and XML objects [1]. So if there is no `respond_to` call within the controller action it will return these formats if there is a relevant view file. Without a view file a `ActionController::UnknownFormat` is raised. If a different format is requested without being added to a `respond_to` block a `ActionController::UnknownFormat` will also be raised.

The correct way to limit the JSON, JS and XML format does seem to be by using `respond_to` within the controller action. While having `respond_to :html` in a before action callback isn't probably the standard thing to do it does mean we can explicit about which formats we respond to.

[1] https://github.com/rails/rails/blob/48661542a2607d55f436438fe21001d262e61fec/actionpack/lib/action_controller/metal/renderers.rb#L26-L28

garethrees (approved):
Thanks for looking in to this further. I must say, I'm surprised Rails doesn't make this cleaner, but it's fine and will fix a nice class of problem for us 👍

PR #6481, opened 2021-08-17
[#4160] Uncaught duplicate requests

### LINKED ISSUE #4160: Uncaught duplicate requests

gbp:
## Relevant issue(s)

Fixes #4160

## What does this do?

Trim the leading and trailing whitespace from the request title when
checking for existing requests.

## Why was this needed?

When a user is directed to signin before a request is made it results in
a `PostRedirect` being created and an email is sent to the requester to
confirm their email address. This mail includes a link to `/new` which
will finish creating their request.

The `PostRedirect` includes the all form data as it was entered into the
new `InfoRequest` form (including any extra whitespace). The comparison
to existing records can then break. We already handle whitespace in the
body to account for line endings difference but the title was a direct
comparison which would fail. Existing records have whitespace stripped
when saving into the database.

## Notes to reviewer

Includes a method rename which cause confusion while debugging this issue.

### DIFF
--- a/app/controllers/request_controller.rb
+++ b/app/controllers/request_controller.rb
@@ -236,9 +236,9 @@ class RequestController < ApplicationController
                                                      params[:outgoing_message][:body],
                                                      params[:public_body_ids])
 
-    @info_request = InfoRequest.create_from_attributes(info_request_params(@batch),
-                                                       outgoing_message_params,
-                                                       authenticated_user)
+    @info_request = InfoRequest.build_from_attributes(info_request_params(@batch),
+                                                      outgoing_message_params,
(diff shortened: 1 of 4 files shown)


garethrees (approved):
Haven't reviewed thoroughly but looks good in principle 👍

Issue #5851, opened 2020-08-04
Review new framework defaults


gbp:
When updating Rails, doing `rails app:update`, we get a new framework defaults file and something like following added to `config/application.rb`

```ruby
# Initialize configuration defaults for originally generated Rails version.
config.load_defaults x.x # <- rails version number
```

We need to go through the `config/initializers/new_framework_defaults*.rb` files and enable these new defaults. This hasn't been done for any of the recent Rails upgrades - since 5.0 -> 5.1 I think.

Some of these changes will require changes to the codebase in order to be enabled.

Rails 5.0
- [ ] Enable per-form CSRF tokens
- [ ] Enable origin-checking CSRF mitigation
- [ ] Make Ruby 2.4 preserve the timezone of the receiver when calling `to_time`
- [ ] Require `belongs_to` associations by default.

Rails 5.1
- [ ] Make `form_with` generate non-remote forms.
- [ ] Unknown asset fallback will return the path passed in when the given asset is not present in the asset pipeline.

Rails 5.2
- [ ] Make Active Record use stable #cache_key alongside new #cache_version method.
- [ ] Use AES-256-GCM authenticated encryption for encrypted cookies.
- [ ] Use AES-256-GCM authenticated encryption as default cipher for encrypting messages instead of AES-256-CBC
- [ ] Add default protection from forgery to ActionController::Base instead of in ApplicationController.
- [ ] Store boolean values are in sqlite3 databases as 1 and 0 instead of 't' and 'f' after migrating old data.
- [ ] Use SHA-1 instead of MD5 to generate non-sensitive digests, such as the ETag header.
- [ ] Make `form_with` generate id attributes for any generated HTML tags.

Rails 6.0
- [ ] Don't force requests from old versions of IE to be UTF-8 encoded.
- [ ] Embed purpose and expiry metadata inside signed and encrypted cookies for increased security.
- [ ] Change the return value of `ActionDispatch::Response#content_type` to Content-Type header without modification.
- [ ] Return false instead of self when enqueuing is aborted from a callback.
- [ ] Send Active Storage analysis and purge jobs to dedicated queues.
- [ ] When assigning to a collection of attachments declared via `has_many_attached`, replace existing attachments instead of appending. Use #attach to add new attachments without replacing existing ones.
- [ ] Use ActionMailer::MailDeliveryJob for sending parameterized and normal mail.
[...]

PR #6485, opened 2021-08-18
[#5159] Rails 6 upgrade

### LINKED ISSUE #5159: Upgrade to Rails 6.0
Requires https://github.com/mysociety/alaveteli/issues/5069.
Requires Ruby 2.5 https://blog.bigbinary.com/2019/04/09/rails-6-requires-ruby-2-5-or-newer.html
Requires Postgres 9.4+ https://blog.bigbinary.com/2019/04/10/rails-6-drops-support-for-postgresql-less-than-9-3.html

gbp:
## Relevant issue(s)

Fixes #5159 

## What does this do?

1. Upgrade Alaveteli to Rails 6.0
2. Refactor gettext/locale loading to resolve app booting issue in development environment. 

## Notes to reviewer

This PR doesn't prep for the Rails 6.1 release as the CI will be red and with how Github actions works we can't allow failures without it putting red crosses everywhere and emailing us. We don't really want that so lets prep in a separate PR and fix CI there.


lib/alaveteli_localization.rb
```diff
@@ -1,8 +1,35 @@
...
+      repos = paths.map do |path|
+        FastGettext::TranslationRepository.build('app', path: path, type: :po)
+      end
+      AlaveteliLocalization.set_default_text_domain('app', repos)
+
+      I18n::Backend::Simple.send(:include, I18n::Backend::Fallbacks)
+
+      AlaveteliLocalization.set_default_locale_urls(
+        AlaveteliConfiguration.include_default_locale_in_urls
+      )
```
gbp, line 24: There was some minor refactoring here but should be equivalent to what was in `config/initializers/fast_gettext.rb`

garethrees (commented):
> 0ca6063

Nice!

What are we doing about the new [autoloading](https://guides.rubyonrails.org/v6.0/upgrading_ruby_on_rails.html#autoloading) stuff? Is that getting handled in https://github.com/mysociety/alaveteli/issues/5851 or do they need to be addressed here?

Not much change so fine in principle. I've booted the app successfully but haven't tested further than that.

lib/alaveteli_localization.rb
```diff
@@ -1,8 +1,35 @@
...
-require 'alaveteli_localization/underscorred_locale'
-
 class AlaveteliLocalization
+  class Railtie < Rails::Railtie
+    config.before_configuration do
+      require 'alaveteli_localization/locale'
+      require 'alaveteli_localization/hyphenated_locale'
+      require 'alaveteli_localization/underscorred_locale'
+
+      require 'routing_filters'
```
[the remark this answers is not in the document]
gbp: https://github.com/mysociety/alaveteli/blob/36c1c748bd2466a9366288cedffacf8438ace1e5/lib/alaveteli_localization.rb#L100

This is called within `set_locales`.

lib/alaveteli_localization.rb
gbp: Is!

gbp:
> What are we doing about the new [autoloading](https://guides.rubyonrails.org/v6.0/upgrading_ruby_on_rails.html#autoloading) stuff? Is that getting handled in #5851 or do they need to be addressed here?

This isn't using Zeitwork yet. We would need load defaults for Rails "6.0" in `config/application.rb` to enable

PR #6506, opened 2021-09-03
[#6500] Refactor utility search path

### LINKED ISSUE #6500: PDF images use disk path rather than web path
For example, https://www.whatdotheyknow.com/request/779754/response/1851171/attach/html/3/FOI%20Acknowledgement%2025265.pdf.html refers to https://www.whatdotheyknow.com/data/vhost/www.whatdotheyknow.com/alaveteli-2021-08-04T11-59-57/cache/views/request/779/779754/response/1851171/attach/html/3/foiextract20210810-12843-1a0biew-1_1.jpg rather than https://www.whatdotheyknow.com/request/779754/response/1851171/attach/html/3/foiextract20210810-12843-1a0biew-1_1.jpg
Over >10k requests for these type of URLs each day.

Looks like the cause is the change made in https://github.com/mysociety/alaveteli/pull/6373 - as you're no longer switching to the temp directory before converting, any image paths in the conversion use the absolute path, rather than relative, and so no longer work. (Presumably there's a reason you don't use $PATH, but can you not have a local override instead of forcing local paths?)

gbp:
## Relevant issue(s)

Fixes #6500

## What does this do?

- Revert previous to utility search path
- Update the default utility search path setting
- Allow empty setting, which means the PATH environment variable will be used.

## Why was this needed?

Broken link to images in `pdftohtml` output.


garethrees (commented):
AIUI:

* 5fa2598cde2840a4289c33d454005b68284731ad fixes https://github.com/mysociety/alaveteli/issues/6500 
* 90f935596978747e2e86bfe9a5f24841e84edc8e adds a spec so that we'd have noticed this regression in the build on https://github.com/mysociety/alaveteli/pull/6373
* e72a868dd9b760fc3d2546f5efb26df777a14747 is some cleanup
* bf2297fb6bfba95e41510c5e52436d4161bb926f resolves the original issue that https://github.com/mysociety/alaveteli/pull/6373 was attempting to resolve

I think we should merge and deploy 5fa2598cde2840a4289c33d454005b68284731ad, 90f935596978747e2e86bfe9a5f24841e84edc8e and e72a868dd9b760fc3d2546f5efb26df777a14747 immediately.

bf2297fb6bfba95e41510c5e52436d4161bb926f seems like a better default, though I wonder why we even make this configurable via `config/general.yml`. Wouldn't one just add to the app user's `$PATH`? 384f4b0d75753ad8e3a7ee405156433c8930a6ec doesn't indicate _why_ we wanted to resolve the command within Alaveteli. Perhaps this is due to the variability of `$PATH` in login/non-login/interactive/non-interactive shells, or the crontab `$PATH`? I'm a bit rusty on that these days.

I think we should take the opportunity to remove complexity here, if possible. If `UTILITY_SEARCH_PATH` isn't solving a problem for anyone (looks like all our installs use the default) then I don't see why we're maintaining it. As such, probably worth extracting bf2297fb6bfba95e41510c5e52436d4161bb926f to its own PR while we give that some thought?

gbp:
I've dropped commit bf2297f

PR #6511, opened 2021-09-09
Update authentication to expire other sessions

gbp:
## Relevant issue(s)

Depends on #6510 
Fixes https://github.com/mysociety/alaveteli_security/issues/22
Fixes https://github.com/mysociety/alaveteli_security/issues/23

## What does this do?

Update authentication to expire other sessions

Adds User#last_login_at which is assigned to the user when they login and store in the session.
Also checks the session value matches the user when authenticating.

## Why was this needed?

This will ensure users are logged out of other sessions when they've logged in.

## Implementation notes

Originally attempted to use `#updated_at` instead of adding another database column but this doesn't allow user to be sign in. We also had concerns this would change which the user was actively using the site and then be logged out.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -147,12 +147,14 @@ class ApplicationController < ActionController::Base
     remember_me ||= session[:remember_me]
     clear_session_credentials
     session[:user_id] = user.id
+    session[:user_login_token] = user.login_token
     session[:remember_me] = remember_me
   end
 
   # Logout form
   def clear_session_credentials
     session[:user_id] = nil
+    session[:user_login_token] = nil
     session[:user_circumstance] = nil
     session[:remember_me] = false
     session[:using_admin] = nil
@@ -278,7 +280,9 @@ class ApplicationController < ActionController::Base
   def authenticated_user
     return unless session[:user_id]
 
-    @user ||= User.find_by(id: session[:user_id])
+    @user ||= User.find_by(
+      id: session[:user_id], login_token: session[:user_login_token]
+    )
   end
 
   # For CanCanCan and other libs which need a Devise-like current_user method
--- a/app/controllers/password_changes_controller.rb
+++ b/app/controllers/password_changes_controller.rb
@@ -85,7 +85,7 @@ class PasswordChangesController < ApplicationController
       end
 
       if @password_change_user.save
-        session[:user_id] ||= @password_change_user.id
+        sign_in(@password_change_user)
 
         if @pretoken_redirect
           if otp_enabled?(@password_change_user)
--- a/app/controllers/user_controller.rb
+++ b/app/controllers/user_controller.rb
@@ -245,6 +245,8 @@ class UserController < ApplicationController
     @user.email = @signchangeemail.new_email
     @user.save!
 
+    sign_in(@user)
+
     # Now clear the circumstance
     session[:user_circumstance] = nil
     flash[:notice] = _("You have now changed your email address used on {{site_name}}",:site_name=>site_name)
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -1,5 +1,5 @@
 # == Schema Information
-# Schema version: 20210114161442
+# Schema version: 20210921094059
 #
 # Table name: users
 #
@@ -34,12 +34,14 @@
 #  daily_summary_hour                :integer
 #  daily_summary_minute              :integer
 #  closed_at                         :datetime
+#  login_token                       :string
 #
 
 class User < ApplicationRecord
   include AlaveteliFeatures::Helpers
   include AlaveteliPro::PhaseCounts
   include User::Authentication
+  include User::LoginToken
   include User::Survey
 
   rolify before_add: :setup_pro_account
--- a/app/models/user/login_token.rb
+++ b/app/models/user/login_token.rb
@@ -1,0 +1,20 @@
+# Updates the current users login token based on their current password and
+# email
+module User::LoginToken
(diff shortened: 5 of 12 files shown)


app/controllers/application_controller.rb
```diff
@@ -145,13 +145,19 @@ def persist_session_timestamp
 
   def sign_in(user, remember_me: false)
     clear_session_credentials
+
+    now = Time.zone.now
+    user.update_column(:last_login_at, now)
```
gbp, line 150: Probably a spec missing here

app/controllers/application_controller.rb
```diff
@@ -275,9 +281,12 @@ def authenticated_as_user?(user, reason_params = {})
 
   # Return logged in user
   def authenticated_user
-    return unless session[:user_id]
+    return unless session[:user_id] && session[:user_last_login_at]
 
-    @user ||= User.find_by(id: session[:user_id])
+    @user ||= (
+      user = User.find_by(id: session[:user_id])
+      user if user&.last_login_at.to_s == session[:user_last_login_at]
```
gbp, line 288: Spec needed to show what is returned is these do and don't match

gbp: Comparing string representation of the timestamp, should we compare actual timestamps?

I went for strings here as this quickly dealt with milliseconds differences.

gbp: Do we want to do the check as part of the ActiveRecord query?

db/migrate/20210909105614_add_last_login_at_to_users.rb
```diff
@@ -0,0 +1,5 @@
+class AddLastLoginAtToUsers < ActiveRecord::Migration[5.2]
+  def change
+    add_column :users, :last_login_at, :timestamp
```
gbp, line 3: - Is is okay to record this about our users? Do we need to update privacy policy or anything else.
- Might be useful addition to for admins?
- Could we switch to some other column?

app/controllers/application_controller.rb
garethrees: > Do we want to do the check as part of the ActiveRecord query?

Something like that seems quite nice:

```ruby
# A
@user ||= User.find_logged_in(session[:user_id], session[:user_last_login_at])

# B
@user ||= User.logged_in(session[:user_last_login_at]).find(session[:user_id])
```

A is better in the context of the controller since it reduces the method chain, but it's definition in `User` could be:

```ruby
def self.find_logged_in(id, last_login_at)
  logged_in(last_login_at).find(id)
end
```

garethrees (commented):
Looks great in principle (have not reviewed https://github.com/mysociety/alaveteli/pull/6510 but happy with that after a quick glance).

Main thing to consider is whether we can do basically the same thing without storing additional PII.

app/controllers/application_controller.rb
gbp: Doing time comparison in PG is such a pain. Though I had is sorted a couple of times but with either milliseconds or time zone issues it would break in different environments.

I've gone with doing the comparison in Ruby but have extracted to a `User.find_logged_in` method instead of doing it in the `ApplicationController`.

db/migrate/20210909105614_add_last_login_at_to_users.rb
gbp: Discussed on standup and we felt there wasn't any issues with storing this - @garethrees am I recalling our discussion on Friday correct?

garethrees (changes requested):
Seems to completely prevent users being logged in?

last-login-at

garethrees:
Something I didn't spot before is that I _think_ this approach would result in the following being true?

* Log in in browser A
* Log in in a private window (browser B)
* Browser A is now logged out because the login timestamp stored does not match the updated timestamp from browser B

I wasn't able to check this because of the issue with not being able to log in at all, but if this is true then I think we might want to reconsider that?

gbp:
@garethrees this has been reworked so we generate a login token based on the current email and password. The token is rereshed when either is updated and stored in the session on login and verified when the user authenticated.

By default the login token is `nil` and will get populated the first time these attributes are changed. This keeps current sessions active. We could create a rake task to generate these tokens and add validation to ensure this is present but its not actually needed to fix the linked issues.

app/controllers/password_changes_controller.rb
```diff
@@ -85,7 +85,7 @@ def update
       end
 
       if @password_change_user.save
-        session[:user_id] ||= @password_change_user.id
+        sign_in(@password_change_user, remember_me: session[:remember_me])
```
gbp, line 88: Made a change in #6510 which means we can just do:

```suggestion
        sign_in(@password_change_user)
```

app/controllers/user_controller.rb
```diff
@@ -245,6 +245,8 @@ def signchangeemail
     @user.email = @signchangeemail.new_email
     @user.save!
 
+    sign_in(@user, remember_me: session[:remember_me])
```
gbp, line 248: Made a change in #6510 which means we can just do:

```suggestion
    sign_in(@user)
```

garethrees (approved):
Really nice simple implementation and fix – love this 👍

garethrees:
> We could create a rake task to generate these tokens and add validation to ensure this is present but its not actually needed to fix the linked issues.

Don't think this is necessary. Happy to just let them populate over time.

gbp:
Merged f46bfe11fa101e285cab75c3839462ba6821fb5f

PR #6512, opened 2021-09-10
Expire used and old change password/email tokens

gbp:
## Relevant issue(s)

~~(Fixes) https://github.com/mysociety/alaveteli_security/issues/23~~ – this PR is about token re-use, not session expiry
~~(Fixes) https://github.com/mysociety/alaveteli_security/issues/25~~ – collection issue

Fixes https://github.com/mysociety/alaveteli_security/issues/26 – appears reproducible because the generated token is the same in each attempt, but this PR correctly prevents the previous token being used after a successful reset.
Fixes https://github.com/mysociety/alaveteli_security/issues/27 – prevents re-use of token.
Fixes https://github.com/mysociety/alaveteli_security/issues/24

## What does this do?

Use ActiveSupport::MessageVerifier and generate token with the current
value so if the password or email is changed the token then won't work
anymore.

### DIFF
--- a/app/controllers/user_controller.rb
+++ b/app/controllers/user_controller.rb
@@ -225,12 +225,13 @@ class UserController < ApplicationController
     if (not session[:user_circumstance]) or (session[:user_circumstance] != "change_email")
       # don't store the password in the db
       params[:signchangeemail].delete(:password)
-      post_redirect = PostRedirect.new(:uri => signchangeemail_url,
-                                       :post_params => params,
-                                       :circumstance => "change_email" # special login that lets you change your email
-                                       )
-      post_redirect.user = @user
-      post_redirect.save!
+
+      post_redirect = PostRedirect.create!(
+        uri: signchangeemail_url,
+        post_params: params,
+        user: @user,
+        circumstance: 'change_email'
+      )
 
       url = confirm_url(:email_token => post_redirect.email_token)
       UserMailer.
--- a/app/controllers/users/confirmations_controller.rb
+++ b/app/controllers/users/confirmations_controller.rb
@@ -3,7 +3,7 @@ class Users::ConfirmationsController < UserController
   def confirm
     post_redirect = PostRedirect.find_by_email_token(params[:email_token])
 
-    if post_redirect.nil?
+    if post_redirect.nil? || !post_redirect.email_token_valid?
       render :template => 'user/bad_token'
       return
     end
--- a/app/models/post_redirect.rb
+++ b/app/models/post_redirect.rb
@@ -26,6 +26,7 @@
 # Copyright (c) 2007 UK Citizens Online Democracy. All rights reserved.
 # Email: hello@mysociety.org; WWW: http://www.mysociety.org/
 
+require 'digest'
 require 'openssl' # for random bytes function
 
 class PostRedirect < ApplicationRecord
@@ -39,6 +40,17 @@ class PostRedirect < ApplicationRecord
 
   after_initialize :generate_token
   after_initialize :generate_email_token
+
+  def self.verifier
+    Rails.application.message_verifier(to_s)
+  end
+
+  def self.generate_verifiable_token(user:, circumstance:)
+    verifier.generate(
+      { user_id: user.id, login_token: user.login_token },
+      purpose: circumstance
+    )
+  end
 
   # Makes a random token, suitable for using in URLs e.g confirmation
   # messages.
@@ -91,6 +103,12 @@ class PostRedirect < ApplicationRecord
   # There is a separate token to use in the URL if we send a confirmation
   # email.
   def generate_email_token
-    self.email_token = PostRedirect.generate_random_token unless email_token
+    if !user || circumstance == 'normal'
+      self.email_token ||= PostRedirect.generate_random_token
+    end
+
+    self.email_token ||= PostRedirect.generate_verifiable_token(
+      user: user, circumstance: circumstance
+    )
   end
 end
--- a/app/models/post_redirect.rb
+++ b/app/models/post_redirect.rb
@@ -92,6 +92,13 @@ class PostRedirect < ApplicationRecord
     $1
   end
 
+  def email_token_valid?
+    return true unless PostRedirect.verifier.valid_message?(email_token)
+
+    data = PostRedirect.verifier.verify(email_token, purpose: circumstance)
+    user.id == data[:user_id] && user.login_token == data[:login_token]
+  end
+
   private
 
   # The token is used to return you to what you are doing after the login
--- a/app/views/user/bad_token.html.erb
+++ b/app/views/user/bad_token.html.erb
@@ -1,19 +1,20 @@
 <h1>
 <%= _('Please check the URL (i.e. the long code of letters and numbers) is ' \
-      'copied correctly from your email.')%>
+      'copied correctly from your email.') %>
 </h1>
 
 <p>
 <%= _("If you can't click on it in the email, you'll have to <strong>select " \
       "and copy it</strong> from the email.  Then <strong>paste it into your " \
       "browser</strong>, into the place you would type the address of any " \
-      "other webpage.")%>
+      "other webpage.") %>
 </p>
 
 <p>
-<%= _("If you got the email <strong>more than two months ago</strong>, then " \
-      "this login link won't work any more. Please try doing what you were " \
-      "doing from the beginning.")%>
+<%= _("If you got the email <strong>more than two months ago</strong> or " \
+      "<strong>updated your user profile</strong>, then this login link " \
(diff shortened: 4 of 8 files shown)


app/models/post_redirect.rb
```diff
@@ -40,6 +41,14 @@ class PostRedirect < ApplicationRecord
   after_initialize :generate_token
   after_initialize :generate_email_token
 
+  def self.verifier
+    Rails.application.message_verifier(to_s)
+  end
+
+  def self.generate_verifiable_token(user:, value:)
+    verifier.generate(user_id: user.id, nonce: Digest::MD5.hexdigest(value))
```
gbp, line 49: Why am I converting the value to MD5... admins would be able to retrive the old values until the PostRedirect is deleted after two months. Or from the params in the app logs

gbp: Not really a `nonce`... is there a better name

app/models/post_redirect.rb
```diff
@@ -80,6 +89,22 @@ def local_part_uri
...
+    return true if circumstance == 'normal'
+
+    case circumstance
+    when 'change_email'
+      value = user.email
+    when 'change_password'
+      value = user.hashed_password
+    end
+
+    data = PostRedirect.verifier.verify(email_token)
```
gbp, line 102: Handle exceptions. EG `ActiveSupport::MessageVerifier::InvalidSignature`

app/controllers/user_controller.rb
```diff
@@ -223,9 +223,9 @@ def signchangeemail
       params[:signchangeemail].delete(:password)
       post_redirect = PostRedirect.new(:uri => signchangeemail_url,
                                        :post_params => params,
+                                       :user => @user,
```
gbp, line 226: Need user here so the `after_initialize` `#generate_email_token` method has access to it.

app/controllers/users/confirmations_controller.rb
```diff
@@ -3,7 +3,7 @@ class Users::ConfirmationsController < UserController
   def confirm
     post_redirect = PostRedirect.find_by_email_token(params[:email_token])
 
-    if post_redirect.nil?
+    if post_redirect.nil? || !post_redirect.email_token_valid?
       render :template => 'user/bad_token'
```
gbp, line 7: `user/bad_token` template would need some copy changes.

app/models/post_redirect.rb
```diff
@@ -80,6 +89,22 @@ def local_part_uri
...
+    nonce = Digest::MD5.hexdigest(value)
+
+    user.id == data[:user_id] && nonce == data[:nonce]
```
gbp, line 105: Do we need to compare the `user.id` or even store in the token?

app/models/post_redirect.rb
gbp: Need to URL encode the token.

app/models/post_redirect.rb
```diff
@@ -91,6 +116,17 @@ def generate_token
...
+    when 'change_email'
+      self.email_token ||= PostRedirect.generate_verifiable_token(
+        user: user, value: user.email
+      )
+    when 'change_password'
+      self.email_token ||= PostRedirect.generate_verifiable_token(
+        user: user, value: user.hashed_password
+      )
+    else
+      self.email_token ||= PostRedirect.generate_random_token
```
garethrees, line 129: Could this be generated from an _unchangeable_ attribute – `User#id` say – so that we do away with the verifiable/random conditionals?

```ruby
CIRCUMSTANCE_ATTRIBUTES = {
  change_email: :email,
  change_password: :hashed_password,
  normal: :id
}

def generate_email_token
  self.email_token ||= PostRedirect.generate_verifiable_token(
    user: user, value: user.send(CIRCUMSTANCE_ATTRIBUTES[circumstance])
  )
end
```

If something like that works then I imagine `email_token_valid?` could be simplified in a similar way.

gbp: I've kinda implemented this... if we use `User#id` then the email token will always be the same. This could meant multiple `PostRedirect` instances with the same email token. I think we would need to expire tokens some other way when they've been used.

I've kept the fallback to the old generator for now.

app/models/post_redirect.rb
gbp: Confirmed the URL token is encoded correctly

app/models/post_redirect.rb
gbp: I'm going to keep this in. In case the digest value is identical for different users (shouldn't happen for the hashed password but could for the email address I guess or a MD5 collision)

app/views/user/bad_token.html.erb
```diff
@@ -11,9 +11,10 @@
 </p>
 
 <p>
-<%= _("If you got the email <strong>more than two months ago</strong>, then " \
-      "this login link won't work any more. Please try doing what you were " \
-      "doing from the beginning.")%>
+<%= _("If you got the email <strong>more than two months ago</strong> or " \
+      "<strong>updated your user profile</strong>, then this login link " \
+      "won't work any more. Please try doing what you were doing from the " \
+      "beginning.")%>
```
garethrees, line 17: Could add a minor fix here:

```diff
- )%>
+ ) %>
```

garethrees (changes requested):
I could replicate https://github.com/mysociety/alaveteli_security/issues/24 at b2cb5bb656f3d407c9e0ee794e336d81fbab7fc0 so I'm not sure this is ready to go yet.

gbp:
@garethrees so with the `User#login_token` from changes #6511 we can simplify this PR too to remove the `CIRCUMSTANCE_ATTRIBUTES`. Instead we can use `User#login_token` for all types of `PostRedirect` records instead of generating our own digest.

This means https://github.com/mysociety/alaveteli_security/issues/24 is also fixed.

app/models/post_redirect.rb
```diff
@@ -91,6 +107,8 @@ def generate_token
   # There is a separate token to use in the URL if we send a confirmation
   # email.
   def generate_email_token
-    self.email_token = PostRedirect.generate_random_token unless email_token
+    return self.email_token = PostRedirect.generate_random_token unless user
```
gbp, line 110:
```suggestion
    self.email_token = PostRedirect.generate_random_token unless user
```

garethrees (commented):
Let's merge https://github.com/mysociety/alaveteli/pull/6511 and rebase this with the simplifications then I'll review.

garethrees (changes requested):
Left a couple of minor comments on implementation, but more concerning is that this does not fix at least https://github.com/mysociety/alaveteli_security/issues/24 as it introduces a significant bug in its own right. (I haven't checked the other issues because the fixes required may or may not impact the results of testing of them).

Here are my notes from following the reproduction steps, recording the email content and confirmation link at each:

```
PASSWORD RESET TOKEN 1

Please click on the link below to confirm your email address. Then you can change your password on [DEV] WhatDoTheyKnow

http://10.10.10.30:3000/c/BAh7BzoMdXNlcl9pZGkoOhBsb2dpbl90b2tlbjA=--1b0883b4070ce04a8360298f8f98508946e537f2

EMAIL RESET TOKEN 1

Please click on the link below to confirm that you want to change the email address that you use for [DEV] WhatDoTheyKnow from 123@example.com to 456@example.com

http://10.10.10.30:3000/c/BAh7BzoMdXNlcl9pZGkoOhBsb2dpbl90b2tlbjA=--1b0883b4070ce04a8360298f8f98508946e537f2
```

I cannot progress past the step of verifying the new email address because the URL redirects me to the password change form. Note that both tokens are identical.

I suspect what's happening here is that in [dropping the `CIRCUMSTANCE_ATTRIBUTES`](https://github.com/mysociety/alaveteli/pull/6512/#issuecomment-925671893) the same token is being generated for both uses, and then we're rendering the change password form just because it happens to be the first in line when deciding what to do based on the `PostRedirect` lookup. That said, I think it should be a fairly trivial issue to resolve.

I'd assumed that you were going to encode the circumstance in the token through `MessageVerifier`'s `purpose:` option, along the lines of `verifier.generate(user_id: user.id, login_token: user.login_token, purpose: circumstance)` and then handle that on the verification side (`def email_token_valid?(purpose = 'normal')`).

gbp:
> more concerning is that this does not fix at least [mysociety/alaveteli_security#24](https://github.com/mysociety/alaveteli_security/issues/24) as it introduces a significant bug in its own right. (I haven't checked the other issues because the fixes required may or may not impact the results of testing of them).

Ugh sorry - I totally meant to pass in the circumstance into `PostRedirect.generate_verifiable_token`. Of course without it generates the same token. Sigh. Should be an easy fix for the morning.

garethrees (approved):

PR #6524, opened 2021-09-23
Deprecate class-based User role viewing permissions

garethrees:
Deprecates class-based User role viewing permissions in favour of instance methods.

Removes unnecessary complexity now that we have the safe navigation operator.

### DIFF
--- a/app/models/ability.rb
+++ b/app/models/ability.rb
@@ -187,9 +187,10 @@ class Ability
     if info_request.embargo
       case prominence
       when 'hidden'
-        User.view_hidden_and_embargoed?(user)
+        user&.view_hidden_and_embargoed?
       when 'requester_only'
-        info_request.is_actual_owning_user?(user) || User.view_hidden_and_embargoed?(user)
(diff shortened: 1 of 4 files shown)


gbp (commented):
I agree with the deprecations. For the "ability" methods, I think it would be good to either:
1. extract these methods into a concern/module eg `User::Ability`
2. rename these methods as `User#can_view_*?`

garethrees:
> For the "ability" methods, I think it would be good to either:

Hmm, yeah, I agree that current state isn't perfect, but I think this might take a little more thought.

Like, should we even have these methods? Shouldn't `Ability` handle all that? Is the `can_` prefix somewhat confusing in relation to the `can` provided by CanCanCan? What about other methods in `User` not touched here that are related to "ability" that we should move?

In general I'm not particularly happy with the way we do authorisation across the app, but that's quite a _can_ of worms. 🥁

I'd rather not get sucked into that right now, so will just merge as is, but yes, I agree it's a good next step.

PR #6528, opened 2021-09-27
Refactor PublicBody#reindex_requested_from

garethrees:
Extract the task of reindexing all events from `reindex_requested_from`
to make the behaviour a little clearer and allow us to call
`reindex_all_events` manually from the console
(`@public_body.send(:reindex_all_events`) if we need to reindex a
specific body.

Also adds missing specs for the behaviour.

Extracted from a case [1] where we needed to manually reindex a body
because the UI timed out due to the large number of requests.

[1] https://github.com/mysociety/whatdotheyknow-theme/issues/879#issuecomment-907130512

### DIFF
--- a/app/models/public_body.rb
+++ b/app/models/public_body.rb
@@ -917,17 +917,10 @@ class PublicBody < ApplicationRecord
 
   private
 
-  # if the URL name has changed, then all requested_from: queries
-  # will break unless we update index for every event for every
-  # request linked to it
+  # If the url_name has changed, then all requested_from: queries will break
+  # unless we update index for every event for every request linked to it.
   def reindex_requested_from
-    return unless saved_change_to_attribute?(:url_name)
-
-    info_requests.find_each do |info_request|
-      info_request.info_request_events.find_each do |info_request_event|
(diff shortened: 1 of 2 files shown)


app/models/public_body.rb
```diff
@@ -921,8 +921,10 @@ def questions
   # will break unless we update index for every event for every
   # request linked to it
   def reindex_requested_from
-    return unless saved_change_to_attribute?(:url_name)
+    reindex_all_events if saved_change_to_attribute?(:url_name)
+  end
 
+  def reindex_all_events
```
garethrees, line 927: This is basically the same as `PublicBody#expire_requests`. Expiring also clears some caches, but perhaps we should just take the hit on additional processing here to remove some complexity? The new specs pass with `expire_requests if saved_change_to_attribute?(:url_name)`

gbp: Yep I think it makes sense to switch to `expire_requests`

gbp (changes requested):

garethrees:
Updated; will merge when green.

PR #6538, opened 2021-09-28
WIP Add Docker provider for Vagrant

gbp:
## What does this do?

POC attempting to show if its possible to test our supported OS installation using Vagrant with a Docker provider

## Why was this needed?

Migration to M1 Macs mean we need a replacement to VirtualBox


gbp:
While this would be good to support I don't think this is going to be taken forwarded. Because:
1. we have , and are using Docker development environment now.
2. would required reworking of crons and services to work correctly in a container.

PR #6543, opened 2021-10-01
Replace unneeded calls to #write_attribute and #write_attribute_without_type_cast

gbp:
## Relevant issue(s)

Needed for #6011 

## What does this do?

Replace unneeded calls to #write_attribute and #write_attribute_without_type_cast

## Why was this needed?

These methods aren't really meant to be called directly, we can replace with direct assignment and newer public methods.


garethrees (commented):
Have you considered https://github.com/mysociety/alaveteli/commit/92d5d2bcdaca56a6508504a8aa86e56f3630d301 and the problem the introduction of `write_attribute` was working around?

Given the tests pass, I _assume_ it's no longer a problem, but this change should reference back to 92d5d2b and explain why the workaround is no longer required.

gbp:
> Given the tests pass, I _assume_ it's no longer a problem, but this change should reference back to [92d5d2b](https://github.com/mysociety/alaveteli/commit/92d5d2bcdaca56a6508504a8aa86e56f3630d301) and explain why the workaround is no longer required.

TBH I'm not even sure why 92d5d2b was even required. I have gone through hoops trying to get a working Rails 3.2 app but can't replicate the issue it supposedly fixed. Suspect there was something else happening, maybe with gem used at that time. As the specs pass and I've tested in development, I don't think there is anything else we need to worry about.

gbp:
> I've had a bit more of a look into this. Many (all? haven't checked) of the attributes where we've used `write_attribute` are re-defined to call `parse_raw_email!`.

Yeah I see that. The thing is, in a new Rails 3.2 app, calling the setters doesn't call the getters. There must be something else going on when 92d5d2bcdaca56a6508504a8aa86e56f3630d301 was committed.

As originally suggested I had hope to improve the commit message by identifying a commit in Rails where this had changed but at this point I'm not sure the old behaviour was caused by Rails or some other gem. I'm not going to spend any more time attempting to track this down as I would need to run an old version of Alaveteli thats not really going to be possible for me on a M1 mac.

The change, as is, in my opinion is go to merge.

PR #6544, opened 2021-10-01
Fix spec run using white label theme

gbp:
When using the default white label theme, running:
  `rspec ./spec/script/mailin_spec.rb:24`

Would return:
  `shlib/deployfns read_conf: error calling PHP to load config file config/general`

This change ensures `config/general.yml` is present before attempting to
reading the file.


script/mailin
```diff
@@ -11,7 +11,7 @@ cat >"$INPUT"
 cd "$(dirname "${BASH_SOURCE[0]}")"/..
 source commonlib/shlib/deployfns
 
-read_conf config/general
+[ -f config/general.yml ] && read_conf config/general.yml
```
gbp, line 14: Probably want:

```suggestion
[ -e config/general.yml ] && read_conf config/general.yml
```

script/load-mail-server-logs
```diff
@@ -4,7 +4,7 @@ LOC=`dirname "$0"`
 
 cd "$LOC"/..
 source commonlib/shlib/deployfns
-read_conf config/general
+[ -f config/general.yml ] && read_conf config/general.yml
```
gbp, line 7: Probably want:

```suggestion
[ -e config/general.yml ] && read_conf config/general.yml
```

PR #6601, opened 2021-10-25
Force SSL

### LINKED ISSUE #6545: Support for Cloudflare SSL
Currently if a domain is protected using Cloudflare - such that the SSL certs are provided by Cloudflare. The Rails app has to be run with `FORCE_SSL: false` this means URLs to assets in the HTML link to the non-http versions and then rewritten/redirected by Cloudflare.

This works fine for assets but for AJAX requests such as how our Stripe checkout process works means we get errors in the web console:

> Blocked loading mixed active content “http://imamopravoznati.org/profile/subscriptions/sub_123/authorise”

​This breaks the checkout process and means a user is charged for a subscription but not granted the Pro abilities.

gbp:
## Relevant issue(s)

Fixes #6545

## What does this do?

Switches to Rails' `force_ssl` option.

## Notes to reviewer

~~Untested. Might help with issues a partner site is seeing.~~

Was needed on IPZ to get Stripe payments and SSL working correctly.


gbp:
Have verified default url options are using `https` by using a [guide](https://blog.revathskumar.com/2020/07/using-ssl-in-local-development.html) to enable SSL in my local development env.

Note if following the guide I also required switching out `thin` for `puma` in our Gemfile.

garethrees (commented):
Haven't run this but looks fine.

What I don't understand is _why_ this fixes https://github.com/mysociety/alaveteli/issues/6545. Doesn't seem to be any mention of the original problem and how the changes fix it.

lib/attachment_to_html/adapters/google_docs_viewer.rb
```diff
@@ -23,11 +23,7 @@ def initialize(attachment, opts = {})
       private
 
       def parse_body
-        %Q(<iframe src="#{ protocol }://docs.google.com/viewer?url=#{ attachment_url }&amp;embedded=true" width="100%" height="100%" style="border: none;"></iframe>)
-      end
-
-      def protocol
-        AlaveteliConfiguration.force_ssl ? 'https' : 'http'
+        %Q(<iframe src="https://docs.google.com/viewer?url=#{ attachment_url }&amp;embedded=true" width="100%" height="100%" style="border: none;"></iframe>)
```
garethrees, line 26: I don't disagree, but _why_? Commit message should explain.

gbp: I have updated the comment

Issue #6347, opened 2021-07-05
Release 0.40


gbp:
- [x] #5159
- [x] #5440
- [x] #5637 
- [x] #6118
- [x] #5608 
- [x] #6344
- [x] #6294 

Announce 
- [x] #5224
- [x] #5225 
- [x] #5638
- [x] #5227

PR #6618, opened 2021-10-29
Fix PATH setting for crontab

gbp:
## What does this do?

Fix PATH setting for crontab

## Why was this needed?

This change was required to get a jobs to run correctly when Ruby is
installed via rbenv.

## Notes to reviewer

This has been tested and put in place on both Sweden and Romania installs.

### DIFF
--- a/config/crontab-example
+++ b/config/crontab-example
@@ -12,8 +12,7 @@ PATH=/home/<%= user %>/.gem/ruby/<%= ruby_version %>/bin:/usr/local/bin:/usr/bin
 PATH=/usr/local/bin:/usr/bin:/bin
 
 <% if use_rbenv? %>
-RBENV_ROOT="/home/<%= user %>/.rbenv"
-PATH="$RBENV_ROOT/shims:$PATH"
+PATH="/home/<%= user %>/.rbenv/shims:$PATH"
 <% else %>
 PATH="/home/<%= user %>/.gem/ruby/<%= ruby_version %>/bin:$PATH"
 <% end %>
--- a/config/crontab-example
+++ b/config/crontab-example
@@ -5,9 +5,9 @@
 # Email: hello@mysociety.org. WWW: http://www.mysociety.org/
 
 <% if use_rbenv? %>
-PATH="/home/<%= user %>/.rbenv/shims:/usr/local/bin:/usr/bin:/bin"
+PATH=/home/<%= user %>/.rbenv/shims:/usr/local/bin:/usr/bin:/bin
 <% else %>
-PATH="/home/<%= user %>/.gem/ruby/<%= ruby_version %>/bin:/usr/local/bin:/usr/bin:/bin"
+PATH=/home/<%= user %>/.gem/ruby/<%= ruby_version %>/bin:/usr/local/bin:/usr/bin:/bin
 <% end %>
 
 MAILTO=<%= mailto %>
--- a/config/crontab-example
+++ b/config/crontab-example
@@ -14,7 +14,6 @@ PATH=/usr/local/bin:/usr/bin:/bin
 <% if use_rbenv? %>
 RBENV_ROOT="/home/<%= user %>/.rbenv"
 PATH="$RBENV_ROOT/bin:$RBENV_ROOT/shims:$PATH"
-SET_PATH="cd <%= vhost_dir %>/<%= vcspath %>; rbenv rehash; rbenv local <%= ruby_version %>"
 <% else %>
 PATH="/home/<%= user %>/.gem/ruby/<%= ruby_version %>/bin:$PATH"
 <% end %>
--- a/config/crontab-example
+++ b/config/crontab-example
@@ -13,7 +13,7 @@ PATH=/usr/local/bin:/usr/bin:/bin
 
 <% if use_rbenv? %>
 RBENV_ROOT="/home/<%= user %>/.rbenv"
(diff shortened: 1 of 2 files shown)


gbp:
@garethrees we should get this into #6347 too

garethrees:
> @garethrees we should get this into #6347 too

Happy to look but I'm struggling to understand what the original problem was and _why_ the diff fixes it.

After unpicking the refactoring part of the change, I _think_ in the `if use_rbenv? #=> true` branch we get left with a diff of:

```diff
- PATH=/home/<%= user %>/.rbenv/bin:/home/<%= user %>/.rbenv/shims:/usr/local/bin:/usr/bin:/bin
+ PATH=/home/<%= user %>/.rbenv/shims:/usr/local/bin:/usr/bin:/bin
```

So dropping `…/.rbenv/bin` from the front of the `PATH` fixes some problem? Why? I don't "why" in the sense of "why should we fix something that's broken", but why was the previous version not working – what was the symptom? why did the old version cause that? – and then why does the new version actually solve the issue?

It looks like the new version is identical to the suggested `PATH` to uncomment on L10.

Looks like https://github.com/mysociety/alaveteli/commit/4a9601c2ac3799b9efada5bd6b5b63f4409371a5 might have added an extra `PATH` line https://github.com/mysociety/alaveteli/blame/4a9601c2ac3799b9efada5bd6b5b63f4409371a5/config/crontab-example#L7-L12 which might be complicating things here?  Prior to that commit (L7) we had a default `PATH` of `PATH=/home/<%= user %>/.gem/ruby/<%= ruby_version %>/bin:/usr/local/bin:/usr/bin:/bin`, but then that default gets overridden by the new addition on L12 (`PATH=/usr/local/bin:/usr/bin:/bin`) – we then go into the RBENV stuff.

I don't doubt that you understand what's actually happening here, but as a reviewer I don't, and without including that context in the commit messages and PR body, it makes it really difficult to review a change like this.

gbp:
I've not done a great job at explaining or even understanding the issue, let me work it out...:

I've spun up a Debian 9 box, with old Ruby so our rbenv install script will install a compatible Ruby versions.

To get a site running: 
1. I run the install script
2. add a rails_env.rb file to set the app to production env
3. migrate and seed the database
4. asset precompile
5. as root `service alaveteli start`

Site is now running and loadable in the browser.

Wait for the cronjobs to be run, and root is mailed an cron error, with:
```
/var/www/server/alaveteli/commonlib/bin/run-with-lockfile.sh: 30: /var/www/server/alaveteli/commonlib/bin/run-with-lockfile.sh: lockfile-check: not found
/var/www/server/alaveteli/commonlib/bin/run-with-lockfile.sh: 37: /var/www/server/alaveteli/commonlib/bin/run-with-lockfile.sh: lockfile-create: not found
```

Strange those commands exist in `/usr/bin`. So I update `run-with-lockfile.sh` with full paths to those commands. Wait for cron, it errors again, this time with:
```
/var/www/ala.gbp.io/alaveteli/commonlib/bin/run-with-lockfile.sh: 30: /var/www/ala.gbp.io/alaveteli/commonlib/bin/run-with-lockfile.sh: lockfile-check: not found
/var/www/ala.gbp.io/alaveteli/script/update-xapian-index: line 6: dirname: command not found
/var/www/ala.gbp.io/alaveteli/script/update-xapian-index: line 7: bundle: command not found
```

Right so, ignore the lockfile command issue, seems like the correct Ruby version isn't being used.

Update the cronfile head as this PR is suggesting. No further errors being generated, both the lockfile command issue and using the correct Ruby version has been resolved.

What fixed the issue? Lets revert back the change one part at a time.

1. Add back the SET_PATH line. No errors
2. Add `/home/alaveteli/.rbenv/bin` to the PATH. No errors
3. Revert the PATH, to use `$RBENV_ROOT` and `$PATH` variables. Errors again
4. Remove `$RBENV_ROOT` from PATH. Errors again
5. Add `$RBENV_ROOT`, remove `$PATH` from PATH. Different error (`Your bundle is locked to rake (13.0.3)` looks like it is using system Ruby)

Conclusions

1. We can keep the SET_PATH line - not sure what it does but it doesn't seem to be needed.
2. We can add `~/.rbenv/bin` back to the PATH - this directory only contains the `rbenv` command so isn't needed
3. We can't have variables in the PATH

gbp:
Based on the above I don't think any changes are required to this PR apart from improving the commit message. @garethrees ?

garethrees:
> I've not done a great job at explaining or even understanding the issue, let me work it out...:

Nice, this was really helpful in figuring out what's going on.


> * We can keep the SET_PATH line - not sure what it does but it doesn't seem to be needed.

I couldn't figure out what that was doing either. Searching for `cron "SET_PATH"` returns our crontab as the first result, so doesn't seem like a standard thing. Are we looking for this variable in our infra scripts? Maybe have a quick look through the git history to see if it becomes clear. Happy to remove it if we can't find a clear reason for it existing, though might be worth a separate commit to do so in case we want to revert it or reference it in some future issue.

> * We can add `~/.rbenv/bin` back to the PATH - this directory only contains the `rbenv` command so isn't needed

Happy to remove – just worth explaining why it's removed.

> * We can't have variables in the PATH

Makes sense.

> Based on the above I don't think any changes are required to this PR apart from improving the commit message.

I think the final diff looks good – much simpler. May just want to tell a clearer story with the commits and commit messages with the above simplifications.

Issue #6628, opened 2021-11-01
The "login as" feature isn't working


RichardTaylor:
There is a [dev@](https://groups.google.com/a/mysociety.org/g/alaveteli-dev-private/c/m7Fk0Fuf9sw) thread on this issue.

PR #6629, opened 2021-11-02
[#6628] Fix login as feature

gbp:
## Relevant issue(s)

Fixes: #6628

## What does this do?

Fix login as feature

## Why was this needed?

Since #6511 we check users login token when authenticating. This has
broken our login as feature as the login token present in the session
belongs to the admin user.

This change updates the session and introduces a spec to show the
authenticated user is updated.

## Implementation notes

I thought about using our new-ish `sign_in` helper method but this clears the session resulting in the black admin bar from being removed. This might be fine as to get back to a fully functional admin user we do need to sign out and sign back in again already.

### DIFF
--- a/app/controllers/admin_users_sessions_controller.rb
+++ b/app/controllers/admin_users_sessions_controller.rb
@@ -18,6 +18,7 @@ class AdminUsersSessionsController < AdminController
     @admin_user.confirm!
 
     session[:user_id] = @admin_user.id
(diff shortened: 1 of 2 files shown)


garethrees (approved):
👍 Fixes the issue. Could either tweak to use `sign_in_as` or find/create issue for removing the admin menu.

Issue #6630, opened 2021-11-02
CI builds failing to install theme


gbp:
CI builds are broken with:

```
Run cp config/general.yml-example config/general.yml
Installing theme alavetelitheme using deprecated THEME_URL from git://github.com/mysociety/alavetelitheme.git
Cloning into '/home/runner/work/alaveteli/alaveteli/lib/themes/alavetelitheme'...
fatal: remote error: 
  The unauthenticated git protocol on port 9418 is no longer supported.
Please see https://github.blog/2021-09-01-improving-git-protocol-security-github/ for more information.
rake aborted!
Cloning from git://github.com/mysociety/alavetelitheme.git to /home/runner/work/alaveteli/alaveteli/lib/themes/alavetelitheme failed
/home/runner/work/alaveteli/alaveteli/lib/tasks/themes.rake:110:in `install_theme'
/home/runner/work/alaveteli/alaveteli/lib/tasks/themes.rake:145:in `block (2 levels) in <top (required)>'
/home/runner/work/alaveteli/alaveteli/vendor/bundle/ruby/2.5.0/gems/rake-13.0.6/exe/rake:27:in `<top (required)>'
/opt/hostedtoolcache/Ruby/2.5.9/x64/bin/bundle:23:in `load'
/opt/hostedtoolcache/Ruby/2.5.9/x64/bin/bundle:23:in `<main>'
Tasks: TOP => themes:install
(See full trace by running task with --trace)
Error: Process completed with exit code 1.
```

We should switch from `git` protocol to `https`.

PR #6631, opened 2021-11-02
[#6630] Fix CI builds

gbp:
## Relevant issue(s)

Fixes: #6630

## What does this do?

Update theme repo URLs

## Why was this needed?

Switch from the `git` protocol to `https`. This is required as GitHub
are strengthening their security around the use of the git protocol.

See: https://github.blog/2021-09-01-improving-git-protocol-security-github/


gbp:
It seems Code Climate fetch is also failing due to this

gbp:
Have tested and made this change to our servers repo.

PR #6644, opened 2021-11-03
Refactor User#reindex_referencing_models

garethrees:
* Move `expire_requests` to be closer to conceptually similar
  `reindex_referencing_models`
* Extract `expire_comments` for clarity

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -320,14 +320,18 @@ class User < ApplicationRecord
     return if no_xapian_reindex == true
     return unless saved_change_to_attribute?(:url_name)
 
-    comments.find_each do |comment|
-      comment.info_request_events.find_each do |info_request_event|
-        info_request_event.xapian_mark_needs_index
-      end
-    end
+    expire_comments
+    expire_requests
+  end
 
-    info_requests.find_each do |info_request|
-      info_request.info_request_events.find_each do |info_request_event|
+  def expire_requests
+    info_requests.find_each(&:expire)
+  end
+
+  def expire_comments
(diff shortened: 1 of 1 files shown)


app/models/user.rb
```diff
@@ -320,17 +320,21 @@ def reindex_referencing_models
     return if no_xapian_reindex == true
     return unless saved_change_to_attribute?(:url_name)
 
+    expire_comments
+    expire_requests
+  end
+
+  def expire_requests
+    info_requests.find_each(&:expire)
```
gbp, line 328: `InfoRequest#expire` will clear fragment and database cache. Are you sure it is okay to do this every time the user is updated? Seems like this could cause a lot of extra processing.

garethrees: Sorry, clearly I should have been more explicit about this change in the commit message.

We [decided to do this](https://github.com/mysociety/alaveteli/pull/6528#discussion_r716576957) for the analogous `PublicBody` method. This will only get called when there's been a change to `url_name`, and given we don't rename users often I don't think it's much of an issue.

I agree that all this cache expiry stuff isn't great, but my aim here has been to at least reduce the complexity – even at the expense of a little extra processing – in order to hopefully extract a common request expiry mechanism down the line.

gbp: Ah yes. I missed https://github.com/mysociety/alaveteli/pull/6644/files#diff-9802ca3c9c4cf89904fd44bc114e35ebdf2c5dd3d5b645491e2b253e1afef29bR321

I think this is fine to merge then

gbp (approved):

PR #6658, opened 2021-11-11
Improve spacing of pro request navigation folders

garethrees:
On mobile devices the lack of bottom margin results in other UI elements
being directly attached to the dashboard navigation.

The two main places this happens are:

1. On the pro dashboard the announcement element is directly against the
   nav.
2. On the pro requests list the "All requests" header is directly
   against the nav.

## Screenshots

BEFORE

Screenshot 2021-11-11 at 11 58 58

Screenshot 2021-11-11 at 11 58 12

Screenshot 2021-11-11 at 11 58 02

AFTER

Screenshot 2021-11-11 at 11 58 47

Screenshot 2021-11-11 at 11 58 35

Screenshot 2021-11-11 at 11 57 28

### DIFF
--- a/app/assets/stylesheets/responsive/alaveteli_pro/_dashboard_layout.scss
+++ b/app/assets/stylesheets/responsive/alaveteli_pro/_dashboard_layout.scss
@@ -43,7 +43,7 @@ $dashboard-collapse: 50em; //where the dashboard layout needs to change
 
 .dashboard-folders,
 .dashboard-sub-folders {
-  margin: 0;
+  margin: 0 0 1em 0;
   padding: 0;
 }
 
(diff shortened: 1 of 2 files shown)


gbp (commented):
Looks like this has added space above the Draft sidebar anchor - was that intentional?

garethrees:
> Looks like this has added space above the Drafts sidebar anchor - was that intentional?

As in here?

Screenshot 2021-11-11 at 11 58 47

Yeah intentional as they're separate `<ul>`s. I think the "drafts" folder is conceptually different than a filter of "all requests", so feels like it should be its own thing. I thought that was originally intended given the distinct lists.

Issue #6624, opened 2021-11-01
Create /tor page for redirecting blocked signups


garethrees:
* Plain text response
* `403` status code
* long cache headers
* Translated text, something like:

```
Signups from Tor have been blocked due to extensive misuse. Please contact us if this is a problem for you.
```

Don't include the contact form URL because if its a genuine user they'll find it. Don't want to encourage abuse of that too.

PR #6666, opened 2021-11-15
[#6624] Add /tor route for blocked Tor signups

garethrees:
Adds an action that renders a simple message stating that signups from
Tor have been blocked. Alaveteli itself doesn't use this action.
Instead, use a webserver-level redirect to `/tor` if this blocking is in
place.

Opted to add to `UserController` since it's conceptually related to
`UserController#signup`.

Fixes https://github.com/mysociety/alaveteli/issues/6624

Screenshot 2021-11-15 at 11 42 59

```
$ curl -I http://192.168.56.30:3000/tor
HTTP/1.1 403 Forbidden
Content-Type: text/html; charset=utf-8
Vary: Cookie
Cache-Control: max-age=86400, public
X-Request-Id: 7fd8e240-1860-4f15-a003-259dc70f9dac
X-Runtime: 0.829214
X-Frame-Options: sameorigin
X-Content-Type-Options: nosniff
X-XSS-Protection: 1
X-Permitted-Cross-Domain-Policies: none
Connection: close
Server: thin
```

### DIFF
--- a/app/controllers/user_controller.rb
+++ b/app/controllers/user_controller.rb
@@ -180,6 +180,17 @@ class UserController < ApplicationController
     render action: :sign
   end
 
+  # A webserver level redirect can be used to redirect from the signup action to
+  # prevent spam signups from Tor.
+  def tor
+    long_cache
+
+    msg = _('Signups from Tor have been blocked due to extensive misuse. ' \
+            'Please contact us if this is a problem for you.')
+
+    render plain: msg, status: :forbidden
+  end
+
   def ip_rate_limiter
     @ip_rate_limiter ||= AlaveteliRateLimiter::IPRateLimiter.new(:signup)
   end
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -245,6 +245,9 @@ Rails.application.routes.draw do
         :via => :get
   match '/profile/sign_up' => 'user#signup',
         :as => :signup, :via => :post
+
+  match '/tor' => 'user#tor', via: :get
+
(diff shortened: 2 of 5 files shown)


gbp (commented):
This works fine. Questioning if this should be in core or in the theme. If in core maybe we should provide re-users sample config or some documentation on alaveteli.org

garethrees:
> This works fine. Questioning if this should be in core or in the theme.

We want to enable this on all our hosted Alavetelis so easier in core. Doesn't really affect others who don't want to enable it.

Pushed an extra change (3c72a5f) to add it to `robots.txt`.

> If in core maybe we should provide re-users sample config or some documentation on alaveteli.org

@sagepe what's the config for blocking signups from Tor like? Is it something we could provide as an easy-to-include config file, or is there a bit more to it?

sagepe:
> ...what's the config for blocking signups from Tor like? Is it something we could provide as an easy-to-include config file, or is there a bit more to it?

I'm not completely certain yet; I will need to set aside some time to look at this. I think my original idea might need to be changed a bit. I think it might be possible to do this in Nginx right at the edge, which would be better from our perspective anyway. Once we've got something that works we could document and share this but we'll also need to discuss how to automate the generation and regular updating of the list of exit nodes.

garethrees:
> I'm not completely certain yet; I will need to set aside some time to look at this…

Okay, I've [created a new issue](https://github.com/mysociety/alaveteli/issues/6693) for this so that we can leave notes as we figure out what to say.

Issue #6669, opened 2021-11-15
Comments are not reindexed when using bulk update


garethrees:
When comment visibility is changed through the bulk update action we call `update_all`, which is just a direct SQL command (no callbacks, etc):

https://github.com/mysociety/alaveteli/blob/0.39.1.8/app/controllers/admin_user_controller.rb#L115-L116

Need to explicitly reindex after the update:

```ruby
# Could do this…
comments = Comment.where(id: params[:comment_ids])
comments.update_all(visible: !params[:hide_selected])
comments.find_each(&:reindex)
```

Could probably tidy this up through an association extension.

PR #6670, opened 2021-11-15
[#6669] Reindex comments after bulk visibility change

garethrees:
## Relevant issue(s)

Fixes https://github.com/mysociety/alaveteli/issues/6669

## What does this do?

Minor refactoring
Reindex comments after bulk visibility change

## Why was this needed?

Comments must be reindexed after a visibility change otherwise they
appear in search-driven lists. The most obvious is the user profile
page.

## Implementation notes

The behaviour could be improved since with this implementation every
comment given is reindexed, rather than only those where the visibility
actually changes. This would require more extensive changes to not use
`update_all`.

### DIFF
--- a/app/controllers/admin_user_controller.rb
+++ b/app/controllers/admin_user_controller.rb
@@ -6,16 +6,16 @@
 
 class AdminUserController < AdminController
 
-  before_action :set_admin_user, :only => [ :show,
-                                            :edit,
-                                            :update,
-                                            :show_bounce_message,
-                                            :clear_bounce,
-                                            :clear_profile_photo ]
+  before_action :set_admin_user, only: %i[show
+                                          edit
+                                          update
+                                          show_bounce_message
+                                          clear_bounce
+                                          clear_profile_photo]
 
   before_action :clear_roles,
                 :check_role_authorisation,
-                :check_role_requirements, :only => [ :update ]
+                :check_role_requirements, only: %i[update]
 
   def index
     @query = params[:query].try(:strip)
--- a/app/controllers/admin_user_controller.rb
+++ b/app/controllers/admin_user_controller.rb
@@ -112,8 +112,13 @@ class AdminUserController < AdminController
   end
 
   def modify_comment_visibility
-    Comment.where(:id => params[:comment_ids]).
-      update_all(:visible => !params[:hide_selected])
+    desired_visibility = params[:hide_selected] ? false : true
+
+    Comment.
+      where(id: params[:comment_ids]).
+      where(visible: !desired_visibility).
+      find_each { |comment| comment.toggle!(:visible) }
+
     redirect_back(fallback_location: admin_users_url)
   end
 
--- a/app/controllers/admin_user_controller.rb
+++ b/app/controllers/admin_user_controller.rb
@@ -11,7 +11,8 @@ class AdminUserController < AdminController
                                           update
                                           show_bounce_message
                                           clear_bounce
-                                          clear_profile_photo]
+                                          clear_profile_photo
+                                          modify_comment_visibility]
 
   before_action :clear_roles,
                 :check_role_authorisation,
@@ -114,7 +115,8 @@ class AdminUserController < AdminController
   def modify_comment_visibility
     desired_visibility = params[:hide_selected] ? false : true
 
-    Comment.
+    @admin_user.
+      comments.
       where(id: params[:comment_ids]).
       where(visible: !desired_visibility).
       find_each { |comment| comment.toggle!(:visible) }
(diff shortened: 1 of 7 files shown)


spec/models/user_spec.rb
```diff
@@ -776,6 +776,22 @@ def create_user(options = {})
     end
   end
 
+  describe '#expire_comments' do
+    it 'calls reindex_request_events on all associated requests' do
+      user = FactoryBot.build(:user)
+
+      comment_1, comment_2 = double(:comment), double(:comment)
+
```
garethrees, line 784: > [Correctable] Do not use parallel assignment.

Disagree! Perfectly fine here.

app/controllers/admin_user_controller.rb
```diff
@@ -112,8 +112,9 @@ def clear_profile_photo
   end
 
   def modify_comment_visibility
-    Comment.where(:id => params[:comment_ids]).
-      update_all(:visible => !params[:hide_selected])
+    comments = Comment.where(id: params[:comment_ids])
+    comments.update_all(visible: !params[:hide_selected])
+    comments.find_each(&:reindex_request_events)
```
gbp, line 117: Ideally we shouldn't need to call this in the controller. Issue is we're calling `update_all` I would be tempted to switch to `#update` (or `#toggle`?)

```suggestion
    comments = Comment.where(id: params[:comment_ids], visible: params[:hide_selected])
    comments.find_each(&:toggle, :visible)
```

Would result in `#updated_at` changing and the callbacks being fired too.

gbp: Also this query should be scoped to the user.

gbp (commented):
Minor comments.

app/controllers/admin_user_controller.rb
garethrees: > I would be tempted to switch to `#update`…

Yeah, this passes all the specs, so that's a good move. 

```diff
diff --git a/app/controllers/admin_user_controller.rb b/app/controllers/admin_user_controller.rb
index 30e23efcd..f7842fb24 100644
--- a/app/controllers/admin_user_controller.rb
+++ b/app/controllers/admin_user_controller.rb
@@ -113,8 +113,11 @@ def clear_profile_photo

   def modify_comment_visibility
     comments = Comment.where(id: params[:comment_ids])
-    comments.update_all(visible: !params[:hide_selected])
-    comments.find_each(&:reindex_request_events)
+
+    comments.find_each do |comment|
+      comment.update!(visible: !params[:hide_selected])
+    end
+
     redirect_back(fallback_location: admin_users_url)
   end
```

> …or `#toggle`

We're not _toggling_ the visibility – we're setting a specific state for each comment based on the button pressed – so I don't think that would work.

We'd have to add an extra step to filter the comments which aren't already in the state requested by the button press, _then_ we could `toggle!` (n.b. need the bang version to save).

This passes:

```diff
diff --git a/app/controllers/admin_user_controller.rb b/app/controllers/admin_user_controller.rb
index 30e23efcd..fe6750c67 100644
--- a/app/controllers/admin_user_controller.rb
+++ b/app/controllers/admin_user_controller.rb
@@ -112,9 +112,11 @@ def clear_profile_photo
   end

   def modify_comment_visibility
-    comments = Comment.where(id: params[:comment_ids])
-    comments.update_all(visible: !params[:hide_selected])
-    comments.find_each(&:reindex_request_events)
+    comments = Comment.where(id: params[:comment_ids]).
+        where(visible: !params[:hide_selected])
+
+    comments.find_each { |comment| comment.toggle!(:visible) }
+
     redirect_back(fallback_location: admin_users_url)
   end
```

I quite like this, since it's now only updating records which actually need to be updated, rather than updating others already in the correct state as a consequence of the implementation details. It's a little messy as above, but with an extra fix to scope to the user's comments it's much better:

```ruby
  def modify_comment_visibility
    @admin_user.
      comments.
      where(id: params[:comment_ids]).
      where(visible: !params[:hide_selected]).
      find_each { |comment| comment.toggle!(:visible) }

    redirect_back(fallback_location: admin_users_url)
  end
```

garethrees: Doh, as soon as I hit enter I realise I made a mistake. The `toggle!` version above won't work because the of the way we negate only one of the params and that doesn't work in the same way with the `where`.

This is what works:

```ruby
    desired_visibility = params[:hide_selected] ? false : true

    comments =
      Comment.where(id: params[:comment_ids]).
      where(visible: !desired_visibility).
      find_each do |comment|
        comment.toggle!(:visible)
      end
```

gbp (approved):

Issue #6627, opened 2021-11-02
Only render `User#about_me` for logged in users


garethrees:
Might be one way of reducing spam users (#3180), since most of it looks like SEO junk.

PR #6688, opened 2021-11-25
[#6627] Reduce prominence of User#about_me

garethrees:
We get lots of spam signups which mainly populate the `about_me` with
links to some form of junk site. Many of these spam signups originate
through another spam user profile, so presumably they're searching for
particular terms in the spam text in order to find sites that can be
targeted.

This commit aims to reduce the attractiveness of Alaveteli to spammers
by hiding the `about_me` text from the wider web, except for when the
user has been manually confirmed as genuine.

We'll show the text to logged in users, unless the user being viewed has
been banned or closed as before.

Fixes https://github.com/mysociety/alaveteli/issues/6627.

### DIFF
--- a/app/controllers/user_controller.rb
+++ b/app/controllers/user_controller.rb
@@ -34,6 +34,7 @@ class UserController < ApplicationController
     set_view_instance_variables
     @same_name_users = User.find_similar_named_users(@display_user)
     @is_you = current_user_is_display_user
+    @show_about_me = show_about_me?
 
     set_show_requests if @show_requests
 
@@ -546,6 +547,15 @@ class UserController < ApplicationController
 
   def current_user_is_display_user
     @user.try(:id) == @display_user.id
+  end
+
+  def show_about_me?
+    return true if @is_you
+    return false unless @display_user.get_about_me_for_html_display.present?
+    return false unless @display_user.active?
+    return true if @display_user.confirmed_not_spam?
+    return true if @user
+    false
   end
 
   # Redirects to front page later if nothing else specified
--- a/app/views/user/_show_user_info.html.erb
+++ b/app/views/user/_show_user_info.html.erb
@@ -1,4 +1,4 @@
-<% if (@display_user.active? && !@display_user.get_about_me_for_html_display.empty?) || @is_you %>
+<% if @show_about_me %>
   <div class="user_about_me">
     <%= image_tag "quote-marks.png", :class => "comment_quote" %>
     <%= @display_user.get_about_me_for_html_display %>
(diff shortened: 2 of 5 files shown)


app/controllers/user_controller.rb
```diff
@@ -537,6 +538,15 @@ def current_user_is_display_user
     @user.try(:id) == @display_user.id
   end
 
+  def show_about_me?
```
garethrees, line 541: Code Climate is unhappy with the complexity and number of returns here. I somewhat agree, but this felt more readable than a big combination of `||`, `&&` and `!`s.

I should add a comment to the method that these statements are order-dependent.

gbp: If you change the code climate warning to wontfix then this PR will go green

garethrees: Yeah I know that – It's more that it's a valid criticism so wanted to sense-check what you thought and whether there's a nicer alternative approach.

app/controllers/user_controller.rb
```diff
@@ -537,6 +538,15 @@ def current_user_is_display_user
...
+    return true if @is_you
+    return false unless @display_user.get_about_me_for_html_display.present?
+    return false unless @display_user.active?
+    return true if @display_user.confirmed_not_spam?
+    return true if @user
+    false
+  end
```
gbp, line 548: Can be simplified as:

```suggestion
  def show_about_me?
    return true if @is_you
    return false unless @display_user.get_about_me_for_html_display.present?
    return false unless @display_user.active?
    @display_user.confirmed_not_spam?
  end
```

gbp (changes requested):

app/controllers/user_controller.rb
gbp: Or maybe a spec if missing?

garethrees: Ah needed this to check that final `return true if @user` – always want to show the `about_me` when logged in, unless there's a specific reason not to (i.e. one of the first 3 conditions).

```diff
diff --git a/spec/controllers/user_controller_spec.rb b/spec/controllers/user_controller_spec.rb
index 3e0600ded..75e6c62c9 100644
--- a/spec/controllers/user_controller_spec.rb
+++ b/spec/controllers/user_controller_spec.rb
@@ -621,7 +621,7 @@ def make_request
     end

     context 'when logged in as a different in user' do
-      before { user.update(about_me: 'x') }
+      before { user.update(about_me: 'x', confirmed_not_spam: false) }
       before { sign_in(FactoryBot.create(:user)) }
```

PR #6692, opened 2021-11-26
Adding videos from TICTeC RTK day

MyfanwyNixon:
- Added the TICTeC RTK videos
- Changed the titling from the name of the service to the language that it's in, to reflect user needs rather than our own subdivisions


garethrees:
Excellent, thanks!

Issue #6487, opened 2021-08-20
Canned CensorRule redaction replacements


garethrees:
When redacting requests there are a handful of phrases we frequently use:

* `[Personally Identifiable Information removed]`
* `[Extraneous material removed]`
* `[Potentially defamatory material removed]`

It would be good to incorporate these canned phrases into the admin interface to both save typing, and use it as a tiny teaching tool for newer admins by way of it showing some of the use cases for the feature.

To avoid too much work, we can just use a select field that loads in the canned replacements, and use [javascript to update the input field if one is selected](https://stackoverflow.com/a/10301561). It doesn't matter that these can be changed once inserted into the text field as we don't need a hard set of options – this is just a helper for common cases. Changing the selected option should change the text field text irrespective of what's currently in it.

![Screenshot 2021-08-20 at 11 27 20](https://user-images.githubusercontent.com/282788/130220304-583c1e25-67f3-48dc-807a-86735097acae.png)

![Screenshot 2021-08-20 at 11 27 28](https://user-images.githubusercontent.com/282788/130220264-4eb3933e-2e70-4f33-a2e9-bd6a00fdb28b.png)

![Screenshot 2021-08-20 at 11 35 57](https://user-images.githubusercontent.com/282788/130221094-4fdb3899-0634-44b9-a59b-4f346200c009.png)

Might want to add a prompt like "canned replacements" to the select field with an empty value that doesn't update, but no need to go out of our way for this.

It should be configurable per install, but only at a theme level since these won't change frequently. Here's a spike:

```ruby
# CORE/app/models/censor_rule.rb
DEFAULT_CANNED_REPLACEMENTS = [
  _('[Personally Identifiable Information removed]'),
  _('[Extraneous material removed]'),
 # etc
]

def self.canned_replacements
  @canned_replacements ||= DEFAULT_CANNED_REPLACEMENTS
end

def self.canned_replacements=(replacements)
  @canned_replacements = replacements
end

# THEME/lib/model_patches.rb
custom_replacements = [
  # …
]
CensorRule.canned_replacements = CensorRule.canned_replacements + custom_replacements
```

garethrees:
Just came across [`<datalist>`](https://blog.corsego.com/interesting-html-tags) which made this [super easy](https://github.com/mysociety/alaveteli/pull/6700)!

RichardTaylor:
There's a list of commonly used replacement text introduced via censor rules at:

[...]

Issue #6699, opened 2021-12-02
Canned User#ban_text reasons


RichardTaylor:
The reason for this proposal is to improve record keeping on why we have banned users, enabling better transparency reporting, and management / oversight.  

Ban reasons:

* Misuse of the service [in breach of the house rules]
* Account suspended as it has been closed and anonymised (Currently this can be a manual, or an automated process, slightly different actions can be taken depending on circumstances)
* Spamming
* Account created to evade ban on another account
* Account suspended to prompt user to engage with admin team advice prior to further use of the service

A review of ban reasons currently in use might help spotting any which don't come into the above categories. 

The downside would be the loss of specifically tailored messages eg. 

"You have persistently misused the service to send non-FOI correspondence despite repeated warnings" 

Having both categories and an optional free-text embellishment is an option but it gets complicated. 

Issue derives from:
https://github.com/mysociety/whatdotheyknow-theme/issues/925#issuecomment-984443864
https://github.com/mysociety/whatdotheyknow-theme/issues/938#issuecomment-984996006

mdeuk:
> Having both categories and an optional free-text embellishment is an option but it gets complicated.

My thinking was that it could work in the same way as setting a prominence reason on a message, e.g. dropdown containing a list of standard options, followed by a free-text message box where the tailored message could be set.

I'd assume, for the most part, that the message could be optional, meaning for most it might actually make this administrative task slightly more efficient, as it would become 'point and click'.

RichardTaylor:
>My thinking was that it could work in the same way as setting a prominence reason on a message

So I think this is an extension of the proposal to include, optionally, sending a message to the user informing them of the ban from within the system.

If we follow the pattern for one-click hiding of requests, the message content would expand on the reason for the ban, but that wouldn't be recorded in the system or shown publicly.

I'll edit the title of this issue.

Related:  Bring user support and takedown requests within the system #419

garethrees:
> My thinking was that it could work in the same way as setting a prominence reason on a message, e.g. dropdown containing a list of standard options, followed by a free-text message box where the tailored message could be set.

The first, easiest, step here would be to use the same implementation mechanism as https://github.com/mysociety/alaveteli/issues/6487.

Wouldn't handle anything in addition to making it easier to create standardised text, but that's quite a big leg up in itself.

garethrees:
> The first, easiest, step here would be to use the same implementation mechanism as #6487.

[...]

PR #6700, opened 2021-12-03
[#6487] Canned censor rule replacements

garethrees:
## Relevant issue(s)

Fixes https://github.com/mysociety/alaveteli/issues/6487

## What does this do?

Adds canned Censor Rule replacements

## Why was this needed?

We frequently write the same text into the `CensorRule#replacement` form
field, so this commit adds a set of defaults that can be selected
through a `<datalist>` element.

Reusers can add to the defaults or set their own in the theme:

    # Add a reason
    CensorRule.canned_replacements << '[A new reason]'

    # Use a custom set of replacements
    CensorRule.canned_replacements = [
      '[A new reason]',
      '[Another reason]'
    ]

## Implementation notes

Used a `cattr_accessor` so haven't added specs. Open to thoughts.

## Screenshots

Screenshot 2021-12-03 at 17 49 42

Screenshot 2021-12-03 at 17 49 47

Screenshot 2021-12-03 at 17 49 52

Screenshot 2021-12-03 at 17 50 06

No placeholder or datalist when there aren't any canned replacements:

Screenshot 2021-12-10 at 17 37 27

## Notes to reviewer

Screenshots a tiny bit out of date:

* I opted to make _both_ fields wider after taking them
* I capitalised "**S**elect or add your own"

### DIFF
--- a/app/models/censor_rule.rb
+++ b/app/models/censor_rule.rb
@@ -23,7 +23,16 @@
 # Email: hello@mysociety.org; WWW: http://www.mysociety.org/
 
 class CensorRule < ApplicationRecord
+  DEFAULT_CANNED_REPLACEMENTS = [
+    _('[Personally Identifiable Information removed]'),
+    _('[Name removed]'),
+    _('[Extraneous material removed]'),
+    _('[Potentially defamatory material removed]'),
+    _('[Extraneous and potentially defamatory material removed]')
+  ].freeze
+
   include AdminColumn
+
   belongs_to :info_request,
              :inverse_of => :censor_rules
   belongs_to :user,
@@ -43,6 +52,10 @@ class CensorRule < ApplicationRecord
           user_id: nil,
           public_body_id: nil)
   }
+
+  cattr_accessor :canned_replacements,
+                 instance_writer: false,
+                 default: DEFAULT_CANNED_REPLACEMENTS.dup
 
   def apply_to_text(text_to_censor)
     return nil if text_to_censor.nil?
--- a/app/views/admin_censor_rule/_form.html.erb
+++ b/app/views/admin_censor_rule/_form.html.erb
@@ -39,7 +39,23 @@
 <div class="control-group">
   <label for="censor_rule_replacement" class="control-label">Replacement</label>
   <div class="controls">
-    <%= text_field 'censor_rule', 'replacement', class: 'span6'  %>
+    <% placeholder =
+      if @censor_rule.canned_replacements.any?
+        'Select or add your own'
+      end
+    %>
+
+    <%= text_field 'censor_rule',
+                   'replacement',
+                   class: 'span6',
+                   list: 'canned-replacements',
+                   autocomplete: 'off',
+                   placeholder: placeholder %>
+
+    <datalist id="canned-replacements">
+      <%= options_for_select(@censor_rule.canned_replacements) %>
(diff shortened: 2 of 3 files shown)


app/views/admin_censor_rule/_form.html.erb
```diff
@@ -39,7 +39,22 @@
...
+      end
+    %>
+
+    <%= text_field 'censor_rule',
+                   'replacement',
+                   class: 'span6',
+                   list: 'canned-replacements',
+                   placeholder: placeholder %>
+
+    <datalist id="canned-replacements">
```
gbp, line 54: TIL datalist

app/models/censor_rule.rb
```diff
@@ -44,6 +53,10 @@ class CensorRule < ApplicationRecord
           public_body_id: nil)
   }
 
+  cattr_accessor :canned_replacements,
+                 instance_writer: false,
+                 default: DEFAULT_CANNED_REPLACEMENTS.dup
```
gbp, line 58: Could add a spec in `alavetelitheme` showing these are customisable

app/views/admin_censor_rule/_form.html.erb
gbp: Not seeing the dropdown arrow on Firefox.

I have to either start typing or click anywhere on the input (while it is focused) to see the options.

Also shows any previous values which have been entered into this field.

app/views/admin_censor_rule/_form.html.erb
```diff
@@ -39,9 +39,27 @@
...
+      <%= options_for_select(@censor_rule.canned_replacements) %>
+    </datalist>
+
     <div class="help-block">
-      put it in <strong>[square brackets]</strong>, e.g. [personal information removed]. applies to text in emails and HTML conversions of binaries; binaries themselves must stay the same length and the replacement is just a bunch of 'x's
+      put it in <strong>[square brackets]</strong>, e.g. [personal information
```
gbp, line 59: Should the `[personal information removed]` example match one of the defaults?

gbp (commented):
Seems to work.

Just need to consider usability issues. Not sure if I have ever encountered a datalist element so not sure if there is a styling issue with the missing dropdown arrow.

Also are other attributes we can include on the HTML tag to prevent previous entered values from being suggested?

app/views/admin_censor_rule/_form.html.erb
garethrees: > Not seeing the dropdown arrow on Firefox.

I'm sure they'll resolve this at some point, or FF users will get used to its behaviour for this field. Not going to put further energy into it for admin interface, but if we come across a trivial fix would be happy to consider it.

> Also shows any previous values which have been entered into this field.

Added `autocomplete: 'off'` to prevent.

app/views/admin_censor_rule/_form.html.erb
garethrees: Good idea. Added in 580a1d6

app/models/censor_rule.rb
garethrees: Hmm yeah, not sure where the best place to document this is. Going to add [more like it](https://github.com/mysociety/alaveteli/issues/6699) so will consider in future when we have a bit more idea of where's appropriate.

PR #6702, opened 2021-12-06
[#4326] Improve event logging for admin comments updates 

### LINKED ISSUE #4326: [ERROR] admin_general#index (NoMethodError) "undefined method `in_time_zone' for nil:NilClass"
Marking a request as "needing attention" through the admin interface breaks the summary.

The error occurs because a `report_comment` event doesn't get created, and the dashboard calls `last_reported_at`.

When updating `attention_requested` through `AdminCommentController#update`, we should call `Comment#report!`.

```
A NoMethodError occurred in admin_general#index:

undefined method `in_time_zone' for nil:NilClass
app/helpers/date_time_helper.rb:39:in `simple_date_html'


Request:

* URL : https://www.whatdotheyknow.com/admin
* HTTP Method: GET
* IP address : FILTERED
* Parameters : {"controller"=>"admin_general", "action"=>"index"}
* Timestamp : 2017-10-27 10:24:39 +0100
* Server : FILTERED
* Rails root : /FILTERED/www.whatdotheyknow.com/alaveteli-2017-10-26T19-38-38
* Process: 19257

Backtrace:

app/helpers/date_time_helper.rb:39:in `simple_date_html'
app/helpers/date_time_helper.rb:26:in `simple_date'
app/views/admin_general/_to_do_list.html.erb:44:in `block in _app_views_admin_general__to_do_list_html_erb___2208061734204758292_70011832342620'
app/views/admin_general/_to_do_list.html.erb:15:in `each'
app/views/admin_general/_to_do_list.html.erb:15:in `_app_views_admin_general__to_do_list_html_erb___2208061734204758292_70011832342620'
app/views/admin_general/index.html.erb:169:in `_app_views_admin_general_index_html_erb__233494462981526963_70011848313640'
```

gbp:
## Relevant issue(s)

Fixes #4326

## What does this do?

Instead of creating a single event when a comment is updated split into three distinct events for changes to the comment's body, visibility and if attention requested.

## Why was this needed?

Reporting a comment in the admin UI would cause the `/admin` page to break due to there being no `report_comment` event.


app/controllers/admin_comment_controller.rb
```diff
@@ -42,21 +42,24 @@ def update
...
-                      :old_visible => old_visible,
-                      :visible => @comment.visible,
-                      :old_attention_requested => old_attention,
-                      :attention_requested => @comment.attention_requested })
+
+      @comment.info_request.log_event(
+        update_type, {
+          comment_id: @comment.id,
+          editor: admin_current_user,
+          old_body: @old_body,
```
garethrees: This has changed to an instance variable, which I don't think is defined?

app/controllers/admin_comment_controller.rb
```diff
@@ -42,21 +42,24 @@ def update
...
+          body: @comment.body,
+          old_visible: @old_visible,
```
garethrees: This has changed to an instance variable, which I don't think is defined?

app/controllers/admin_comment_controller.rb
```diff
@@ -42,21 +42,24 @@ def update
...
+          visible: @comment.visible,
+          old_attention_requested: @old_attention,
```
garethrees: This has changed to an instance variable, which I don't think is defined?

garethrees: Woo this one at least actually fails a spec:

```
$ bin/rspec spec/controllers/admin_comment_controller_spec.rb

Randomized with seed 62448
......................F...

Failures:

  1) AdminCommentController PUT update on valid data submission the attention_requested flag is the only change captures the old and new attention_requested values
     Failure/Error:
       expect(most_recent_event.params).
         to include(old_attention_requested: false)

       expected {:attention_requested => true, :body => "This a wise and helpful annotation.", :comment_id => 24, :editor => "*unknown*", :old_attention_requested => nil, :old_body => nil, :old_visible => nil, :visible => true} to include {:old_attention_requested => false}
       Diff:
       @@ -1,8 +1,15 @@
       -:old_attention_requested => false,
       +:attention_requested => true,
       +:body => "This a wise and helpful annotation.",
       +:comment_id => 24,
       +:editor => "*unknown*",
       +:old_attention_requested => nil,
       +:old_body => nil,
       +:old_visible => nil,
       +:visible => true,

     # ./spec/controllers/admin_comment_controller_spec.rb:187:in `block (5 levels) in <top (required)>'
     # /home/vagrant/bundle/ruby/2.7.0/gems/webmock-3.14.0/lib/webmock/rspec.rb:37:in `block (2 levels) in <top (required)>'

Finished in 11.23 seconds (files took 6.42 seconds to load)
26 examples, 1 failure

Failed examples:

rspec ./spec/controllers/admin_comment_controller_spec.rb:185 # AdminCommentController PUT update on valid data submission the attention_requested flag is the only change captures the old and new attention_requested values

Randomized with seed 62448
```

app/controllers/admin_comment_controller.rb
```diff
@@ -81,7 +78,7 @@ def set_comment
     @comment = Comment.find(params[:id])
   end
 
-  def comment_hidden?(old_visibility, old_body)
-    !@comment.visible && old_visibility && old_body == @comment.body
+  def comment_hidden?
+    @comment.previous_changes[:visible] && !@comment.previous_changes[:body]
```
garethrees: We now get a `hide_comment` event when you change `visible` from `false` to `true`. Given the next commit, I think we should preserve the existing behaviour here. Could either re-add `!@comment.visible` or change to:

```ruby
@comment.previous_changes[:visible] == [true, false] &&
  !@comment.previous_changes[:body]
```

app/controllers/admin_comment_controller.rb
```diff
@@ -38,24 +38,39 @@ def update
...
-          editor: admin_current_user,
-          old_body: @comment.body_previously_was,
-          body: @comment.body,
-          old_visible: @comment.visible_previously_was,
-          visible: @comment.visible,
-          old_attention_requested: @comment.attention_requested_previously_was,
-          attention_requested: @comment.attention_requested
-        }
-      )
+      if @comment.previous_changes[:visible]
```
garethrees, line 52: Do we still want a `hide_comment` event when going from hidden to visible? I'm not against it, but it's a change from what happens now so we might want to consider what we do for other hidable records and whether we need to do anything with historical events.

app/controllers/admin_comment_controller.rb
```diff
@@ -38,24 +38,39 @@ def update
...
+          'hide_comment', {
+            comment_id: @comment.id,
+            editor: admin_current_user,
+            old_visible: @comment.visible_previously_was,
+            visible: @comment.visible
+          }
+        )
+      end
+
+      if @comment.previous_changes[:attention_requested]
```
garethrees, line 63: Similar to `visible`, we currently only create `report_comment` when `attention_requested` goes from `false` to `true`, but not the other way around.

app/controllers/admin_comment_controller.rb
```diff
@@ -38,24 +38,39 @@ def update
     end
 
     if @comment.update(comment_params)
-      update_type = if comment_hidden?
-        'hide_comment'
-      else
-        'edit_comment'
+      if @comment.previous_changes[:body]
```
garethrees, line 41: One really minor improvement we could make here is extracting the details to private methods:

```ruby
log_edit_comment(@comment) if @comment.previous_changes[:body]
log_hide_comment(@comment) if @comment.previous_changes[:visible]
log_report_comment(@comment) if @comment.previous_changes[:attention_requested]
```

garethrees (changes requested):
This is great! 😎 

There's some behaviour changes which have been introduced over and above the creation of independent events which I think we should reflect on.

Issue #6341, opened 2021-07-02
Upgrade to Rails 7.0


garethrees:
Guessing at a Q4 2021 / Q1 2022 release date.

- [x] Requires https://github.com/mysociety/alaveteli/issues/6011
- [x] Requires https://github.com/mysociety/alaveteli/issues/5440
- [x] Requires https://github.com/mysociety/alaveteli/issues/5224 
- [x] Requires https://github.com/mysociety/alaveteli/issues/5225
- [x] Requires a new release of globalize with this PR https://github.com/globalize/globalize/pull/792
- [x] Requires https://github.com/mysociety/alaveteli/pull/6944
- [x] #5485 
- [ ] #6949

gbp:
- [ ] Requires #5224 
- [ ] Requires #5225

gbp:
Looks like this will be released before Christmas https://twitter.com/rafaelfranca/status/1467974458015158273

gbp:
Out now https://rubyonrails.org/2021/12/15/Rails-7-fulfilling-a-vision

gbp:
Requires a new release of globalize with this PR https://github.com/globalize/globalize/pull/792

PR #6703, opened 2021-12-07
[#6341] Prepare for Rails 7 upgrade

gbp:
## Relevant issue(s)

Connected to #6341 

## What does this do?

Bumps Rails in our next Gemfile

## Why was this needed?

Lets see what specs fail


garethrees (approved):
haven't run but all looks sensible

PR #6705, opened 2021-12-07
[#6532] Migrate FoiAttachment to use ActiveStorage

### LINKED ISSUE #6532: Store FoiAttachment#body in Active Storage
Requires https://github.com/mysociety/alaveteli/issues/6530.

* Migrate extracting and storing the attachment to `attach` provided by Active Storage.
* Files on disk should continue to be stored in `APP_ROOT/cache/attachments_production` under directories namespaced based on the first few characters of the hexdigest. Looks like this is [possible](https://github.com/rails/rails/issues/32790#issuecomment-642128038) if a little bit of a hack.
* Once this ticket is completed, create an issue for migrating to a non-namespaced directory.

gbp:
## Relevant issue(s)

Fixes #6532

## What does this do?

Migrate FoiAttachment to use ActiveStorage

## Why was this needed?

When the `ActiveStorage::Blob` file isn't attached then use the old file
store.

### DIFF
--- a/app/models/foi_attachment.rb
+++ b/app/models/foi_attachment.rb
@@ -90,8 +90,8 @@ class FoiAttachment < ApplicationRecord
       content_type: content_type
     )
 
-    update_display_size!
     @cached_body = d.force_encoding("ASCII-8BIT")
+    update_display_size!
   end
 
   # raw body, encoded as binary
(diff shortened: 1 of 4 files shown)


gbp:
Discussed with @garethrees last week and we agreed this was good to merge

PR #6734, opened 2022-01-14
Activestorage debug

gbp:

### DIFF
--- a/spec/support/xapian_index.rb
+++ b/spec/support/xapian_index.rb
@@ -9,7 +9,9 @@ def destroy_and_rebuild_xapian_index(terms = true, values = true, texts = true, 
     ActsAsXapian.writable_init
     ActsAsXapian.writable_db.close
   end
+  load_raw_emails_data
   parse_all_incoming_messages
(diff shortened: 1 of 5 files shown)


spec/support/xapian_index.rb
```diff
@@ -9,48 +9,20 @@ def destroy_and_rebuild_xapian_index(terms = true, values = true, texts = true,
...
   # safe_rebuild=true, which involves forking to avoid memory leaks, doesn't work well with rspec.
   # unsafe is significantly faster, and we can afford possible memory leaks while testing.
   models = [PublicBody, User, InfoRequestEvent]
   ActsAsXapian.destroy_and_rebuild_index(models, verbose=false, terms, values, texts, safe_rebuild=false)
 end
 
 def update_xapian_index
-  setup_xapian_index unless @xapian_setup
-  @xapian_setup = true
+  if $xapian_index_setup.nil?
```
gbp, line 22: I switched to global variable here and rebased - but this has cause specs to be red. Revert and investigate next week

gbp:
We have removed the Xapian spec setup in #6747

PR #6735, opened 2022-01-14
Remove unnecessary silly comment fixture use

garethrees:
These specs don't need to use the fixture "world", so convert them to
use factories.

Can't remove `silly_comment` just yet because the remaining uses are
linked to the `InfoRequestEvent` records that reference it.


gbp (approved):
lgtm

PR #6736, opened 2022-01-14
Remove sensible comment fixture

garethrees:
Not used anywhere. We'll lose some sample data, but getting rid of
fixtures seems like a bigger win.

### DIFF
--- a/spec/fixtures/comments.yml
+++ b/spec/fixtures/comments.yml
@@ -22,13 +22,3 @@ silly_comment:
   info_request_id: "101"
   user_id: "2"
   created_at: 2008-08-13 01:25:17.486939
-
-sensible_comment:
-  id: 2
-  body: This a wise and helpful annotation.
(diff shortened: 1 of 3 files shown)


garethrees:
Failing because the spec for [`Statistics::Leaderboard#all_time_commenters`](https://github.com/mysociety/alaveteli/blob/0.40.0.0/spec/models/statistics/leaderboard_spec.rb#L47-L54) relies on the fixture data. There's a `FIXME` to move that spec to factories anyway, so that's the resolution path. Shouldn't be that difficult to refactor.

gbp (approved):
Always happy to remove fixture data from specs where we can. If we get to a point where we need sample data for development then we could use a strategy like: https://dalezak.medium.com/a-strategy-for-rails-for-better-structured-seeds-for-all-and-specific-environments-16855367932c

PR #6747, opened 2022-01-21
Remove custom test Xapian db setup

gbp:
## What does this do?

Remove custom test Xapian db setup

## Why was this needed?

We had been attempting to restore the test Xapian db to a know state
between each spec but this is causing issues making the some specs
flaky.

The purpose of this setup was to prevent the spec suite from needing to
recreate the database from scratch for each spec. As we use Xapian
extensively throughout the application this reduced the total time to
run the spec suite up significantly but unfortunately there are still
some confusing issues so lets remove for now so we're unblocked. We can
attempt a setup like this again in the future.


gbp:
This is basically what @garethrees and I came up with together to get the specs running consistently so #6532 passes CI

garethrees (approved):

PR #6757, opened 2022-01-26
[#6756] External request handling for event digest

### LINKED ISSUE #6756: Alert track failing for external (anonymous user) requests
Alert track daemon exiting with:

```
./app/views/track_mailer/event_digest.text.erb:23:in `block (2 levels) in _app_views_track_mailer_event_digest_text_erb___217851957946
830175_76540': undefined method `html_safe' for nil:NilClass (NoMethodError)
```

This seems to be due to an old request from 2015 which was made by an anonymous user has been updated triggering an email alert.

gbp:
## Relevant issue(s)

Fixes #6756

## What does this do?

Handle external requests in event digest

## Why was this needed?

If a external request, which doesn't have an associated user record, is
updated and matches a TrackThing then an exception will be raise causing
the alert track daemon to fail.

This change check to see if the request is external and substitute "An
anonymous user" in place of the user name. This is the same string which
appears on the request#show view.

### DIFF
--- a/app/views/track_mailer/event_digest.text.erb
+++ b/app/views/track_mailer/event_digest.text.erb
@@ -17,10 +17,15 @@
 
         # e.g. Julian Burgess sent a request to Royal Mail Group (15 May 2008)
         if event.event_type == 'response'
+          if event.info_request.is_external?
+            user_name = _('An anonymous user')
+          else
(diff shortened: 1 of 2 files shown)


crowbot (commented):
Looks alright to me fwiw

PR #6762, opened 2022-02-01
Fixed overflowing problem on "Get help on refusal"

### LINKED ISSUE #6732: Refusal reasons overflowing container
Not sure what's caused this, since it's a relatively new feature and I don't think anything in particular has changed on this page since it was launched without this issue present.

Screenshot 2022-01-14 at 09 07 21

lucascumsille:
Fixes #6732
Changed the property from max-height to height, apparently there is a bug in chrome where the overflow:hidden on a fieldset element does not work properly.
Sorry @gbp I'm not really sure who should I ask for review, now that @garethrees is on leave.
Screenshot 2022-02-01 at 12 27 38

Screenshot 2022-02-01 at 12 27 53

### DIFF
--- a/app/assets/stylesheets/responsive/_wizard_layout.scss
+++ b/app/assets/stylesheets/responsive/_wizard_layout.scss
@@ -80,12 +80,11 @@
    */
 
   [data-block="exemption"] fieldset {
-    max-height: 16em;
+    height: 16em;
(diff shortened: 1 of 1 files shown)


gbp (approved):
Nice one.

This appears to work correctly on Chrome, Firefox and Safari now

PR #6773, opened 2022-02-10
[#6769] Cache from email address on incoming message

### LINKED ISSUE #6769: Improve/limit amount of data download from S3
Now that we're storing raw emails in S3 we need to consider how often we access raw emails. I've done some quick analyst of the production log of which actions cause raw emails to be downloaded from S3. 

```
AdminIncomingMessageController#redeliver
AdminRawEmailController#show
FollowupsController#create
FollowupsController#new
FollowupsController#preview
RequestController#show
UserController#show
```

gbp:
## Relevant issue(s)

Fixes #6769

## What does this do?

Cache from email address on incoming message

## Why was this needed?

This will reduce disk IO operations or cloud storage downloads by not
accessing the raw email when users perform common actions such as when
composing followup or internal reviews.

## Implementation notes

It [was suggested](https://github.com/mysociety/alaveteli/issues/6769#issuecomment-1034749314) adding this column to the `RawEmail` model but I decided to add to `IncomingMessage` for two reasons:
1. there are also corresponding `#mail_from`, `#safe_mail_from` and `#mail_from_domain` methods already on `IncomingMessage` - seporate PR incoming to rename these existing fields for consistency. 
2. `RawEmail` only has `id` and timestamp columns and nothing else.


gbp:
@garethrees and I have discussed this and we are happy to merge as is. 

This is basically the solution which is the simplest and quickest way to get this done without too many changes. In the future there is an argument to be made for either:

1. Refactor to migrating some `IncomingMessage` columns to the `RawEmail` model

The fact `RawEmail` only has `id` seems like a huge code smell, and lots of the refactoring that introduced those delegate calls was intended to push the complexity down to the correct layer.

2. Migrating the `has_one_attached` from `RawEmail` to `IncomingMessage`, and removing the `RawEmail` model

It could be viewed that the main reason for the `RawEmail` was to encapsulated the writing and reading of the data from disk. This is now done via `ActiveStorage`'s `has_one_attached` method so it there much point of the model at all anymore.

PR #6774, opened 2022-02-10
[#6769] Refactor mail from attributes

gbp:
## Relevant issue(s)

Refactor after #6773

## What does this do?

Refactor mail from attributes

## Why was this needed?

Consistency and will save confusion about what each attribute return.


garethrees (approved):
Nice!

Issue #6556, opened 2021-10-06
Show who made each edit on public body admin pages


RichardTaylor:
Related:

* #6161 
* #1294

RichardTaylor:
A new admin joined WhatDoTheyKnow and they signed their "Comment for this edit" - working around the absence of this feature.

mdeuk:
> A new admin joined WhatDoTheyKnow and they signed their "Comment for this edit" with their name - working around the absence of this feature.

I agree, this would be a useful feature. I usually sign my changes, but having the audit history  right in front of us would be helpful for looking back at things in the future.

PR #6776, opened 2022-02-11
[#6556] Show who made each edit on public body admin pages

garethrees:
Show who made each edit on public body admin pages
We already store `PublicBody::Version#last_edit_editor`, so this just
presents it alongside the relevant comment related to that version.

I've handled the case where we can't find a `User` record for an editor
since a `User#url_name` may change if an admin requests anonymisation
and then the link would break. Ideally we'd use an association between
the records rather than having to do a lookup based on a `String`, but
this adds some value now with what we have.

Fixes https://github.com/mysociety/alaveteli/issues/6556

Screenshot 2022-02-11 at 16 55 24

## Notes to reviewer

I was considering whether to handle `version.last_edit_editor` being `nil` or `blank?`, but that doesn't appear to be a possible case according to WDTK data:

```ruby
PublicBody::Version.where(last_edit_editor: nil).count
# => 0
PublicBody::Version.where(last_edit_editor: '').count
# => 0
```

### DIFF
--- a/app/models/public_body.rb
+++ b/app/models/public_body.rb
@@ -217,6 +217,10 @@ class PublicBody < ApplicationRecord
       end
       changes
     end
+
+    def editor
+      User.find_by(url_name: last_edit_editor)
+    end
   end
 
   # Public: Search for Public Bodies whose name, short_name, request_email or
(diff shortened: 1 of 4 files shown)


gbp (approved):
Nice. Would've split 614b64f158adee6519110414a8a1858aef4cdd0e as the refactoring of the view made the change more confusing but all seems to work well.

Issue #6531, opened 2021-09-27
Store RawEmail#data in Active Storage


garethrees:
Requires https://github.com/mysociety/alaveteli/issues/6530.

* Migrate extracting and storing the raw email to `attach` provided by Active Storage.
* Probably want to make sure we're not accessing the direct file representation outside of `RawEmail` if at all possible.
* Files on disk should continue to be stored in `APP_ROOT/../raw_emails` under directories namespaced based on the request ID. Looks like this is [possible](https://github.com/rails/rails/issues/32790#issuecomment-642128038) if a little bit of a hack.
* Once this ticket is completed, create an issue for migrating to a non-namespaced directory.

gbp:
So I've done some changes and testing locally to ensure we can handle S3 being down. To replicate it being down I changed my work on https://github.com/mysociety/alaveteli/pull/6539 with:

```diff
diff --git a/Gemfile b/Gemfile
index 61e9a9def..0e3dfef56 100644
--- a/Gemfile
+++ b/Gemfile
@@ -161,10 +161,10 @@ gem 'alaveteli_features', :path => 'gems/alaveteli_features'
 gem 'aws-sdk-s3', require: false
 gem 'azure-storage', require: false
 gem 'google-cloud-storage', '~> 1.11', require: false
+gem 'webmock', '~> 3.14.0'
 
 group :test do
   gem 'fivemat', '~> 1.3.7'
-  gem 'webmock', '~> 3.14.0'
   gem 'simplecov', '~> 0.17.1'
   gem 'simplecov-lcov', '~> 0.7.0'
   gem 'capybara', '~> 3.35.3'
diff --git a/config/initializers/webmock.rb b/config/initializers/webmock.rb
new file mode 100644
index 000000000..f29a97915
--- /dev/null
+++ b/config/initializers/webmock.rb
@@ -0,0 +1,6 @@
+require 'webmock'
+include WebMock::API
+
+WebMock.enable!
+
+stub_request(:put, 'wdtk-test-raw-emails.s3.eu-west-2.amazonaws.com').to_timeout
diff --git a/lib/tasks/storage.rake b/lib/tasks/storage.rake
index 40592937c..c4fc1be97 100644
--- a/lib/tasks/storage.rake
+++ b/lib/tasks/storage.rake
@@ -6,20 +6,24 @@
 
   desc 'Mirror files from primary ActiveStorage backend to secondary backends'
   task mirror: :environment do
[...]

PR #6786, opened 2022-02-15
Add storage rake tasks check

gbp:
## Relevant issue(s)

Related to #6531 

## What does this do?

1. Add a `disk_service?` check to ensure we are deleting files from the correct service.
2. Add example cronjob

## Why was this needed?

To ensure raw emails already uploaded to the cloud aren't accidental deleted.

### DIFF
--- a/config/crontab-example
+++ b/config/crontab-example
@@ -37,6 +37,7 @@ MAILTO=<%= mailto %>
 45 3 * * * <%= user %> <%= vhost_dir %>/<%= vcspath %>/commonlib/bin/run-with-lockfile.sh -n <%= vhost_dir %>/stop-new-responses-on-old-requests.lock <%= vhost_dir %>/<%= vcspath %>/script/stop-new-responses-on-old-requests || echo "stalled?"
 55 4 * * * <%= user %> <%= vhost_dir %>/<%= vcspath %>/commonlib/bin/run-with-lockfile.sh -n <%= vhost_dir %>/update-public-body-stats.lock <%= vhost_dir %>/<%= vcspath %>/script/update-public-body-stats || echo "stalled?"
 0 6 * * * <%= user %> <%= vhost_dir %>/<%= vcspath %>/commonlib/bin/run-with-lockfile.sh -n <%= vhost_dir %>/send-webhook-digest.lock <%= vhost_dir %>/<%= vcspath %>/script/send-webhook-digest || echo "stalled?"
+30 4 * * * <%= user %> <%= vhost_dir %>/<%= vcspath %>/commonlib/bin/run-with-lockfile.sh -n <%= vhost_dir %>/storage-raw-emails.lock <%= vhost_dir %>/<%= vcspath %>/bin/rails storage:raw_emails:mirror storage:raw_emails:promote storage:raw_emails:unlink || echo "stalled?"
 
 # Once a day on all servers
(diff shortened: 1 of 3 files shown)


config/crontab-example
```diff
@@ -37,6 +37,7 @@ MAILTO=<%= mailto %>
 45 3 * * * <%= user %> <%= vhost_dir %>/<%= vcspath %>/commonlib/bin/run-with-lockfile.sh -n <%= vhost_dir %>/stop-new-responses-on-old-requests.lock <%= vhost_dir %>/<%= vcspath %>/script/stop-new-responses-on-old-requests || echo "stalled?"
 55 4 * * * <%= user %> <%= vhost_dir %>/<%= vcspath %>/commonlib/bin/run-with-lockfile.sh -n <%= vhost_dir %>/update-public-body-stats.lock <%= vhost_dir %>/<%= vcspath %>/script/update-public-body-stats || echo "stalled?"
 0 6 * * * <%= user %> <%= vhost_dir %>/<%= vcspath %>/commonlib/bin/run-with-lockfile.sh -n <%= vhost_dir %>/send-webhook-digest.lock <%= vhost_dir %>/<%= vcspath %>/script/send-webhook-digest || echo "stalled?"
+30 4 * * * <%= user %> <%= vhost_dir %>/<%= vcspath %>/commonlib/bin/run-with-lockfile.sh -n <%= vhost_dir %>/storage-raw-emails.lock <%= vhost_dir %>/<%= vcspath %>/bin/rails storage:raw_emails:mirror storage:raw_emails:promote storage:raw_emails:unlink || echo "stalled?"
```
garethrees, line 40: Typically we create a wrapper script for rake tasks. I assume this was because of some `bundle exec` weirdness that's resolved by binstubs. Just want to double check that we know this is going to work successfully.

lib/tasks/storage/storage.rb
```diff
@@ -149,10 +152,20 @@ def not_a_mirror
     "correct."
   end
 
-  def primary_service
+  def mirror_primary_not_disk_service
+    "#{prefix}: Mirror primary service is not a disk service, ensure " \
+    "config/storage.yml is correct."
```
garethrees, line 157: Ought to be more specific about what "correct" means in this context.

Need to rephrase, but something like?

> Mirror primary service is not an on-disk service. Records are only pushed to off-site storage after 7 days because {REASONS} and only then do we promote the secondary mirror to be the source of truth and delete the on-disk file.

config/crontab-example
garethrees: Should also note this in the changelog.

garethrees (commented):
Looks fine in principle. Couple of minor suggestions/tweaks but nothing that warrants re-review. 👍

lib/tasks/storage/storage.rb
gbp: Am going to link to a page one configuring storage on alaveteli.org

See #6796

config/crontab-example
gbp: I don't think wrapper scripts are necessary any more and ideally we wouldn't use them here as once #6532 is merged this might change from `storage:raw_emails:mirror` to `storage:mirror`.

I think it might be best to document on alaveteli.org. Will link to the new page (#6796) from the changelog.

PR #6792, opened 2022-02-16
User Clean & Refactor

garethrees:
Various minor cleanup and refactoring of methods in `User`

Nothing controversial; just had half an hour at the end of the day and wanted to do something that didn't need too much brainpower.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -205,20 +205,17 @@ class User < ApplicationRecord
 
   # The "internal admin" is a special user for internal use.
   def self.internal_admin_user
-    user = User.find_by_email(AlaveteliConfiguration.contact_email)
+    user = find_by(email: AlaveteliConfiguration.contact_email)
+    return user if user
 
-    if user.nil?
-      password = PostRedirect.generate_random_token
-      user = User.new(
-        name: 'Internal admin user',
-        email: AlaveteliConfiguration.contact_email,
-        password: password,
-        password_confirmation: password
-      )
-      user.save!
-    end
+    password = PostRedirect.generate_random_token
 
-    user
+    create!(
+      name: 'Internal admin user',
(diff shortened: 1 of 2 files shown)


app/models/user.rb
```diff
@@ -371,10 +369,9 @@ def name_and_email
   # Returns list of requests which the user hasn't described (and last
   # changed more than a day ago)
   def get_undescribed_requests
-    info_requests.where(
-      "awaiting_description = ? and #{ InfoRequest.last_event_time_clause } < ?",
-      true, 1.day.ago
-    )
+    info_requests.
+      where(awaiting_description: true).
+      where("#{ InfoRequest.last_event_time_clause } < ?", 1.day.ago)
```
gbp, line 374:
```suggestion
      where(InfoRequest.last_event_time_clause => (...1.day.ago))
```

app/models/user.rb
```diff
@@ -659,5 +659,4 @@ def setup_pro_account(role)
   def update_pro_account
     pro_account.update_stripe_customer if pro_account
   end
-
```
gbp, line 662: Move from 9e0485dac6c651e667862764a31a462df8128547 to 86fc1801a011f1ab0bf47b805cf2f52219ab5e6f

gbp (commented):
Couple of suggestions. Should also resolve RuboCop warnings or update its configuration.

app/models/user.rb
garethrees: Oops, fixed.

app/models/user.rb
garethrees: Fancy! I'm not sure I fully understand what this does, so I'm going to merge this PR as is. Plz do make a quick PR with this so that you can school me and also get the credit for it 😄

PR #6794, opened 2022-02-16
Update structure of our custom lib directories

gbp:
## What does this do?

Update structure of our custom lib directories

## Why was this needed?

Ensures constants are loadable with Zeitwerk code loader. This will be
required for switching to Rails 7.

Fixes errors such as:
```
  Zeitwerk::NameError: expected file lib/health_checks/health_checkable.rb to define constant HealthCheckable, but didn't
  expected file lib/mail_handler/reply_handler.rb to define constant ReplyHandler
  expected file lib/attachment_to_html/adapter.rb to define constant Adapter
```


garethrees (commented):
Not overly concerned, but `lib/health_checks/health_checks/*.rb` is a bit gross? 

Could we do `mv lib/health_checks/health_checks.rb lib/` and e.g. `mv lib/attachment_to_html/attachment_to_html.rb lib/`? I assume that would mean we could remove the [custom autoloaders](https://github.com/mysociety/alaveteli/blob/0.40.1.0/config/application.rb#L82-L84) then?

Happy either way tbh

garethrees (approved):

PR #6797, opened 2022-02-18
Fix storage cronjob

gbp:
## What does this do?

Fix storage cronjob

## Why was this needed?

Needed to:
1. add quotes so only one argumnet if passed to `run-with-lockfile`
2. cd in the application directory so the `config.ru` is found.

## Notes to reviewer

I merged the original commit too soon and need to make these changes on the production server to get this working.


garethrees (approved):
👍

Issue #5368, opened 2019-09-12
Fully bring the process of adding and editing new authorities into the web-application


RichardTaylor:
We are already close to dealing with requests to add or edit bodies via the web system, rather than relying on email.

 The outstanding issues are

- [x] we need to bring the notes field from the submission form into the admin interface at https://www.whatdotheyknow.com/admin
- [ ] make some minor tweaks to the "close and respond" form so it includes the user name, signature etc. as the "Response to change request (will be emailed to user):" element of the body edit form.
- [ ] Add a way for admins to add a note / discuss without actually making a change. We could use the changelog "Comment for this edit" for this field for this.

Related: #4686  "Add more fields to the public - Ask us to add an authority - form "

This is part of #419 "Bring user support and takedown requests within the system"

It might be a good way to start tackling that large ticket and establish an approach. 

At the moment (during transition away from a reliance on email) I think we'd still want emails to alert admins that there is a new change/add request, and for copies of responses to be sent to admins [ideally in the same email thread](https://github.com/mysociety/alaveteli/issues/579#issuecomment-190190284). 

The downside of starting with this issue when tackling #419 might be it's not addressing an area of high-volume correspondence where the benefits of bringing correspondence into the system are expected to be greatest.

PR #6799, opened 2022-02-18
[#5368] Show change request notes on body edit page

garethrees:
If we're editing a public body via a change request, show the additional
notes when editing the body. This provides the admin with better inline
context about the edit and allows quicker copy-pasting of details added
to the notes field.

A quick win from https://github.com/mysociety/alaveteli/issues/5368.

Won't get the notes until https://github.com/mysociety/alaveteli/pull/6798 is merged, but there's no dependency in either direction.

## Screenshots

Screenshot 2022-02-18 at 17 40 37

Screenshot 2022-02-18 at 17 41 14

### DIFF
--- a/app/views/admin_public_body/edit.html.erb
+++ b/app/views/admin_public_body/edit.html.erb
@@ -36,6 +36,11 @@
 
   </div>
   <div class="span4">
+    <% if @change_request %>
+      <h3>Change Request</h3>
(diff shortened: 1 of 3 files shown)


mdeuk:
This looks fab, a definite improvement over the existing workflow 🥳

gbp (approved):

PR #6807, opened 2022-02-21
[#6800] Tweak heading used for public body questions

### LINKED ISSUE #6800: Amend heading for public body questions
While working on https://github.com/mysociety/whatdotheyknow-theme/issues/722 and https://github.com/mysociety/whatdotheyknow-theme/issues/745 I've noted that the way we present the public body questions on relevant bodies (currently on WDTK the Home Office) is slightly cryptic. This was implemented in #5013.

We begin by stating "Are you asking about your, or others', personal information or personal situation?" - which seems clear enough, to someone with a good grasp of English - but it might not be if you aren't a native speaker. It may also cause reuse problems, as it means the question posed has to tie into the existing heading.

### Proposal

To make this easier for our users to understand, would suggest that we modify the heading to something more generic (and reusable).

A simple way of doing this would be to go for the bare minimum of asking: "**What would  you like to do?**"

### Why?

Two main reasons:

- Some of our users do not have a strong grasp of the language, meaning they might not fully understand why we're asking them the question, and will continue to misuse the site.
- As @RichardTaylor has noted in https://github.com/mysociety/alaveteli/issues/3515#issuecomment-457208430, we've had advice from the regulator about using _simple language_.

mdeuk:
## Relevant issue(s)

Fixes #6800 

## What does this do?
This replaces the generic header text used when _public body questions_ are rendered, and ensures that the correct semantic tagging is in place.

## Why was this needed?

To aid reuse on WDTK theme, and improve legibility - a simpler heading allows us more flexibility.

## Implementation notes
Nothing of note - thanks to @garethrees for help with the translation parameter, hopefully I've done it right!

## Notes to reviewer
This is best tested against WDTK theme - a PR will follow there to amend the existing public body questions for Home Office (and add some others) imminently.

### DIFF
--- a/app/views/request/new.html.erb
+++ b/app/views/request/new.html.erb
@@ -235,7 +235,7 @@
         <%= render 'form', f: f %>
       <% else %>
         <div id="request_form_questions">
-          <p><%= _('Are you asking about your, or others\', personal information or personal situation?') %></p>
+          <h2><%= _('What would you like to do?') %></h2>
 
           <% @info_request.public_body.questions.each do |question| %>
             <label for="<%= question.key %>">


gbp (approved):

mdeuk:
> Before: before
> 
> After: after

It’s quite obvious, in hindsight, that screenshots should have been in the PR first time - apologies 😳

PR #6809, opened 2022-02-22
Check OutgoingMessage#body for spam content

garethrees:
Some spammers are sneaky and disguise their spam by using an innocuous
title followed by a load of spam in the request body. This adds a quick
check against that.


garethrees:
Closing this for now as the immediate need has passed.

PR #6820, opened 2022-02-24
Add prominence_icon to admin user listings

garethrees:
Add prominence_icon to admin user listings

Allows an at-a-glance look at whether a User is banned, inactive or
otherwise in a compact method. When banned a user gets “Account
Suspended” prepended to their name, and in some cases an orange “banned”
label, but in other more information-dense listings we either don’t show
this or it’s hard to parse out.

This makes it easy to tie banned users to hidden content or hidden
content to unbanned users.

Main benefit is in a listing like this where we can get a quick overview of a group of content:

Screenshot 2022-02-24 at 17 32 08

Also tweaks the users index. Not much more information than before, but marginally less messy.

Screenshot 2022-02-24 at 17 31 07

### DIFF
--- a/app/helpers/admin_helper.rb
+++ b/app/helpers/admin_helper.rb
@@ -32,8 +32,8 @@ module AdminHelper
   end
 
   def user_both_links(user)
-    link_to(eye, user_path(user), :title => "view user's page on public website") + " " +
-      link_to(h(user.name), admin_user_path(user), :title => "view full details")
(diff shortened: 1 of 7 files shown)


WilliamWDTK:
>When banned a user gets “Account Suspended” prepended to their name,

Isn't it appended/postpended?

garethrees:
> Isn't it appended/postpended?

🤦 yes, oops.

Issue #2691, opened 2015-07-29
Moderation / banning tools suggestions


skenaja:
Some suggestions for improved moderation tools - feel free to break out into separate issues:

Use cases:
- admin user tracks down and bans sockpuppet accounts (eg for previously banned users)
- admin user tracks down human spammers (who can bypass captchas) - eg to find those registering accounts to add link spam to their user profile - and bans or deletes their accounts

*Store / search IP address to track activity on the site*

 - [ ] save IP addresses in the DB 
 - [ ] allow search by IP to view list of actions taken recently by that IP 
 - [ ] allow filtering of list by activity (Eg register, login, edit profile, send message to user), by userid
 - [ ] show IP(s) against on user profile (admin) - allow initiate 

*Ability to completely delete user accounts*
-  This would need to take into account any tracks, requests, outbound replies, annotations, profile photos.

kingqueen3065:
I support this as a useful tool. Of particular note recently has been a number of attempts to sign up with false details and then automatically reply to the confirmation email with a spam, which hits the team@whatdotheyknow address - Helen has zapped a large number of these - and the accounts Alex has mentioned that are created solely to create links from our reputable site to a spammer's, to improve its page rank.

I am somewhat nervous of IP tracking based solutions because many of our users, being data rights savvy, may be understandably nervous of any ability to track users by IP. Our user profiles are great at the moment as they are so simple, they can comprise just a name and email address, and they have the minimum of extra information. I value this, it circumvents some privacy based concerns, and as an administrator I personally would rather not have the ability to track by IP.

FOIMonkey:
There has been another recent uptick in the creation of accounts in order to post spam in the bio section. I've edited quite a few of these but the process is time consuming. It'd be good to have the option to easily delete the accounts entirely.

RichardTaylor:
Would logging "IP address last logged in from" and making that available on a user admin page be a possible first step?

Perhaps if we did something like that we shouldn't hold it more than 30 days in line with the server log retention period?

On Wikipedia only 38 users have what they there call "CheckUser" privileges which let them see IP addresses for logged in users. https://en.wikipedia.org/wiki/Wikipedia:CheckUser  Alaveteli might want to follow that model and control access to such privacy sensitive information?

There is a current case on WhatDoTheyKnow where tools to identify people circumventing bans would be useful which has prompted this comment.

RichardTaylor:
Just adding a +1

[...]

PR #6822, opened 2022-02-25
[#2691] Track IP addresses associated with User signins

garethrees:
If a retention period is configured, create a record associated with the
user of the signin and the IP address used to do so.

A rake/cron task is provided to purge records older than the retention
period.

A version of https://github.com/mysociety/alaveteli/issues/2691

---

View/search all recorded sign ins:

Screenshot 2022-02-25 at 21 36 47

View a given user's sign ins. Check how many others have signed in with the IP:

Screenshot 2022-02-25 at 21 36 27

Click the IP address to go through to the main listing filtered by that IP (or country, added after screenshot):

Screenshot 2022-02-25 at 21 36 38

Filter by partial IP:

Screenshot 2022-02-25 at 21 55 32

Filter by country:

Screenshot 2022-02-25 at 21 55 23

### DIFF
--- a/app/controllers/admin/users/sign_ins_controller.rb
+++ b/app/controllers/admin/users/sign_ins_controller.rb
@@ -1,0 +1,15 @@
+# Display information about User::SignIn attempts
+class Admin::Users::SignInsController < AdminController
+  layout 'admin/users'
+
+  def index
+    @title = 'Listing user sign ins'
+
+    @query = params[:query]
+
+    sign_ins = User::SignIn
+    sign_ins = sign_ins.search(@query) if @query
+
+    @sign_ins = sign_ins.paginate(page: params[:page], per_page: 100)
+  end
+end
--- a/app/models/user/sign_in.rb
+++ b/app/models/user/sign_in.rb
@@ -22,6 +22,14 @@ class User::SignIn < ApplicationRecord
     where('created_at < ?', retention_days.days.ago).destroy_all
   end
 
+  def self.search(query)
+    joins(:user).references(:users).where(<<~SQL, query: query)
+      lower(user_sign_ins.ip::text) LIKE lower('%'||:query||'%') OR
+      lower(users.name) LIKE lower('%'||:query||'%') OR
+      lower(users.email) LIKE lower('%'||:query||'%')
+    SQL
+  end
+
   def self.retain_signins?
     retention_days >= 1
   end
--- a/app/models/user/sign_in.rb
+++ b/app/models/user/sign_in.rb
@@ -38,6 +38,12 @@ class User::SignIn < ApplicationRecord
     AlaveteliConfiguration.user_sign_in_activity_retention_days
   end
 
+  def other_users
+    User.distinct.joins(:sign_ins).
+      where(user_sign_ins: { ip: ip }).
+      where.not(id: user_id)
+  end
+
   private
 
   def create?
--- a/app/models/user/sign_in.rb
+++ b/app/models/user/sign_in.rb
@@ -1,0 +1,38 @@
+# == Schema Information
+# Schema version: 20220225094330
+#
+# Table name: user_sign_ins
+#
+#  id         :bigint           not null, primary key
+#  user_id    :bigint
+#  ip         :inet
+#  created_at :datetime         not null
+#  updated_at :datetime         not null
+#
+
+# Record medadata about User sign in activity
+class User::SignIn < ApplicationRecord
+  default_scope { order(created_at: :desc) }
+
+  belongs_to :user, inverse_of: :sign_ins
+
+  before_create :create?
+
+  def self.purge
+    where('created_at < ?', retention_days.days.ago).destroy_all
+  end
+
+  def self.retain_signins?
+    retention_days >= 1
+  end
+
+  def self.retention_days
+    AlaveteliConfiguration.user_sign_in_activity_retention_days
+  end
+
+  private
+
+  def create?
+    throw :abort unless self.class.retain_signins?
+  end
+end
--- a/app/models/user/sign_in.rb
+++ b/app/models/user/sign_in.rb
@@ -1,5 +1,5 @@
 # == Schema Information
-# Schema version: 20220225094330
+# Schema version: 20220225214524
 #
 # Table name: user_sign_ins
 #
@@ -8,6 +8,7 @@
 #  ip         :inet
 #  created_at :datetime         not null
 #  updated_at :datetime         not null
+#  country    :string
 #
 
 # Record medadata about User sign in activity
@@ -25,6 +26,7 @@ class User::SignIn < ApplicationRecord
   def self.search(query)
     joins(:user).references(:users).where(<<~SQL, query: query)
       lower(user_sign_ins.ip::text) LIKE lower('%'||:query||'%') OR
+      lower(user_sign_ins.country) LIKE lower('%'||:query||'%') OR
       lower(users.name) LIKE lower('%'||:query||'%') OR
       lower(users.email) LIKE lower('%'||:query||'%')
     SQL
--- a/app/views/admin/users/_sign_in_table.html.erb
+++ b/app/views/admin/users/_sign_in_table.html.erb
@@ -22,6 +22,11 @@
           </td>
 
           <td><%= admin_date(sign_in.created_at, ago_only: true) %></td>
+
+          <td>
+            <strong><%= sign_in.other_users.size %></strong> others using this
+            IP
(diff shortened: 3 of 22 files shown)


app/controllers/admin/users/sign_ins_controller.rb
```diff
@@ -0,0 +1,17 @@
+class Admin::Users::SignInsController < AdminController
```
garethrees, line 1: Will add class-level docs on merge.

config/routes.rb
```diff
@@ -652,6 +652,14 @@
   end
   ####
 
+  #### Admin::Users::SignIns controller
+  namespace :admin do
```
garethrees, line 656: Couldn't quite decide on how to route this.

Alternative was…

```ruby
namespace :admin do
  namespace: :users do
    resources :sign_ins, only: [:index] #=> /admin/users/sign_ins
  end
end
```

…but wanted to avoid nesting hell. Could possibly do the above and use shallow routes if more actions added. Open to suggestion.

app/views/admin/users/_sign_in_table.html.erb
```diff
@@ -10,6 +10,10 @@
       </td>
 
       <td><%= admin_date(sign_in.created_at, ago_only: true) %></td>
+
+      <td>
+        <strong><%= sign_in.other_users.size %></strong> others using this IP
```
garethrees, line 27: Probably okay for now, but I think we'll want to handle [Apple Private Relay](https://developer.apple.com/support/prepare-your-network-for-icloud-private-relay/). Probably annotate the stored country with a note so that a high count + APR country wouldn't create suspicion?

garethrees: Might want to add some conditional for when we have no recorded sign ins

app/models/user/sign_in.rb
```diff
@@ -0,0 +1,47 @@
...
+      lower(users.email) LIKE lower('%'||:query||'%')
+    SQL
+  end
+
+  def self.retention_days
+    AlaveteliConfiguration.user_sign_in_activity_retention_days
+  end
+
+  def other_users
+    self.class.where(ip: ip).where.not(user_id: user_id)
```
gbp, line 39: Should return distinct User instances not User::SIgnIn instances.

```suggestion
    User.distinct.joins(:sign_ins).
      where(user_sign_ins: { ip: ip }).
      where.not(user_sign_ins: { user_id: user_id })
```

gbp: Or:

```suggestion
    User.distinct.joins(:sign_ins).
      where(user_sign_ins: { ip: ip }).
      where.not(id: user_id)
```

app/controllers/admin/users/sign_ins_controller.rb
```diff
@@ -0,0 +1,17 @@
...
+  def index
+    @title = 'Listing user sign ins'
+    @query = params[:query]
+
+    sign_ins =
+      if @query
+        User::SignIn.search(@query)
+      else
+        User::SignIn
+      end
```
gbp, line 13: Reduce to two lines:

```suggestion
    sign_ins = User::SignIn.search(@query) if @query
    sign_ins ||= User::SignIn
```


or:

```suggestion
    sign_ins = User::SignIn
    sign_ins = sign_ins.search(@query) if @query
```

db/migrate/20220225214524_add_country_to_user_sign_ins.rb
```diff
@@ -0,0 +1,5 @@
+class AddCountryToUserSignIns < ActiveRecord::Migration[6.1]
+  def change
+    add_column :user_sign_ins, :country, :string
```
gbp, line 3: Combine into one migration?

app/views/admin/users/_sign_in_table.html.erb
gbp:
```suggestion
        <strong><%= pluralize sign_in.other_users.size, 'other' %></strong> using this IP
```

gbp (changes requested):
This is good, I think there is still some work to do tho.

Does this work with both IPv4 and IPv6 addresses. I think we also need to consider what we do when a user is closed and anonymised.

app/models/user/sign_in.rb
```diff
@@ -0,0 +1,47 @@
...
+# Record medadata about User sign in activity.
+class User::SignIn < ApplicationRecord
+  default_scope { order(created_at: :desc) }
+
+  belongs_to :user, inverse_of: :sign_ins
+
+  before_create :create?
+
+  def self.purge
+    where('created_at < ?', retention_days.days.ago).destroy_all
```
gbp, line 22:
```suggestion
    where(created_at: (...retention_days.days.ago)).destroy_all
```

gbp:
How about making this more generic and recording IP on sign ins and sign ups?

Could then link the "rate limited signup from" email to the index action with IP already filtered.

app/controllers/admin/users/sign_ins_controller.rb
garethrees: Prefer the latter as it's clearer it's an additive filter.

app/views/admin/users/_sign_in_table.html.erb
garethrees: > `pluralize sign_in.other_users.size`

I thought about this, but I wanted only the number to be `<strong>`.

Could enhance further with only bolding if number is > 1, but not clear that's useful yet.

app/views/admin/users/_sign_in_table.html.erb
```diff
@@ -0,0 +1,9 @@
+<table class="table table-condensed table-striped">
```
garethrees, line 1: Add "none yet" conditional

app/views/admin/users/_sign_in_table.html.erb
garethrees: > Might want to add some conditional for when we have no recorded sign ins

Noted elsewhere.

> Need to return distinct users

Noted elsewhere.

app/models/user/sign_in.rb
garethrees: Also add a failing spec

app/models/user/sign_in.rb
garethrees: To me the former is much clearer. Just resolving for now but happy for a PR later to change.

app/views/admin_user/show.html.erb
```diff
@@ -156,6 +156,19 @@
   <hr>
 <% end %>
 
+<h2>Sign Ins</h2>
+
+<table class="table table-condensed table-striped">
```
garethrees: Add "none yet" conditional in this commit

garethrees:
> Does this work with both IPv4 and IPv6 addresses.

I've now used postgres inet type and added a basic test, so _should_ do, though I don't know much about IPv6 tbh.

> I think we also need to consider what we do when a user is closed and anonymised.

Good point. I think the regular purge would have been fine, but also added on-demand purging via `User#close_and_anonymise` in 97c6d71815f14af4f7f320d66e9c6e6a295384b4.

> How about making this more generic and recording IP on sign ins and sign ups? Could then link the "rate limited signup from" email to the index action with IP already filtered.

I did think about that, but this is the more immediate problem. To consider in relation to #6931.

garethrees:
Couple of things left to do, but in order to do so I'll want to squash these commits:

- [ ] Move the addition of `default_scope` to the previous (first) commit
- [ ] Add "none yet" conditional to the sign in table
- [ ] Squash eab805a and 73e939d into a single commit (~ "Render user sign ins in the admin interface")
- [ ] Handle `USER_SIGN_IN_ACTIVITY_RETENTION_DAYS == 0`. Instead of hiding, I'll use a notice saying "sign in tracking is not configured; see general.yml", or something to that effect.

garethrees:
@gbp a quick pair of eyes over the fixups would be great before I squash them down to complete the [remaining tasks](https://github.com/mysociety/alaveteli/pull/6822#issuecomment-1112092653).

gbp:
> @gbp a quick pair of eyes over the fixups would be great before I squash them down to complete the [remaining tasks](https://github.com/mysociety/alaveteli/pull/6822#issuecomment-1112092653).

Fixups look good me, feel free to squash and rebase

garethrees:
Think this is good to go now 🥳

gbp (approved):
LGTM

garethrees:
Linking to https://github.com/mysociety/alaveteli/pull/7592.

PR #6827, opened 2022-02-28
[WIP] Detect signups via banned user profiles

garethrees:
A common spam pattern is for accounts to arrive to the site via another
spam user profile. Following the post redirects of accounts can help
identify this pattern.

For example

User `247700` has a post redirect `uri` of `/user/gerami_george/profile` -> User `247693`
User `247693` has a post redirect `uri` of `/user/sandra_drago` -> User `203473` 
User `203473` has no post redirects and is banned for spamming.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -622,6 +622,36 @@ class User < ApplicationRecord
     "User;#{id}"
   end
 
+  def signed_up_via_banned_user_profile?
+    user = follow_post_redirect_chain
+    user&.banned?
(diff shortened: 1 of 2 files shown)


garethrees:
IDK if this is actually useful, because it relies on us banning a user. If the user is banned, then it might not be used as a signup source (because the spam about me text will no longer be present).

garethrees:
Closing this spike because we're not going to actively work on it soon.

PR #6834, opened 2022-03-01
[#6833] Improve storage rake logging

### LINKED ISSUE #6833: Reduce storage rake task output
We're being emailed progress bar output from the storage rake task.

We should run with the `--quiet` argument.

gbp:
## Relevant issue(s)

Fixes #6833

## What does this do?

Refactoring logging and changed to only output when the `--quiet` argument isn't used

## Why was this needed?

We're receiving emails with progress bar output

## Implementation notes

Only stdout is silenced, errors are still outputted


config/crontab-example
```diff
@@ -37,7 +37,7 @@ MAILTO=<%= mailto %>
...
 55 4 * * * <%= user %> <%= vhost_dir %>/<%= vcspath %>/commonlib/bin/run-with-lockfile.sh -n <%= vhost_dir %>/update-public-body-stats.lock <%= vhost_dir %>/<%= vcspath %>/script/update-public-body-stats || echo "stalled?"
 0 6 * * * <%= user %> <%= vhost_dir %>/<%= vcspath %>/commonlib/bin/run-with-lockfile.sh -n <%= vhost_dir %>/send-webhook-digest.lock <%= vhost_dir %>/<%= vcspath %>/script/send-webhook-digest || echo "stalled?"
-30 4 * * * <%= user %> <%= vhost_dir %>/<%= vcspath %>/commonlib/bin/run-with-lockfile.sh -n <%= vhost_dir %>/storage-raw-emails.lock "cd <%= vhost_dir %>/<%= vcspath %> && bin/rails storage:raw_emails:mirror storage:raw_emails:promote storage:raw_emails:unlink" || echo "stalled?"
+30 4 * * * <%= user %> <%= vhost_dir %>/<%= vcspath %>/commonlib/bin/run-with-lockfile.sh -n <%= vhost_dir %>/storage-raw-emails.lock "cd <%= vhost_dir %>/<%= vcspath %> && bin/rails storage:raw_emails:mirror storage:raw_emails:promote storage:raw_emails:unlink --quiet" || echo "stalled?"
```
garethrees, line 40: Would be worth adding this option before the arguments as that's a little more common practice?

```
… bin/rails --quiet storage:raw_emails:mirror …
```

No big deal if that doesn't work.

garethrees (approved):
Haven't run but looks sensible and trust it fixes the issue.

PR #6839, opened 2022-03-02
Add Taggable concern to Info Request

gbp:
This sets up `has_tag_string` but includes the `with_tag` and
`without_tag` query methods to allow filtering of requests.


garethrees (approved):
Not sure if this was explicitly ready for review but lgtm

PR #6855, opened 2022-03-04
WIP Extract User::AnonymousUser

garethrees:
* Reduce complexity of InfoRequest
* Allow the return value of InfoRequest#user to duck-type with methods
  that expect an object (e.g. AdminHelper#prominence_icon)


garethrees:
Oh no durr – sorry. The problem is that now `InfoRequest#user` returns a `User::AnonymousUser` _instead of_ `nil`, and so  when we try to pass that to `RequestSummary.create` it can't set `nil` for `RequestSummary#user_id` because ``User::AnonymousUser` doesn't have an `id` method / attribute.

app/models/alaveteli_pro/request_summary.rb
```diff
@@ -76,12 +76,14 @@ def self.create_from(request)
   end
 
   def self.attributes_from_request(request)
+    user = request.user.is_a?(User::AnonymousUser) ? nil : request.user
```
garethrees, line 79: Just to add to the commit message, `AlaveteliPro::RequestSummary` has a `belongs_to :user` association, so when we pass a `User::AnonymousUser` that obviously doesn't work because the association setter is expecting an `ActiveRecord` object.

On WhatDoTheyKnow we do have request summaries without an associated user:

```ruby
AlaveteliPro::RequestSummary.where(user_id: nil).count
# => 1002
```

garethrees:
> @garethrees can we get this ready to review asap as I'm hitting this often on staging and local development environment

Sure, can work on this today I think. Happy with the general approach?

Issue #852, opened 2013-03-01
Are all the states `error_message`, `requires_admin` and `attention_requested` needed?


mlandauer:
They kind of seem to do the similar things which is say "hey, this request needs admin attention". Also the messages displayed to the user are similar but not quite the same. Also, as a user if I'm presented with two very similar options I might find myself like the dog stuck between two bowls of food.

What am I missing?

HelenWDTK:
This issue has been automatically closed due to a lack of discussion or resolution for over 12 months.
Should we decide to revisit this issue in the future, it can be reopened.

PR #6856, opened 2022-03-04
[WIP] Don't show awaiting classification requests that require attention

garethrees:
…TODO…

### DIFF
--- a/app/controllers/admin_general_controller.rb
+++ b/app/controllers/admin_general_controller.rb
@@ -8,14 +8,22 @@ class AdminGeneralController < AdminController
 
   def index
     # Tasks to do
-    @requires_admin_requests = InfoRequest.
-      find_in_state('requires_admin').
+    @requires_admin_requests =
+      InfoRequest.
+        find_in_state('requires_admin').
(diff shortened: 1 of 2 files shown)


garethrees:

app/controllers/admin_general_controller.rb
```diff
@@ -8,14 +8,22 @@ class AdminGeneralController < AdminController
...
+        find_in_state('error_message').
+        where(awaiting_description: false).
         not_embargoed
-    @attention_requests = InfoRequest.
-      find_in_state('attention_requested').
+
+    @attention_requests =
+      InfoRequest.
+        find_in_state('attention_requested').
+        where(awaiting_description: false).
```
garethrees, line 26: I think there's some subtlety to address with the requires_admin/attention_requested cases.

Note to self: A request might be awaiting classification by the author because it's received an auto-response, but then a different user might report it because there's a different issue with it (offensive, or whatever). Does that retain the `awaiting_description: false` state?

Also check notes in https://github.com/mysociety/alaveteli/issues/852.

PR #6868, opened 2022-03-08
[#6532] Update FoiAttachment storage migrate task

gbp:
## Relevant issue(s)

See #6532
See #6705

## What does this do?

Adds a checks if the `FoiAttachment` exists on disk before migrating.

## Why was this needed?

It is fine for `FoiAttachment` to be missing from disk as we deal with
that within the `FoiAttachment#body` method.

We previously needed to `#reload` the instance (adding in [1]) if the
file was missing but now this isn't necessary any more.

[1] https://github.com/mysociety/alaveteli/pull/6705

### DIFF
--- a/lib/tasks/storage/attachments.rake
+++ b/lib/tasks/storage/attachments.rake
@@ -3,7 +3,12 @@ namespace :storage do
     require_relative 'storage'
 
     def attachment_storage
-      Storage.new(FoiAttachment, :file, setter: :body=, getter: :body)
+      Storage.new(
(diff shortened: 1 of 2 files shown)


garethrees (changes requested):
Looks like the build is genuinely failing after this change?

gbp:
> Looks like the build is genuinely failing after this change?

Sorry yeah, hadn't expected that. Should be okay now... 🤞

garethrees (approved):
Haven't run but looks good in principle.

PR #6877, opened 2022-03-11
WIP Fix reading attachments inside transaction

gbp:
When parsing incoming messages we use to reload the `foi_attactments`
association as sometimes attachments might be removed. This happens
rarely but normally due to code changes.

The reload of the attachments broke the Xapian update index tasks since
migrating to `ActiveStorage`.

...


spec/models/foi_attachment_spec.rb
```diff
@@ -67,6 +67,15 @@
       expect(main.body).to eq(orig_body)
     end
 
+    it 'can parse raw email and read attachment inside DB transaction' do
```
garethrees, line 70: No commits should break the build; should fix this up into the commit that fixes the failing spec.

app/models/incoming_message.rb
```diff
@@ -625,14 +621,7 @@ def extract_attachments
...
 
-    attachment_ids = attachments.map { |attachment| attachment.id }
-    # now get rid of any attachments we no longer have
-    FoiAttachment.
-      where(
-        ["id NOT IN (?) AND incoming_message_id = ?",
-         attachment_ids,
-         self.id]
-      ).destroy_all
+    (foi_attachments - attachments).each(&:mark_for_destruction)
```
garethrees, line 619: Given this method is still rather complicated, feel like it warrants preservation of the comment (with some minor rewording)? Something like:

```diff
- # now get rid of any attachments we no longer have
+ # Purge old attachments that have been rebuilt with a new hexdigest
```

app/models/incoming_message.rb
```diff
@@ -599,17 +600,12 @@ def extract_attachments
       memo
     end
 
-    attachments = []
-    attachment_attributes.each do |attrs|
-      attachment = foi_attachments.find_or_create_by(hexdigest: attrs[:hexdigest])
-      attachment.update(attrs)
-      attachment.save!
-      attachments << attachment
+    attachments = attachment_attributes.map do |hexdigest, attrs|
```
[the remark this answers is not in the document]
gbp: Above we are calling `attachment_attributes.inject({})` to make it a hash with `hexdigest` keys I need to change commit 799e07e2cde44c5df1f7f18278dab1fbfde5c375 to update the block arguments here.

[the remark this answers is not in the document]
garethrees: > make it a hash with `hexdigest` keys

Ah gotcha.

> I need to change commit 799e07e to update the block arguments here.

 👍

spec/models/foi_attachment_spec.rb
gbp: Okay. Think it warrants having this commit first as it shows the issue which we're fixing. I will change this commit to `xit` and change it back to `it` when fixed.

PR #6885, opened 2022-03-16
Add SFTP ActiveStorage service

gbp:
## Relevant issue(s)

https://github.com/mysociety/sysadmin/issues/1527

## What does this do?

Add SFTP ActiveStorage service

## Why was this needed?

Allows mirroring to remote servers of attached files. This will be
useful when migrating to a new application server.

## Implementation notes

Can be configured in `config/storage.yml`, with:

```
attachments_sftp:
  service: SFTP
  root: remote/path
  user: username
  host: host.name.tld
```

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -163,6 +163,8 @@ gem 'azure-storage', require: false
 gem 'google-cloud-storage', '~> 1.36', require: false
 gem 'activestorage-sftp', git: 'https://github.com/treenewbee/activestorage-sftp.git',
                           ref: '679fbe1'
+gem 'ed25519'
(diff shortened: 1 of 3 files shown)


gbp:
Have tested this on staging and all now seems to be working. Initial attempt used the published version of the gem which didn't include [a fix](https://github.com/treenewbee/activestorage-sftp/pull/10) and my quick attempt to patch the library didn't work as expected. I have switch to a more recent upstream commit and all seems to be fine now.

garethrees (approved):
Let's give it a go!

PR #6848, opened 2022-03-03
Show citations on admin request page

garethrees:
Makes it easier for admins to see when a Citation was added to a request
and who added it.

Request with Citations:

Screenshot 2022-03-03 at 12 20 15

Request with no Citations (fixed the missing full stop after the screenshot):

Screenshot 2022-03-03 at 12 20 51

PR #6888, opened 2022-03-18
Show citations on admin pages

garethrees:
Makes it easier for admins to see when a citation was added to a request
and who added it, or what citations a user has added and to which
requests.

Supersedes https://github.com/mysociety/alaveteli/pull/6848.

Instead of a table format, use the "Thing" → "By" → "On" trio used in other lists. This makes the partial more flexible in that it makes more sense when rendered in multiple views.

In the future, if `Citation` gets its own admin/show page, we can link internally rather than the `Citation#source_url`, add a `prominence_icon`, etc.

I've currently handled batches by giving `InfoRequestBatch` a hard-coded prominence value so we can continue the pattern of using `prominence_icon` to link to the public record page, and then not rendered a link to the admin batch page, which can be added if we decide to add an admin view of a batch.

## Single Request page

Screenshot 2022-03-18 at 14 38 09

## Request that's part of a batch

We don't have an admin page for a batch request, so only render the unlinked title.

Screenshot 2022-03-18 at 14 45 39

## User page

Screenshot 2022-03-18 at 14 37 43

## Notes to Reviewer

Open to options on the CodeClimate suggestions – yes duplication, but alternative is pretty ugly `link_to` calls.

### DIFF
--- a/app/helpers/admin_helper.rb
+++ b/app/helpers/admin_helper.rb
@@ -22,26 +22,9 @@ module AdminHelper
     icon("arrow-right")
   end
 
-  def request_both_links(info_request)
-    link_to(prominence_icon(info_request), request_path(info_request), title: 'View request on public website') + ' ' +
-      link_to(info_request.title, admin_request_path(info_request), title: 'View full details')
-  end
-
-  def public_body_both_links(public_body)
-    link_to(eye, public_body_path(public_body), :title => "view authority on public website") + " " +
-      link_to(h(public_body.name), admin_body_path(public_body), :title => "view full details")
-  end
-
-  def user_both_links(user)
-    link_to(prominence_icon(user.prominence), user_path(user), title: 'View user on public website') + ' ' +
-      link_to(h(user.name), admin_user_path(user), title: 'View full details')
-  end
-
-  def comment_both_links(comment)
-    link_to(prominence_icon(comment), comment_path(comment),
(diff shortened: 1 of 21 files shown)


app/helpers/admin_helper.rb
```diff
@@ -22,26 +22,9 @@ def arrow_right
...
-  end
-
-  def comment_both_links(comment)
-    link_to(prominence_icon(comment), comment_path(comment),
-            :title => "view comment on public website") + " " +
-      link_to(h(truncate(comment.body)), edit_admin_comment_path(comment),
-              :title => "view full details")
+  def both_links(record)
+    method = "#{record.class.to_s.underscore}_both_links"
+    send(method, record)
```
gbp: Could check to see if method exists. if not raise argument error

gbp (approved):
This is great. One comment but happy to merge as is.

app/helpers/admin_helper.rb
garethrees: I was wondering about this, but eventually thought that the `NoMethodError` is kinda more descriptive (if you want to pass a record of class `X`, add the method named `x_both_links`) and it avoids a line of code. I'm happy that this breaks if you pass an "unexpected" record, as the only time someone will do that is while working on the codebase (as opposed to having to handle unknown user-submitted data).

Issue #6875, opened 2022-03-10
[ERROR] admin_public_body#show (ActionView::Template::Error) "undefined method `prominence' for nil:NilClass"


gbp:
```
An ActionView::Template::Error occurred in admin_public_body#show:

  undefined method `prominence' for nil:NilClass
  app/helpers/admin_helper.rb:36:in `user_both_links'


-------------------------------
Request:
-------------------------------

  * URL        : https://staging.alaveteli.org/admin/bodies/2
  * HTTP Method: GET
  * Parameters : {"controller"=>"admin_public_body", "action"=>"show", "id"=>"2"}


-------------------------------
Backtrace:
-------------------------------

  app/helpers/admin_helper.rb:36:in `user_both_links'
  app/views/admin_request/_some_requests.html.erb:14
  app/views/admin_request/_some_requests.html.erb:2
  app/views/admin_public_body/show.html.erb:118
  app/controllers/admin_public_body_controller.rb:28:in `block in show'
  lib/alaveteli_localization.rb:46:in `with_locale'
  app/controllers/admin_public_body_controller.rb:19:in `show'
```

gbp:
Seeing these since #6820

gbp:
@garethrees I think you mentioned a new issues with external users in standup - was it this?

garethrees:
Yeah, that looks about right. I assume body 2 has an external request (which has a `nil` user)?

garethrees:
Closing this as we've now removed all external requests from WDTK and complete removal of the code is underway in #6893.

Issue #6446, opened 2021-08-08
Delete features


RichardTaylor:
Inspired by Elon Musk who regularly says: ~"the best part is no part. The best process is no process"[*](https://en.wikiquote.org/wiki/Elon_Musk) consider what we can delete from the system. 

* "Download a zip file of all correspondence" may be a contender? Is it used?
* Collapsing correspondence isn't useful as it is, but could help improve the readability of threads if the collapsed state persisted.

garethrees:
> "Download a zip file of all correspondence" may be a contender? Is it used?

This a key way of getting data _out_ of Alaveteli and into something else. I don't know how much it's used (either overall or by Pros, who are probably a bit more interested in this), but I imagine when it is used it's very important.

> Collapsing correspondence isn't useful as it is

I actually use it a fair amount (not "collapse all", but clicking each bubble). This is a key stepping stone to https://github.com/mysociety/alaveteli/issues/4558.

> "the best part is no part"

I do agree though, the less we have to maintain the easier the rest of it is to maintain.

RichardTaylor:
Turning off email redaction has been suggested previously

* https://github.com/mysociety/alaveteli/issues/10#issuecomment-3812719 

We almost certainly shouldn't do that in respect of past requests though, and doing it from a point in time would probably add complexity now.

garethrees:
Bunch of things to consider:

- https://github.com/mysociety/alaveteli/issues/6339
- https://github.com/mysociety/alaveteli/issues/214
- https://github.com/mysociety/alaveteli/issues/5833
- https://github.com/mysociety/alaveteli/issues/3692
- Remove Gaze (https://github.com/mysociety/alaveteli/issues/2905, https://github.com/mysociety/alaveteli/issues/5518)
- "External" requests (vaguely related issue https://github.com/mysociety/alaveteli/issues/2875)
- Site health checks. Are these useful? (vaguely related issue https://github.com/mysociety/alaveteli/issues/6062)
- Capistrano
- Admin stats graphs? Do we use these?

garethrees:
* https://github.com/mysociety/alaveteli/issues/7015

RichardTaylor:
>As an alternative to https://github.com/mysociety/alaveteli/issues/7079, we could remove the user to user messaging system

[...]

PR #6891, opened 2022-03-21
[#6875] Extract User::ExternalUser

garethrees:
## Relevant issue(s)

Fixes https://github.com/mysociety/alaveteli/issues/6875.

## What does this do?

* Reduce complexity of `AdminRequestController#destroy`
* Improve `InfoRequest#external_user_name` test data
* Extract `User::ExternalUser`

## Why was this needed?

https://github.com/mysociety/alaveteli/pull/6820 didn't handle the case that a `nil` user would be passed when an `InfoRequest` is external. Rather than adding a load of complexity around the application, extract a `User` duck-type for when we have an external user.

## Notes to reviewer

I haven't gone wild on the specs since this is an area of the codebase I'd like to get rid of somehow (https://github.com/mysociety/alaveteli/issues/6446).

### DIFF
--- a/app/models/user/external_user.rb
+++ b/app/models/user/external_user.rb
@@ -1,0 +1,44 @@
+# Specialisations for handling requests made externally and mirrored to
+# Alaveteli.
+class User::ExternalUser < User
+  def initialize(info_request:)
+    @info_request = info_request
+  end
+
+  def name
+    info_request.external_user_name || _('Anonymous user')
+  end
+
+  def url_name
+    fake_slug = MySociety::Format.simplify_url_part(name, 'external_user', 32)
+    (info_request&.public_body&.url_name || '') + '_' + fake_slug
+  end
+
+  def prominence
+    'backpage'
+  end
+
+  def censor_rules
+    []
+  end
+
+  def flipper_id
+    'User;external'
+  end
+
+  def is_pro?
+    false
+  end
+
+  def json_for_api
+    { name: name }
+  end
+
+  def external?
+    true
+  end
+
+  protected
+
+  attr_reader :info_request
+end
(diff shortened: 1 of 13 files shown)


app/models/user/external_user.rb
```diff
@@ -0,0 +1,44 @@
...
+    @info_request = info_request
+  end
+
+  def name
+    info_request.external_user_name || _('Anonymous user')
+  end
+
+  def url_name
+    fake_slug = MySociety::Format.simplify_url_part(name, 'external_user', 32)
+    (info_request&.public_body&.url_name || '') + '_' + fake_slug
```
garethrees, line 14: Looks like there's an issue with this actually

garethrees: Production request:

```ruby
r = InfoRequest.find 187945
r.external_user_name
# => nil
# r.user_name_slug
# => "brighton_and_hove_city_council_anonymous"
```

Dev request:

```
r = InfoRequest.find 115
r.external_user_name
# => nil
r.user_name_slug
# => "tgq_anonymous_user"
```

The problem is that we always return a `name` now, whereas before we were doing:

```
      if external_user_name.nil?
        fake_slug = "anonymous"
      else
        fake_slug = MySociety::Format.simplify_url_part(external_user_name, 'external_user', 32)
      end
```

garethrees: That said, I'm not sure this even worked previously: https://whatdotheyknow.com/user/brighton_and_hove_city_council_anonymous => 404

garethrees: If we don't think we're going to break URLs we could just hard code this.

```ruby
def url_name
  'anonymous'
end
```

garethrees: omg its all over the place:

* `app/views/admin_request/show.html.erb:86`
* `LinkToHelper#user_link_for_request`
* `LinkToHelper#user_admin_link_for_request`
* `LinkToHelper#request_user_link_absolute`
* `LinkToHelper#request_user_link`

PR #6910, opened 2022-03-29
Add AdminUser scopes

gbp:
## Relevant issue(s)

Following on from https://github.com/mysociety/alaveteli/pull/6823

## What does this do?

Combine `AdminUser#banned` with the `#index` action which includes searching, filtering and ordering so these can be done to within the banned user scope.

Also adds scopes for `active` and `closed` users because why not.

## Screenshots

image

### DIFF
--- a/app/controllers/admin_user_controller.rb
+++ b/app/controllers/admin_user_controller.rb
@@ -27,7 +27,7 @@ class AdminUserController < AdminController
     @sort_order =
       @sort_options.key?(params[:sort_order]) ? params[:sort_order] : 'name_asc'
 
-    users = User
+    users = @base_scope || User
 
     if @query.present?
       users = users.where(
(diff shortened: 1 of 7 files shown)


app/controllers/admin_user_controller.rb
```diff
@@ -88,12 +88,24 @@ def update
     end
   end
 
+  def active
+    @title = 'Active users'
+    @base_scope = User.active
```
garethrees, line 93: Not caused by this PR, but adding this prompted me to create https://github.com/mysociety/alaveteli/issues/6924. No problem for getting this merged, but just noting for posterity.

garethrees (approved):
Niiiice!

Added a suggestion in https://github.com/mysociety/alaveteli/pull/6925, so feel free to merge that in to this branch as `ff-only` and then merge the whole lot.

PR #6913, opened 2022-03-29
[sysadmin#1440] Add health check for Xapian job queue length

gbp:
## Relevant issue(s)

https://github.com/mysociety/sysadmin/issues/1440

## What does this do?

Add health check for Xapian job queue length

## Why was this needed?

If the job queue isn't being processed we need to alerted so the issue
can be resolved.

To prevent batch requests from triggering this check we ignore jobs
created within the last 30 minutes.

### DIFF
--- a/config/initializers/health_checks.rb
+++ b/config/initializers/health_checks.rb
@@ -17,7 +17,19 @@ Rails.application.config.after_initialize do
                                       OutgoingMessage.last.created_at
                                   end
 
+  xapian_queue_limit = 500
+  xapian_queue_check = HealthChecks::Checks::LimitCheck.new(
+    limit: xapian_queue_limit,
+    failure_message: _('The number of Xapian index jobs, older than 30 ' \
+                       'minutes, is over {{limit}}', limit: xapian_queue_limit),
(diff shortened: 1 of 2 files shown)


config/initializers/health_checks.rb
```diff
@@ -17,7 +17,19 @@
...
 
+  xapian_queue_limit = 500
+  xapian_queue_check = HealthChecks::Checks::LimitCheck.new(
+    limit: xapian_queue_limit,
+    failure_message: _('The number of Xapian index jobs, older than 30 ' \
+                       'minutes, is over {{limit}}', limit: xapian_queue_limit),
+    success_message: _('The number of Xapian index jobs, older than 30 ' \
+                       'minutes, is under {{limit}}', limit: xapian_queue_limit)
+  ) do
+    ActsAsXapian::ActsAsXapianJob.where(created_at: (...30.minutes.ago)).count
```
sagepe, line 28: I assume this does what it appears to - I'm not familiar with the underlying code though!

sagepe (approved):
Looks good to me from what I can see.

garethrees (approved):

PR #6918, opened 2022-03-31
[#1785] Allow admins to revert back to their own account

### LINKED ISSUE #1785: Button for admins to revert to their own account after logging in as a user
Currently when superusers who have logged in as another user to do something want to return to their own account they need to either log out and back in again or get to their account's admin page and click on the "log in as this user" button there.

There could be a button on the admin toolbar for admins to "become themselves" again in one click. 

The presence of such a button might also help make clear what's going on when you're logged in as another user clearer.

gbp:
## Relevant issue(s)

Fixes #1785

## What does this do?

When an admin has logged in as another user, this adds the ability to revert back to their own account.

## Why was this needed?

Too many incidences where I've forgotten I was signed in as someone else and thinking the admin has broken.

## Screenshots

image

## Notes to reviewer

Todo: specs ?

### DIFF
--- a/app/controllers/admin_users_sessions_controller.rb
+++ b/app/controllers/admin_users_sessions_controller.rb
@@ -7,20 +7,20 @@
 class AdminUsersSessionsController < AdminController
   def create
     # Don't use @user as that is any logged in user
-    @admin_user = User.find(params[:id])
+    @user_to_login_as = User.find(params[:id])
 
-    if cannot? :login_as, @admin_user
+    if cannot? :login_as, @user_to_login_as
       flash[:error] =
-        "You don't have permission to log in as #{ @admin_user.name }"
-      return redirect_to admin_user_path(@admin_user)
+        "You don't have permission to log in as #{ @user_to_login_as.name }"
+      return redirect_to admin_user_path(@user_to_login_as)
     end
 
-    @admin_user.confirm!
+    @user_to_login_as.confirm!
 
-    session[:user_id] = @admin_user.id
-    session[:user_login_token] = @admin_user.login_token
+    session[:user_id] = @user_to_login_as.id
+    session[:user_login_token] = @user_to_login_as.login_token
     session[:user_circumstance] = 'login_as'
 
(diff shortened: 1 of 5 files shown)


app/controllers/admin_users_sessions_controller.rb
```diff
@@ -24,4 +24,12 @@ def create
 
     redirect_to user_path(@user_to_login_as)
   end
+
+  def destroy
+    @admin_to_revert_to = User.find(session[:admin_id])
+
+    sign_in(@admin_to_revert_to)
+
+    redirect_to admin_general_index_path
```
garethrees, line 33: Might be nicer to redirect back to the `@user_to_login_as` user admin page, since that's where you "started" from to get to the login as mode.

app/controllers/admin_users_sessions_controller.rb
```diff
@@ -17,6 +17,7 @@ def create
 
     @user_to_login_as.confirm!
 
+    session[:admin_id] = current_user.id
```
garethrees, line 20: Worth adding spec:

```ruby
    it 'remembers the current admin' do
      post :create, params: { id: target_user.id }
      expect(session[:admin_id]).to eq(admin_user.id)
    end
```

app/controllers/admin_users_sessions_controller.rb
```diff
@@ -24,4 +24,12 @@ def create
 
     redirect_to user_path(@user_to_login_as)
   end
+
+  def destroy
```
garethrees, line 28: Worth adding some basic specs to this.

- Check that we're now logged in as the correct admin
- Check that we redirected to the right place

garethrees (approved):
This is amazing!

Couple of suggestions. No need for re-review.

Also add a changelog note!

RichardTaylor:
Excellent. 

I suspect this may also fix: "Make it easier for admins to add an annotation as the WhatDotheyKnow user" https://github.com/mysociety/whatdotheyknow-theme/issues/956

PR #6923, opened 2022-04-04
Add InfoRequestEvent#pro

gbp:
## What does this do?

Add `InfoRequestEvent#pro`

## Why was this needed?

This will be useful for showing impact of the Pro service.

For example showing how many citations have been created for pro/non-pro requests.

## Implementation notes

Initial implementation was to add a completely separate event which was created when a `InfoRequest` was first created.

Instead now, I've opted to record if the user was a pro on each `InfoRequestEvent`, as this might give us more insight in to actions carried out - for example if an embargo expires while the Pro is still a subscriber or if the Pro is following up or classifying their requests after their subscription has lasped.

### DIFF
--- a/app/models/info_request.rb
+++ b/app/models/info_request.rb
@@ -1136,6 +1136,7 @@ class InfoRequest < ApplicationRecord
   # Log an event to the history of some things that have happened to this request
   def log_event(type, params, options = {})
     event_attributes = { event_type: type, params: params }
+    event_attributes[:pro] = true if user.is_pro?
     event_attributes[:created_at] = options[:created_at] if options[:created_at]
     event = info_request_events.create!(event_attributes)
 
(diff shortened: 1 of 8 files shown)


garethrees:
I haven't looked through this in depth yet, but this, alongside the thought of tracking sign_ups_ as well as sign_ins_ (https://github.com/mysociety/alaveteli/pull/6822#issuecomment-1080345357) did prompt me to write up some thoughts on extracting some more generic domain components (https://github.com/mysociety/alaveteli/issues/6931).

One of those is an `Event` class that can be applied to _any_ appropriate model. With that in mind, I'm not totally sure that adding a new attribute directly on `InfoRequestEvent` makes sense. I'm not totally against it as it solves a problem for now, but it's worth us thinking about if we can avoid some pain here.

Did you look in to adding a new key in the `params_yaml` (https://github.com/mysociety/alaveteli/pull/6594#pullrequestreview-785625848)? It feels like whether an event was generated while the user was a pro is more incidental, but I appreciate that might currently make the querying more difficult. With the above approach of a generic `Event` class I think we'd want to move the params to a `jsonb` column or [`ActiveRecord::Store`](https://api.rubyonrails.org/classes/ActiveRecord/Store.html), though again I appreciate that doesn't help us right now.

garethrees:

PR #6928, opened 2022-04-06
Reposition the admin user list heading

garethrees:
Move it above the subnav tabs – feels wrong for a main page heading to
be "in between" elements.

This also moves the setting of `@title` from the view to the controller
to remove logic from views. In the `index` action we need to use
conditional assignment in case its been set via one of the filter lists
(banned, active, etc).

reposition-admin-user-list-title


gbp (approved):
Not run. Looks fine

PR #6944, opened 2022-04-08
Update query order/reorder syntax 

gbp:
## What does this do?

Update query order/reorder syntax

## Why was this needed?

Remove string based ordering and replace with symbols/hashes. This has
the benefit of allowing `ActiveRecord` to automatically adding the
correct database table prefix before each column. This is important when
using joins as sometimes there will be column name clashes. In Rails 7
this appears to be an issue with some of the `has_many` condition procs.

For ordering on joins tables we can't use symbols/hashes, so instead of:
  `Klass.joins(:join_klass).order('join_klasses.column')`
we can use the:
  `Klass.joins(:join_klass).merge(JoinKlass.order(:column))`

When doing this I needed to switch from using `includes` which adds the
join tables columns to query select, to `joins` and `preload`, so an
single extra database query is run.

## Notes to reviewer

Where we used to have `reorder` I have checked the new implementation matches the old.

### DIFF
--- a/app/controllers/admin_comment_controller.rb
+++ b/app/controllers/admin_comment_controller.rb
@@ -14,9 +14,9 @@ class AdminCommentController < AdminController
 
     comments = if @query
       Comment.where(["lower(body) LIKE lower('%'||?||'%')", @query]).
-        order('created_at DESC')
+        order(created_at: :desc)
     else
-      Comment.order('created_at DESC')
+      Comment.order(created_at: :desc)
     end
 
     if cannot? :admin, AlaveteliPro::Embargo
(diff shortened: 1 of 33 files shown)


gbp:
When running Rails 7 this fixes:

```
InfoRequest .....................................................................................................................................................................................................................................................................................F........................................................................................................................

  2) InfoRequest#foi_attachments when there are incoming messages with attachments is expected to be many
     Failure/Error: it { is_expected.to be_many }

     ActiveRecord::StatementInvalid:
       PG::AmbiguousColumn: ERROR:  column reference "id" is ambiguous
       LINE 1: ...ncoming_messages"."info_request_id" = $1 ORDER BY id, create...
                                                                    ^
     # ./spec/models/info_request_spec.rb:87:in `block (4 levels) in <top (required)>'
     # ------------------
     # --- Caused by: ---
     # PG::AmbiguousColumn:
     #   ERROR:  column reference "id" is ambiguous
     #   LINE 1: ...ncoming_messages"."info_request_id" = $1 ORDER BY id, create...
     #                                                                ^
     #   ./spec/models/info_request_spec.rb:87:in `block (4 levels) in <top (required)>'
```

garethrees (approved):
Haven't run but LGTM. All seems sensible and the tests pass 🥳

Issue #6493, opened 2021-08-24
Change password email address box too small / emails truncated


everdha:
In "change your password" page of Alaveteli (tested on the French site https://madada.fr/profile/change_password and on the UK one https://www.whatdotheyknow.com/profile/change_password/new?pretoken=[REDACTED]) the email address box is too short / small for showing not so long emails. Maximum seems to be 16 characters seen on screen, after that characters are written, but shown truncated, not visible to user. 

Typical user could think that the email is not written properly, and cannot check if it was written properly (unless he erases charcaters, then he can see what was written next), and has to figure out that he can ignore what he sees and go on by typing blindly the right address.

Try for example 123@567891123456789, this email will be visibly truncated after the 16th character, the second 6. The rest is there but no way to see it unless you erase characters.

EdaNano:
Issue opened on the French Alaveteli platform also MaDada : https://gitlab.com/madada-team/dada-core/-/issues/60

PR #6946, opened 2022-04-08
Increased width of email box when changing password

lucascumsille:
This PR increases the width of email box, so it can fit longer emails.
Tested on mobile so there is no overflow on small screen sizes.

Screenshot 2022-04-08 at 14 25 28

@garethrees let me know if you prefer this fix to go in _global_styling.scss instead of creating a new file, like I have done in this PR.
As I saw good list of well organised files, I thought it would be better this way, but maybe it's not worthy it for such small file.

Fixes: #6493

### DIFF
--- a/app/assets/stylesheets/responsive/_password_changes_style.scss
+++ b/app/assets/stylesheets/responsive/_password_changes_style.scss
@@ -1,0 +1,5 @@
+#change_password form{
+  input[type="email"] {
+    width: 280px;
+  }
+}
--- a/app/assets/stylesheets/responsive/all.scss
+++ b/app/assets/stylesheets/responsive/all.scss
@@ -84,6 +84,8 @@
 
 @import "responsive/_time_series";
 
+@import "responsive/_password_changes_style";
+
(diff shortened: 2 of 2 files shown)


garethrees (approved):
This looks great, thanks!

I've pushed a couple of fixup commits with some feedback notes within as they were so minor and easy enough to just do rather than explain and then put back on your hands. Here's what I tweaked:

* Use 2 space indentation
* Use `_style` suffix as per existing conventions of using a `_layout` and `_style` file per view.
* Tweak filename in line with controller action name (`PasswordChangesController`)

lucascumsille:
> This looks great, thanks!
> 
> I've pushed a couple of fixup commits with some feedback notes within as they were so minor and easy enough to just do rather than explain and then put back on your hands. Here's what I tweaked:
> 
> * Use 2 space indentation
> * Use `_style` suffix as per existing conventions of using a `_layout` and `_style` file per view.
> * Tweak filename in line with controller action name (`PasswordChangesController`)

Thanks for the explanation Gareth, much appreciated it.

Issue #5485, opened 2019-12-03
Sanitise all YAML stored in the database


gbp:
Following on from #5462 

We should consider removing all ruby objects from YAML stored in the DB and switch to using primary keys or converting objects to strings. 

Some InfoRequestEvent records currently store users data which could easily be overlooked when removing user accounts.

gbp:
We are seeing `DEPRECATION WARNING: YAML loading from legacy format older than Rails 5.0 is deprecated and will be removed in Rails 6.2. (called from load at /home/runner/work/alaveteli/alaveteli/lib/yaml_compatibility.rb:14)` in the spec suite now.

This is required before #6341

gbp:
I have gone through the WDTK production DB to see which ruby objects (and how many) are stored as YAML in the database:

```ruby
pp InfoRequestEvent.where("params_yaml LIKE '%!ruby/object%'").pluck(:params_yaml).inject([]) { |m, y| m + y.scan(/!ruby\/object.*/) }.inject(Hash.new(0)) { |t, o| t[o] += 1; t }

{"!ruby/object:ActiveSupport::TimeWithZone"=>209902,
 "!ruby/object:ActiveSupport::TimeZone"=>209900,
 "!ruby/object:User"=>6490,
 "!ruby/object:ActiveRecord::Attribute::FromDatabase"=>70722,
 "!ruby/object:ActiveRecord::AttributeSet"=>1880,
 "!ruby/object:ActiveRecord::LazyAttributeHash"=>1880,
 "!ruby/object:ActiveModel::Type::Integer"=>1880,
 "!ruby/object:ActiveRecord::Type::String"=>1525,
 "!ruby/object:ActiveRecord::AttributeMethods::TimeZoneConversion::TimeZoneConverter"=> 7100,
 "!ruby/object:ActiveRecord::ConnectionAdapters::PostgreSQL::OID::DateTime"=> 1880,
 "!ruby/object:ActiveRecord::Type::Boolean"=>1525,
 "!ruby/object:ActiveRecord::Type::Text"=>1880,
 "!ruby/object:ActiveModel::Type::String"=>355,
 "!ruby/object:ActiveModel::Type::Boolean"=>355,
 "!ruby/object:IncomingMessage "=>52,
 "!ruby/object:InfoRequest "=>52,
 "!ruby/object:TMail::Mail "=>139,
 "!ruby/object:TMail::StringPort "=>191,
 "!ruby/object:TMail::Config "=>44,
[...]

PR #6949, opened 2022-04-08
Event params jsonb

gbp:
Friday spike for changing `InfoRequesEvent#params` to a JSONB column.

One issue we're going to hit is storing model instances in the database. We need to remove these as JSON values can only be a number of limited types. We want to remove all instances (see #5485) so this spike also is an initial attempt at that (for application models at least). I'm thinking the migration process to JSONB would be a good time to sanitise these instances.

### DIFF
--- a/spec/controllers/reports_controller_spec.rb
+++ b/spec/controllers/reports_controller_spec.rb
@@ -61,7 +61,7 @@ RSpec.describe ReportsController do
         expect(last_event.params).
           to match(
             request_id: info_request.id,
-            editor: user,
+            editor: user.url_name,
             reason: 'my reason',
             message: '',
             old_attention_requested: false,
--- a/app/models/info_request_event.rb
+++ b/app/models/info_request_event.rb
@@ -364,7 +364,6 @@ class InfoRequestEvent < ApplicationRecord
     ignore = {}
     for key, value in params
       key = key.to_s
-      value = value.url_name if value.is_a?(User)
       if key.match(/^old_(.*)$/)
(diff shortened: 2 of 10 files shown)


garethrees:
> I'm thinking the migration process to JSONB would be a good time to sanitise these instances.

💯

spec/controllers/reports_controller_spec.rb
```diff
@@ -61,7 +61,7 @@
         expect(last_event.params).
           to match(
             request_id: info_request.id,
-            editor: user,
+            editor: user.url_name,
```
gbp, line 64: Think we should include a bit more information for these instance objects, like: 

```
editor: { type: 'User', id: 1, url_name: 'foo' },
authority: { type: 'PublicBody', id: 1, title: 'Ministry of ...', version: 4 }
```

garethrees: > `editor: { type: 'User', id: 1, url_name: 'foo' }`

I was thinking that we'd have an explicit `Event#editor` relation rather than storing that in params.

> `authority: { type: 'PublicBody', id: 1, title: 'Ministry of ...', version: 4 }`

+1, nice. Certainly the record class and ID, as then we can constantise and find the record to render out more information. FWIW, BC stores this as an [`Event::Details`](https://github.com/garethrees/on_writing_software_well/blob/master/bc3/app/models/recording/eventable.rb#L16-L19) associated with the `Event`.

garethrees:

Issue #6568, opened 2021-10-12
Make embargo status clear to Pro users and encourage publication


RichardTaylor:
Issue raised having seen[ a Tweet](https://twitter.com/librarieshacked/status/1447876576780161033) by a Pro user who appears unsure about the embargo they have set: 

>I just started using their pro 'batch' functionality and _I think_ they'll get published after 3 months.

My _emphasis_.

Related: #6051

garethrees:
Possible quick fix here:

![Screenshot 2021-10-14 at 13 11 58](https://user-images.githubusercontent.com/282788/137315102-908269a8-3caf-4a4b-80ae-9849e040a1ba.png)

lucascumsille:
> Possible quick fix here:
> 
> ![Screenshot 2021-10-14 at 13 11 58](https://user-images.githubusercontent.com/282788/137315102-908269a8-3caf-4a4b-80ae-9849e040a1ba.png)

Do we want to encourage the user to make it public or should it have the same level of importance as the "Change privacy" button?

garethrees:
a bit bolder would be nice, perhaps: [Change privacy or **publish now**]()

PR #6953, opened 2022-04-11
Added styling for "Publish now" button

lucascumsille:
Added bold class to "Publish now" button. (The font-weight for this button has been increased so it pops out
more than his sibling element)

Fixes: #6568 
Screenshot 2022-04-11 at 07 43 30

### DIFF
--- a/app/views/alaveteli_pro/info_requests/_embargo_form.html.erb
+++ b/app/views/alaveteli_pro/info_requests/_embargo_form.html.erb
@@ -7,7 +7,7 @@
   <%= render partial: "alaveteli_pro/info_requests/embargo_info",
              locals: {embargo: info_request.embargo, tense: :present} %>
 
-  <label class="houdini-label" for="input1"><%= _("Change privacy") %></label>
+  <label class="houdini-label" for="input1"><%= _("Change privacy or <strong>publish now</strong>") %></label>
   <input class="houdini-input" type="checkbox" id="input1">
   <div class="houdini-target extend-embargo-sidebar">
   <% if info_request.embargo && can?(:update, info_request.embargo) %>


app/views/alaveteli_pro/info_requests/_embargo_form.html.erb
```diff
@@ -8,6 +8,8 @@
              locals: {embargo: info_request.embargo, tense: :present} %>
 
   <label class="houdini-label" for="input1"><%= _("Change privacy") %></label>
+  <span style="font-size:13px">or</span>
+  <label class="houdini-label is--bold" for="input2"><%= _("Publish now") %></label>
```
garethrees, line 12: How about just doing?

```diff
- <label class="houdini-label" for="input1"><%= _("Change privacy") %></label>
+ <label class="houdini-label" for="input1"><%= _("Change privacy or <strong>publish now</strong>") %></label>
```

garethrees: We just decided on standup to keep these actions as two individual links that should unfold the correct options (i.e. clicking "publish now" only unfolds the "publish now" button). We wanted to keep the two-step process so that it's clear the user is taking an intentional step to publish as it's not easy to revert.

lucascumsille:
@garethrees I got rid of the .is--bold class and just use the <strong> instead. 

I also added the "Publish now" input, however I wasn't able to test if the content inside the `<div class="houdini-target extend-embargo-sidebar">` was correct. 
The way this solution has been implemented is:
"Change privacy" should work as usual, it should still display the "publish" option, because is still part of the privacy settings.
While the "Publish now" should only display the option to publish the request.

Correct me if I'm wrong, but the `_sidebar_layout.scss` files is it part of the Alaveteli core?

garethrees:
> I also added the "Publish now" input, however I wasn't able to test if the content inside the `<div class="houdini-target extend-embargo-sidebar">` was correct.

You _should_ get these options. Maybe try creating a new private request in case its a weird issue with the data of the sample request you're using?

> The way this solution has been implemented is: "Change privacy" should work as usual, it should still display the "publish" option, because is still part of the privacy settings. While the "Publish now" should only display the option to publish the request.

Hmm, this sounds a little confusing to me. I'd either have one button; one thing getting unfolded, or just have a single link that unfolds both options.

> Correct me if I'm wrong, but the `_sidebar_layout.scss` files is it part of the Alaveteli core?

Yep

lucascumsille:
Thanks for the feedback Gareth
I tried adding a new request, but again nothing was unfolding. If you want I can check with @gbp when he is back.

At the end I ended up implementing the initial solution(one link unfolding all the content), mainly for two reasons, yesterday checking how the Houdini labels worked I realised that when unfolding the "Change privacy" it pushed "Publish now" right after the unfolded div, no ideal, but no terrible. And also having both option on one achieves the goal with less intervention on the code.

garethrees (approved):

PR #6966, opened 2022-04-14
Improve Xapian jobs health check

gbp:
## Relevant issue(s)

https://github.com/mysociety/sysadmin/issues/1440

## What does this do?

Improve Xapian jobs health check

## Why was this needed?

False triggering the health check when lots of jobs build up. Say from changing a public body's name.


config/initializers/health_checks.rb
```diff
@@ -20,15 +20,15 @@
...
+  xapian_queue_check = HealthChecks::Checks::PeriodCheck.new(
+    period: 30.minutes,
+    failure_message: _('The oldest Xapian index job, has been idle for more ' \
+                       'than 30 minutes'),
+    success_message: _('The oldest Xapian index job, hasn\'t been idle for ' \
+                       'more than 30 minutes')
   ) do
-    ActsAsXapian::ActsAsXapianJob.where(created_at: (...30.minutes.ago)).count
+    oldest_job = ActsAsXapian::ActsAsXapianJob.order(:created_at).first
+    oldest_job&.created_at || Time.now
```
garethrees, line 31: `Time.zone.now`?

garethrees (approved):
Only given it a very quick skim but looks great!

PR #6967, opened 2022-04-14
Update PublicBody#expire_requests delegation

gbp:
## Relevant issue(s)

#7101

## What does this do?

Update PublicBody#expire_requests delegation

## Why was this needed?

This method is called when a `PublicBody#name` is changed as the
`#url_name` will also get updated.

Delegating to `InfoRequest#expire` isn't ideal as this method will drop
database cache columns for any incoming messages.

This means all messages received from a given `PublicBody` will be
reparsed when `#name`/`#url_name` is changed. This very costly and
something admins should be able to do without impacting the site.

### DIFF
--- a/app/models/public_body.rb
+++ b/app/models/public_body.rb
@@ -714,7 +714,7 @@ class PublicBody < ApplicationRecord
   end
 
   def expire_requests
-    info_requests.find_each(&:expire)
+    info_requests.find_each(&:reindex_request_events)
   end
 
   def self.where_clause_for_stats(minimum_requests, total_column)


app/models/public_body.rb
```diff
@@ -714,7 +714,7 @@ def json_for_api
   end
 
   def expire_requests
-    info_requests.find_each(&:expire)
+    info_requests.find_each(&:reindex_request_events)
```
garethrees, line 717: -1 to this particular implementation. I'd expect a method called `expire_requests` to call `expire` on the associated requests.

If we need two different reindexing methods in different situations we should create clear and consistent domain language for that.

garethrees:

Issue #6844, opened 2022-03-02
Make canned censor rule redaction replacements start with a lowercase letter


RichardTaylor:
Censor rules often end up mid-sentence, so shouldn't start with a capital letter. 

If they should have a capital letter when they appear other than mid-sentence I'm not sure, I suspect perhaps they should. 

The introduction of Canned CensorRule redaction replacements #6487 led to a change in practice.  

This is clearly a very minor issue; the reason for raising it is to seek to make the site/service look professional and smart, which I hope might contribute to encouraging good quality use of the service. 

Related:
https://github.com/mysociety/alaveteli/issues/4626#issuecomment-431662093

garethrees:
I personally don't have a strong opinion on this, but then I capitalised the canned replacements so maybe I shouldn't weigh in.

I assume we use them more mid-sentence than where a capitalised version _would_ be correct?

The canned replacements are only a starting point so can be edited prior to form submission, but I agree we want to have it work without modification for our most frequent case.

PR #7036, opened 2022-05-18
[#6844] Lowercase canned CensorRule replacements

garethrees:
> Censor rules often end up mid-sentence, so shouldn't start with a
> capital letter

I've left "Personally Identifiable Information" capitalised as that's a
proper noun and commonly referred to as PII.

Fixes https://github.com/mysociety/alaveteli/issues/6844.


garethrees:
The canned censor rules haven't been included in a release yet so no translation issues to be concerned about.

garethrees:
The canned censor rules haven't been included in a release yet so no translation issues to be concerned about.

PR #7044, opened 2022-05-23
Update Apache config

gbp:
## Relevant issue(s)

Fixes https://github.com/mysociety/sysadmin/issues/1624

## What does this do?

Configure Passenger/Ruby to require bundler. Our Gemfile now contains
different versions of gems which shipped with Ruby 2.7. 

## Why was this needed?

Without this change the application couldn't be launched.

## Notes to reviewer

This has been run in production since March 30th


garethrees (approved):
LGTM. Warrants an upgrade note in the changelog.

Issue #6999, opened 2022-04-29
Admin links to embargoed individual responses work for non-pro-admins


RichardTaylor:
This issue has been reported by another admin. 

I'm not sure how general it is, nor can I precisely describe the issue.

FOIMonkey:
I don't have admin rights to manage pro requests. If I know the request/response number, I can still access the admin pages for individual responses to embargoed requests eg: https://www.whatdotheyknow.com/admin/incoming_messages/2028904/edit 
![2022-04-29](https://user-images.githubusercontent.com/11962965/165961838-124eb68a-6fd6-452f-92e5-094cc1e5f03b.png)

FOIMonkey:
The admin page for the entire request displays as expected. 
![2022-04-29 (2)](https://user-images.githubusercontent.com/11962965/165962923-9ab24449-dfb3-4519-bf24-ce6b1139119b.png)

PR #7047, opened 2022-05-25
Prevent admin users from accessing embargoed content

laurentS:
## Relevant issue(s)
#6999 

## What does this do?

This PR (tries to) prevent (non-pro) admin users from having access to embargoed requests content (metadata, incoming and outgoing emails).

## Why was this needed?

As described in the linked issue, it is currently possible for an admin user to craft a url that will bypass auth checks and access embargoed content.

## Implementation notes

I am not very familiar with rails auth system and CanCanCan's mechanism, so I've mostly relied on copy/pasting what is done in other parts of the code. I've tried to add tests as well. I'm not sure any of this is optimal.

## Notes to reviewer

Feel free to take over the PR content and adjust as needed.

### DIFF
--- a/app/controllers/admin_incoming_message_controller.rb
+++ b/app/controllers/admin_incoming_message_controller.rb
@@ -3,6 +3,9 @@ class AdminIncomingMessageController < AdminController
   before_action :set_incoming_message, :only => [:edit, :update, :destroy, :redeliver]
 
   def edit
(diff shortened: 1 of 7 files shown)


gbp:
@laurentS thanks for starting this. I have continued the work (with the aim to get in to the release 0.41 next week). PR #7123 has the latest work.

PR #6707, opened 2021-12-10
Refactor incoming message parse raw email

garethrees:
## What does this do?

* Add `IncomingMessage#parse_raw_email`
* Remove `IncomingMessage#parse_raw_email!` options

## Why was this needed?

Reduce complexity and improve readability.

It's more conventional to have a bang and non-bang version for this type of behaviour.

PR #7053, opened 2022-05-27
Add IncomingMessage::CacheAttributesFromRawEmail

garethrees:
Extract duplication when caching RawEmail attrs

All these methods do the same thing – override the accessor to ensure
`parse_raw_email!` is called, and then call the original accessor.

`parse_raw_email!` caches each attribute from the `RawEmail` on
`IncomingMessage`.

This commit extracts that duplication to a concern, with an explicit
DSL method for defining which attributes should be cached. This makes it
clear that there's "something different" about these methods, rather
than invisibly overriding them in the concern directly.

Existing specs on IncomingMessage cover this behaviour.

### DIFF
--- a/app/models/incoming_message.rb
+++ b/app/models/incoming_message.rb
@@ -39,6 +39,7 @@ require 'zip'
 class IncomingMessage < ApplicationRecord
   include AdminColumn
   include MessageProminence
+  include CacheAttributesFromRawEmail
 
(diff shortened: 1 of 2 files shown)


garethrees:
@gbp have a look and see what you think. If you're happy I'll tidy the commit up. Existing specs should cover it.

gbp:
> @gbp have a look and see what you think. If you're happy I'll tidy the commit up. Existing specs should cover it.

Yes looks good. Could we get this done with #6707 too

PR #7075, opened 2022-06-08
[#7074] Add taggable module to incoming message

### LINKED ISSUE #7074: Make messages taggable
Requests (in other words correspondence threads) are taggable and there is a ticket for making those tags, and the tagging system, public at https://github.com/mysociety/alaveteli/issues/2302

Some things we might want to tag will relate to a specific message. 

One could get even more specific, it may be useful to relate a tag to specific text, or a specific attachment.

gbp:
## Relevant issue(s)

Fixes #7074

## What does this do?

Allow individual message to be tagged within the admin UI.

### DIFF
--- a/app/controllers/admin_incoming_message_controller.rb
+++ b/app/controllers/admin_incoming_message_controller.rb
@@ -118,7 +118,9 @@ class AdminIncomingMessageController < AdminController
 
   def incoming_message_params
     if params[:incoming_message]
-      params.require(:incoming_message).permit(:prominence, :prominence_reason)
+      params.require(:incoming_message).permit(
+        :prominence, :prominence_reason, :tag_string
+      )
     else
       {}
     end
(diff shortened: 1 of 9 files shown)


garethrees (changes requested):
Nice!

Looks like we don't handle the change of tag string, which makes for a bit of a confusing edit event. In the case of only modifying the tag string and not prominence, I don't think we want to record an empty prominence change. I expect this is down to not calling `strip_attributes` in `IncomingMessage`.

Screenshot 2022-06-17 at 12 55 20

Should also add the tag string on the dropdown of attributes:

Screenshot 2022-06-17 at 12 56 07
…snip…
Screenshot 2022-06-17 at 12 55 57

Also one issue inline around the UI of the form.

It's always helpful to include screenshots when there are UI changes 🙂

PR #7077, opened 2022-06-08
[#6696] Add warning to help prevent duplicate batches

### LINKED ISSUE #6696: Prevent duplicate submissions in AlaveteliPro::InfoRequestBatchesController
When creating an `InfoRequestBatch` through the legacy public interface we [check for duplicate submissions](https://github.com/mysociety/alaveteli/blob/0.40.0.0/app/controllers/request_controller.rb#L234-L237).

There's no call to `InfoRequestBatch.find_existing` in [`AlaveteliPro::InfoRequestBatchesController`](https://github.com/mysociety/alaveteli/blob/0.40.0.0/app/controllers/alaveteli_pro/info_request_batches_controller.rb) and from a quick scan there's no sign that this duplicate prevention is handled at the model layer.

We've noted [one instance](https://github.com/mysociety/alaveteli/issues/4160#issuecomment-976950890) of a duplicate batch being created.

gbp:
## Relevant issue(s)

Fixes #6696

## What does this do?

Add warning to help prevent duplicate batches

## Why was this needed?

Sometimes Pro users are making the same batch to request to the same bodies.

## Implementation notes

We may want to change the `InfoRequestBatch.find_existing` to return an existing batch if any of the new batch's public bodies were included in the previous batch. Currently the existing implementation (which I haven't changed) won't match if extra new bodies have been added to a batch are has already been sent.

## Screenshots

image

### DIFF
--- a/app/assets/javascripts/alaveteli_pro/alaveteli_pro.js
+++ b/app/assets/javascripts/alaveteli_pro/alaveteli_pro.js
@@ -4,6 +4,7 @@
 //= require alaveteli_pro/embargo_dropdown
 //= require alaveteli_pro/status_dropdown
 //= require alaveteli_pro/marketing
+//= require alaveteli_pro/existing_batch
 
 // These modules must be initialised first, because later sub components may
 // need access to things defined in either one of them.
--- a/app/assets/javascripts/alaveteli_pro/existing_batch.js
+++ b/app/assets/javascripts/alaveteli_pro/existing_batch.js
@@ -1,0 +1,9 @@
+$(function() {
+  $('#ignore_existing_batch').change(function() {
+    if ($(this).prop('checked')) {
+      $('#submit_button').removeAttr('disabled', '');
+    } else {
+      $('#submit_button').attr('disabled', 'disabled');
+    }
+  }).change();
+});


app/views/alaveteli_pro/info_request_batches/_message_preview.html.erb
```diff
@@ -27,7 +27,18 @@
...
+        <%= _('As a Pro user, you can choose to ignore the existing batch by ' \
+              'checking the box below and send your requests. Please ' \
+              'remember you are still bound by our <a href="{{url}}">House ' \
+              'Rules on making responsible requests</a>.',
+              url: help_requesting_path(anchor: 'responsible')) %>
       </div>
+
+      <%= label_tag :ignore_existing_batch do %>
+        <%= check_box_tag :ignore_existing_batch %>
+        <%= _('Ignore existing batch') %>
```
garethrees, line 40: - [ ] Should disable the submit tag unless this is checked.
- [ ] Change wording to "Send this batch anyway?"

app/views/alaveteli_pro/info_request_batches/_message_preview.html.erb
```diff
@@ -27,7 +27,18 @@
               'request to make a new but similar batch of requests.',
               date: simple_date(@info_request_batch.existing_batch.created_at),
               existing_batch: info_request_batch_path(@info_request_batch.existing_batch)) %>
+
+        <%= _('As a Pro user, you can choose to ignore the existing batch by ' \
+              'checking the box below and send your requests. Please ' \
+              'remember you are still bound by our <a href="{{url}}">House ' \
+              'Rules on making responsible requests</a>.',
+              url: help_requesting_path(anchor: 'responsible')) %>
```
garethrees, line 35: > We may want to change the `InfoRequestBatch.find_existing` to return an existing batch if any of the new batch's public bodies were included in the previous batch.

I agree, I think we should do this, which would then allow the full `errorExplanation` text to read:

> **You have already sent this request** to some authorities in this batch on 13 June 2022.
>
> You can [view the existing batch](), [edit your request](), or ignore this warning by checking the box below.
>
> Please remember that our [terms and conditions]() include **making responsible requests**.

`help_requesting_path(anchor: 'responsible')` isn't a [required section](https://alaveteli.org/docs/customising/themes/#customising-the-help-pages), but `/pro/pages/legal` is when pro is enabled.

Would ideally make each line a `<p>` – should work, but no big deal if it introduces lots of formatting hassle.

garethrees (commented):
This is great!

> We may want to change the `InfoRequestBatch.find_existing` to return an existing batch if any of the new batch's public bodies were included in the previous batch.

I think we should do this as it would allow the warning message to be tightened up and make sense in more use cases (e.g. a requester making a pilot batch, then lazily using the category picker to include them again in the "full" batch).

Don't think I need to re-review after this change, but feel free to raise if you want a second pair of eyes.

app/views/alaveteli_pro/info_request_batches/_message_preview.html.erb
```diff
@@ -32,7 +32,8 @@
           <%=
             _('You can view the ' \
               '<a href="{{existing_batch_url}}">existing batch</a>, or ' \
-              '<a href="{{edit_batch_url}}">edit you request</a>.',
```
garethrees: Couple of minor tweaks:

* Drop the "or"  on L34
* Spelling typo: you -> you**r** request on L35
* tweak the final option to "or ignore this warning by confirming you want to proceed anyway."

So it should read:

> You can view the existing batch, edit your request, or ignore this warning by confirming you want to proceed anyway.

garethrees (commented):
Aw yeaaa! 😎 Looking _great_! Tiny tweaks suggested but good to go once resolved.

gbp:
> > We may want to change the `InfoRequestBatch.find_existing` to return an existing batch if any of the new batch's public bodies were included in the previous batch.
> 
> I think we should do this

Turns out this the existing behaviour. So all good :)

Issue #1294, opened 2013-12-18
Show changes to tags in public body version history 


crowbot:
Spun out from discussion on https://github.com/mysociety/alaveteli/issues/521

garethrees:
This is desirable, but unlikely to be worked on in the next 12 months so closing for now.

PR #7100, opened 2022-06-13
[#7050] Add authority remove tag button

### LINKED ISSUE #7050: Ability to delete a tag from a subset of authorities
As an admin I would like to use the Authorities list function on admin GUI to list authorities with a specific tag and be able to delete that tag for all those authorities just like I can add a tag to all authorities in a filtered list

gbp:
## Relevant issue(s)

Fixes #7050

## What does this do?

Add authority remove tag button to the admin public bodies index UI.

## Why was this needed?

For easier management of authorities' tags.

## Implementation notes

Need to use existing search for to find authorities which have a current tag, then the remove button with be displayed along side the add new tag button.

## Screenshots

image


app/views/admin_public_body/_one_list.html.erb
```diff
@@ -28,14 +28,14 @@
...
-  <p>
-    <%= text_field_tag 'new_tag', params[:new_tag], { :size => 15, :id => "mass_add_tag_new_tag_" + table_name } %>
-    <%= hidden_field_tag(:query, params[:query], { :id => "mass_add_tag_query_" + table_name } ) %>
-    <%= hidden_field_tag(:page, params[:page], { :id => "mass_add_page_" + table_name } ) %>
-    <%= hidden_field_tag(:table_name, table_name, { :id => "mass_add_tag_table_name_" + table_name } ) %>
-    <%= submit_tag "Add tag to all", :class => "btn btn-primary" %>
-    <span class="help-inline">(in table just above)</span>
-  </p>
-<% end %>
+<p>
```
garethrees, line 31: Should be a `<div>` I think?

app/controllers/admin_public_body_controller.rb
```diff
@@ -135,10 +135,19 @@ def mass_tag
...
+        bodies.each do |body|
+          body.add_tag_if_not_already_present(params[:tag])
+        end
+        flash[:notice] = "Added tag to table of bodies."
+
+      elsif request.delete?
+        bodies.each do |body|
+          body.remove_tag(params[:tag])
+        end
+        flash[:notice] = "Removed tag from table of bodies."
```
garethrees, line 149: Minor tweaks:

```ruby
      if request.post?
        bodies.each { |body| body.add_tag_if_not_already_present(params[:tag]) }
        flash[:notice] = 'Added tag to table of bodies.'
      elsif request.delete?
        bodies.each { |body| body.remove_tag(params[:tag]) }
        flash[:notice] = 'Removed tag from table of bodies.'
      end
```

* Block style
* Remove newline before `elsif`
* Single quote strings

gbp:
Merged in 5165a47ddf07e5bd58e081de1c938a4ed8448c73

garethrees:
> I also noticed in testing that adding and removing tags in this way doesn't create a `Version`, though editing tags through the body edit form does. Probably one to ticket up since it's an existing omission.

Already ticketed as https://github.com/mysociety/alaveteli/issues/1294.

Issue #7102, opened 2022-06-13
Make clearer the fact the WhatDoTheyKnow request form is not a public body's general, direct, contact form


RichardTaylor:

garethrees:
Made an improvement at https://github.com/mysociety/alaveteli/pull/7103

RichardTaylor:
User feedback on WhatDoTheyKnow :

>I am not sure how your organisation even got involved.

garethrees:
We could add a much more prominent banner that we're a third-party FOI archive for non-logged in visitors, and get rid of the additional material once logged in, or once logged in and a user has more than one FOI request (indicating they understand what's happening) – via https://github.com/mysociety/alaveteli/issues/7517

garethrees:
Is it clear who "notes" are from? https://github.com/mysociety/alaveteli/issues/7543

garethrees:
Closing as too general.

Issue #6220, opened 2021-04-29
Support translated strings for custom legislation


gbp:
We current have some translated strings with interpolation like:
```ruby
_("Your {{law_used_full}} request has been sent", law_used_full: legislation.to_s(:full))
```

This works fine for the English locale but for others, such as French, we can't accurately subsitute the law used into the string. EG if we had `_("{{law_used_full}} request made", ...)` the translation wouldn't match existing strings in the PO files:

```gettext
msgid "Freedom of Information"
msgstr "Liberté d'accès à l'information"

msgid "Freedom of Information requests made"
msgstr "Demande d'accès à l'information envoyée"
```

We should conside using the `p_` gettext method to add context to the strings:
```ruby
p_('for', 'Law used') #=> 'Freedom of Information'
p_('foi', 'Your request has been sent') #=> 'Your Freedom of Information request has been sent'

p_('eir', 'Law used') #=> 'Environmental Information Regulations'
p_('eir', 'Your request has been sent') #=> 'Your Environmental Information Regulations request has been sent'
```

This would open up the option to support translated strings for custom legislations.

We can't hardcode the legislation context into the `p_` calls so a more relistic example would be:
 ```ruby
legislation = @public_body.legislation.first
p_(legislation, 'Law used')
```

This raises a problem as the `gettext:po:update` rake command won't find these string as understandally it doesn't run the whole application and wouldn't know what `@public_body.legislation.first` evaulated to.

Instead we would need to do something similar to what `gettext_i18n_rails` gem does for model attributes where it generates a `locale/model_attributes.rb` file https://github.com/mysociety/alaveteli/blob/ee3699b0041f49699b8700cbb10e3bc59a86cb8d/locale/model_attributes.rb#L1-L3

We would need to grep `.rb` and `.erb` files for `p_(legislation....` and for each known legislation generate a file, EG:
`locale/legislation/foi.rb`:
```ruby
#DO NOT MODIFY! AUTOMATICALLY GENERATED FILE!
p_('for', 'Law used')
p_('foi', 'Your request has been sent')
```
`locale/legislation/custom.rb`:
```ruby
#DO NOT MODIFY! AUTOMATICALLY GENERATED FILE!
p_('custom', 'Law used')
p_('custom', 'Your request has been sent')
```

HelenWDTK:
This issue has been automatically closed due to a lack of discussion or resolution for over 12 months.
Should we decide to revisit this issue in the future, it can be reopened.

PR #7103, opened 2022-06-15
[#7102] Make it clearer that users are making info requests

garethrees:
Sometimes users use Alaveteli to make general or unsuitable requests. It
is likely they are less digitally literate and believe the authority
page to be an official or general contact form.

This tweak makes it clearer that users will be making an info request.

Part of https://github.com/mysociety/alaveteli/issues/7102

Screenshot 2022-06-15 at 09 41 57

Screenshot 2022-06-15 at 09 58 33


gbp (commented):
This is fine for now. Once we have #6220 then this can be refactored.

PR #7110, opened 2022-06-20
Add drag and drop handles to admin category UI

garethrees:
Make it more obvious that headings and categories can be reordered
through drag and drop.

Had to tweak the JavaScript so that it didn't turn the "move" icon into
a chevron.

Screenshot 2022-06-20 at 17 50 34


garethrees:
Hmm this failed to get marked as merged by GitHub. It _was_ merged as a177aa3.

Issue #6995, opened 2022-04-28
Update foi_no special tag description


garethrees:
Currently:

> `foi_no`: if FOI does not apply but the authority is still willing to accept requests

We're also considering an `foi_voluntary` tag in https://github.com/mysociety/whatdotheyknow-theme/issues/1143:

> `foi_voluntary`: Body volunteers to act as if it’s subject to FOI law.

To me, these sound basically the same ("willing to accept"), but `foi_no` is also used even if they do not voluntarily comply:

> Many of the `foi_no` bodies very much do not voluntarily comply!  foi_no is for when a body isn't subject to FOI but we think it should be.  They may or may not be happy with that position!
>
> – volunteer comment

This would be easier with https://github.com/mysociety/alaveteli/issues/6994.

garethrees:
Suggestion to kick things off:

```diff
- foi_no: if FOI does not apply but the authority is still willing to accept requests
+ foi_no: FOI does not apply but we list the authority to campaign for it to be subject to FOI
```

RichardTaylor:
I agree with the proposal. 

Related ticket on improving the statement the foi_no tags cause to be added to body pages, and suggesting a note appears on request pages too:

https://github.com/mysociety/whatdotheyknow-theme/issues/914

PR #7115, opened 2022-06-21
[#6995] Update foi_no special tag description

garethrees:
Clarify that we also use `foi_no` even when an authority does not
voluntarily comply.

Fixes https://github.com/mysociety/alaveteli/issues/6995

Screenshot 2022-06-21 at 17 34 17


garethrees:
Ugh github didn't auto-close again

PR #7117, opened 2022-06-22
[#7018] Add batch authority count

### LINKED ISSUE #7018: Give some indication of the size of the batch
Moved from private issue at https://github.com/mysociety/alaveteli-professional/issues/237.

From @crowbot:
* So that users are hyper-aware that this is going to send emails to lots of authorities who will have to deal with them at some cost.
* eg. make the preview look like it is a stack of papers, have some kind of cost estimate total.

From @zarino:

Random thought:

tm_starfield

From me:
* We should do this:
  * On the authority selection interface (https://github.com/mysociety/alaveteli-professional/issues/364).
  * On the preview page.
* Should do a basic implementation of this before we roll out the category picker, because that allows you to easily add loads of authorities.

gbp:
## Relevant issue(s)

Fixes #7018

## What does this do?

Display the number of authorities added to a draft batch. Dynamically update this as authorities and added or removed.

## Why was this needed?

To make pro users more aware of the amount of requests they are sending and the amount of work their batch is going to generate. 

## Implementation notes

It seemed logical to display along side the "Write Request" button. This button is in different locations on the search/browse modes so now this count is too.

Attempted to handle plural terms other locales have more but limited to `zero`, `one`, `many` given the effort required (either lots of duplication or adding some sort of gettext JS library) to implement them all.

## Screenshots

https://user-images.githubusercontent.com/5426/174986565-8c5e1e76-fbf5-4eb7-93b3-6d45fe609cc5.mov

image

### DIFF
--- a/app/assets/javascripts/alaveteli_pro/alaveteli_pro.js
+++ b/app/assets/javascripts/alaveteli_pro/alaveteli_pro.js
@@ -23,5 +23,6 @@
 //= require alaveteli_pro/batch_authority_search/results
 //= require alaveteli_pro/batch_authority_search/pagination
 //= require alaveteli_pro/batch_authority_search/result
+//= require alaveteli_pro/batch_authority_search/count
 
 //= require alaveteli_pro/request_navigation
--- a/app/assets/javascripts/alaveteli_pro/batch_authority_search/count.js
+++ b/app/assets/javascripts/alaveteli_pro/batch_authority_search/count.js
@@ -1,0 +1,44 @@
+// Handles updating the batch authority search authority count
+(function($, BatchAuthoritySearch, DraftBatchSummary) {
+  var DraftEvents = DraftBatchSummary.Events;
+
+  var $search,
+      $draft,
+      $count,
+      messageTemplateZero,
+      messageTemplateOne,
+      messageTemplateMany;
+
+  // Update the count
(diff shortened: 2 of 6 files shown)


garethrees (approved):
Looks great!

> It seemed logical to display along side the "Write Request" button

Yep, good choice.

> Attempted to handle plural terms other locales

Hmmmm interesting! In English I'd actually _always_ use the plural term (if reading aloud, I'd say "1 of 500 authorit_ies_"), but I don't know if that holds for other languages.

Perhaps worth asking in `#general` or `alaveteli-users` to see if others know the answer to this? Or, AIUI @FOIMonkey knows a few! Always nice not to have to maintain more code than we need, but I don't object to it as is as to solve the problem – I think zero, one, many is a fine compromise and covers the majority of cases.

FOIMonkey:
> Hmmmm interesting! In English I'd actually _always_ use the plural term (if reading aloud, I'd say "1 of 500 authorit_ies_"), but I don't know if that holds for other languages.

Not all languages have plural nouns, and some that do, don't always modify the noun if the adjectives/articles show that it would be in the plural form. See Hungarian: https://en.wikipedia.org/wiki/Hungarian_noun_phrase#Plurality

It's always best to give an option.

Issue #7138, opened 2022-07-04
None pro admins can work out private share link for embargoed requests


FOIMonkey:
The token for accessing embargoed requests is displayed in the timeline. All you need to do is add it to https://www.whatdotheyknow.com/r/ to access the request.

![image](https://user-images.githubusercontent.com/11962965/177171617-52db0205-6be4-4899-b881-efbae1cec103.png)

PR #7139, opened 2022-07-05
[#7138] Don't expose private event params

garethrees:
Don't expose params of events on embargoed requests to those who can't
admin the request.

The issue that prompted this was private link tokens being exposed via
the params to non-pro admins, which would enable them to then view the
request.

Fixes https://github.com/mysociety/alaveteli/issues/7138

**Pro admin:**

Screenshot 2022-07-05 at 18 03 54

**Standard admin:**

Screenshot 2022-07-05 at 18 05 28


gbp (approved):
LGTM

PR #7153, opened 2022-07-13
fix: cp should take two arguments and does not write output to new file

jpatters:
## What does this do?
Fixes a bug with the installer

## Why was this needed?
There was a bug in the installer


gbp (approved):
Thanks for spotting this.

Issue #4808, opened 2018-08-14
Download of public bodies should include internal id


ajparsons:
On https://www.whatdotheyknow.com/body/list/all, there is the option to download a csv of all public bodies. This doesn't currently include the internal id, but should as name and slug are flexible. This will help anyone trying to build a lookup to other databases where names can change over time (e.g. matching our names to canonical names for local gov, etc).

HelenWDTK:
This issue has been automatically closed due to a lack of discussion or resolution for over 12 months.
Should we decide to revisit this issue in the future, it can be reopened.

PR #7171, opened 2022-07-18
[#4808] Add the internal ID number to the CSV download

ajparsons:
## Related Issue

https://github.com/mysociety/alaveteli/issues/4808

## What does this do?

This adds a new column to the CSV download that includes the internal database ID number for the authority. 

## Why was this needed?

The authorities csv has no columns that are guaranteed not to change if an authority is renamed. 

I'm also using the id number to pass information to the survey platform, and I need to be able to get from that back the ID name and categories . I *could* reconcile this using information from the database, but this CSV is otherwise 99% of what I need, and it's already in public. 

## Notes to reviewer

My ruby is rusty, I *think* this should work, but advice is good.

### DIFF
--- a/doc/CHANGES.md
+++ b/doc/CHANGES.md
@@ -1,3 +1,10 @@
+# develop
+
+## Highlighted Features
+
+* Add internal ID number to authority CSV download (Alex Pearson, Graeme
+  Porteous)
+
 # 0.44.0.0
 
 ## Highlighted Features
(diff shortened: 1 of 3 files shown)


ajparsons:
Hmm, the remaining failing test seems to have a different ID every time...

garethrees:
Need to check how this interacts with our Download -> Modify -> Re-upload workflow. I can't remember if the importer just ignores columns that it doesn't recognise, but it does (for reasons unknown to me before my time) expect `#name` to be the first column header.

ajparsons:
Not super important to me ID is at the start.

ajparsons:
@gbp : does giving the other authority generation ticket give any insight on my test issue above?

gbp (approved):

PR #5462, opened 2019-11-12
YAML compatibility class

gbp:
## Relevant issue(s)

Fixes #5263 
Fixes #5265 
Fixes #5464 
Fixes #5463
Fixes #5465

## What does this do?

This helps load old InfoRequestEvent instances which contain YAML from
older Rails version or includes old classes which don't exist in the app
any more.

## Why was this needed?

To retain compatibility with old Alaveteli versions.

## Implementation notes

This uses the a similar method to Ruby's `YAML.safe_load` method [1] but we are own `ClassLoader` which can map to different classes instead of restricting which classes can be loaded [2].

[1] https://github.com/ruby/psych/blob/0910ae5575786d57783eafd4d03ebc0d077cd2ed/lib/psych.rb#L352-L360
[2] https://github.com/ruby/psych/blob/0910ae5575786d57783eafd4d03ebc0d077cd2ed/lib/psych/class_loader.rb#L93-L99

PR #7173, opened 2022-07-19
[#5485] Add `InfoRequestEvent#param` JSONB column

### LINKED ISSUE #5485: Sanitise all YAML stored in the database

gbp:
## Relevant issue(s)

Fixes #5485
See #6949 

## What does this do?

1. Adds a `InfoRequestEvent#param` JSONB column - this will replace the `#params_yaml` column.
2. Updates event param assignment to better handle model object instances and model object IDs so both get stored as an global ID, EG: `user: { gid: 'gid://alaveteli/User/1' }`
3. Adds rake task to sanitise and populate the new JSONB column from the old YAML column.

## Why was this needed?

YAML column could contain full ruby objects which we shouldn't be storing - [CVE-2022-32224](https://github.com/advisories/GHSA-3hhc-qp5v-9p2j). The latest Ruby patch release has fixed this by limiting the permitted classes which could be decoded. We don't need to store any objects really, in fact we shouldn't as this has the potential to cause issues for GDPR requests as user information could be stored.

Also removed the need for https://github.com/mysociety/alaveteli/pull/5462 which was only meant to be temporary.

### DIFF
--- a/app/models/info_request_event.rb
+++ b/app/models/info_request_event.rb
@@ -1,5 +1,5 @@
 # == Schema Information
-# Schema version: 20210114161442
+# Schema version: 20220408125559
 #
 # Table name: info_request_events
 #
@@ -15,6 +15,7 @@
 #  outgoing_message_id :integer
 #  comment_id          :integer
 #  updated_at          :datetime
+#  params              :jsonb
 #
 
 # models/info_request_event.rb:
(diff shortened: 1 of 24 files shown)


gbp:
@garethrees are you able to find time to review this. I want to get this merged before doing the latest Rails bump release as that has changes to the YAML loading - we would need to enable `use_yaml_unsafe_load` in order to upgrade without this.

gbp (commented):
Reviewed with @garethrees made a few changes: 1. to keep updates to the `params_yaml` column for now. 2. change the model storage to use global IDs.

This is ready to be merged/deployed.

garethrees:
Thanks for all the work on this and sorry it took so long to get over the line! 🚚

PR #7191, opened 2022-07-26
[#7190] Over one million requests!

### LINKED ISSUE #7190: Request count on Alaveteli.org is out-of-date
@FOIMonkey has very helpfully noted that there are now well over **one million** transparency requests on Alaveteli powered sites around the world. This is worth noting.

mdeuk:
## Relevant issue(s)
Fixes #7190 

## What does this do?
This changes the request counter on alaveteli.org from "800,000+" to "1,000,000+".

## Why was this needed?
@FOIMonkey noted the milestone - this is worth celebrating!

## Implementation notes
Nothing fancy - a rainbow would have been nice, mind you! 🌈
## Screenshots
Screenshot from alaveteli.org/deployments. The mySociety logo is in the top right, followed directly by a purple bar with the Alavateli logo, language picker, and navigation options. Directly below, in a slightly different shade of purple, is a map of the world,with text overloid in white: '20+ languages, 25+ jurisdictions 1,000,000+ requests for information, Alaveteli can help open up government in any country, in any language, and within any legislation'

## Notes to reviewer
N/A

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -35,5 +35,5 @@ exclude:
 plugins:
   - jekyll-redirect-from
 jurisdictions: "25+"
-requests: "800,000+"
+requests: "1,000,000+"
 languages: "20+"


mdeuk:
I should add, the style guide makes the following comment on numbers:

> Digits or letters?
Numbers between one and ten should be spelled out. Numbers above ten are generally represented in digits.
However, large, round numbers may be represented in words or figures: it’s usually pretty obvious which style will read better in the context of your piece. 

My preference would have been to spell it out (e.g. One million+), but I _**haven't**_ done so, as this would need to be translated, and owing to the way the numbering is currently setup, I don't think this can be done without some changes elsewhere. Keeping the existing format and just iterating the number seems to make sense.

garethrees (approved):
Oof, amazing! Thanks for spotting – this is useful timing. Agree on just using numbers.

mdeuk:
> Oof, amazing! Thanks for spotting – this is useful timing. Agree on just using numbers.

All credit to @FOIMonkey, I just did the patch.

it really is an impressive number though!

PR #7200, opened 2022-08-02
[#7184] Fix unknown notification mailer method

### LINKED ISSUE #7184: [ERROR] undefined method `expire_embargo_notification' for #<NotificationMailer>
We're seeing errors for:

```
undefined method `expire_embargo_notification' for #<NotificationMailer:0x00000000015108>
Did you mean?  embargo_expired_notification
```

This results in the `send-notifications` daemon crashing.

Pro users who unsubscribe used to keep the daily notifications digest, but due to recent changes for the batch category picker work we did some work on the feature flags which resulted in Pro all features being revoked, including the daily notifications. The `expire_embargo` events don't work as instant notification and returns the error above.

This error only seems to occur when embargos expire for requests belonging to former-Pro users, who've unsubscribed since we deployed #7034.

We need to:
1. rename either the event_type from `expire_embargo` to `embargo_expired` (this more be more consistent with the other events) 
2. rename the the `NotificationMailer#embargo_expired_notification` to `NotificationMailer#expire_embargo_notification` (this is quicker/easier to do)

gbp:
## Relevant issue(s)

Fixes #7184

## What does this do?

Rename `expire_embargo_notification` to `embargo_expired_notification`
as this method was misnamed.

## Why was this needed?

Exceptions occurred when an embargo expires and the requester uses
instant notifications. When on daily notification no exceptions would
occur.

## Implementation notes

Opted to not rename `factories` and `lets` variables as these read better as `embargo_expired_...`.


garethrees (approved):
Haven't run but seems sensible 👍

PR #7212, opened 2022-08-10
[#4478] New notes model and admin controller

### LINKED ISSUE #4478: Extract PublicBody#notes to model
At the moment we store notes as translated attributes of `PublicBody`.

We're needing to do more complex things with notes, particularly in the Pro batch interface (https://github.com/mysociety/alaveteli-professional/issues/356).

The main things we'd like to be able to do are:

* Distinguish between an "FYI" note vs a "Must Read"
* Be able to keep track of when a note was last updated

## Decisions

* Should we create a polymorphic `Note` model? It could then be used for things like `PublicBodyCategory`, `PublicBodyHeading`, maybe even `InfoRequest`…
* What levels should we use? The [Bootstrap alerts](https://getbootstrap.com/docs/4.0/components/alerts/) seem like good general cases (info, warning, danger, success), but maybe we only need `info` and `warning`?

## Main tasks

* Create a polymorphic `Note` model with attrs:
  * content (i18n)
  * severity (default to lowest severity)
  * notable_id
  * noteable_type
  * created_at
  * updated_at

* Add `has_many :notes` to `PublicBody`.

* Add admin interface for creating and editing notes for a `PublicBody`.

* Consider adding interface and nested attribute handling for creating notes in the `AdminPublicBodyCOntroller#edit` form.

* Display notes in their own table in `AdminPublicBodyController#show`, rather than in the main attributes table.

* Display all notes with relevant styling in `PublicBodyController#show`. (We could also add "last updated" to the notes now, which would be useful).

## Consequences

* CSV importer will need reworking – maybe the default is to support only one default severity note column (i18n options as usual). Additional notes can get added in the interface later.

* Notes are indexed by Xapian – we'll need to tweak that so that it finds all notes and joins the content together in to one string.

## Migration

Write a rake task to:

* Find authorities with notes
* Create a `Note` for each translation, defaulted to "info"

### LINKED ISSUE #6994: Allow tags to have notes
A tag should be "notable" (#6931) so that we can display additional context in cases where we're trying to make consistent use of tags (https://github.com/mysociety/whatdotheyknow-theme/issues/1158, https://github.com/mysociety/whatdotheyknow-theme/issues/1143, etc)

As a concrete example, I'd expect to see the tag `poisoned` with a counterpart note saying we use it when the "Body isn’t subject to FOI due to minority non-public sector owner(s)/member(s)".

gbp:
## Relevant issue(s)

Fixes #4478
Fixes #6994

## What does this do?

Adds new notes model and admin controller. Associated as a "concrete note" with `PublicBody` via a new `Notable` concern. Also can be assoicated as a "tagged note" via a `notable_tag` column.

## Why was this needed?

Easier handling of notes, especially via tags, so notes don't need to be manually added to multiple authorities.

## Implementation notes

Have extract out these new issues from #4478
- [ ] #7216 
- [ ] #7217 
- [ ] #7218 

These will be done as separate PRs.

### DIFF
--- a/app/helpers/admin/tag_helper.rb
+++ b/app/helpers/admin/tag_helper.rb
@@ -9,6 +9,10 @@ module Admin::TagHelper
 
   def render_tag(record_tag)
     tag.span class: 'label label-info tag' do
+      if record_tag.is_a?(String)
+        record_tag = HasTagString::HasTagStringTag.from_string(record_tag)
+      end
+
       render_tag_href(record_tag)
     end
   end
--- a/app/models/concerns/notable.rb
+++ b/app/models/concerns/notable.rb
@@ -1,0 +1,15 @@
+module Notable
+  extend ActiveSupport::Concern
+
+  included do
+    has_many :concrete_notes,
+             class_name: 'Note',
+             as: :notable,
+             inverse_of: :notable,
+             dependent: :destroy
+  end
+
+  def all_notes
+    concrete_notes.with_translations
+  end
+end
--- a/app/models/concerns/notable.rb
+++ b/app/models/concerns/notable.rb
@@ -10,6 +10,16 @@ module Notable
   end
 
   def all_notes
-    concrete_notes.with_translations
+    concrete_notes.with_translations + tagged_notes.with_translations
+  end
+
+  def tagged_notes
+    Note.where(notable_tag: notable_tags)
+  end
+
+  private
+
+  def notable_tags
+    tags.map(&:name_and_value)
   end
 end
--- a/app/views/admin/tags/show.html.erb
+++ b/app/views/admin/tags/show.html.erb
@@ -1,5 +1,7 @@
 <h1><%= @title = "Tag – #{@tag}" %></h1>
 
+<%= render partial: 'notes', locals: { notes: @notes } %>
+
 <div class="row">
   <div class="span12">
     <h2>Taggings</h2>
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -485,6 +485,16 @@ Rails.application.routes.draw do
   end
   ####
 
+  #### AdminNote controller
+  namespace :admin do
+    resources :notes, except: [:index, :show]
+  end
+
+  direct :admin_note_parent do |note|
+    admin_general_index_path
+  end
+  ####
+
   #### AdminPublicBody controller
   scope '/admin', :as => 'admin' do
     resources :bodies,
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -491,7 +491,11 @@ Rails.application.routes.draw do
   end
 
   direct :admin_note_parent do |note|
-    admin_general_index_path
+    if note.notable
+      url_for([:admin, note.notable])
+    else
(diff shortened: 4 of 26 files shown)


app/helpers/admin/tag_helper.rb
```diff
@@ -9,6 +9,15 @@ def render_tags(tags)
 
   def render_tag(record_tag)
     tag.span class: 'label label-info tag' do
+      if record_tag.is_a?(String)
+        name, value = HasTagString::HasTagStringTag.split_tag_into_name_value(
```
garethrees, line 13: Made https://github.com/mysociety/alaveteli/pull/7215 to simplify this.

app/models/concerns/notable.rb
```diff
@@ -0,0 +1,15 @@
+module Notable
+  extend ActiveSupport::Concern
+
+  included do
+    has_many :concrete_notes,
```
garethrees, line 5: Is this intended to be temporary (and will be called "notes" once we get rid of the existing column?). I'm wondering whether "`directly_attached_notes`" would be clearer, but if this is only a temporary, internal thing then doesn't matter much.

gbp: Yeah we should rename to `notes`. Thinking about it a bit more maybe we should rename the current column to `notes_old` or something and rename this now as this would save waiting for a release before renaming.

garethrees: Would that prevent the existing notes from displaying? We'd have to update all calls (e.g in views to `public_body.notes_old`) and then change it back?

gbp: Lets ignore this for now and get this PR reviewed and merged. We can deal with this in https://github.com/mysociety/alaveteli/issues/7216

config/routes.rb
```diff
@@ -485,6 +485,22 @@
   end
   ####
 
+  #### AdminNote controller
```
garethrees, line 488: be nicer to have shallow routes for new so that we can do e.g. `/admin/bodies/3/notes/new` (which automatically sets `notable_type` & `notable_id`), but then editing the created note is simply `/admin/note/:id/edit`. not essential this minute though

app/views/admin/tags/show.html.erb
```diff
@@ -1,5 +1,7 @@
 <h1><%= @title = "Tag – #{@tag}" %></h1>
 
+<%= render partial: 'notes', locals: { notes: @notes, title: 'Notes' } %>
```
garethrees, line 3: doesn't work – not sure how to fix. might need to be `'admin/notes/show'` but then the locals don't quite work

Screenshot 2022-08-24 at 11 12 44

garethrees (commented):
looks good in principle; admin ui seems to break so can't test fully

app/views/admin/tags/show.html.erb
gbp: Ugh. I think this got excluded/missed due to my global git ignore file containing `tags` - I'll recreate.

garethrees (approved):

Issue #3500, opened 2016-09-24
Add ability for administrator to delete an unconfirmed request


RichardTaylor:
Currently a request drafted is held until a users' email address is confirmed. 

On a couple of occasions recently inappropriate draft requests (eg. requests for personal information, or general correspondence) have come to administrators' attention and ideally it would have been possible to delete them to prevent them being sent.

garethrees:
This would need something like a moderation queue (#75)

garethrees:
Or a delete button for post redirects.

garethrees:
Just noting that in the last WDTK catchup we mentioned that given its infrequent volunteers can email developers to do this. If it becomes more frequent we can consider some more UI for it.

garethrees:
Can probably find out how many of these there are with something like

```ruby
User.
  where(email_confirmed: false).
  map { |user| [user.id, user.post_redirects.where(uri: '/new').count] }
```

Currently have lots of unconfirmed users, so doing this in raw SQL would be much faster:

```ruby
User.where(email_confirmed: false).count
# => 12751
```

RichardTaylor:
Today a WhatDoTheyKnow user specifically requested we delete a draft request on the grounds they'd accidentally included their real name in it, when they were intending to make a request under a pseudonym.

RichardTaylor:
+1 There was an issue yesterday with a user whose draft request would have breached the house rules and potentially been problematic. 

There is a risk in these cases that an admin follows the confirmation link and a problematic message gets sent - by an admin.

RichardTaylor:
It has been suggested that this issue might be related to 

https://github.com/mysociety/alaveteli/issues/7049

garethrees:
> It has been suggested that this issue might be related to deleting drafts (#7049)

These are different. This ticket is essentially about being able to delete a `PostRedirect`. #7049 is about deleting `AlaveteliPro::DraftInfoRequest` and `AlaveteliPro::DraftInfoRequestBatch` records.

FOIMonkey:
I have just done some testing around this. I created a new request to the test quango and registered a new account. I logged in to the system and banned the account that I had just created. I then went into the email account linked to the test account and clicked on the confirmation link. The request was sent to the test Quango despite the fact that the account had been banned. Had an admin clicked on the link it is likely the same thing would happen.

tl;dr banning a user in advance is not a solution.

PR #7233, opened 2022-08-19
[#3500] Allow admins to destroy post redirects

garethrees:
We often spot user accounts with unconfirmed requests that don't meet
our house rules. The only way to prevent a user going on to confirm the
request is by destroying its associated `PostRedirect`, which until this
commit could not be performed through the admin UI.

This action currently only handles post redirects with an associated
`User`. We do create post redirects _without_ associated an associated
`User` record, but those are not displayed in the admin UI anywhere, so
I don't think that's an immediate issue.

Fixes https://github.com/mysociety/alaveteli/issues/3500

Screenshot 2022-08-19 at 09 23 49

Screenshot 2022-08-19 at 09 23 54

Screenshot 2022-08-19 at 09 24 02

### DIFF
--- a/app/controllers/admin/post_redirects_controller.rb
+++ b/app/controllers/admin/post_redirects_controller.rb
@@ -1,0 +1,17 @@
+# Controller for managing PostRedirects
+class Admin::PostRedirectsController < AdminController
+  before_action :set_post_redirect, only: %i[destroy]
+
+  # DELETE /admin/post_redirects
+  def destroy
+    @post_redirect.destroy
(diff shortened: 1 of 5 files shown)


gbp (commented):
Might want to consider admin/user roles. With the current implementation non-pro admins can delete pro users' post redirects.

Happy to merge as is if this isn't a concern.

garethrees:
> With the current implementation non-pro admins can delete pro users' post redirects.

That's fine; non-pro admins can already admin pro users. The segregation is intended to be around the private requests, not the account in general.

PR #7241, opened 2022-08-24
Rationalise message admin links

garethrees:
Add incoming & outgoing message to both_links & Rationalise links in incoming & outgoing edit view 

When administering messages it’s quite annoying to have different ways –
or no way at all – of getting back to the public content or the admin
page for the request.

This commit rationalises the two admin views so that there’s the ability
and consistency of being able to navigate to the right content.

The one thing that’s been dropped is the link to “List all”, but that’s
easily available from the main admin nav bar.

The use of both_links also adds richer information in the form of
showing the current prominence for both request and message.

Screenshot 2022-08-24 at 10 15 04

Screenshot 2022-08-24 at 10 16 44

Screenshot 2022-08-24 at 10 17 04

### DIFF
--- a/app/views/admin_incoming_message/_intro.html.erb
+++ b/app/views/admin_incoming_message/_intro.html.erb
@@ -1,3 +1,7 @@
 <% @title = "Incoming message #{incoming_message.id} of FOI request '#{incoming_message.info_request.title}'" %>
 <h1>Incoming message <%= incoming_message.id %></h1>
-<p>FOI request: <%= both_links(incoming_message.info_request) %></p>
+
+<ul class="breadcrumb">
+  <li>FOI request: <%= both_links(incoming_message.info_request) %> <span class="divider">/</span></li>
+  <li class="active">Message: <%= both_links(incoming_message) %></li>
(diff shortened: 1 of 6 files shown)


app/views/admin_incoming_message/_intro.html.erb
```diff
@@ -1,3 +1,7 @@
 <% @title = "Incoming message #{incoming_message.id} of FOI request '#{incoming_message.info_request.title}'" %>
 <h1>Incoming message <%= incoming_message.id %></h1>
-<p>FOI request: <%= both_links(incoming_message.info_request) %></p>
+
+<p>
+  FOI request: <%= both_links(incoming_message.info_request) %> |
+  Message: <%= both_links(incoming_message) %>
```
gbp, line 6: Feels like we could use a bootstrap breadcrumb nav here

gbp (approved):

garethrees:
> Feels like we could use a bootstrap breadcrumb nav here

Yeah that worked nice!

Screenshot 2022-09-01 at 11 56 51

Issue #387, opened 2012-01-25
Show note based on a public body's tags


robinhouston:
**[Imported from https://github.com/mysociety/whatdotheyknow/issues/13, reported by TomSteinberg on January 12, 2011]**

The original issue had no text & the following comments:

**RichardTaylor commented May 30, 2011**
We already can add public notes to the authorities.

What would be useful is the ability to add text to all bodies with a particular tag. For example then we could keep the standard bit of text we add to Housing Associations in one place rather than replicating it manually over all the Housing Associations. We could also write a note about say the kind of information a police authority would be expected to hold, and have that automatically appear on every police authority's page.

**confirmordeny commented June 10, 2011**
By way of an update we have started adding descriptions for some of the bodies on the site: http://www.whatdotheyknow.com/body/list/hasdescription

The Twiki has guidance on writing good concise descriptions so we should get some consistency and over time the quality of descriptions should improve.

**RichardTaylor commented November 25, 2011**
I have posted a related Alaveteli issue at:

https://github.com/sebbacon/alaveteli/issues/296

**confirmordeny commented November 25, 2011**
Richard
Good issue. I think I could slightly more myself when adding new
bodies even within the constraints of the current set up.  You are
right though to put this forward.

regards,

John

RichardTaylor:
This was intended as a development issue. 

Take for example:

https://www.whatdotheyknow.com/body/list/housing_association

Currently we have manually put similar public notes in for each housing association (varying if Scottish or not). If we could add a note which appears whenever a body is tagged with a certain tag we could enter such notes once. 

In this case a link we use has disappeared from the web, so we ought update all these notes. The tag specific note requested here would mean this update could be done via one (or two given Scotland) edits rather than over a hundred. 

I can envisage we might similarly want to add a note to all schools, or Police and Crime Commissioners, in this manner if Alaveteli had this feature.

kingqueen3065:
this is still very much desired

RichardTaylor:
I just added a note to all 11 Northern Ireland local councils; they're all tagged 

https://www.whatdotheyknow.com/body?tag=ni_local_council

it would have been good to be able to add the note to all of them at once. 

I'll retitle this issue to reflect the proposed new feature.

While adding the note I wondered how notes are shown to those using the batch request feature; drawing such requestors' attention to, for example in this case that NI local councils have different responsibilities than local councils generally do in the UK might be useful and responsible.

RichardTaylor:
[...]

PR #7242, opened 2022-08-25
Refactor notes rendering

### LINKED ISSUE #7218: Update `PublicBody` Xapian indexing for notes
Existing `PublicBody#notes` column is indexed in Xapian. With notes moving to their own model (#7216) we should index notes from models. Easiest way would be to concat the notes' body together before indexing.

gbp:
## Issues

Fixes #7218

## What does this do?

Refactor notes rendering

## Why was this needed?

General improvements to help with implementing #387 & #4478

## Implementation notes

Should be independent to other PRs and doesn't actually change how notes are rendered.


garethrees:
Nice! 🧹

PR #7243, opened 2022-08-26
[#387] Notes by tag

### LINKED ISSUE #387: Show note based on a public body's tags

gbp:
## Relevant issue(s)

Depends on #7212 
Depends on #7242 
Fixes #387

## What does this do?

Show public body notes by tag

## Implementation notes

This updates the markup but doesn't really change how the notes look yet - design work will be done in a separate branch.

## Notes to reviewer

This only has two commits, after dependent branches are merged:
- 0a2fc5a32c1a89366fd8643bde715a4c0d184177
- 92223f33e2d0363a0b17e32483408ca19e7875ee

### DIFF
--- a/app/models/public_body.rb
+++ b/app/models/public_body.rb
@@ -668,12 +668,28 @@ class PublicBody < ApplicationRecord
     $1.nil? ? nil : $1.downcase
   end
 
-  def has_notes?
-    notes.present?
+  def notes
+    [legacy_note].compact + all_notes
   end
 
   def notes_as_string
-    notes.to_s
+    notes.map(&:body).join(' ')
+  end
+
+  def legacy_note
+    return unless read_attribute(:notes).present?
+
+    Note.new(notable: self) do |note|
+      AlaveteliLocalization.available_locales.each do |locale|
+        AlaveteliLocalization.with_locale(locale) do
+          note.body = read_attribute(:notes)
+        end
+      end
+    end
+  end
+
+  def has_notes?
+    notes.present?
   end
 
   def json_for_api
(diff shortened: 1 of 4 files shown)


app/models/public_body.rb
```diff
@@ -668,12 +668,28 @@ def self.extract_domain_from_email(email)
     $1.nil? ? nil : $1.downcase
   end
 
-  def has_notes?
-    notes.present?
+  def notes
```
garethrees, line 671: This borks the table view. Might not be an issue depending on other PRs

Screenshot 2022-08-30 at 11 42 12

garethrees (commented):
Final markup & rendering looks good in principle.

Probably want @lucascumsille's review on the specific markup elements and classes.

app/models/public_body.rb
gbp: This is addressed in #7245

gbp:
@lucascumsille are you happy with the markup used here - this was taken from the concept document as discussed last week.

lucascumsille (approved):

Issue #3477, opened 2016-09-08
Add a way for users to find missing authority email addresses


henare:
In the admin backend you can see authorities where we're missing an email but this isn't exposed in the frontend. I can see volunteers potentially helping find email addresses if there was something like the request categorisation game but for missing email addresses.

garethrees:
Once done, create ticket for adding it to https://www.whatdotheyknow.com/help/volunteers

RichardTaylor:
If implementing this authorities tagged defunct should be excluded. 

Ideally users would only get points in the game if admins accepted their suggestions. 

We'd want to encourage good quality suggestions with links to sources or other justifications. 

This isn't currently a big issue for WhatDoTheyKnow but the presence of this feature could change the way we operate - eg. we decided not to add large numbers of bodies we don't have email addresses for, but with this feature we could. 

WhatDoTheyKnow started with crowdsourced addresses on a shared GoogleSheets document, with this feature new sites could be started "on-site" rather than having to direct helpers to a spreadsheet.

garethrees:
Linking to admin version of this issue https://github.com/mysociety/alaveteli/issues/6424.

RichardTaylor:
>something like the request categorisation game but for missing email addresses.

The request classification game only has an overall target. 

It would be good to have more achievable per body / per category targets / reports for both an email finding "game" and the request classification game.

garethrees:
A way of achieving this would be to use a tag and link to a tag-based search `/search/tag:missing_email/bodies`.

Would have to manually tag the bodies then remember to remove the tag once an address is accepted.

Would be quite easy to mass-apply the tag for current bodies missing an email:

```ruby
PublicBody.without_request_email.each { |b| b.add_tag_if_not_already_present('missing_email') }
```

Could hack something in the theme to automatically keep the tag up to date:

```ruby
class PublicBody
  after_save :update_missing_email_tag

  def update_missing_email_tag
    if missing_email?
      add_tag_if_not_already_present('missing_email')
    else
      remove_tag('missing_email')
    end
  end

  def missing_email?
    !has_request_email?
  end
```

RichardTaylor:
Can we "follow success" and perhaps eg. improve 

https://www.whatdotheyknow.com/body/list/missing_email

* Can we mix up the results so different ones come to the top? 
* Can we gamify and do a leaderboard?  (Perhaps combining points with classification 'points')?
* Blog about the new feature 
[...]

PR #7250, opened 2022-08-30
[#3477] Automatically tag bodies missing a request address

garethrees:
Automatically apply and remove the `missing_email` tag to bodies that
have a missing email so that users can find them via a public search and
help to source the addresses.

The changelog suggests a one-shot script to mass-apply the tag to
existing bodies that are missing an email.

Fixes https://github.com/mysociety/alaveteli/issues/3477.

### DIFF
--- a/doc/CHANGES.md
+++ b/doc/CHANGES.md
@@ -2,6 +2,8 @@
 
 ## Highlighted Features
 
+* Automatically apply `missing_email` tag to bodies who are missing a request
+  email so that they can be found in a public list (Gareth Rees)
 * Improve linking from outgoing & incoming message admin pages (Gareth Rees)
 * Allow admins to destroy user post redirects (Gareth Rees)
 * Use correct mime type for cached CSV attachments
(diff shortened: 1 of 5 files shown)


doc/CHANGES.md
```diff
@@ -2,12 +2,20 @@
...
 
 ## Highlighted Pro Features
 
 ## Upgrade Notes
 
+* _Optional:_ Bodies missing a request email will automatically get tagged
+  `missing_email` as they are updated. If you want to automatically tag them all
+  in one go, run the following from the app root directory:
+
+    bin/rails runner "PublicBody.without_request_email.each(&:save)"
```
gbp, line 17: Render as a code block with:

```suggestion
      bin/rails runner "PublicBody.without_request_email.each(&:save)"
```

gbp (approved):

doc/CHANGES.md
garethrees: I think it's correct as-is – 4 spaces from the start of the line, not relative to the lines above, which are already indented 2 because of the bullet point?

Screenshot 2022-09-01 at 16 10 43

garethrees: Oh looks like GitHub prefers your suggestion. Will fix on merge.

PR #7257, opened 2022-08-30
[#7074] Taggable incoming/outgoing messages

### LINKED ISSUE #7074: Make messages taggable

gbp:
## Relevant issue(s)

Supersedes https://github.com/mysociety/alaveteli/pull/7075
Fixes #7074

## What does this do?

1. Refactoring of existing spec model concerns
2. Extraction of `Taggable` concern specs
3. Allow individual message to be tagged within the admin UI.

### DIFF
--- a/app/controllers/admin_incoming_message_controller.rb
+++ b/app/controllers/admin_incoming_message_controller.rb
@@ -9,6 +9,7 @@ class AdminIncomingMessageController < AdminController
   def update
     old_prominence = @incoming_message.prominence
     old_prominence_reason = @incoming_message.prominence_reason
+    old_tag_string = @incoming_message.tag_string
     if @incoming_message.update(incoming_message_params)
       @incoming_message.info_request.log_event(
         'edit_incoming',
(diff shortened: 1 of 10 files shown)


garethrees (commented):
Love the spec cleanup!

I think my comments from https://github.com/mysociety/alaveteli/pull/7075#pullrequestreview-1010397365 still stand.

I _think_ we can fix the issue of showing `tag_string` in the unfolded attributes with https://github.com/mysociety/alaveteli/commit/0b35c96a4534b706d5baf70e216fff8e8cb41a11, though this is not super critical.

Any ideas what we'd need to do to resolve the issue of an empty edit event when changing the tags?

Also, I think this is only part of #7074. Adding `Taggable` to `OutgoingMessage` would also be needed to close it.

PR #7273, opened 2022-09-02
Admin raw email view

garethrees:
## What does this do?

Improve the admin raw email & holding pen views

## Why was this needed?

Just some small quality of life and code cleanup improvements.

## Screenshots

Improved raw email page header and breadcrumbs:

Screenshot 2022-09-02 at 17 39 23

Improved holding pen details:

Screenshot 2022-09-02 at 17 44 15

## Notes to reviewer

Loads more cleanup to be done, but at least this now extracts the holding pen partial.

### DIFF
--- a/app/assets/stylesheets/admin.scss
+++ b/app/assets/stylesheets/admin.scss
@@ -269,6 +269,14 @@ body.admin {
   }
 }
 
+/* RawEmails */
+
+.holding_pen {
+  border: 12px solid #c09853;
+  border-radius: 5px;
+  padding: 2em;
+}
+
 /* Bootstrap Extensions */
 /* These must come last because we @import bootstrap in .admin  */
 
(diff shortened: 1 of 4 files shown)


garethrees:
Could also add a `both_links` for `RawEmail`. Dunno what to use for the admin text. Could go for the request name again:

Screenshot 2022-09-02 at 17 50 31


…or the raw email ID:

Screenshot 2022-09-02 at 17 49 20

If we do this then we should add `RawEmail#prominence` for the icon; should always return `"normal"` since we only change prominence at the incoming message level.

mdeuk:
> …or the raw email ID:

I think that is probably the **better** one to use, otherwise you've got the same phrase to read three times, so although it is _**obvious**_ which one you'd want, the mind isn't always as clear.

garethrees:
Yeah, I had the same thought.

garethrees:
> Could also add a `both_links` for `RawEmail`

Next time!

PR #7316, opened 2022-09-22
Mark share private link header for translation

laurentS:
## Relevant issue(s)
None found

## What does this do?
Marks a string for translation

## Why was this needed?
Foreign languages are great :)

### DIFF
--- a/app/views/request/_share_by_link.html.erb
+++ b/app/views/request/_share_by_link.html.erb
@@ -1,6 +1,6 @@
 <% if can? :share, info_request %>
   <div class="sidebar__section share-private-url">
-    <h2>Share with private link</h2>
+    <h2>_('Share with private link')</h2>
 
     <% if info_request.public_token? %>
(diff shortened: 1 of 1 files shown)


garethrees (approved):
Oops, thanks!

app/views/request/_share_by_link.html.erb
```diff
@@ -1,6 +1,6 @@
 <% if can? :share, info_request %>
   <div class="sidebar__section share-private-url">
-    <h2>Share with private link</h2>
+    <h2>_('Share with private link')</h2>
```
gbp, line 3: Will need to be:

```suggestion
    <h2><%= _('Share with private link') %></h2>
```

garethrees: oops, shouldn't code review while on a call 🤦

gbp:
Merged 201f38f05f11233f3b10dba6b270075c7e9afa18

PR #7324, opened 2022-09-26
[#1005] Preparation work before adding indivdual attachment prominence

gbp:
## Relevant issue(s)

Connected to #1005

## What does this do?

1. Made changes to `app/views/request/_bubble.html.erb` partial, now assumed it will be passed an incoming message (instead of also handling outgoing messages).
2. Refactors anchors to incoming/outgoing messages
3. Adds anchors for attachments so these can be linked to from the admin UI

## Why was this needed?

These changes will makes the changes required for #1005 smaller and simpler to review.


app/helpers/link_to_helper.rb
```diff
@@ -311,4 +308,15 @@ def message_path(message, options = {})
     message_url(message, options.merge(:only_path => true))
   end
 
+  def dom_id(record, prefix = nil)
+    case record
+    when IncomingMessage, OutgoingMessage
+      param_key = record_or_class.class.to_s.gsub('Message', '').downcase
+    else
+      return super
```
garethrees, line 320: On one request (a public request made by pro that's part of a batch) I get an error here:


```
ActionView::Template::Error (undefined local variable or method `record_or_class' for #<ActionView::Base:0x000000000414b0>):
    1: <div class="outgoing correspondence js-collapsable" id="<%= dom_id(outgoing_message) %>">
    2:   <%- if cannot?(:read, outgoing_message) %>
    3:     <%= render :partial => 'request/hidden_correspondence', :locals => { :message => outgoing_message } %>
    4:   <%- else %>

app/helpers/link_to_helper.rb:314:in `dom_id'
app/views/request/_outgoing_correspondence.html.erb:1
app/views/request/events/_sent.html.erb:1
app/views/request/_correspondence.html.erb:5
app/views/request/show.html.erb:42
app/views/request/show.html.erb:40
app/controllers/request_controller.rb:100:in `block (3 levels) in show'
app/controllers/request_controller.rb:99:in `block in show'
lib/alaveteli_localization.rb:46:in `with_locale'
app/controllers/request_controller.rb:50:in `show'
app/controllers/application_controller.rb:116:in `record_memory'
lib/deeply_nested_params.rb:15:in `call'
lib/strip_empty_sessions.rb:14:in `call'
```

82d9d2e20d61b1ac370e545fe82350c29c4b9a2b then works, and then 1f3734967592eede73fa98c7965336adff2e464e generates a different error:

```
ActionView::Template::Error (undefined method `to_key' for nil:NilClass
Did you mean?  to_query):
    36: <% end %>
    37:
    38: <%= tag.div incoming_message.get_body_for_html_display(@collapse_quotes),
    39:             id: dom_id(incoming_message.get_main_body_text_part) %>

app/helpers/link_to_helper.rb:320:in `dom_id'
app/views/request/_bubble.html.erb:39
app/views/request/_incoming_correspondence.html.erb:28
app/views/request/events/_response.html.erb:1
app/views/request/_correspondence.html.erb:5
app/views/request/show.html.erb:42
app/views/request/show.html.erb:40
app/controllers/request_controller.rb:100:in `block (3 levels) in show'
app/controllers/request_controller.rb:99:in `block in show'
lib/alaveteli_localization.rb:46:in `with_locale'
app/controllers/request_controller.rb:50:in `show'
app/controllers/application_controller.rb:116:in `record_memory'
lib/deeply_nested_params.rb:15:in `call'
lib/strip_empty_sessions.rb:14:in `call'
```

It is a request that I've been testing destroying the underlying raw email on so there's no content in one of the incoming messages, but it works on develop.

Screenshot 2022-09-27 at 11 04 30

app/helpers/link_to_helper.rb
```diff
@@ -310,13 +310,14 @@ def message_path(message, options = {})
 
   def dom_id(record, prefix = nil)
     case record
-    when IncomingMessage, OutgoingMessage
-      param_key = record_or_class.class.to_s.gsub('Message', '').downcase
+    when IncomingMessage
```
garethrees, line 313: just merge into the previous commit so that it's right first time :)

garethrees (changes requested):
Looks good in principle.

Couple of small suggestions, but one breakage that probably needs a fix.

app/helpers/link_to_helper.rb
gbp: Calling `dom_id(nil)` without the override will return the same error so once we come to implement destroying the underlying raw email then we will need to add a conditional here to handle. Should need to handle this case for this PR thought.

garethrees: In the case of 15d1d5864240dc95f855aa5d2793971ed05fdbb9 it looks like there's an error on the `OutgoingMessage`?

```
ActionView::Template::Error (undefined local variable or method `record_or_class' for #<ActionView::Base:0x000000000414b0>):
    1: <div class="outgoing correspondence js-collapsable" id="<%= dom_id(outgoing_message) %>">
```

Having looked a bit more, it looks like the following change is causing the breakage…

```diff
- <%= tag.div incoming_message.get_body_for_html_display(@collapse_quotes) %>
+ <%= tag.div incoming_message.get_body_for_html_display(@collapse_quotes),
+             id: dom_id(incoming_message.get_main_body_text_part) %>
```

…which looks like it's being used to wrap the main body text?

```html
<div class="correspondence_text">
  <div id="attachment-2">
    <p>Dear “Bob”…</p>
  </div>
</div>
```

Why do we need this?

app/views/request/_attachments.html.erb
```diff
@@ -0,0 +1,39 @@
...
+          </p>
+        <% end %>
+      <% end %>
+    </ul>
+
+    <a href="#" class="attachments__show-more" style="display: none;" data-show-all="<%= _('Show all attachments') %>" data-show-fewer="<%= _('Show fewer attachments') %>"></a>
+  </div>
+<% end %>
+
+<%= tag.div incoming_message.get_body_for_html_display(@collapse_quotes),
```
garethrees, line 38: Should this be rendered in `app/views/request/_incoming_correspondence.html.erb` below the `render partial: 'request/attachments'` call?

I guess this is technically an attachment… but conceptually… IDK?

app/helpers/link_to_helper.rb
gbp: Needed to link directly to the attachment, see: https://github.com/mysociety/alaveteli/pull/7327#discussion_r982225290

app/views/request/_attachments.html.erb
gbp: Well I could've left the partial named `bubble` then this wouldn't be a problem ;) I agree technically an attachment but its stored in the DB and presented differently. I'm leaning towards leaving where it is tho.

garethrees: > Well I could've left the partial named `bubble` then this wouldn't be a problem ;)

I know I know 😔 ! I actually thought of the "bubble" as the whole thing, not a subsection of the correspondence "box". Whatevs.

> I agree technically an attachment but its stored in the DB and presented differently. I'm leaning towards leaving where it is tho.

I don't think it matters at this point, so yeah let's just leave it.

I think it's just a bit semantically weird in `request/_incoming_correspondence.html.erb` to have:

```erb
    <div class="correspondence_text">
      <%= render partial: 'request/attachments',
                 locals: { incoming_message: incoming_message } %>
    </div>
```

Feels like it should be:

```erb
  <%= render partial: 'request/attachments', locals: { incoming_message: incoming_message } %>
  
  <div class="correspondence_text">
     <!-- reduced prominence conditionals omitted -->
      <%= incoming_message.get_body_for_html_display(@collapse_quotes) %>
  </div>
```

garethrees: We're currently rendering attachments within "correspondence_text" though, so yeah let's just leave it and unpick later.

PR #7325, opened 2022-09-26
Translate delivery statuses

laurentS:
## Relevant issue(s)

None

## What does this do?

This PR fixes a small annoyance for non-English sites whereby the status of outgoing email delivery is not translated.

Before :cry: 
image

After :grinning: :fr: :croissant: 
image

## Why was this needed?
As this is a publicly visible item made quite prominent by colours/icons, it's better if it translates.

## Implementation notes

The template was not actually calling `to_s!` as implemented in `app/models/mail_server_log/delivery_status/translated_constants.rb`. This should fix the problem. There is a possible loophole though, as `humanize` does not seem to correctly adjust accented characters in French, for instance `échoué` should become `Échoué`, but appears to be left untouched.


garethrees (approved):
Good catch! Sorry we missed that.

PR #7335, opened 2022-09-29
[#6746] Add InfoRequest prominence reason

### LINKED ISSUE #6746: Add prominence reason to entire request
When hiding specific messages you can manually add a free text prominence reason. 

It would be useful if this was also available when hiding whole requests, alongside the pre set prominence reason that we have now. So that you only need to add 'free text'  should you want to.

Sally

gbp:
## Relevant issue(s)

Fixes #6746

## What does this do?

Adds prominence reason to `InfoRequest`

## Why was this needed?

Easier to recall why the prominence has been set the way it has. Good for historical cases.

### Request admin - show
Screenshot 2022-10-10 at 14-37-01 FOI request – Cheap v1agra - WhatDoTheyKnow admin

### Request admin - edit form
Screenshot 2022-10-10 at 14-37-23 WhatDoTheyKnow admin

### Request admin - show with reason entered
Screenshot 2022-10-10 at 14-52-06 FOI request – Cheap v1agra - WhatDoTheyKnow admin

### Request admin - quick hide form
Screenshot 2022-10-10 at 14-41-18 FOI request – Cheap v1agra - WhatDoTheyKnow admin

### Request show - prominence notice (as admin)
Screenshot 2022-10-10 at 14-52-47 Cheap v1agra - a Freedom of Information request to Ministry of Sensible Walks

### Loading attachment in response to a request which is hidden/requester only
Screenshot 2022-10-10 at 14 54 18

### DIFF
--- a/app/controllers/admin_request_controller.rb
+++ b/app/controllers/admin_request_controller.rb
@@ -167,14 +167,22 @@ class AdminRequestController < AdminController
     ActiveRecord::Base.transaction do
       subject = params[:subject]
       explanation = params[:explanation]
+
+      old_prominence = @info_request.prominence
+      old_prominence_reason = @info_request.prominence_reason
       @info_request.prominence = "requester_only"
+      @info_request.prominence_reason = params[:prominence_reason]
 
       @info_request.log_event(
         'hide',
         editor: admin_current_user,
         reason: params[:reason],
         subject: subject,
-        explanation: explanation
+        explanation: explanation,
+        old_prominence: old_prominence,
+        prominence: @info_request.prominence,
+        old_prominence_reason: old_prominence_reason,
+        prominence_reason: @info_request.prominence_reason
       )
 
       @info_request.set_described_state(params[:reason])
--- a/app/controllers/admin_request_controller.rb
+++ b/app/controllers/admin_request_controller.rb
@@ -36,6 +36,7 @@ class AdminRequestController < AdminController
   def update
     old_title = @info_request.title
     old_prominence = @info_request.prominence
+    old_prominence_reason = @info_request.prominence_reason
     old_described_state = @info_request.described_state
     old_awaiting_description = @info_request.awaiting_description
     old_allow_new_responses_from = @info_request.allow_new_responses_from
@@ -52,6 +53,8 @@ class AdminRequestController < AdminController
         title: @info_request.title,
         old_prominence: old_prominence,
         prominence: @info_request.prominence,
+        old_prominence_reason: old_prominence_reason,
+        prominence_reason: @info_request.prominence_reason,
         old_described_state: old_described_state,
         described_state: params[:info_request][:described_state],
         old_awaiting_description: old_awaiting_description,
@@ -203,6 +206,7 @@ class AdminRequestController < AdminController
     if params[:info_request]
       params.require(:info_request).permit(:title,
                                            :prominence,
+                                           :prominence_reason,
                                            :described_state,
                                            :awaiting_description,
                                            :allow_new_responses_from,
--- a/app/views/admin_request/edit.html.erb
+++ b/app/views/admin_request/edit.html.erb
@@ -66,6 +66,11 @@
 
   <p><label for="info_request_prominence_reason"><strong>Reason for prominence</strong></label>
     <%= text_area 'info_request', 'prominence_reason', rows: 5, class: 'span6' %>
+
+    <span class="help-block">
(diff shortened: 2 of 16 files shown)


app/controllers/admin_request_controller.rb
```diff
@@ -193,6 +193,7 @@ def info_request_params
     if params[:info_request]
       params.require(:info_request).permit(:title,
                                            :prominence,
+                                           :prominence_reason,
```
garethrees, line 196: Need to also record this in the hide/edit event on change as per https://github.com/mysociety/alaveteli/issues/6746#issuecomment-1133038351:

> When `InfoRequest#prominence_reason` is updated, it should log the prominence reason on the associated `edit` `InfoRequestEvent`.

Screenshot 2022-09-30 at 12 04 25

Note that we'll also want to record this on `hide` events too.

https://github.com/mysociety/alaveteli/blob/0.41.1.0/app/controllers/admin_request_controller.rb#L48

https://github.com/mysociety/alaveteli/blob/0.41.1.0/app/controllers/admin_request_controller.rb#L160

app/views/admin_request/edit.html.erb
```diff
@@ -64,6 +64,10 @@
     </span>
   </p>
 
+  <p><label for="info_request_prominence_reason"><strong>Reason for prominence</strong></label>
+    <%= text_area 'info_request', 'prominence_reason', rows: 5, class: 'span6' %>
+  </p>
+
```
garethrees, line 70: Let's add some help text here:

```diff
diff --git a/app/views/admin_request/edit.html.erb b/app/views/admin_request/edit.html.erb
index 7bbd773bb..056ed1b26 100644
--- a/app/views/admin_request/edit.html.erb
+++ b/app/views/admin_request/edit.html.erb
@@ -66,6 +66,11 @@
 
   <p><label for="info_request_prominence_reason"><strong>Reason for prominence</strong></label>
     <%= text_area 'info_request', 'prominence_reason', rows: 5, class: 'span6' %>
+
+    <span class="help-block">
+      This reason is shown in public. If left blank a generic reason of
+      "<%= default_prominence_reason %>" will be shown.
+    </span>
   </p>
 
   <p><label for="info_request_described_state"><strong>Described state</strong></label>
```

Will need the following change, which I think ought to be fine. Don't think it matters if admins generally see this text?

```diff
diff --git a/app/helpers/prominence_helper.rb b/app/helpers/prominence_helper.rb
index f92bbf13d..479f16de0 100644
--- a/app/helpers/prominence_helper.rb
+++ b/app/helpers/prominence_helper.rb
@@ -17,6 +17,11 @@ def render_prominence(object, format: :html)
       join(' ').squish.html_safe
   end
 
+  def default_prominence_reason
+    _("There are various reasons why we might have done this, sorry we can't " \
+      "be more specific here.")
+  end
+
   private
 
   def prominence_notice(object)
@@ -100,12 +105,6 @@ def prominence_reason_for_object(object)
     object.prominence_reason.presence || default_prominence_reason
   end
 
-  def default_prominence_reason
-    return '' if current_user&.is_admin?
-    _("There are various reasons why we might have done this, sorry we can't " \
-      "be more specific here.")
-  end
-
   def prominence_contact_us
     return if current_user&.is_admin?
     _('Please {{contact_us_link}} if you have any questions.',
```

garethrees (changes requested):
Nice!

Still needs to record the event and one minor suggestion for the form.

Also a bit annoying that prominence and prominence reason are so disconnected in the request metadata:

Screenshot 2022-09-30 at 12 20 03

Not sure there's an easy way around this at present?

app/controllers/admin_request_controller.rb
gbp: Both hide and edit events record the prominence and reason changes now.

gbp:
@garethrees I have fixed up and split the commit into 4 separate ones. I think its good to go now.

garethrees (approved):
Looks great! Particularly like the quick hide form improvement.

Two tiny tweaks but no need to re-review them 🚀

app/views/admin_request/show.html.erb
```diff
@@ -51,15 +51,28 @@
               </tr>
             <% end %>
 
-            <% @info_request.for_admin_column do |name, value| %>
```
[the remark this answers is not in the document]
garethrees: Given this would add duplication, I couldn't resist having a go at this. Added better fix in https://github.com/mysociety/alaveteli/pull/7364.

PR #7363, opened 2022-10-14
[#7361] Automatically tag bodies with few public requests

### LINKED ISSUE #7361: Auto tag authorities that have received no/very few requests
Similar to the auto tagging of missing emails, this would enable us to suggest users make requests to these authorities and/or offer suggestions as to what they could ask for.

garethrees:
## Relevant issue(s)

Fixes #7361.

## What does this do?

Automatically apply and remove a tag to requestable bodies that have a low number of public requests.

## Why was this needed?

Allows users to find them via a public search to help bootstrap basic transparency/site content, and admins can add a tag-based note to suggest generally useful first requests.

## Implementation notes

The number considered "not many requests" is configurable in the theme by setting `PublicBody.not_many_public_requests_size`.

The changelog suggests a one-shot script to mass-apply the tag to existing bodies where applicable.

It's a bit annoying that I needed to add the `InfoRequest#notify_public_body` hook. Can't think of a less obtrusive way of doing this though.

## Screenshots

Added some admin guidance and separated out the "special" and "automatically managed" tags:

Screenshot 2022-10-14 at 09 19 16

## Notes to reviewer

Left the spec DRYing up as fixups just in case there's a better approach.

### DIFF
--- a/app/models/info_request.rb
+++ b/app/models/info_request.rb
@@ -188,6 +188,7 @@ class InfoRequest < ApplicationRecord
   before_create :set_use_notifications
   before_validation :compute_idhash
   before_validation :set_law_used, on: :create
+  after_create :notify_public_body
   after_save :update_counter_cache
   after_update :reindex_request_events, if: :reindexable_attribute_changed?
   before_destroy :expire
@@ -1934,5 +1935,9 @@ class InfoRequest < ApplicationRecord
     %i[url_title prominence user_id].any? do |attr|
       saved_change_to_attribute?(attr)
     end
+  end
+
+  def notify_public_body
+    public_body.request_created
   end
 end
--- a/app/models/public_body.rb
+++ b/app/models/public_body.rb
@@ -984,10 +984,14 @@ class PublicBody < ApplicationRecord
   end
 
   def update_not_many_requests_tag
-    if info_requests.is_searchable.size < not_many_public_requests_size
+    if is_requestable? && not_many_public_requests?
       add_tag_if_not_already_present('not_many_requests')
     else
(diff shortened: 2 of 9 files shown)


garethrees:
> Implemented well, but wondering why we need to specify `not_many_requests_tag: true` on some specs now and if there is an alterative.

Err, it's been a while… 😅 IIRC it was that in those cases the specs were intermittently failing because depending on ordering of runs an authority may or may not meet the criteria to have the tag auto-applied, and as such breaks the expectations (e.g. sometimes the tag string will be `"foo"`, sometimes `"foo not_many_requests"`). Very open to a better way if you can think of one.

gbp (approved):
Not a fan of using RSpec example metadata to setup the specs (yeah I know I added that for features but that is different because those are optional for re-users)

I have pushed a fixup to be more explicit, adding the around example block for the 4 times the metadata was used.

garethrees:
Okay! might see if we can refactor d9834d0 to be more like:

```rb
  describe "DELETE #mass_tag" do
    around do |example|
      disable_not_many_requests_auto_tagging { example.run }
    end
```

garethrees:
> Okay! might see if we can refactor [d9834d0](https://github.com/mysociety/alaveteli/commit/d9834d08a3cf05a3022a056f838c7fb026eb386f) to be more like:

Done and squashed into https://github.com/mysociety/alaveteli/pull/7363/commits/b12b788ce7dda109166cc507bf92785f95e6f085#diff-8799477776f9e46de8e471992f769f6c8fe603d9c671c12472446948c8e57256. Will merge on green.

PR #7364, opened 2022-10-14
[#6746] Ensure prominence_reason follows prominence

garethrees:
Ensure the `prominence_reason` attribute always follows `prominence` in
`admin_columns` listings.

This is needed for models where other columns have been added to the
database table between `prominence` and `prominence_reason`. Rails'
`content_columns` returns the columns in the order they appear in
Postgres, which we have no control over.

This commit repositions the `prominence_reason` column if necessary.

    # BEFORE: prominence_reason at the end of the Array
    InfoRequest.admin_columns
    => ["created_at", "updated_at", "described_state",
        "awaiting_description", "prominence", "law_used", # snip…
        "prominence_reason", "rejected_incoming_count"]

    # AFTER: prominence_reason follows prominence
    InfoRequest.admin_columns
    => ["created_at", "updated_at", "described_state",
        "awaiting_description", "prominence", "prominence_reason",
        "law_used", "allow_new_responses_from", # snip…]

Screenshot 2022-10-14 at 13 10 51

### DIFF
--- a/app/models/concerns/admin_column.rb
+++ b/app/models/concerns/admin_column.rb
@@ -6,14 +6,16 @@ module AdminColumn
       def admin_columns(exclude: nil, include: nil)
         @excluded_admin_columns = exclude || @excluded_admin_columns
         @included_admin_columns = include || @included_admin_columns
-        sorted_columns
+        ordered_columns
       end
 
       # Ensure prominence_reason immediately follows prominence
-      def sorted_columns
+      def ordered_columns
         return all_columns unless prominenceable_admin_columns?
-        index = all_columns.index('prominence') + 1
-        all_columns.insert(index, all_columns.delete('prominence_reason'))
+
+        columns = all_columns
+        index = columns.index('prominence') + 1
+        columns.insert(index, columns.delete('prominence_reason'))
       end
 
       def prominenceable_admin_columns?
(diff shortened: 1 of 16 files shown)


app/models/concerns/admin_column.rb
```diff
@@ -6,11 +6,25 @@ class << self
...
         @excluded_admin_columns = exclude || @excluded_admin_columns
         @included_admin_columns = include || @included_admin_columns
+        sorted_columns
+      end
+
+      # Ensure prominence_reason immediately follows prominence
+      def sorted_columns
+        return all_columns unless prominenceable_admin_columns?
+        index = all_columns.index('prominence') + 1
+        all_columns.insert(index, all_columns.delete('prominence_reason'))
```
gbp, line 16: This keeps the original 'prominence_reason' in the array so the column is now displayed twice. Will have to update to delete the duplicate.

app/models/concerns/admin_column.rb
```diff
@@ -6,11 +6,25 @@ class << self
       def admin_columns(exclude: nil, include: nil)
         @excluded_admin_columns = exclude || @excluded_admin_columns
         @included_admin_columns = include || @included_admin_columns
+        sorted_columns
+      end
+
+      # Ensure prominence_reason immediately follows prominence
+      def sorted_columns
```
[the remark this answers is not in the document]
garethrees: +1 better!

app/models/concerns/admin_column.rb
garethrees: I thought `delete` removed it? Maybe because we're doing it in one shot?

gbp: Calling `delete` on `all_columns` doesn't because it's a method which builds/returns a different array instance.

PR #7377, opened 2022-10-21
[#7376] Fix overflow in admin UI

### LINKED ISSUE #7376: Wrap `post redirect` tokens in `/admin/users`
Password reset post-redirect tokens are unusually large (in comparison to other tokens used) - which results in a rather large token being displayed on user admin pages.

There's no doubt a good reason for that, but it makes the page rather unwieldy, as the token isn't wrapping  at a reasonable breakpoint, meaning the admin has a lot of scrolling to do.
Redacted segment of admin page showing a table of 'post redirect' tokens which has spilled off the screen

Could the `Email token` column have some styling added so as to wrap the text?

gbp:
## Relevant issue(s)

Fixes #7376

## What does this do?

Truncate `PostRedirect#email_token` so the admin UI doesn't overflow.

## Screenshots

image

### DIFF
--- a/app/views/admin_user/show.html.erb
+++ b/app/views/admin_user/show.html.erb
@@ -164,7 +164,7 @@
         <td><%=h post_redirect.id %></td>
         <% for column in PostRedirect.content_columns.map { |c| c.name } %>
           <% if column == 'email_token' %>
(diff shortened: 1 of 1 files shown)


mdeuk:
Thank you! 🙇

PR #7378, opened 2022-10-21
[#3830] Update request base calculated status

### LINKED ISSUE #3830: Update status form does not include 'I'm still waiting for the internal review' when a response is received to a request where an internal review has been requested
I think this is because [`InfoRequest::State::Calculator.transitions`](https://github.com/mysociety/alaveteli/blob/develop/app/models/info_request/state/calculator.rb#L87-L99) uses `@info_request.calculate_status(cached_value_ok)`, which will return `waiting_classification` when the request is awaiting description (as it is when a response has been received). I think we could resolve this by switching it to use `@info_request.described_state`.

gbp:
## Relevant issue(s)

Fixes #3830

## What does this do?

Update request base calculated status

## Why was this needed?

Once a requester has asked for a internal review their request's base status shouldn't ever be `waiting_classification` as this limits the possible classification statuses.

Returning `internal_review` allows the requester to classify the request as `I'm still waiting for the internal review` after they receive an acknowledgement of their review.

### DIFF
--- a/app/models/info_request.rb
+++ b/app/models/info_request.rb
@@ -758,6 +758,10 @@ class InfoRequest < ApplicationRecord
     end
   end
 
+  def internal_review_requested?
+    outgoing_messages.where(what_doing: 'internal_review').any?
+  end
+
   def is_external?
     external_url.nil? ? false : true
   end
--- a/app/models/info_request/state/calculator.rb
+++ b/app/models/info_request/state/calculator.rb
@@ -93,7 +93,9 @@ class InfoRequest
             other: {}
           }
         end
-        opts.merge!(in_internal_review: state == 'internal_review')
+        opts[:in_internal_review] = state == 'internal_review'
+        opts[:internal_review_requested] =
+          @info_request.internal_review_requested?
         build_transitions_hash(opts)
       end
 
@@ -132,7 +134,11 @@ class InfoRequest
         if opts.fetch(:in_internal_review, false)
           states = %w[internal_review gone_postal]
         else
-          states = %w[
+          states = []
+          if opts.fetch(:internal_review_requested, false)
+            states += ['internal_review']
+          end
+          states += %w[
             waiting_response
             waiting_clarification
             gone_postal
@@ -141,7 +147,7 @@ class InfoRequest
             states += ['internal_review']
           end
         end
-        states
+        states.uniq
       end
 
       def complete_states(_opts = {})
--- a/app/models/info_request/state/transitions.rb
+++ b/app/models/info_request/state/transitions.rb
@@ -86,7 +86,8 @@ class InfoRequest
       end
 
       def self.owner_internal_review_transition_label(opts = {})
-        if opts.fetch(:in_internal_review, false)
+        if opts.fetch(:in_internal_review, false) ||
+           opts.fetch(:internal_review_requested, false)
           _("I'm still <strong>waiting</strong> for the internal review")
         else
           _("I'm waiting for an <strong>internal review</strong> response")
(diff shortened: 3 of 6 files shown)


app/models/info_request.rb
```diff
@@ -744,6 +744,10 @@ def must_be_internal_or_external
     end
   end
 
+  def internally_reviewed?
```
gbp, line 747: A better method name `internal_review_sent?` ??

garethrees (commented):
(For future PRs), it's really helpful to add before/after screenshots to demonstrate the before/after so that it's clear that we have reproduced the problem behaviour and that the commit fixes it, and that the reviewer understands what they're looking for when assessing the fix.

Screenshot 2022-10-21 at 17 05 42

---

> Once a requester has asked for a internal review their request's base status shouldn't ever be `waiting_classification` 

This logic doesn't _sound_ right. Waiting classification is where the user must update the state, which they could well be asked to do after an internal review. For example:

1. some correspondence has happened and the user is not happy
2. they request an internal review
3. the authority sends an auto-acknowledgement of their correspondence

At step 3 the request _would_ need classification. Obviously the fix has rendered the correct option, but I'm not sure I understand why.

---

The original issue (https://github.com/mysociety/alaveteli/issues/3830#issue-214392730) noted:

> I think we could resolve this by switching it to use `@info_request.described_state`

This implementation obviously adds the option, but would be good for this PR to note why we didn't just switch to `#described_state`

---

This returns the "internal review" variation of the form if the request has _ever_ had an internal review requested. I _think_ that's okay? What happens if we get a chain like:

* normal outgoing
* internal review
* normal outgoing
* normal outgoing

Should it _always_ be treated as internal review in those later messages?

app/models/info_request.rb
garethrees: I'd go with `internal_review_requested?`. `internal_review_sent?` is also okay and says similar, but feels more about technical message delivery than user intent.

`internally_reviewed?` implies the authority _did_ a review, which may not be true.

gbp:
Rebased without doing any of the changes required to move this forward.

gbp:
> > Once a requester has asked for a internal review their request's base status shouldn't ever be `waiting_classification`
> 
> This logic doesn't _sound_ right. Waiting classification is where the user must update the state, which they could well be asked to do after an internal review.

It's been awhile but I think what I was getting on here was the `internal_review` state is equivalent to `waiting_classification` once a review has been requested, but yes, I agree that sounds wrong.

Looking back on it now, it seems like I was thinking this because it was an easy way to get the correct options displayed on the describe state form. I probably was thinking short of adding new states (like `internal_review_waiting_classification`/`internal_review_done`) there is no real way to track if an internal review had been request or completed.

gbp:
> The original issue ([#3830 (comment)](https://github.com/mysociety/alaveteli/issues/3830#issue-214392730)) noted:
> 
> > I think we could resolve this by switching it to use `@info_request.described_state`
> 
> This implementation obviously adds the option, but would be good for this PR to note why we didn't just switch to `#described_state`

I have looked at this method but using it causes other issues, for example the overdue and very overdue phases aren't returned any more. There is extra logic in `calculate_status` to handle these states. I'm happy that sticking with `calculate_status` is the easier option.

gbp:
> This returns the "internal review" variation of the form if the request has _ever_ had an internal review requested. I _think_ that's okay? ...

There is a way we can still have both variations of the form. I have pushed a fixup which reverts the changes to `calculate_status`, instead now using the `internal_review_requested?` conditional in the state calculator to show the `internal_review` option (but bumped to the top of the list to be more prominent), we already show this option on the main (non-internal review variation) form if the user has selected 'Update the status of this request' from the action popup. I think this is all that is needed to fixed this issue/PR.

There are really 4 variations; screenshots of each are below:
7378-1-request-response
7378-2-request-manual
7378-3-ir-response
7378-4-ir-manual

garethrees (approved):
Nice – seems to look good! 🙏

PR #7379, opened 2022-10-21
[#7370] Update Pro classification controller auth

### LINKED ISSUE #7370: Ex pro user hitting "This page is only accessible to WhatDoTheyKnow Pro users" message when trying to follow-up correspondence
Specific details shared with dev@

gbp:
## Relevant issue(s)

Fixes #7370

## What does this do?

Update Pro classification controller auth

## Why was this needed?

Former pro users still need to be able to update their request statuses.

This change removes the `before_action` callback to authenticate the pro users. This means all requesters can use this endpoint to classify their request but in reality only current and former Pro users will as this is only linked to from the Pro UI.


garethrees (commented):
Haven't run this but looks about right 👍

Issue #7372, opened 2022-10-19
Don't show `missing_email` note on `defunct` bodies


RichardTaylor:

WilliamWDTK:
Or, don't apply the `missing_email` tag to `defunct` bodies.

mdeuk:
> Or, don't apply the `missing_email` tag to `defunct` bodies.

@confirmordeny recently suggested that we could use a `defunct@<sitedomain>` email address for these entities, perhaps that would be a solution?

gbp:
Deployed and ran `PublicBody.with_tag('missing_email').with_tag('defunct').each { |b| b.send :update_missing_email_tag }` to update the 6 bodies which had both tags.

PR #7381, opened 2022-10-21
[#7372] Don't tag defunct bodies with missing email tag

garethrees:
If an authority is defunct we don't want suggestions for a request email.

Fixes https://github.com/mysociety/alaveteli/issues/7372.


gbp (approved):
Great, looks like this will also remove the tag if a body without an email gets tagged as defunct.

PR #7391, opened 2022-10-24
[#1005] Hide individual attachments

### LINKED ISSUE #1005: Allow individual attachments to incoming messages to be hidden

gbp:
## Relevant issue(s)

Supersedes #7327 
Fixes #1005

## What does this do?

Allows admins to edit attachment prominence and reason.

Add prominence checks for attachments when:
- viewing requests/messages
- viewing attachments
- ZIP downloads
- Xapian indexing

## Why was this needed?

Easier admin of responses and attachments.

## Notes to reviewer

### DIFF
--- a/app/controllers/admin/foi_attachments_controller.rb
+++ b/app/controllers/admin/foi_attachments_controller.rb
@@ -9,10 +9,31 @@ class Admin::FoiAttachmentsController < AdminController
   end
 
   def update
-    redirect_to edit_admin_incoming_message_path(@incoming_message)
+    if @foi_attachment.update(foi_attachment_params)
+      @info_request.log_event(
+        'edit_attachment',
+        attachment_id: @foi_attachment.id,
+        editor: admin_current_user,
+        old_prominence: @foi_attachment.prominence_previously_was,
+        prominence: @foi_attachment.prominence,
+        old_prominence_reason: @foi_attachment.prominence_reason_previously_was,
+        prominence_reason: @foi_attachment.prominence_reason
+      )
+      @info_request.expire
+
+      flash[:notice] = 'Attachment successfully updated.'
+      redirect_to edit_admin_incoming_message_path(@incoming_message)
+
+    else
+      render action: 'edit'
+    end
   end
 
   private
+
+  def foi_attachment_params
+    params.require(:foi_attachment).permit(:prominence, :prominence_reason)
+  end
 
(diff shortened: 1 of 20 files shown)


gbp:
Need to address points made in previous review https://github.com/mysociety/alaveteli/pull/7327#pullrequestreview-1151112343

gbp:
From https://github.com/mysociety/alaveteli/pull/7327#pullrequestreview-1151112343
> This doesn't cover the case where the `FoiAttachment` is "normal", but it's `IncomingMessage` is reduced prominence. In this situation you can still view the attachment if you know its direct URL (which search engines will). I think it needs to be like:
> 
> ```ruby
> can_view_with_prominence?(attachment.prominence, attachment.incoming_message.info_request) &&
>   can?(:read, attachment.incoming_message)
> ```
> 
> This is pretty critical behaviour, so I think it needs tests, somewhere….

This has now been fixed and separated out into its own commit in: https://github.com/mysociety/alaveteli/pull/7391/commits/d606bf6a807e205dbc5ec6b5bb5f1464b8e56b54

gbp:
From https://github.com/mysociety/alaveteli/pull/7327#pullrequestreview-1151112343
> As an admin user I though I should be able to see reduced-prominence main body parts below the prominence reason?

Fixed in https://github.com/mysociety/alaveteli/pull/7391/commits/6c302bfe097998c9453bbb8fe6f19a6daf5fad9b

garethrees (approved):
Looks great! So much work to get here, but everything much nicer as a result 🥷

I haven't been able to test every permutation of every visibility options per action, but given the much clearer `Ability` checks I'm pretty confident.

PR #7409, opened 2022-11-02
[#1005] Hide individual attachment fixes 

gbp:
## Relevant issue(s)

Connected to #1005

## What does this do?

1. Removes CanCanCan Ability scopes
2. Fixes exception rendering delivery statues


gbp:
I have tested `ability.rb` on production using the Rails console. Prior to these changes running `Ability.guest` would hang, now it returns near instantly.

gbp:
Running `Ability.new(pro_admin_user)`, `Ability.new(admin_user)` and `Ability.new(other_user)` all work correctly too.

garethrees (approved):
Haven't run but looks fine in principle and the specs all pass.

Issue #7416, opened 2022-11-03
Couple of in-line admin page help tips related to the ability to hide individual attachments


RichardTaylor:
Suggestions:

* Note saying the first? attachment, or one with type text/plain is in-fact the message body and not what would generally be considered an attachment. 
  * Can the relevant attachment be marked "message body"?
* Warning on the edit attachment page that changing the prominence of the attachment doesn't affect the visibility of the filename
  * Could go a step further and note file names are impacted by censor rules on requests as another related tip

We want to make things easy for new admins to understand.  

[Sorry this is one of those tickets/ticket notes that arises from a great new feature being deployed]

garethrees:
> Can the relevant attachment be marked "message body"?

Yep! https://github.com/mysociety/alaveteli/pull/7417

> Warning on the edit attachment page that changing the prominence of the attachment doesn't affect the visibility of the filename

It does hide it for those who _can't_ view the attachment:

![Screenshot 2022-11-04 at 09 48 56](https://user-images.githubusercontent.com/282788/199943629-dc60a1e4-d0b3-408f-8f6f-4a062c70474a.png)

So in this case (`requester_only`):

* Admins _can_ see the filename
* The requester _can_ see the filename
* A different logged in user _cannot_ see the filename
* A non-logged in visitor _cannot_ see the filename

Is there somewhere else we've missed the reducing prominence of filename?

> Sorry this is one of those tickets/ticket notes that arises from a great new feature being deployed

No need to apologise! This is how software works! ☺️

RichardTaylor:
>Is there somewhere else we've missed the reducing prominence of filename?

Closing pending an example of any filename related issue. 

I suspect the issue would have been one of surprise related to "Admins can see the filename", and the difference in behaviour between this system and censor rules - censor rules do make material disappear even for admins

garethrees:
> the difference in behaviour between this system and censor rules - censor rules do make material disappear even for admins

Ah, yeah, this is based on the "prominence" system, which _does_ allow admins to view all content. I'd like to nudge the Censor Rules towards using prominence too (https://github.com/mysociety/alaveteli/issues/5412#issuecomment-1227432934) so that it's all a bit more aligned.

PR #7417, opened 2022-11-04
[#7416] Label the main body attachment

garethrees:
Make it clear which attachment is the main body part when viewing lists of attachments associated with an incoming message.

A part of https://github.com/mysociety/alaveteli/issues/7416.

Screenshot 2022-11-04 at 09 44 30


garethrees:
Could use "Message body" as suggested in https://github.com/mysociety/alaveteli/issues/7416, but "Main body" is closer to the existing code language and is a shorter label so takes a little less space.

Issue #7456, opened 2022-12-08
Ex-pro users can't follow up to still-private requests


garethrees:
Bug reported via [dev list](https://groups.google.com/a/mysociety.org/g/alaveteli-dev-private/c/h-mY9r7Wukw/m/7_bNJ1MxAQAJ).

> A handful of requests in the batch are still active due to internal reviews or delays. However when I try to use Write Reply (or other response actions) using the buttons in the page for these I just get taken to a 'Sorry we couldn't find that page' error.

@gbp did some initial investigation and found that it looks like the bug is here: 

https://github.com/mysociety/alaveteli/blob/481790fa6494abb1a125c9a30e91fdaf6ec1c848/app/controllers/followups_controller.rb#L196

> second part of the conditional shouldn't be there I think

The code was introduced in https://github.com/mysociety/alaveteli/commit/085b1b2bc21f504aecac36cea9e253527ad5a257.

PR #7457, opened 2022-12-08
[#7456] Allow ex-pro users to followup to private requests

garethrees:
Ex-pro users should still be able to manage requests even if they're still private; they just can't extend privacy periods or create new ones.

We don't need to check that the `current_user` is a Pro here; we first search the user's requests, irrespective of privacy state, and assign to the instance variable. If that isn't found – which means the user is trying to follow up to someone else's request – we only search public requests in order to display the notice to sign in to the request author's account.

Fixes https://github.com/mysociety/alaveteli/issues/7456.

### DIFF
--- a/app/controllers/followups_controller.rb
+++ b/app/controllers/followups_controller.rb
@@ -188,10 +188,11 @@ class FollowupsController < ApplicationController
   end
 
   def set_info_request
-    if current_user && current_user.is_pro?
+    if current_user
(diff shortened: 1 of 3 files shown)


garethrees:
Haven't tested in the browser yet; just checking the full spec run first.

gbp (approved):
I think this is good to go.

Issue #7466, opened 2022-12-14
Add a link to the tag page from the admin navbar


HelenWDTK:
At the moment you have to follow the link from the authorities page to reach the tags page. We use this a lot to add notes, so it'd be nice to make this easier to find.

garethrees:
IIRC we intentionally didn't add this – partly because it wasn't obvious where to put it, and partly because we figured it was easy enough to just navigate via other content – but I agree; I've been finding myself typing `/admin/tags` fairly frequently these days.

garethrees:
Fixed by 8db57c4b5.

PR #7467, opened 2022-12-14
[#7466] Add link to tags to the admin navbar

HelenWDTK:
Add link to tags to the navbar

## Relevant issue(s)
#7466 
## What does this do?
Adds a link to the tag page from the navbar
## Why was this needed?
At the moment you have to follow the link from the authorities page to reach the tags page. We use this a lot to add notes, so it'd be good to make this easier to find.
## Screenshots
Screenshot 2022-12-14 at 06 48 36

## Notes to reviewer
I put this under authorities because that's where we go to find it at the moment, but there's no reason not to move it somewhere else, seeing as we tag lots of things now.

### DIFF
--- a/app/views/admin_general/_admin_navbar.html.erb
+++ b/app/views/admin_general/_admin_navbar.html.erb
@@ -27,6 +27,7 @@
             <ul class="dropdown-menu" role="menu">
               <li><%= link_to 'Authorities', admin_bodies_path %></li>
               <li><%= link_to 'Categories', admin_categories_path %></li>
+              <li><%= link_to 'View tags', admin_tags_path %></li>
             </ul>
           </li>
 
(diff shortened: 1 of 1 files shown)


garethrees (approved):
Nice! Happy to merge as is, but here are some extra thoughts (that can totally be tackled later).

Adding under authorities is fine. I think we could also add under requests too, scoping the tag filter to records of the given class.

"Authorities" menu:

```erb
<li><%= link_to 'View tags', admin_tags_path(model_type: 'PublicBody') %></li>
```

"Requests" menu:

```erb
<li><%= link_to 'View tags', admin_tags_path(model_type: 'InfoRequest') %></li>
```

I'd have probably gone with "Tags" rather than "View tags", as we just use the record name in all other items, but wouldn't object to merging and tidying up later.

HelenWDTK:
OK, so I encorporated the suggested changes. They look fine. I've no idea what I'm doing though, so will not press any more buttons, lest I break something 🙊 

Screenshots here:
Screenshot 2022-12-14 at 13 32 58
Screenshot 2022-12-14 at 13 44 23

garethrees:
Amazing, thanks. Merged as 8db57c4b5 (squashed into one commit and made a tiny tweak to set the `model_type` parameter for the authority link).

PR #7471, opened 2022-12-15
[#7470] Add rake task to convert legacy YAML

### LINKED ISSUE #7470: Migration to JSONB params failing
When running `temp:sanitise_and_populate_events_params_json` we are seeing errors as it would seem the `params_yaml` stored for old WDTK events could be invalid YAML and not able to be parsed by Psych - the Ruby default YAML parser.

The YAML seems to have been generated by the older Syck library - this has now been removed from Ruby stdlib and extracted into a gem.

We should fallback to the old library when an `Psych::SyntaxError` exception is raised.

gbp:
## Relevant issue(s)

Fixes #7470

## What does this do?

[Add rake task to convert legacy YAML](https://github.com/mysociety/alaveteli/commit/28991ccd9fd16932977e22fb7efb50b6ef196f54)

## Why was this needed?

In WDTK we have YAML stored which was generate by the old `Syck` library
which can't be parsed by the newer `Psych` library.

This rake task fixes the YAML so it can be parsed.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -129,6 +129,7 @@ else
   gem 'strip_attributes', :git => 'https://github.com/mysociety/strip_attributes.git', :branch => 'globalize3-rails5.2'
 end
 gem 'stripe', '~> 5.55.0'
+gem 'syck', '~> 1.4.1', require: false
 gem 'syslog_protocol', '~> 0.9.0'
 gem 'thin', '~> 1.8.1'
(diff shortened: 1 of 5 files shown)


gbp (commented):
I have run a variation of these tasks in the Rails production console to ensure the remaining events can now be migrated. With exception on 4 records which hit https://github.com/mysociety/alaveteli/issues/7478

PR #7490, opened 2023-01-03
Add ability to paginate through batch requests

garethrees:
Add "Next" and "Prev" pagination links to the sidebar of info requests that belong to a batch request.

This allows users to navigate through requests in the batch without having to click "in and out" via the batch page.

To minimise implementation complexity the pagination doesn't have a beginning and end. We just cycle back around to the first request in the batch when we hit the last one; same for paginating in reverse. If we didn't do this we'd have to have conditionals for disabling the particular navigation option when we hit an end.

Screenshot 2023-01-03 at 09 59 26

### DIFF
--- a/app/models/info_request/batch_pagination.rb
+++ b/app/models/info_request/batch_pagination.rb
@@ -1,0 +1,26 @@
+# Methods to handle pagination within an associated InfoRequestBatch
+module InfoRequest::BatchPagination
+  def next_in_batch
+    return nil unless info_request_batch
+    batch_sibling_requests_ordered_by_id[index_in_batch + 1] || first_in_batch
+  end
+
+  def prev_in_batch
+    return nil unless info_request_batch
+    batch_sibling_requests_ordered_by_id[index_in_batch - 1]
+  end
+
+  private
+
+  def first_in_batch
+    batch_sibling_requests_ordered_by_id.first
+  end
+
+  def index_in_batch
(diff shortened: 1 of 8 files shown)


app/models/info_request/batch_pagination.rb
```diff
@@ -0,0 +1,22 @@
+# Methods to handle pagination within an associated InfoRequestBatch
+module InfoRequest::BatchPagination
+  def next_in_batch
+    return nil unless info_request_batch
+    info_request_batch.info_requests[index_in_batch + 1] || first_in_batch
```
gbp, line 5: Are we certain the info_requests are always order the same way?

app/models/info_request/batch_pagination.rb
```diff
@@ -0,0 +1,22 @@
...
+  end
+
+  private
+
+  def first_in_batch
+    info_request_batch.info_requests.first
+  end
+
+  def index_in_batch
+    info_request_batch.info_requests.index(self)
```
gbp, line 20: Potentially loading 500 `InfoRequest` objects, pluck the IDs and find the index of `self.id`?

app/models/info_request/batch_pagination.rb
garethrees: Pretty sure, yeah. Could explicitly order by ID here?

gbp (approved):

PR #7503, opened 2023-01-05
Drop Vagrant development boxes

gbp:
## What does this do?

Drop Vagrant development boxes

## Why was this needed?

We are now using Docker for development instead.

## Notes to reviewer

@garethrees are you okay with remove the Vagrantfile? As its of no use on M1 Macs I see no reason to keep it around as we can't test it. Going forward we need a plan on how to test OSs.


garethrees:
Need to think about this a bit. It looks like it [may be possible](https://osxdaily.com/2022/10/22/you-can-now-run-virtualbox-on-apple-silicon-m1-m2/) to use VirtualBox on M1 soon.

I'd like to both simplify what we maintain, but also get back to a place where we develop with [dev/prod parity](https://12factor.net/dev-prod-parity), which we're not really doing right now, but were much closer with the Vagrant setup.

Also need to consider future mySociety infrastructure direction.

garethrees:
Looks like Rails is going to [start generating a `Dockerfile` by default](https://github.com/rails/rails/pull/46762), so that's probably the way we want to go here.

gbp:
The above `wkhtmltox` package change isn't needed anymore as mySociety provides Arm version of `wkhtmltox` now.

gbp:
With the move to BUSL I think we should get this merged.

garethrees:
Agree! 🧹

PR #7510, opened 2023-01-12
Update text masking for PDF attachments

gbp:
## What does this do?

Update text masking to use tempfiles for PDF attachments when running `pdftk` external commands

## Why was this needed?

Large PDFs attachments, with lots of images are sometimes received.
These will normally will be compressed and expand to much larger files
size. This can cause memory issues due to how we are passing the data
around the application to external commands such as `pdftk`.

Passing this data via STDIN to `pdftk` to seems to be the main issue
here where we are seeing uncompressed PDFs over 900Mb not able to be
re-compressed despite leaving the process running for over 1 hour.

This change switches to use tempfiles when uncompressing or compressing
PDF attachments which, in testing on my development and WDTK production
environment, allows the whole `apply_masks` process which includes
uncompressing and re-compressing, to complete in under 2 minutes.

### DIFF
--- a/lib/alaveteli_text_masker.rb
+++ b/lib/alaveteli_text_masker.rb
@@ -1,3 +1,5 @@
+require 'tempfile'
+
 module AlaveteliTextMasker
   include ConfigHelper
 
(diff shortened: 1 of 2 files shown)


gbp:
Tested this on a PDF received on WDTK. File came in at 11.7MB but decompresses to 953.3MB.

The file couldn't be downloaded as the text masking never completed as the processes were killed after around 1 hour.

With this change:

```
$ time bin/rails runner pr7510-test.rb
real	1m52.591s
user	1m12.224s
sys	0m11.967s
```

lib/alaveteli_text_masker.rb
```diff
@@ -47,10 +49,22 @@ def apply_masks(text, content_type, options = {})
   private
 
   def uncompress_pdf(text)
-    AlaveteliExternalCommand.run("pdftk", "-", "output", "-", "uncompress", :stdin_string => text)
+    temp = Tempfile.new(encoding: 'ascii-8bit')
```
garethrees, line 52: Perhaps these ought to get created within the app's `tmp` dir (with a name prefix), so that it's clear what they relate to:

```diff
- temp = Tempfile.new(encoding: 'ascii-8bit')
+ temp = Tempfile.new('pdftk', './tmp', encoding: 'ascii-8bit')
```

garethrees (approved):
Haven't run but looks sensible to me.

PR #7521, opened 2023-01-17
Add list of batch requests to admin user page

garethrees:
## What does this do?

Make it easier to see distinct batches when viewing a user record.

## Why was this needed?

Currently you have to open up requests to get to the batch record, and
there's always the possibility that the user has created to batches with
identical titles – maybe after missing some authorities from the initial
list.

## Implementation notes

Fixes some minor code style issues and adds `InfoRequestBatch.not_embargoed` scope.

## Screenshots

Screenshot 2023-01-17 at 09 51 06

The batches are not yet sent – "Sent at" is not yet populated.

Screenshot 2023-01-17 at 09 43 52

The batches get sent, and now "Sent at" is populated

Screenshot 2023-01-17 at 09 50 16

Viewing the user's page as Joe Admin, who doesn't have `pro_admin` role:

Screenshot 2023-01-17 at 09 52 58

### DIFF
--- a/app/models/info_request_batch.rb
+++ b/app/models/info_request_batch.rb
@@ -46,6 +46,8 @@ class InfoRequestBatch < ApplicationRecord
 
   strip_attributes only: %i[embargo_duration]
 
+  scope :not_embargoed, -> { where(embargo_duration: nil) }
+
   def self.send_batches
     where(sent_at: nil).find_each do |info_request_batch|
       AlaveteliLocalization.with_locale(info_request_batch.user.locale) do
--- a/app/controllers/admin_user_controller.rb
+++ b/app/controllers/admin_user_controller.rb
@@ -48,10 +48,12 @@ class AdminUserController < AdminController
 
   def show
     @info_requests = @admin_user.info_requests
+    @info_request_batches = @admin_user.info_request_batches
     @comments = @admin_user.comments
 
     if cannot? :admin, AlaveteliPro::Embargo
       @info_requests = @info_requests.not_embargoed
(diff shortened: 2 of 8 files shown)


app/models/info_request_batch.rb
```diff
@@ -46,6 +46,8 @@ class InfoRequestBatch < ApplicationRecord
 
   strip_attributes only: %i[embargo_duration]
 
+  scope :not_embargoed, -> { where(embargo_duration: nil) }
```
gbp, line 49: Is this always true?, would a better check be to see if any of the batch's info requests have embargo:

```suggestion
  scope :not_embargoed, -> { left_joins(info_requests: :embargo).where(embargoes: { id: nil }) }
```

garethrees: It's how we define `InfoRequestBatch#embargoed?`. The value gets cleared in [`AlaveteliPro::Embargo.expire_publishable`](https://github.com/mysociety/alaveteli/blob/0.42.0.0/app/models/alaveteli_pro/embargo.rb#L103-L105), although it looks like we might _not_ do so when a user clicks "publish now". I'd _thought_ we'd fixed that, so will have to check.

In general I do think we should be able to rely on it; just need to make sure in practice that we can.

garethrees: > it looks like we might _not_ do so when a user clicks "publish now". I'd _thought_ we'd fixed that, so will have to check.
> 
> In general I do think we should be able to rely on it; just need to make sure in practice that we can.

It _does_ get cleared when clicking "publish requests", so I think this is okay as-is.

PR #7544, opened 2023-01-24
Add support for Mail v2.8.x

gbp:
## Relevant issue(s)

See: https://github.com/mikel/mail/issues/1538
See: https://github.com/rails/rails/pull/46650

## What does this do?

Add support for Mail v2.8.x

## Why was this needed?

Update sendmail settings to fix argument error.


gbp (commented):
Tested on staging and this fixes the issue we've seen deploying the develop branch.

PR #7584, opened 2023-02-13
Set STAGING_SITE when generating initial config

### LINKED ISSUE #7583: Problem launching site after initial script install
After installing Alaveteli without the [development flag](https://github.com/mysociety/commonlib/blob/8b638d674badcb1dc6a4867d9f49571caecc923c/bin/install-site.sh#L546) the application won't boot due to the wrong Rails ENV being used.

In the install process the `config/rails_env.rb` file is only [created](https://github.com/mysociety/alaveteli/blob/44cc939dd65fcbb78f3b29fde608bdbd989a8553/script/rails-deploy-before-down#L80-L96) when `STAGING_SITE = 0` but this defaults to `1` so is only done is the installer edits `config/general.yml` and reruns the install script.

We could detect the development flag and set `STAGING_SITE` automatically.

gbp:
## Relevant issue(s)

Fixes #7583

## What does this do?

Set `STAGING_SITE` when generating initial config

## Why was this needed?

We need this configuration value in `script/rails-deploy-before-down` to correctly set `RAILS_ENV` to `production` when not doing a development install.

### DIFF
--- a/script/install-as-user
+++ b/script/install-as-user
@@ -141,6 +141,12 @@ RANDOM_EMAIL_SECRET=$(random_alphanumerics 32)
 RANDOM_EMERGENCY_PASSWORD=$(random_alphanumerics 10)
 RANDOM_COOKIE_SECRET=$(random_alphanumerics 100)
 
+if [ "$DEVELOPMENT_INSTALL" = true ]; then
+  STAGING_SITE=1
+else
+  STAGING_SITE=0
+fi
+
 if ! [ -f config/general.yml ]
 then
     sed -r \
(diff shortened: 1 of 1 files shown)


script/install-as-user
```diff
@@ -158,6 +164,7 @@ then
         -e "s,^( *CONTACT_EMAIL:).*,\\1 'postmaster@$HOST'," \
         -e "s,^( *TRACK_SENDER_EMAIL:).*,\\1 'postmaster@$HOST'," \
         -e "s,^( *SECRET_KEY_BASE:).*,\\1 '$RANDOM_COOKIE_SECRET'," \
+        -e "s,^( *STAGING_SITE:).*,\\1 '$STAGING_SITE'," \
```
garethrees, line 167: Isn't the intention that `STAGING_SITE` is set in `general.yml`. AIUI, this would overwrite that value?

Or are we saying that irrespective of the `general.yml` configuration, `STAGING_SITE` should be set to `1` if `DEVELOPMENT_INSTALL` is `true`? This seems a bit confusing in that the expected config would be ignored. Could it error out instead (with a message like "Incompatible config options! `STAGING_SITE` must be `1` for a `DEVELOPMENT_INSTALL`"?)

gbp: This only gets run when `general.yml` doesn't exist so is setting the default value based on if the install script has the `--development` flag or not.

gbp: If the install script is rerun then yes maybe we should warn of "Incompatible config options". Would you like me to add that now?

garethrees: Ah, yeah, sounds good then.

garethrees: > "Incompatible config options". Would you like me to add that now?

If it's a quick change it would likely avoid confusion.

gbp: See: https://github.com/mysociety/alaveteli/pull/7622

PR #7587, opened 2023-02-14
Update setting RAILS_ENV in install scripts

gbp:
## What does this do?

When re/running the installer:
1. update `RAILS_ENV` file based on the `STAGING_SITE` configuration
2. don't nuke customised settings in `config/rails_env.rb`.

## Why was this needed?

When switching to Puma #7581 there are other ENV variables which might need to be customised in  `config/rails_env.rb`, such as:
- `RAILS_MAX_THREADS`
- `RAILS_MIN_THREADS`
- `WEB_CONCURRENCY`
- `PORT`

The `config/rails_env.rb` file seems the correct place for over `config/general.yml` as we don't really need to expose these settings for a majority of re-users.

Without these changes we risk accidental removing these when the installer is re-run.

### DIFF
--- a/script/rails-deploy-before-down
+++ b/script/rails-deploy-before-down
@@ -77,9 +77,8 @@ OPTION_STAGING_SITE=$(bin/config STAGING_SITE)
 STAGING_SITE="${OPTION_STAGING_SITE:-1}"
 
 # Force appropriate environment in production
-if [ "$STAGING_SITE" = "0" ]
-then
-    cat <<-END
+if [ ! -f config/rails_env.rb ]; then
+    [ "$STAGING_SITE" = "0" ] && cat <<-END
 
     *****************************************************************
     WARNING: About to make config/rails_env.rb which, via special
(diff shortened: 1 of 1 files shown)


script/rails-deploy-before-down
```diff
@@ -92,7 +91,17 @@ then
...
 
 END
-    echo "ENV['RAILS_ENV'] ||= 'production'" > config/rails_env.rb
+fi
+
+RAILS_ENV=$(ruby -r ./config/rails_env.rb -e "puts ENV['RAILS_ENV']" 2> /dev/null)
+if [ "$STAGING_SITE" = "0" ] && [ "$RAILS_ENV" != "production" ] ||
+   [ "$STAGING_SITE" = "1" ] && [ "$RAILS_ENV" == "production" ]; then
+    # STAGING_SITE and RAILS_ENV mismatch - remove line
+    sed -i "/^ENV\['RAILS_ENV'\]/d" config/rails_env.rb 2> /dev/null
```
garethrees, line 100: Should it error rather than silently change human-driven config?

garethrees (commented):
Don't have a strong opinion and haven't reviewed thoroughly; left a thought, but basically – up to you.

script/rails-deploy-before-down
gbp: Error here I was avoiding was when the `config/rails_env.rb` doesn't exist. I could prefix this line with `[ -f config/rails_env.rb ] &&` instead.

garethrees: I'm not sure what makes most sense for this one – happy to go with whichever you think is clearest

PR #7592, opened 2023-02-17
Reduce default visibility of sign-in tracking

garethrees:
IP addresses are quite identifying, and in WhatDoTheyKnow's privacy policy [1] we say that we don't ordinarily look at them while administering the site.

Making it an active decision by an admin to reveal these preserves this position of default privacy, but makes it much easier to investigate issues without developer involvement.

[1] https://www.whatdotheyknow.com/help/privacy#our_logging

Revealing on a user page:

https://user-images.githubusercontent.com/282788/219610817-38c0efdf-8f0e-48e1-b0e6-ab08df9a152b.mov

Revealing on the main sign-ins list. Note that you have to reveal on every page-load, so have to click as you paginate. Hopefully we won't need to use this enough for it to be too annoying.

https://user-images.githubusercontent.com/282788/219612211-385b1b72-af3b-4611-9aa7-990a64fdf626.mov

Enabled, with sign ins recorded (click to reveal):

Screenshot 2023-02-16 at 17 56 57

Sign ins revealed:

Screenshot 2023-02-16 at 17 57 03

Enabled, but no sign ins recorded for a user:

Screenshot 2023-02-16 at 17 57 07

Disabled, but with previously recorded sign ins:

Screenshot 2023-02-16 at 17 59 34

Disabled:

Screenshot 2023-02-16 at 17 59 23

### DIFF
--- a/app/views/admin/users/_sign_in_table.html.erb
+++ b/app/views/admin/users/_sign_in_table.html.erb
@@ -10,38 +10,63 @@
   <% end %>
 
   <% if sign_ins.any? %>
-    <table class="table table-condensed table-striped">
-      <% sign_ins.each do |sign_in| %>
-        <tr class="<%= cycle('odd', 'even') %>">
-          <td><%= user_both_links(sign_in.user) %></td>
+    <div class="accordion" id="sign-ins-accordion">
+      <div class="accordion-group">
+        <div class="accordion-heading">
+          <p>
+            Sign in tracking records the IP address and date of sign in, as well
+            as a geolocation estimation, and links this IP address to sign ins
+            from other user accounts using the same IP address. This can be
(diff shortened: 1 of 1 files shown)


garethrees:
Merged as 8da34dbe9a4963a571d85d2884bb3837a627c505

RichardTaylor:
While this technically and literally results in compliance with the privacy policy statement: "It will not normally be seen by anyone" it isn't far off highlighting black text in black which we generally decry as a means of redaction. 

Hiding the data is better than not doing so, it stops an admin from accidentally remembering / noticing an IP associated with different accounts and making a connection, but we should be careful not to think that this feature for reducing the visibility of this material substantively controls access to this data.

PR #7598, opened 2023-02-20
Rubocop style linting

alexander-griffen:
## Relevant issue(s)
N/A
## What does this do?
The following RuboCops:
- [AndOr](https://rubystyle.guide#no-and-or-or)
- [NumericLiteralPrefix](https://rubystyle.guide#numeric-literal-prefixes)
- [TrailingCommaInArrayLiteral](https://rubystyle.guide#no-trailing-array-commas)
- IfInsideElse
- [PreferredHashMethods](https://rubystyle.guide#hash-key)
- ZeroLengthPredicate
- [Lambda](https://rubystyle.guide#lambda-multi-line)
- [RedundantBegin](https://rubystyle.guide#begin-implicit)

### DIFF
--- a/app/models/holiday_import.rb
+++ b/app/models/holiday_import.rb
@@ -66,24 +66,22 @@ class HolidayImport
   end
 
   def populate_from_ical_feed
-    begin
-      cal_file = open(ical_feed_url)
-      cal_parser = Icalendar::Parser.new(cal_file)
-      cals = cal_parser.parse
-      cal = cals.first
-      unless cal
-        errors.add(:ical_feed_url, "Sorry, there's a problem with the format of that feed.")
-        return
-      end
-      cal.events.each { |cal_event| populate_from_ical_event(cal_event) }
-    rescue Errno::ENOENT, Exception => e
-      if e.message == 'Invalid line in calendar string!'
-        errors.add(:ical_feed_url, "Sorry, there's a problem with the format of that feed.")
-      elsif e.message =~ /^No such file or directory/
-        errors.add(:ical_feed_url, "Sorry we couldn't find that feed.")
-      else
-        raise e
-      end
+    cal_file = open(ical_feed_url)
+    cal_parser = Icalendar::Parser.new(cal_file)
+    cals = cal_parser.parse
+    cal = cals.first
+    unless cal
+      errors.add(:ical_feed_url, "Sorry, there's a problem with the format of that feed.")
+      return
+    end
+    cal.events.each { |cal_event| populate_from_ical_event(cal_event) }
+  rescue Errno::ENOENT, Exception => e
+    if e.message == 'Invalid line in calendar string!'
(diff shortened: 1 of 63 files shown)


app/controllers/concerns/fragment_cachable.rb
```diff
@@ -37,6 +37,6 @@ def foi_fragment_cache_write(key_path, content)
     File.atomic_write(key_path) do |f|
       f.write(content)
     end
-    FileUtils.chmod 0644, key_path
+    FileUtils.chmod 0o644, key_path
```
garethrees, line 40: Hmm I don't quite understand this check but -1 without exploring this more.

`FileUtils.chmod 0644` is much clearer than `FileUtils.chmod 0o644` to me.

Some of the changes where we're parsing dates don't look right.

app/models/holiday_import.rb
```diff
@@ -95,14 +95,14 @@ def populate_from_ical_event(cal_event)
...
+    
       holiday_info = Holidays.between(start_date, end_date, @country_code.to_sym, :observed)
       holiday_info.each do |holiday_info_hash|
         holidays << Holiday.new(description: holiday_info_hash[:name],
                                 day: holiday_info_hash[:date])
       end
     rescue Holidays::InvalidRegion
       []
-    end
+
```
garethrees, line 106: Where it removes these redundant begin/end blocks, the contents should also be shifted left to get back to two-spacing.

Simple example:

```rb
# BEFORE
def foo
  begin
    bar
  rescue SomeError
    # do something
  end
end

# AFTER AUTOCORRECT
def foo
    bar
  rescue SomeError
    # do something
end

# WHAT WE WANT
def foo
  bar
rescue SomeError
  # do something
end
```

alexander-griffen: This has been changed and squashed into that commit

lib/mail_handler/backends/mail_backend.rb
```diff
@@ -327,17 +325,15 @@ def _get_attachment_leaves_recursive(part, within_rfc822_attachment, parent_mail
                                                                parent_mail)
             end
           end
-        else
+        elsif part.rfc822_attachment
           # Add all the parts of a decoded attached message
-          if part.rfc822_attachment
-            leaves_found += _get_attachment_leaves_recursive(part.rfc822_attachment,
+          leaves_found += _get_attachment_leaves_recursive(part.rfc822_attachment,
```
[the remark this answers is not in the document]
garethrees: The `else` just needs to be indented -2 to the left

app/controllers/concerns/fragment_cachable.rb
alexander-griffen: The commit has been removed, and the cop disabled (for now).

gbp (approved):
One fixup for indentation for the `IfInsideElse` changes

PR #7602, opened 2023-02-22
Add comment creation interval rate limit

garethrees:
Allow admins to always be able to create comments.

Add comment creation interval rate limit.

Calculates whether too many comments are being created too quickly.

We don't want to allow a user to sign up and smash through their daily
limit in 5 minutes, then create a new account to do the same, and so on.

This ensures that the allocation is used sensibly. It hopefully won't
affect the majority of genuine users.

Uses the existing rate limiting mechanism for the daily cap, with an extra line on the rendered page to account for this addition.

Screenshot 2023-02-22 at 14 32 37

### DIFF
--- a/app/models/comment.rb
+++ b/app/models/comment.rb
@@ -24,6 +24,19 @@ class Comment < ApplicationRecord
   include Rails.application.routes.url_helpers
   include LinkToHelper
 
+  DEFAULT_CREATION_RATE_LIMITS = {
+    1 => 2.seconds,
+    2 => 5.minutes,
+    4 => 30.minutes,
+    6 => 1.hour
+  }.freeze
+
+  cattr_accessor :creation_rate_limits,
+                 instance_reader: false,
+                 instance_writer: false,
+                 instance_accessor: false,
+                 default: DEFAULT_CREATION_RATE_LIMITS
+
   strip_attributes allow_empty: true
 
   belongs_to :user,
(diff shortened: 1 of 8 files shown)


app/models/comment.rb
```diff
@@ -87,6 +94,15 @@ def self.find_existing(info_request_id, body)
     end
   end
 
+  def self.exceeded_creation_rate?(comments)
+    comments = comments.reorder(created_at: :desc)
+
+    CREATION_RATE_INTERVALS.any? do |limit, duration|
+      return false if comments.size < limit
+      comments.limit(limit).all? { |comment| comment.created_at > duration.ago }
```
gbp, line 102:
```suggestion
      comments.where(created_at: duration.ago..).size >= limit
```

app/models/comment.rb
```diff
@@ -24,6 +24,13 @@ class Comment < ApplicationRecord
   include Rails.application.routes.url_helpers
   include LinkToHelper
 
+  CREATION_RATE_INTERVALS = {
+    1 => 2.seconds,
+    2 => 5.minutes,
+    4 => 30.minutes,
+    6 => 1.hour
+  }
```
gbp, line 32: Not clear what the key/values represent here. How about making this an array of hashes with keys `limit` & `duration`

gbp (changes requested):
This will be good. Query could be optimised slightly.

app/models/comment.rb
garethrees: Doh, much better!

app/models/comment.rb
garethrees: Renamed to `DEFAULT_CREATION_RATE_LIMITS` (and set via a `cattr_accessor`).

I think the improved constant name should help – I read it as "Limit of 2 records in 5 minutes". I think this makes a clearer definition than adding extra keys, but happy to reconsider in https://github.com/mysociety/alaveteli/issues/7620.

gbp (approved):

PR #7610, opened 2023-02-27
Rubocop style

alexander-griffen:
## Relevant issue(s)
N/A
## What does this do?
This PR goes through the majority of the Style cops that need changing. Most of these were autocorrectable.
(It's also a very long PR, that should have been multiple smaller ones).


gbp (approved):
Nice. I have pushed a few fixups for some autocorrected code for some indentation and do/end block styles to save adding more issues.

PR #7617, opened 2023-03-01
Show user messages count in admin UI

garethrees:
Gives us a quick way of seeing if a user has sent any user messages.

* Add User#user_messages counter cache
* Show User#user_messages_count in the Admin UI

Screenshot 2023-03-01 at 11 24 58

Need to run the following on deployment to prime our counter caches:

```ruby
UserMessage.find_each { |msg| User.reset_counters(msg.user.id, :user_messages) }
```

Don't need to worry about re-users as they won't have been recording user messages before the addition of the counter cache.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -57,7 +57,8 @@ class User < ApplicationRecord
          after_remove: :assign_role_features
   strip_attributes allow_empty: true
 
-  admin_columns exclude: [:otp_secret_key]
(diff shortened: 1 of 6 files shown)


garethrees:
> Need to run the following on deployment to prime our counter caches:

```rb
UserMessage.find_each { |msg| User.reset_counters(msg.user.id, :user_messages) }
```

I've done this for WDTK.

PR #7619, opened 2023-03-01
[#7618] Fix project request scopes

### LINKED ISSUE #7618: Projects - requests appearing as extracted when they really aren't
When a request has been added to multiple projects it can appear as extracted for all projects but in reality it has only been done once.

This means the extract progress bar for the project is incorrect and the queue of extractable requests is empty so it's not possible to extract data from the request.

gbp:
## Relevant issue(s)

Fixes #7618

## What does this do?

Fix project request scopes

Check the project submissions belong to the current project when calling `Project#extractable` or `Project#extracted`.

## Why was this needed?

This fixes an issue where requests which have been added to more than one project appears to be extracted due to being processed in the other project.

### DIFF
--- a/app/models/project/info_request_extension.rb
+++ b/app/models/project/info_request_extension.rb
@@ -15,14 +15,25 @@ class Project
     end
 
     def extractable
-      where(described_state: EXTRACTABLE_STATES).
+      scope = where(described_state: EXTRACTABLE_STATES).
         left_joins(:extraction_project_submissions).
-        where(project_submissions: { id: nil }).
         classified
+
+      scope.where(project_submissions: { id: nil }).or(
+        scope.where.not(project_submissions: { project: project })
+      )
     end
 
     def extracted
-      joins(:extraction_project_submissions).distinct
+      joins(:extraction_project_submissions).
+        where(project_submissions: { project: project }).
+        distinct
+    end
+
+    private
+
+    def project
+      proxy_association.owner
     end
   end
 end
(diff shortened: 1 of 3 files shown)


app/models/project/info_request_extension.rb
```diff
@@ -14,15 +14,20 @@ def classified
       where(awaiting_description: false)
     end
 
-    def extractable
-      where(described_state: EXTRACTABLE_STATES).
+    def extractable(project)
```
garethrees: Having to go `@project.info_requests.extractable(@project)` is a bit verbose.

Is there a reason we need the argument here? I assume it's added due to this module being used as an association extension.

I think we can use one of the `proxy_association`[^1] methods to get at the current `project` instance, so we'd end up with:

```ruby
# …
# TODO: figure out which proxy_association method is appropriate instead of placeholder "foo"
scope.where(project_submissions: { id: nil }).or(
  scope.where.not(project_submissions: { project: proxy_association.foo })
)
# …
```

I forget whether it's `#owner`, `#target`  or `#reflection` that's appropriate in this case.

[^1]: https://guides.rubyonrails.org/association_basics.html#association-extensions

garethrees (commented):
Haven't run, but the issue description / spec make sense and the fix looks good in principle. Left a comment that would clean up the implementation and result in a smaller diff if what I think will work works.

app/models/project/info_request_extension.rb
gbp: Yes, I came to the same conclusion when I saw the failing specs due to other calls to these method I hadn't updated. Hence the fixup using `@assoication.owner` but will look to replace with `proxy_assoication.onwer`.

Issue #6589, opened 2021-10-20
Import case studies from blog


garethrees:
In general we feel we should make more use of the case studies we write for our blog (https://github.com/mysociety/alaveteli-project/issues/143). 

One way of doing this would be to give them more prominence within the site itself (as a more "manual" version of https://github.com/mysociety/alaveteli/issues/6253).

This could/should be combined with adding more illustration and graphical content (https://github.com/mysociety/alaveteli/issues/6586) by accompanying the posts with a small thumbnail.

As it stands, we only import a limited number of posts via the blog feed (https://github.com/mysociety/alaveteli/issues/3663) and while the link to the blog is in the main navigation, I imagine it's not massively frequented.

## MVP

We have some [custom code on AskTheEU](https://github.com/mysociety/asktheeu-theme/blob/7da62fc251e533a2248ed8c78720e2c6b76ccf47/lib/views/general/frontpage.html.erb#L34-L46) to import the blog posts to the front page. At a minimum we should build this into Alaveteli.

![Screenshot 2021-10-20 at 16 35 39](https://user-images.githubusercontent.com/282788/138124792-911751d7-a703-45e5-8e13-c68da545f781.png)

## Blog posts everywhere

I also want to be able to dynamically pull in blog posts in appropriate places. For example, on [`/pro/pages/researchers`](https://www.whatdotheyknow.com/pro/pages/researchers) I might want to import case studies tagged `#researchers`.

On authority pages, we might key the imported posts based on the public body category or tags. For example, when visiting a "school" body, the case study about [asbestos in schools](https://www.mysociety.org/2016/12/07/freedom-of-information-and-asbestos-in-schools/) might be pulled in.

![Screenshot 2022-12-13 at 11 24 31](https://user-images.githubusercontent.com/282788/207305381-ec5e8853-3a41-44b0-a462-5bf539308ccd.png)

## Pulling in tags

We should build a generic helper that imports posts with given tags. This should only attempt to fetch content if the `BLOG_FEED` is configured.

```ruby
blog_posts(tagged: %w[case-study campaigning])
```

Let's start with fetching the blog title, url, thumbnail and an excerpt of the body content.

In the school example above, the code might be like:

```erb
<!-- a random post that matches the body category -->
<div class="sidebar">
  <% blog_posts(tagged: @public_body.categories.map(&:category_tag)).sample(1).each do |post| %>
    <%= render partial: 'blog/post_sidebar', post: post %>
  <% end %>
</div>

[...]

PR #7632, opened 2023-03-09
[#6589] Blog refactor

gbp:
## Relevant issue(s)

Connected to #6589

## What does this do?

Refactor loading of blog items to happen in a class rather than controller

## Why was this needed?

To continue with the development of the #6589

### DIFF
--- a/app/controllers/general_controller.rb
+++ b/app/controllers/general_controller.rb
@@ -29,9 +29,7 @@ class GeneralController < ApplicationController
 
   # Display blog entries
   def blog
-    if AlaveteliConfiguration.blog_feed.empty?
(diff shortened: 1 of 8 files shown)


garethrees (commented):
This is a huge improvement!!

Could take this a step further and pass `AlaveteliConfiguration` to the initializer, which would reduce some of the necessity to stub calls in the specs:

```rb
before_action :set_blog, only: [:blog]

def blog
  raise(ActiveRecord::RecordNotFound, "Page not enabled") unless @blog.enabled?
  # …
end

def set_blog
  @blog ||= Blog.new(AlaveteliConfiguration)
end
```

Either way, such a nice improvement! Love it! 😍

PR #7633, opened 2023-03-09
Adding localisation to resent notice

schlos:
## What does this do?

Adding localisation for resent notices.

## Why was this needed?

For sites in languages other than English.

## Notes to reviewer

I've merged separate strings before/after ifs into larger sentence in each if which makes it easier to translate and to have necessary context.


gbp:
Thanks @schlos I've opened another PR #7653 with some additional changes. 1. updated the plain text template. 2. Combined the "using new contact address" message 3. fixed spacing.

PR #7637, opened 2023-03-13
Add option for embedded background jobs

gbp:
## What does this do?

Add option for embedded background jobs

## Why was this needed?

This requires Sidekiq 7 and could make it easier for re-users to run the
job queue on smaller VPSs where resources might be more constrained.


gbp:
Unless there are issues with running sidekiq as a daemon we're not going to offer this as an option.

PR #7642, opened 2023-03-16
[#6589] Adds blog post model

gbp:
## Relevant issue(s)

Connected to #6589 
Requires #7632

## What does this do?

Adds blog post model for storing posts in the database

## Why was this needed?

Can't import tags from the WordPress blog feed so we're going to add an admin UI and make the posts taggable.


garethrees (approved):
haven't run, but lgtm

PR #7645, opened 2023-03-16
[#6589] Blog case studies

### LINKED ISSUE #6580: Drop down options do not work on WDTK when using Internet Explorer
Reported by a user (https://mail.google.com/mail/u/1/#inbox/FMfcgzGlkFnGMJLCNVQQjdfbdlhQhcxC) to the WDTK support team and confirmed by myself.

When logged in using Internet Explorer, any drop-down menu within the site will not open.
- User menu (for log out etc)
- Any of the admin menus
- Language switcher
Are the three examples I have readily found and tested. There might be more.

The specific issue caused for the user is the inability to log out.

So few people (IMO) use IE these days (which is being discontinued) that I'm not sure this is worth worrying about, but given an account security implication as noted above I felt it worthy of a ticket.

Checked in Microsoft Edge and it works OK.

gbp:
## Relevant issue(s)

Fixes #6580
Requires #7643 

## What does this do?

Renders blog case studies on authority and request pages.

## Implementation notes

Issue asks for case studies on the homepage, this isn't yet rendered but the `blog_posts_for_tag` helper doesn't require the `taggable` kwarg so can easily be added. Thinking this might be better in the theme. 

## Screenshots

image

image


garethrees:
> Issue asks for case studies on the homepage

Sorry, the case studies was more aimed at the requests and bodies. For the homepage, I think just render the most recent 4 articles as per ask the eu (in core, if blog enabled – not theme)

app/views/blog_posts/_case_studies.html.erb
```diff
@@ -0,0 +1,11 @@
+<% case_studies = blog_posts_with_tag('case-study', taggable: taggable) %>
```
garethrees, line 1: On having got a feel for this and with us now importing all posts in to the app, I think we should drop the requirement for having a 'case-study' tag for posts to appear in the sidebar. Basically, just have `blog_posts = related_blog_posts(taggable)`, which just gives any blog posts that match the taggable's tags.

app/views/blog_posts/_case_studies.html.erb
```diff
@@ -0,0 +1,11 @@
...
+
+<% if case_studies.any? %>
+  <div class="sidebar__section case_studies">
+    <h2><%= _('Related content from our blog') %></h2>
+
+    <div class="sidebar">
```
garethrees, line 7: We're already in the sidebar so don't think we need this element.

app/views/blog_posts/_case_studies.html.erb
```diff
@@ -0,0 +1,11 @@
+<% case_studies = blog_posts_with_tag('case-study', taggable: taggable) %>
+
+<% if case_studies.any? %>
+  <div class="sidebar__section case_studies">
```
garethrees, line 4:
```diff
- sidebar__section case_studies
+ sidebar__section blog_posts
```

app/views/blog_posts/_case_studies.html.erb
```diff
@@ -0,0 +1,11 @@
...
+    <h2><%= _('Related content from our blog') %></h2>
```
garethrees, line 5: Think we could tighten this up to "Related blog posts"

spec/helpers/blog_helper_spec.rb
```diff
@@ -3,6 +3,26 @@
 RSpec.describe BlogHelper do
   include BlogHelper
 
+  describe '#blog_posts_for_frontpage' do
+    subject(:posts) { blog_posts_for_frontpage }
+
+    before do
+      4.times { FactoryBot.create(:blog_post) }
```
garethrees, line 10: Should we make 5 to test that the spec below that's testing the limit isn't e.g. calling `.all`?

garethrees (commented):
Looking amazing!

Couple of minor tweaks and then good to go 🚀 ✍️ 🏷️

garethrees (commented):
One more thought, sorry. I guess more related to https://github.com/mysociety/alaveteli/pull/7662 / https://github.com/mysociety/alaveteli/pull/7642, but it's in using this that I spotted it…

I was looking through our blog posts and saw that some hadn't been imported ([this](https://www.mysociety.org/2022/11/30/have-your-say-on-the-icos-prioritisation-consultation/), [this](https://www.mysociety.org/2022/11/17/1-in-10-have-made-an-freedom-of-information-request-we-want-to-support-you/)) because they don't have the WhatDoTheyKnow category applied. This does happen from time to time, but it's fine; we just apply the category.

At present, this would cause a bit of an issue with the ordering by ID. I assume this would get added as the oldest post, rather than inserted at the correct place. I think we probably ought to add a `published_at` column to cache `pubDate`, and always ordering by that. We were [relying on this before](https://github.com/mysociety/alaveteli/blob/0.42.0.0/app/views/general/blog.html.erb#L9) so shouldn't be an unexpected key for a reusers feed to be providing.

app/views/admin/blog_posts/index.html.erb
```diff
@@ -0,0 +1,19 @@
...
+
+<h1><%= @title %></h1>
+
+<div class="btn-toolbar">
+  <div class="btn-group">
+    <%= link_to 'View tags', admin_tags_path(model_type: Blog::Post), class: 'btn' %>
+  </div>
+</div>
+
+<% if @blog_posts.size > 0 %>
```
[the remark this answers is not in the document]
gbp: Ah this is already on develop as its part of the admin UI.

spec/helpers/blog_helper_spec.rb
gbp: The 5th object is created within the `it` block.

garethrees: 🤦 doh

Issue #7651, opened 2023-03-20
Select authority is listing defunct authorities


schlos:
Similar to Defunct body is listed in batch request authorities list selection #6039, but for this instance, I'm seeing defunct bodies being listed on https://imamopravoznati.org/select_authority using search query "ministarstvo" (i.e. ministry).

![Screenshot from 2023-03-20 08-49-12](https://user-images.githubusercontent.com/1014134/226277768-b95b68b9-7701-4bb6-bcc6-b206e4a06fe8.png)

IPZ version info:

"alaveteli_git_commit": "4c2aa965c7c0e7f6c1c77d14b3ac711e66e12f07",
"alaveteli_version": "0.39.1.6",
"ruby_version": "2.5.1",

Reproducible also on WDTK:
https://www.whatdotheyknow.com/select_authority

![Screenshot from 2023-03-20 08-51-45](https://user-images.githubusercontent.com/1014134/226278349-4e117707-8f65-49e0-b22e-67df0b87c823.png)

WDTK version info:

"alaveteli_git_commit": "0b77ee54061004bbf63a5337f4f6396120fc156b",
"alaveteli_version": "0.42.0.0",
"ruby_version": "3.2.0",

Since on this screen we're tryign to submit a request, I don't see a reason why we should list defunct bodies. Only non-defunct bodies should be shown (i.e. those that can receive a request - not defunct, not internal admin testing, has email address).

garethrees:
These are listed here intentionally since users might not realise a "popular" authority has now been replaced.

Here's a situation:

1. User made an FOI request last year to "Another Public Body"
2. In the meantime, the body gets made defunct
3. A year later, the user wants to request the same data for this year

Right now, the user types "Another Public Body" as they did before, and then gets "notified" that they're defunct.

If we didn't list them, the user would be wondering why they aren't appearing. Did they make a typo? Has the site deleted that authority?

The current behaviour is clearer, though agree the presentation could be better, so closing in favour of https://github.com/mysociety/alaveteli/pull/7655 and https://github.com/mysociety/alaveteli/issues/721.

schlos:
Thanks Gareth.

I don't fully agree with you. Users that go to screen "select authority" are going actually to try to send a new request, so IMHO we should list here only active bodies that can receive requests.

We could always add a small text below or above result list saying:

[...]

Issue #7107, opened 2022-06-18
list active authorities first


confirmordeny:
It would be helpful to show active authorities before defunct authorities in lists of public authorities

For example on https://www.whatdotheyknow.com/body/list/department the Department for Business, Enterprise and Regulatory Reform which is defunct appears before the Department for Education that is still active.

confirmordeny:
This is similar to #721 but not the same (721 is concerned with search results).

HelenWDTK:
This issue has been automatically closed due to a lack of discussion or resolution for over 12 months.
Should we decide to revisit this issue in the future, it can be reopened.

Issue #721, opened 2012-11-23
Promote active bodies over those tagged defunct in search results


RichardTaylor:
- [ ] Currently active bodies ought be promoted over those tagged defunct in search results. 
- [x] Those bodies marked "defunct" should be marked "defunct" in the search results. 
- [x] The defunct tag currently prevents the "make a Freedom of Information request.." button appear. Perhaps instead of the button it should explain "you cannot make a Freedom of Information request as this body has been marked defunct" or similar.

garethrees:
Renamed to reflect outstanding issue.

PR #7655, opened 2023-03-21
Reduce prominence of defunct bodies in lists

garethrees:
The small "Defunct" label at the bottom of the authority listing is easily missable. This adds all tags as CSS classes to hook off, and applies styling to the case where an authority is defunct to make it clearer that it's "gone".

Not a "fix", but mitigates:

* https://github.com/mysociety/alaveteli/issues/721
* https://github.com/mysociety/alaveteli/issues/7107
* https://github.com/mysociety/alaveteli/issues/7651

## Screenshots

In search results:

Screenshot 2023-03-21 at 09 06 18

In listings:

Screenshot 2023-03-21 at 09 08 49

With theme applied:

Screenshot 2023-03-20 at 17 57 12

### DIFF
--- a/app/assets/stylesheets/responsive/_lists_style.scss
+++ b/app/assets/stylesheets/responsive/_lists_style.scss
@@ -15,6 +15,14 @@
   }
 }
 
+.body_listing.tag--defunct {
+  color: grey;
+
+  a {
+    color: grey;
+  }
+}
+
 .request_listing{
   .bottomline {
     background-position:top left;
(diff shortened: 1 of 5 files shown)


schlos:
Thanks! Grayed out title looks good and users can now easily see that there is something special with this body (i.e. small defunct text below description). Although this still does not solve ordering by "defunct" status. So if user searched for a term that has a lot of defunct bodies alphabetically on top, they will have to scroll a lot down to find an active body.

garethrees:
Yeah, search result ordering is a trickier piece of work so can be dealt with separately.

PR #7670, opened 2023-04-01
[#6384] Add admin button to delete annotations

### LINKED ISSUE #6384: Admin button to delete annotations
While site administrators can delete messages and correspondence threads from within the web application they cannot currently delete an an annotation. Admins can only hide annotations, not delete them. 

In very rare cases (related primarily to GDPR right to Erasure Cases, or similar cases where we can't justify holding, or don't want to hold, the content of hidden annotations) hiding an annotation is not sufficient, deletion is required.  

This probably occurs rarely enough that developer action on rare occasion might be satisfactory, and less effort than actually implementing this. Front-line support admins having this feature would allow quicker case resolution, involving fewer people, and fewer emails though.

HelenWDTK:
## Relevant issue(s)
Fixes #6384 

## What does this do?
Adds comment deletion to the request admin and comments admin pages, for single annotations and bulk deletion

## Why was this needed?
Sometimes annotations need deleting entirely to comply with GDPR. This stops admins needing to nag the devs to do it.

## Screenshots
Pro Admin timeline view:
Screenshot 2023-04-01 at 22 41 15
Normal admin view:
Screenshot 2023-04-01 at 22 40 09
Deletion event log:
Screenshot 2023-04-01 at 22 40 55
Comments admin page
Screenshot 2023-04-01 at 22 42 02
Request admin page:
Screenshot 2023-04-01 at 22 42 18
Single annotation edit page
Screenshot 2023-04-01 at 22 42 28
## Notes to reviewer

I think this is almost ready now, but there's undoubtedly room for improvement.

### DIFF
--- a/app/models/comment.rb
+++ b/app/models/comment.rb
@@ -194,6 +194,20 @@ class Comment < ApplicationRecord
     end
   end
 
+  def destroy_and_log_event(event: {})
+    return false unless destroy
+
+    info_request.log_event(
+      'destroy_comment',
+      event.merge(
+        comment: self,
+        comment_user: user,
+        comment_created_at: created_at,
+        comment_updated_at: updated_at
+      )
+    )
+  end
+
   def cached_urls
     [
       request_path(info_request),
--- a/app/controllers/admin_comment_controller.rb
+++ b/app/controllers/admin_comment_controller.rb
@@ -58,6 +58,18 @@ class AdminCommentController < AdminController
     end
   end
 
+  def destroy
+    @comment = Comment.find(params[:id])
+
+    if @comment.destroy_and_log_event(event: { editor: admin_current_user })
+      flash[:notice] = 'Comment successfully destroyed.'
+      redirect_to admin_request_url(@comment.info_request)
+    else
+      flash[:error] = 'Could not destroy the comment.'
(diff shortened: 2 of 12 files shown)


gbp:
Have rebased over current develop and fixed a few linting warnings.

garethrees:
Squashed this down and rebased. Seems to work pretty well.

gbp (changes requested):
The delete multiple button on the index view doesn't seem to work as the submit_tag to going to the wrong controller action.

gbp:
I have refactored the event creation so that 1. we create it in the model instead of the controller 2. store the same params we did when we manually deleted comments on WDTK).

Also stripped out the "delete selected" button for now to get the ability to individual comments merged sooner.

app/models/comment.rb
```diff
@@ -194,6 +194,20 @@ def hide(editor:)
     end
   end
 
+  def destroy_and_log_event(event: {})
+    return false unless destroy
+
+    info_request.log_event(
```
garethrees, line 200: Let's verify that this results in the search index being purged. I think it should, since it should destroy the previous comment events, which causes them to be marked for purging. Still, worth running through manually to be sure.

PR #7678, opened 2023-04-04
Alter project download to include requests with no extractions

### LINKED ISSUE #7669: `Project::Export` should include all requests in the project
At present we only include requests with extractions in the CSV data download of projects.

https://github.com/mysociety/alaveteli/blob/55b04c3e1820bd38a72c35cbf021ff68dd478d5a/app/models/project/export.rb#L16-L20

It would be [useful](https://groups.google.com/a/mysociety.org/g/alaveteli/c/F97HPM-Mz_0/m/sB0kYCtTBAAJ) for all requests in the project to get included, so that e.g. the classification status (and classifier) is present. This would also allow the export to be useful before extraction happens (i.e. to see classification/response progress).

Where there's no extraction data for a request, we can just render empty values.

This might just be as simple as removing the `.extracted` call, but not sure how `Project::Export::InfoRequest` would cope with a request with no extractions.

Related to https://github.com/mysociety/alaveteli/issues/7547.

alexander-griffen:
## Relevant issue(s)
Fixes #7669
## What does this do?
This means that requests which haven't been replied to by the authority are still included in project export.
## Why was this needed?
Requested by CEUK
## Implementation notes
Implemented as suggested in linked issue
## Notes to reviewer
I think this works as expected, but had difficulty figuring out how best to test it against various types/configurations of Projects

### DIFF
--- a/spec/models/project/export_spec.rb
+++ b/spec/models/project/export_spec.rb
@@ -11,11 +11,11 @@ RSpec.describe Project::Export do
     let(:info_request_b) { instance_double('InfoRequest') }
 
     before do
-      allow(project).to receive_message_chain(:info_requests, :extracted).
+      allow(project).to receive_message_chain(:info_requests).
         and_return([info_request_a, info_request_b])
     end
 
-    it 'individualy exports info requests' do
+    it 'individually exports info requests' do
       expect(Project::Export::InfoRequest).to receive(:new).
         with(project, info_request_a).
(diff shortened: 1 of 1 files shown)


garethrees:
Just removing the review requests until the build is passing at least.

gbp (approved):
This looks to be working as expected now. Nice one.

PR #7681, opened 2023-04-05
Attachment background processing

gbp:
## Relevant issue(s)

https://github.com/mysociety/whatdotheyknow-private/issues/207

## What does this do?

Move attachment masking and censor rule processing into a background job.

## Why was this needed?

Running this process inside the web request response sometimes times out.

## Implementation notes

We probably want to consider the message we display on the `done` view. Will we be deleting the stored masked files now they are in `ActiveStorage`, could they instead, use are existing rake tasks to be mirrored to S3 and served from there? 

## Screenshots

processing
download

### DIFF
--- a/app/controllers/attachment_masks_controller.rb
+++ b/app/controllers/attachment_masks_controller.rb
@@ -1,0 +1,36 @@
+##
+# Controller to process FoiAttachment objects before being served publicly by
+# applying masks and censor rules.
+#
+class AttachmentMasksController < ApplicationController
+  before_action :set_no_crawl_headers
+  before_action :find_attachment
+  before_action :ensure_attachment, :ensure_referer
+
+  def wait
+    if @attachment.masked?
+      redirect_to params[:referer]
+
+    else
+      FoiAttachmentMaskJob.perform_later(@attachment)
+    end
+  end
+
+  private
+
+  def set_no_crawl_headers
+    headers['X-Robots-Tag'] = 'noindex'
+  end
+
+  def find_attachment
+    @attachment = GlobalID::Locator.locate_signed(params[:id])
+  end
+
+  def ensure_attachment
+    raise ActiveRecord::RecordNotFound unless @attachment
+  end
+
+  def ensure_referer
+    raise RouteNotFound unless params[:referer].present?
+  end
+end
--- a/app/controllers/attachment_masks_controller.rb
+++ b/app/controllers/attachment_masks_controller.rb
@@ -9,11 +9,25 @@ class AttachmentMasksController < ApplicationController
 
   def wait
     if @attachment.masked?
-      redirect_to params[:referer]
+      redirect_to done_attachment_mask_path(
+        id: @attachment.to_signed_global_id,
+        referer: params[:referer]
+      )
 
     else
       FoiAttachmentMaskJob.perform_later(@attachment)
     end
+  end
+
+  def done
+    unless @attachment.masked?
+      redirect_to wait_for_attachment_mask_path(
+        id: @attachment.to_signed_global_id,
+        referer: params[:referer]
+      )
+    end
+
+    @show_attachment_path = params[:referer]
   end
 
   private
--- a/app/jobs/foi_attachment_mask_job.rb
+++ b/app/jobs/foi_attachment_mask_job.rb
@@ -1,0 +1,44 @@
+##
+# Job to apply masks and censor rules to FoiAttachment objects. Masked file will
+# be stored as FoiAttachment#file ActiveStorage association.
+#
+# Example:
+#   FoiAttachmentMaskJob.perform(FoiAttachment.first)
+#
+class FoiAttachmentMaskJob < ApplicationJob
+  queue_as :default
+  unique :until_and_while_executing, on_conflict: :log
+
+  attr_reader :attachment
+
+  delegate :incoming_message, to: :attachment
+  delegate :info_request, to: :incoming_message
+
+  def perform(attachment)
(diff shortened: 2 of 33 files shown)


app/controllers/attachment_masks_controller.rb
```diff
@@ -0,0 +1,38 @@
...
+  def find_attachment
+    @attachment = GlobalID::Locator.locate_signed(params[:id])
+  end
+
+  def ensure_attachment
+    raise ActiveRecord::RecordNotFound unless @attachment
+  end
+
+  def ensure_referer
+    raise RouteNotFound unless params[:referer].present?
```
garethrees, line 52: Is this supposed to prevent direct access of the `/attachments/:global_id` URL? I can still access it by pasting it into a new private browsing window. Is that what's expected?

app/views/attachment_masks/done.html.erb
```diff
@@ -0,0 +1,12 @@
+<h1><%= _('Attachment available for download.') %></h1>
+
+<p>
+  <%= _('The attachment you\'ve requested has now been processed and is ' \
+        'available for download. The processed attachment will be retained ' \
+        'for up to 7 days. Once 7 days has lasped then the cache might be ' \
+        'cleared again.') %>
```
garethrees, line 7: Where's the 7 days coming from here? I don't think _Alaveteli_ purges attachments (though we've manually hacked this in in the case of some installs we host). I'm not sure we'll need to do that with this setup, so maybe making this  as simple as:

```
<%= _('The attachment has now been processed and is available for download.') %>
```

should be fine. Also worth setting the `@title` instance variable.

app/controllers/attachment_masks_controller.rb
```diff
@@ -11,13 +11,29 @@ def wait
...
+        id: @attachment.to_signed_global_id,
+        referer: params[:referer]
+      )
 
     else
       FoiAttachmentMaskJob.perform_later(@attachment)
     end
   end
 
+  def done
```
garethrees, line 24: Looking at the commit message:

> Once an attachment masked file is available … redirecting to the download.

Redirecting to the actual attachment seems more preferable than requiring an extra click to do so…

> This is so we aren't left on the `wait` view which looks like nothing has happened.

…but I'm not sure I understand the conditions where this happens.

I'd have thought we could do:

```rb
def wait
  if @attachment.masked_file.attached?
    # If we have a cached file, render it
    redirect_to attachment_path(@attachment)
  else
    # otherwise, fire off a job to cache it, render the waiting page, and
    # keep refreshing until `@attachment.masked_file.attached?` returns
    # true
    FoiAttachmentMaskJob.perform_later(@attachment)
    render 'waiting'
  end
end
```

One to demo perhaps.

garethrees (commented):
🎉 omg, huge improvements.

Couple of things to discuss but all really nice in principle.

app/controllers/attachment_masks_controller.rb
gbp: After the `redirect_to attachment_path(@attachment)` the users browser would show the previous view as this will be treated link any other download link.

We could automatically download via JS or meta tag on the `done` view (and change the wording to say "Your download will start automatically...").

Happy to demo

app/views/attachment_masks/done.html.erb
gbp: I'm not sure, I will re-word.

app/models/foi_attachment.rb
```diff
@@ -34,6 +34,7 @@ class FoiAttachment < ApplicationRecord
              inverse_of: :foi_attachments
 
   has_one_attached :file, service: :attachments
+  has_one_attached :masked_file, service: :masked_attachments
```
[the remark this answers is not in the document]
gbp: This has been refactored so file and masked_file the same thing now.

app/jobs/foi_attachment_mask_job.rb
```diff
@@ -0,0 +1,43 @@
...
+  unique :until_and_while_executing, on_conflict: :log
+
+  attr_reader :attachment
+
+  delegate :incoming_message, to: :attachment
+
+  def perform(attachment)
+    @attachment = attachment
+
+    body = AlaveteliTextMasker.apply_masks(
```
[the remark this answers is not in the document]
gbp: Since we've merged `#masked_file` into `#file` I'm going to keep this in a job, at least for now, as having this in `FoiAttachment#body` isn't really optimal as this is called throughout the codebase already - eg indexing, zip downloads. These places currently do there own masking but in the future we can refactor this.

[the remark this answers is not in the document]
garethrees: I didn't necessarily mean to inject it into the current method call chain; I just meant extract the logic into `FoiAttachment` so that it's not as buried inside the job:

```rb
class FoiAttachmentMaskJob
  def perform(attachment)
    attachment.apply_masks
  end
end

class FoiAttachment
  # `self` calls just included for clarity where they replace `attachment` from
  # the current implementation within `FoiAttachmentMaskJob`
  #
  # This only gets called via `FoiAttachmentMaskJob`
  def apply_masks
    body = AlaveteliTextMasker.apply_masks(
      self.unmasked_body,
      self.content_type,
      masks
    )

    if attachment.content_type == 'text/html'
      body =
        Loofah.scrub_document(body, :prune).
        to_html(encoding: 'UTF-8').
        try(:html_safe)
    end

    self.update(body: body, masked_at: Time.zone.now)
  end
end
```

Can do this later, but just wanted to clarify.

PR #7689, opened 2023-04-11
Remove duplicate method

gbp:
## What does this do?

Remove duplicate method

## Why was this needed?

Looks like this was meant to be removed in #7452.

## Notes to reviewer

@garethrees I think this was your original intent?


garethrees (approved):
oops, yeah 👍

PR #7690, opened 2023-04-11
Add previous comment feature in censor rule edit form

### LINKED ISSUE #4625: Make censor rule comments less annoying for administrators
Quite reasonably a comment is required when creating or editing a censor rule.

Often when implementing censor rules it's necessary to go through quite a bit of trial and improvement until a combination of rules, or a single rule, works. Having to enter comments with every edit becomes tedious. 

Often a series of related censor rules are created for the same reason - to implement one type of censoring eg. to remove personal information, or remove a users' name, or remove potential defamation. 

Pre-populating the reason/comment field with the last used comment might help; especially on an edit. 

On a related point admins can't see the a history for edited censor rules; so admins are creating a comment history which is invisible to them.

alexander-griffen:
## Relevant issue(s)
Fixes #4625 
## What does this do?
This change introduces a new JavaScript file, censor-rules.js, to handle the functionality for the admin censor rules section. Specifically, it adds a "Use previous comment" feature to the form for editing censor rules. 
## Why was this needed?
The purpose of this change is to streamline the editing process for admin users. By providing a "Use previous comment" option, admins can easily reuse the last edit comment as the reason for the current update, saving time and effort when making similar or related changes to censor rules.
## Implementation notes
This PR consists of the following changes:
 - Including the new censor-rules.js file in admin.js to make it available for use in the admin section of js files.
 - Creating the censor-rules.js file, where a click event listener is added to the "Use previous comment" link. When the link is clicked, the event handler retrieves the last edit comment from the associated text area's data attribute and sets the text area's value to the comment at the last edit.
 - Updating the _form.html.erb partial for the censor rule form. In the partial, a data attribute is added to the last_edit_comment text area to store the last_edit_comment value from the censor_rule object.

## Screenshots

image

### DIFF
--- a/app/assets/javascripts/admin.js
+++ b/app/assets/javascripts/admin.js
@@ -9,5 +9,6 @@
 //= require bootstrap-tooltip
 //= require admin/admin
 //= require admin/category-order
+//= require admin/censor-rules
 //= require admin/holidays
 //= require jquery_ujs
--- a/app/assets/javascripts/admin/censor-rules.js
+++ b/app/assets/javascripts/admin/censor-rules.js
@@ -1,0 +1,8 @@
+$(function () {
+  $('#js-use-previous-comment').click(function (e) {
+    var textArea = $($(this).data('target'));
+    var lastEditComment = textArea.data('last-edit-comment');
+    textArea.val(lastEditComment);
+    e.preventDefault();
+  })
+});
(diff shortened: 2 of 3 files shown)


garethrees:
Nice! Might be nicer for the link to be `class="btn"` since it's an action, but great work – this will be much appreciated!

gbp (approved):
Nice one, looks good to me. Just need the trailing newline in `censor-rules.js` then you're good to merge.

PR #7691, opened 2023-04-11
Config change to make docker development easier

alexander-griffen:
## Relevant issue(s)
N/A
## What does this do?
Removes the `file_watcher` variable in the development config file, and removes the `listen` gem from the dependencies.

[This change was also made in Rails 7.](https://github.com/rails/rails/pull/42985)
## Why was this needed?
Makes developing for Alaveteli in docker more efficient.


gbp (approved):
This works locally for me when either using Docker or not. Nice to remove a dependency.

PR #7694, opened 2023-04-14
Rubocop linting

alexander-griffen:
## Relevant issue(s)
N/A
## What does this do?
Changes to satisfy the following cops:
 - Lint/IneffectiveAccessModifier
 - Lint/ShadowingOuterVariable
 - Lint/ImplicitStringConcatenation

## Why was this needed?
Conforming to coding standards to help maintainability of Alaveteli code base

### DIFF
--- a/app/controllers/api_controller.rb
+++ b/app/controllers/api_controller.rb
@@ -244,18 +244,18 @@ class ApiController < ApplicationController
       render template: "api/request_events", formats: [:atom], layout: false
     elsif feed_type == "json"
       @event_data = []
-      @events.each do |event|
+      @events.each do |json_event|
 
-        request = event.info_request
+        request = json_event.info_request
         this_event = {
           request_id: request.id,
-          event_id: event.id,
-          created_at: event.created_at.iso8601,
-          event_type: event.event_type,
+          event_id: json_event.id,
+          created_at: json_event.created_at.iso8601,
+          event_type: json_event.event_type,
           request_url: request_url(request),
           request_email: request.incoming_email,
           title: request.title,
-          body: event.outgoing_message.body,
+          body: json_event.outgoing_message.body,
           user_name: request.user_name
         }
         this_event[:user_url] = user_url(request.user) if request.user
--- a/app/models/incoming_message.rb
+++ b/app/models/incoming_message.rb
@@ -235,7 +235,7 @@ class IncomingMessage < ApplicationRecord
     # http://www.whatdotheyknow.com/request/71/response/108
     # http://www.whatdotheyknow.com/request/police_powers_to_inform_car_insu
     # http://www.whatdotheyknow.com/request/secured_convictions_aided_by_cct
-    multiline_original_message = '(' + '''>>>.* \d\d/\d\d/\d\d\d\d\s+\d\d:\d\d(?::\d\d)?\s*>>>''' + ')'
+    multiline_original_message = '(>>>.* \d\d/\d\d/\d\d\d\d\s+\d\d:\d\d(?::\d\d)?\s*>>>)'
     text.gsub!(/^(#{multiline_original_message}\n.*)$/m, replacement)
 
     # On Thu, Nov 28, 2013 at 9:08 AM, A User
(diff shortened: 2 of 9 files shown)


app/models/mail_server_log.rb
```diff
@@ -256,7 +256,7 @@ def delivery_status
     DeliveryStatusSerializer.load(read_attribute(:delivery_status))
   end
 
-  private
+  private_class_method
```
gbp, line 259: Changing this is now causing `Lint/UselessAccessModifier` warning so I think the wrong change has been made here.

Looking at the warning:
```
app/models/mail_server_log.rb:288:3: W: Lint/IneffectiveAccessModifier: private (on line 259) does not make singleton methods private. Use private_class_method or private inside a class << self block instead.
  def self.create_mail_server_logs(emails, line, order, done)
  ^^^
```

It appears this is because the `self.create_mail_server_logs` is under the `private` keyword. This is because it is a class method (EG it starts with `self.`).

We should retain the `private` keyword and resolve the this warning in a different way. This [resource shows how](https://medium.com/rubycademy/3-ways-to-make-class-methods-private-in-ruby-64b970e54613) this can be done.

app/models/public_body.rb
```diff
@@ -911,7 +911,7 @@ def questions
     PublicBodyQuestion.fetch(self)
   end
 
-  private
+  private_class_method
```
gbp, line 914: As above

app/models/info_request.rb
```diff
@@ -1475,12 +1475,12 @@ def applicable_censor_rules
 
   def apply_censor_rules_to_text(text)
     applicable_censor_rules.
-      reduce(text) { |text, rule| rule.apply_to_text(text) }
+      reduce(text) { |uncensored_text, rule| rule.apply_to_text(uncensored_text) }
```
gbp, line 1478: For single line blocks I would lean towards using a single lowercase character for the variables. It's assumed that single line blocks are typically short and simple, so a longer or more descriptive variable name would not provide much benefit.

EG:

```suggestion
      reduce(text) { |t, rule| rule.apply_to_text(t) }
```

or 

```suggestion
      reduce(text) { |t, r| r.apply_to_text(t) }
```

lib/alaveteli_text_masker.rb
```diff
@@ -173,7 +173,7 @@ def apply_text_masks(text, options = {})
       memo.gsub(mask[:to_replace], mask[:replacement])
     end
 
-    censor_rules.reduce(text) { |text, rule| rule.apply_to_text(text) }
+    censor_rules.reduce(text) { |uncensored_text, rule| rule.apply_to_text(uncensored_text) }
```
gbp, line 176:
```suggestion
    censor_rules.reduce(text) { |t, rule| rule.apply_to_text(t) }
```

app/models/outgoing_message.rb
```diff
@@ -191,7 +191,7 @@ def body(options = {})
       info_request.try(:applicable_censor_rules) or []
     end
 
-    censor_rules.reduce(text) { |text, rule| rule.apply_to_text(text) }
+    censor_rules.reduce(text) { |uncensored_text, rule| rule.apply_to_text(uncensored_text) }
```
gbp, line 194:
```suggestion
    censor_rules.reduce(text) { |t, rule| rule.apply_to_text(t) }
```

gbp (changes requested):
There are a few changes needed here.

app/models/incoming_message.rb
```diff
@@ -235,7 +235,7 @@ def self.remove_quoted_sections(text, replacement = "FOLDED_QUOTED_SECTION")
     # http://www.whatdotheyknow.com/request/71/response/108
     # http://www.whatdotheyknow.com/request/police_powers_to_inform_car_insu
     # http://www.whatdotheyknow.com/request/secured_convictions_aided_by_cct
-    multiline_original_message = '(' + '''>>>.* \d\d/\d\d/\d\d\d\d\s+\d\d:\d\d(?::\d\d)?\s*>>>''' + ')'
+    multiline_original_message = '(>>>.* \d\d/\d\d/\d\d\d\d\s+\d\d:\d\d(?::\d\d)?\s*>>>)'
```
gbp, line 238: The commit message (and PR description) needs to be updated now.

app/controllers/api_controller.rb
```diff
@@ -244,18 +244,18 @@ def body_request_events
       render template: "api/request_events", formats: [:atom], layout: false
     elsif feed_type == "json"
       @event_data = []
-      @events.each do |event|
+      @events.each do |e|
```
gbp, line 247: Ah sorry if I wasn't clear, multi line `do`/`end` blocks shouldn't use a single lowercase letter. That was only for single line `{`/`}` blocks.

gbp (changes requested):

app/controllers/api_controller.rb
alexander-griffen: Oh, that makes sense!

app/models/mail_server_log.rb
```diff
@@ -296,6 +296,10 @@ def self.create_mail_server_logs(emails, line, order, done)
     end
   end
 
+  class << self
+    private :create_mail_server_logs
+  end
```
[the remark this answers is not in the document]
alexander-griffen: That makes sense, this is now implemented

gbp (approved):
Code changes are looking alright, if I'm being picky then I would say the commit messages aren't consistent, "Rubocop ...", "... Rubocop" & "... cop".

Good to merge once those are consistent with each other.

PR #7695, opened 2023-04-14
`FoiAttachment` cleanup

garethrees:
Couple of small refactorings I took the opportunity to make while reviewing https://github.com/mysociety/alaveteli/pull/7681.

### DIFF
--- a/app/models/foi_attachment.rb
+++ b/app/models/foi_attachment.rb
@@ -267,10 +267,6 @@ class FoiAttachment < ApplicationRecord
     end
   end
 
-  def has_google_docs_viewer?
-    AttachmentToHTML::Adapters::GoogleDocsViewer.viewable?(content_type)
-  end
-
   # Whether this type has a "View as HTML"
   def has_body_as_html?
     AttachmentToHTML.extractable?(self)
--- a/app/models/foi_attachment.rb
+++ b/app/models/foi_attachment.rb
@@ -241,23 +241,8 @@ class FoiAttachment < ApplicationRecord
     end
   end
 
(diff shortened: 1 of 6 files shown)


app/models/foi_attachment.rb
```diff
@@ -49,6 +49,32 @@ class FoiAttachment < ApplicationRecord
   BODY_MAX_TRIES = 3
   BODY_MAX_DELAY = 5
 
+  # rubocop:disable Style/LineLength
```
gbp, line 52: Commit message: `s/by reducing noise by/by reducing noise/`

app/models/foi_attachment.rb
```diff
@@ -298,4 +295,7 @@ def text_type?
     AlaveteliTextMasker::TextMask.include?(content_type)
   end
 
+  def extractable_to_html?
+    %w[text/plain application/rtf].include?(content_type)
```
gbp, line 299: Maybe not for now... it would be nice if these content types weren't hardcoded here as this could delegate to `AttachmentToHTML`. I believe this is `true` if `AttachmentToHTML.send(:adapter_for, foi_attachment) != AttachmentToHTML::Adapters::CouldNotConvert`

Aside: for PDF files this is `true` because `has_google_docs_viewe?` returns true, but we use `pdftohtml` instead of Google Doc viewer for PDF files.

gbp (approved):
Nice cleanup.

app/models/foi_attachment.rb
garethrees: Great suggestion – done!

gbp (approved):
This is great. Lets merge next week after the release.

Issue #285, opened 2011-11-10
Username change without changing history


RichardTaylor:
There are circumstances where we might want to change a username (effective from a particular point in time) and have the system use the original name for correspondence threads before that date. 

eg. Someone changes their name on marriage

eg. Someone changes their name for any other reason

hsenag:
I'm not too sure about the complexity of this one - at least, no really simple solution jumps out at me. What happens to correspondence threads in progress at the time of the request?

RichardTaylor:
Just a quick note to say we've had another case on WhatDoTheyKnow where this feature would have been useful.

henare:
We just had someone ask how they can change their name. They signed up using only their first name for privacy reasons but after making a request were no longer concerned about that so wanted to add their last name.

In this case they probably would have been fine just updating their display name and not the URL name.

RichardTaylor:
A user of WhatDoTheyKnow just got married and this feature would have been useful to deal with their subsequent request to change their name on WhatDoTheyKnow. We've changed their username for them but this leaves the history a little inconsistent with the new name where the username is shown but the old name in the body of the requests.

RichardTaylor:
There has been another case on WhatDoTheyKnow today where this feature would have been useful.

RichardTaylor:
Another case arose today - where someone who has made hundreds of requests on WhatDoTheyKnow under using a single word name, and had many rejected on the grounds of not providing a full name, wanted a second word added to their name.  Given the history of rejections just adding the second word to the user name could have made the presentation of existing correspondence misleading / hard to understand.  The option to change the name from a point in time would have made the decision on what to do easier.

RichardTaylor:
Another +1 from today. A user making requests under the name of an "organisation" wanted to change that organisation's name. 

We want to encourage organisations to use WhatDoTheyKnow to make requests and from time organisations do change their names so handling such events nicely would be good. 

The situation isn't terrible at the moment as the name used at the time of the request will probably still be in the text of the correspondence even if a user name is changed.

garethrees:
Maybe name should be added on a per-request basis (`InfoRequest#signature`?), with `User#name` being the pre-filled value. That would allow us (or users themselves) to more easily redact the signature, allow user account name changes with less disruption on history, and also allow pro users to use pseudonyms.

Haven't though too much about this, so just noting for when we get to it.

RichardTaylor:
[...]

PR #7708, opened 2023-04-20
[#285] Add OutgoingMessage#from_name prep

gbp:
## Relevant issue(s)

Connected to #285
Commits extracted from #7702 which are ready for review.

## What does this do?

Minor refactorings link to helpers and applying censor rules. These changes will make it easier to implement the #285.

### DIFF
--- a/app/helpers/link_to_helper.rb
+++ b/app/helpers/link_to_helper.rb
@@ -150,23 +150,25 @@ module LinkToHelper
     end
   end
 
-  def external_user_link(request, text)
+  def external_user_link(request, text = nil)
     if request.external_user_name
       request.external_user_name
     else
+      text ||= _("Anonymous user")
       link_to(text, help_privacy_path(anchor: 'anonymous'))
     end
   end
 
-  def external_user_link_absolute(request, text)
+  def external_user_link_absolute(request, text = nil)
     if request.external_user_name
       request.external_user_name
     else
+      text ||= _("Anonymous user")
       link_to(text, help_privacy_url(anchor: 'anonymous'))
     end
   end
 
-  def request_user_link_absolute(request, anonymous_text=_("Anonymous user"))
+  def request_user_link_absolute(request, anonymous_text = nil)
     if request.is_external?
       external_user_link_absolute(request, anonymous_text)
     else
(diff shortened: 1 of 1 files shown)


spec/models/outgoing_message_spec.rb
```diff
@@ -536,40 +536,6 @@ class TestError < StandardError; end
             to include("Please enter your letter requesting information")
         end
 
-        context "when a censor rule changes the default text" do
```
garethrees, line 539: Introduced in https://github.com/mysociety/alaveteli/commit/7619291e4cb7531b6d71ee6dec2dae3818c5933d; looks like it's related to this part of the conversation https://github.com/mysociety/alaveteli/pull/3270#issuecomment-225127010.

garethrees (commented):
Haven't run but had a read through and LGTM in principle. Couple of questions on some of the changes.

gbp:
I've dropped the commits with remaining issues (we will need to go into a more in depth refactoring here to clear up the remaining issues) so I can continue to make progress on https://github.com/mysociety/alaveteli/issues/285

app/models/outgoing_message.rb
```diff
@@ -198,8 +198,9 @@ def raw_body
     read_attribute(:body)
   end
 
-  def apply_masks(text, content_type)
```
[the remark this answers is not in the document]
gbp: Notes for future me; We're not applying all masks to outgoing messages. Only the email mask by calling `#remove_privacy_sensitive_things!` for the HTML display (when calling `#get_body_for_html_display`) this works differently to the same method on incoming messages. We could/should? unify this into a common model concern.

spec/models/outgoing_message_spec.rb
gbp: Notes for future me; still not convinced these specs are testing anything.

Why would the default letter need censor rules applying... for example with a censor rule doing `s/07912345678/00000000000/`, [this URL](http://0.0.0.0:3000/new/msw?default_letter=07912345678+alice@localhost.com) doesn't mask the number or email in the form textarea. Once the user previews the request the default letter isn't being used and the message body is like any other request - the number is redacted (by censor rule) and email masked as normal.

PR #7727, opened 2023-05-05
Independent account closure actions

### LINKED ISSUE #7646: Replace close and anonymise with independent account closure actions
Requires https://github.com/mysociety/alaveteli/pull/7611.

Instead of a single "close and anonymise" action, we should make the set of actions more composable. For example, a user may only want their account closed rather than anonymised.

Screenshot 2023-03-17 at 14 57 22

Each option should document what the action does inline in the UI. Might want to [fold the specifics in a `<details>` tag](https://github.com/mysociety/alaveteli/issues/4910#issuecomment-1130045202) if it starts getting too long.

For close and anonymise we currently state:

> This is disabled for banned users as the email address must be preserved to enforce the ban. First remove the ban text and then close and anonymise the account.

I think this is still true. We should allow the `#close` and `#anonymise` options when the user is banned, but disable `#erase`, as that's the one that will now remove the account email address.

When an account gets closed, we should ensure that no email gets sent to it (https://github.com/mysociety/alaveteli/issues/4603 is likely relevant here). We'll need to be careful about this becoming a rabbit hole. If it starts getting complicated, the easiest solution might be a callback along the lines of `before_action { return false if @recipient.closed? }` in the mailer methods.

Note that https://github.com/mysociety/alaveteli/issues/5052#issuecomment-1345525599 still requires the `User#close_and_anonymise` method – this issue is just to replace the _admin UI_ options.

alexander-griffen:
## Relevant issue(s)
Fixes #7646 
## What does this do?
This separates the 'Close and anonymise' action into three separate actions.
## Why was this needed?
More granular control for admins.
## Implementation notes
I slightly altered the `User::anonymise!` method to create a censor rule even if there were no info requests.

Additionally, I changed the `disable` criteria for the Erase button to be checking if the user was banned rather than suspended (as a closed account is considered suspended, but an account has to be closed to be erased).

## Screenshots
image


app/models/user.rb
```diff
@@ -445,8 +445,6 @@ def erase!
   end
 
   def anonymise!
-    return if info_requests.none?
```
gbp, line 448: This change is breaking some specs, in fact I'm not sure we need this commit at all. Was there a reason why this is needed I'm not seeing?

app/controllers/admin_users_account_suspensions_controller.rb
```diff
@@ -24,11 +24,7 @@ def set_suspended_user
   end
```
gbp, line 24: Do we need this controller/routes at all now. Hasn't it been replaced by the 3 new ones?

spec/controllers/admin_users_account_anonymising_controller_spec.rb
```diff
@@ -0,0 +1,55 @@
+require 'spec_helper'
+
+RSpec.describe AdminUsersAccountAnonymisingController do
+
```
gbp, line 4: The one style was to pad `describe`/`context` blocks with extra spaces - we're not doing that anymore. 
```suggestion
```

spec/controllers/admin_users_account_anonymising_controller_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+
+    context 'with invalid params' do
+      it 'renders a 404' do
+        expect {
+          post :create
+        }.to raise_error(ActiveRecord::RecordNotFound)
+      end
+    end
+
```
gbp, line 53:
```suggestion
```

gbp (changes requested):
I think we're almost there but its hard to see when there are a number of Rubocop warnings which needs addressing. If you could fix those up and take a look at the inline comments/suggestions I've left. Then ask for a re-review. Thanks.

app/models/user.rb
garethrees, line 448: > so that requests going forward are still censored

Ah, this isn't an intended behaviour (as requests would just get rejected by the authority as a name _must_ be present). 

Instead maybe we should disable anonymise button until the account is closed, in the same way that it must be closed before erasure?

app/controllers/admin_users_account_suspensions_controller.rb
alexander-griffen: I think this controller is still needed for the "Ban for Spamming" button before you get to the edit page.

app/models/user.rb
alexander-griffen: Removed - see comments below

alexander-griffen: Okie dokie! I've removed the removal (hah), and I'm going to implement the disabling until the account is closed. If I can figure out a way to do it that makes sense, I might try and put some explanation of that and the clsoed-before-erased need on the page also, so it doesn't look like buttons are disabled at random.

app/views/admin_user/edit.html.erb
```diff
@@ -28,55 +28,92 @@
...
-            <li>Makes a naive attempt to redact <tt>name</tt> from requests<sup>1</sup></li>
           </ul>
         </p>
+      </div>
 
-        <p>Closing the account:
+      <%= form_tag admin_users_account_closing_index_path(user_id: @admin_user.id), class: 'span3 form form-inline' do %>
+        <% submit_class = %w(btn btn-danger) %>
+        <%= submit_tag 'Close',
+          class: submit_class,
```
gbp, line 45: Should disable if account already closed:

```suggestion
        <% disabled = @admin_user.closed? %>
        <% submit_class = %w(btn btn-danger) %>
        <% submit_class << 'disabled' if disabled %>
        <%= submit_tag 'Close',
          class: submit_class,
          disabled: disabled,
```

app/views/admin_user/edit.html.erb
```diff
@@ -28,55 +28,92 @@
...
-          the name</em>, so its worth checking that this has done what you
-          expect. More Censor Rules may need to be added to the user if this
-          has not been sufficient.
-        </p>
       </div>
-      <%= form_tag admin_users_account_suspensions_path(user_id: @admin_user.id, close_and_anonymise: true), class: 'span3 form form-inline' do %>
-        <% disabled = @admin_user.suspended? || @admin_user.is_pro? %>
+
+      <%= form_tag admin_users_account_erasing_index_path(user_id: @admin_user.id), class: 'span3 form form-inline' do %>
+        <% disabled = @admin_user.banned? || @admin_user.is_pro? %>
```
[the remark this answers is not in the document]
gbp: Also need to add "This is disabled for users whose accounts have not been closed. To erase a user, first close the account." to the help text.

gbp (changes requested):
This is so almost there. I have added some suggestions for some changes which would be good to get done.

app/views/admin_user/edit.html.erb
```diff
@@ -106,11 +110,14 @@
...
           and anonymise the account.
+
+          This is disabled for users whose accounts have not been closed.
+          To erase a user, first close the account.
         </p>
       </div>
 
       <%= form_tag admin_users_account_erasing_index_path(user_id: @admin_user.id), class: 'span3 form form-inline' do %>
-        <% disabled = @admin_user.banned? || @admin_user.is_pro? %>
+        <% disabled = !@admin_user.closed? || @admin_user.is_pro? %>
```
gbp, line 120: Sorry. Not quite what I asked for here: https://github.com/mysociety/alaveteli/pull/7727#discussion_r1195040170

Banned users and closed users aren't the same thing. We should retain disabling this form for banned users too.

gbp (approved):
One last change then we can merge.

PR #7735, opened 2023-05-09
[#285] Add Friendly ID gem

gbp:
## Relevant issue(s)

Requires #7702 
Connected to #285

## What does this do?

- Adds the `friendly_id` gem
- Uses this gem over our own implementation for `User#url_name` generation
- Adds migration from the current column to the `FriendlyID::Slug` model.

## Why was this needed?

To allow user to change their names without breaking URLs. The `friendly_id` gem will store old slugs and correctly load the record if the old URL is hit.

## Implementation notes

Unfortunately we need to override some of `friendly_id` internal methods until we do the next release.


lib/tasks/temp.rake
```diff
@@ -1,2 +1,22 @@
 namespace :temp do
+  desc 'Migrate current user slugs into'
```
garethrees, line 2: Unfinished sentence?

garethrees (commented):
Looks good. Could probably do with a quick chat through how it all works, but seems to be working fine in practice 🐌

PR #7770, opened 2023-06-12
Search all email headers for info requests

### LINKED ISSUE #5988: Extract info request address from Received headers
When BCC replies are received the info request address can be contained in one of the `Received` headers we should attempt to extract these and delivery the response to the correct info request instead of the holding pen.

Example mail which should match carlos@example.net:
```ruby
Mail.new(<<~EOF)
  From: "FOI Person" <foiperson@localhost>
  Received: from smtp-out.localhost
          by example.net with esmtps
          (Exim 4.89)
          (envelope-from <foiperson@localhost>)
          id ABC
          for carlos@example.net; Mon, 23 Nov 2020 00:00:00 +0000
  Test
EOF
```

alexander-griffen:
## Relevant issue(s)
Fixes #5988
## What does this do?
These changes ensure that all email headers are looked at to find info-request email addresses. It also checks to see if there's multiple (different) info-request addresses - if there are, these go to the holding pen, still.
## Why was this needed?
Reduce noise in holding pen


app/mailers/request_mailer.rb
```diff
@@ -231,10 +231,10 @@ def self.receive(raw_email, source = :mailin)
 
   # Find which info requests the email is for
   def requests_matching_email(email)
-    # We deliberately don't use Envelope-to here, so ones that are BCC
-    # drop into the holding pen for checking.
-    addresses = ((email.to || []) + (email.cc || [])).compact
-    InfoRequest.matching_incoming_email(addresses)
+    addresses = MailHandler.get_all_addresses(email)
+    addresses.concat (MailHandler.get_header_string('Received', email) || '').
+      scan(/[A-Z0-9+_.-]+@[A-Z0-9.-]+/i)
```
gbp, line 236: I think there is an issue here when an email contains more than one `Received` header. `MailHandler.get_header_string('Received', email)` will return an array of Ruby objects which have been converted as string using `Array#to_s` EG. `[#<Mail::Field ... >", #<Mail::Field ... >"]`. We can't be sure the email address with be matched as this output could be truncated so need to handle multiple headers correctly as this will be the norm.

Internally for email regexp matching we normally use used `MySociety::Validate.email_find_regexp` which would be preferable. 

As it stands its not very clear looking at the code what it is doing, extracting to a helper method with a descriptive name would help but also worth considering the location of the code, I would probably move to be within `MailHandler` and refactor slightly to extract to a `#emails_within_received_headers` method. I think we could call `#email['received']` rather than calling `#get_header_string(...)` as we're doing with `envelope-to`, not sure if those are different or not..?

spec/mailers/request_mailer_spec.rb
```diff
@@ -48,12 +48,29 @@
...
     end
 
-    it "puts messages with the request address in Bcc: in the holding pen" do
-      request = FactoryBot.create(:info_request)
-      receive_incoming_mail('bcc-contact-reply.email',
-                            email_to: 'dummy@localhost',
-                            email_bcc: request.incoming_email)
-      expect(InfoRequest.holding_pen_request.incoming_messages.count).to eq(1)
+
+    it "attaches messages with an info request address in the Received headers to the appropriate request" do
```
gbp, line 52: This is a good test to keep even when the main implementation is moved to `MailHandler` but we should add additional test for the changes made there.

gbp (changes requested):
Nice work this seems to work but as described in inline comments there are cases when it might break so needs some changes and linting fixes.

alexander-griffen:
**PR Update:**
Changes have been added in fixups - obviously will be squashed before a merge, but just to allow incremental changes to be obvious during a review.

Code for checking Received headers *does* work, but might be slightly messy.

gbp (changes requested):
One question and some styling/linting/Rubocop changes required.

app/mailers/request_mailer.rb
```diff
@@ -231,10 +231,9 @@ def self.receive(raw_email, source = :mailin)
 
   # Find which info requests the email is for
   def requests_matching_email(email)
-    # We deliberately don't use Envelope-to here, so ones that are BCC
-    # drop into the holding pen for checking.
-    addresses = ((email.to || []) + (email.cc || [])).compact
-    InfoRequest.matching_incoming_email(addresses)
+    addresses = MailHandler.get_all_addresses(email)
+    addresses.concat MailHandler.get_emails_within_received_headers(email)
+    InfoRequest.matching_incoming_email(addresses.uniq)
```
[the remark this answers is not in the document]
alexander-griffen: I can absolutely do this. The reason I didn't, is I'm not sure if we want to throw a mail that has a spam address in the 'Received' header as spam (and thus not put them in the holding pen)?

app/mailers/request_mailer.rb
```diff
@@ -244,8 +243,8 @@ def receive(email, raw_email, source = :mailin)
     # Find which info requests the email is for
     reply_info_requests = requests_matching_email(email)
 
-    # Nothing found, so save in holding pen
-    if reply_info_requests.empty?
+    # Nothing found OR multiple different info requests, so save in holding pen
+    if reply_info_requests.empty? or reply_info_requests.count > 1
```
alexander-griffen, line 247: @gbp I've just thought of a potential issue. Due to the necessity to put messages with multiple BCC addresses in the holding pen (and I sort of thought that if there were multiple in Received, it should also do the same behaviour), I've simply added a check for multple info requests in general. Should this be the case, given that an array is expected, so I assume in general multiple info requests should be accepted? If not, I can change it to check on BCC addresses specifically.

(If I shouldn't change it though, I will fix the rubocop error to use `||` instead of `or`

gbp: Not sure I follow the issue. Lets find a time to talk today

lib/mail_handler/backends/mail_backend.rb
```diff
@@ -396,6 +397,22 @@ def address_from_string(string)
...
+          if received_headers.class == Mail::Field
+            received_header_string = received_headers.to_s
+          else
+            received_header_string = received_headers.
+              map { | header | header.to_s } * ' '
+          end
+          received_addresses = received_header_string.
+            scan(MySociety::Validate.email_find_regexp).flatten
+        end
+        received_addresses
```
gbp, line 414: This method can be tidied up to fix the Rubocop warnings and for readability.


```suggestion
        received_headers = Array(email['Received'])
        return [] if received_headers.empty?

        received_headers.map(&:to_s).
          join(' ').
          scan(MySociety::Validate.email_find_regexp).
          flatten
```

gbp (approved):
We've demoed this change in a video chat last week and this looking good.

One final change which I have tested locally so if the specs are green feel free to merge.

gbp (approved):

PR #7781, opened 2023-06-19
[#7780] Update User#can_make_comments?

### LINKED ISSUE #7780: Migrate User#can_make_comments? no limit fix
Migrate quick fix https://github.com/mysociety/whatdotheyknow-theme/pull/1670 in to core.

gbp:
## Relevant issue(s)

Fixes #7780

## What does this do?

Update User#can_make_comments?

## Why was this needed?

Allow active users with `no_limit` set to always make comments and bypass the exceeded limit/creation rate checks.

## Implementation notes

Was fixed in WDTK theme although it appears maybe non-active users might currently be able to make comments if this flag was set.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -489,7 +489,7 @@ class User < ApplicationRecord
 
   def can_make_comments?
     return false unless active?
-    return true if is_admin? || is_pro_admin?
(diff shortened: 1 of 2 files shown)


garethrees (approved):
lgtm

PR #7799, opened 2023-06-30
Update User#anonymise!

gbp:
## What does this do?

Update User#anonymise!

## Why was this needed?

We should add a censor rule when the user has left comments on other users requests.

## Notes to reviewer

Raised on standup with @HelenWDTK. Obscure case but possible.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -434,7 +434,7 @@ class User < ApplicationRecord
   end
 
   def anonymise!
-    return if info_requests.none?
+    return if info_requests.none? && comments.none?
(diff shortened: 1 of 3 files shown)


garethrees (changes requested):
LGTM.

Needs to also include an update to the documentation next to the button.

Something like:

```diff
- Makes a naive attempt to redact name from requests ^1
+ Makes a naive attempt to redact name from content ^1

- If the user has any requests, this will create a
+ If the user has created any content that includes their name, this will create a
```

PR #7807, opened 2023-07-04
[#7532] Bump sprockets to unreleased 3.x branch

### LINKED ISSUE #7532: ERB warning when precompiling assets
Since #6348 we're seeing warnings for sprockets v3. We should upgrade to sprockets v4 or take a look at the new options with Rails 7 https://discuss.rubyonrails.org/t/guide-to-rails-7-and-the-asset-pipeline/80851

```
gems/sprockets-3.7.2/lib/sprockets/erb_processor.rb:21: warning: Passing safe_level with the 2nd argument of ERB.new is deprecated. Do not use it, and specify other arguments as keyword arguments.
gems/sprockets-3.7.2/lib/sprockets/erb_processor.rb:21: warning: Passing trim_mode with the 3rd argument of ERB.new is deprecated. Use keyword argument like ERB.new(str, trim_mode: ...) instead.
```

gbp:
## Relevant issue(s)

Fixes #7532

## What does this do?

Bump sprockets to unreleased 3.x branch

## Why was this needed?

This branch is not in active development but contains some backports for ERB deprecations warnings we're seeing.

Full changes: https://github.com/rails/sprockets/compare/v3.7.2...3.x

## Implementation notes

Bumping to the latest 4.x release is not viable as this introduces `Sprockets::DoubleLinkError` which prevents multiple files with the same output path which is exactly how our themes work. In the future we should look to move to `propshaft`.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -154,6 +154,7 @@ gem 'mime-types', '< 4.0.0', require: false
 gem 'bootstrap-sass', '~> 2.3.2.2'
 gem 'mini_racer', '~> 0.8.0'
 gem 'sass-rails', '~> 5.0.8'
+gem 'sprockets', git: 'https://github.com/rails/sprockets', ref: '3.x'
 gem 'uglifier', '~> 4.2.0'
 
 # Feature flags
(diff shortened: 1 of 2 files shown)


gbp:
Tested on our staging ENV and assets compile correctly without any deprecation warnings.

PR #7824, opened 2023-07-13
Docker install guide changes

FozzieHi:
## What does this do?
- Change setup arguments - This is through my limited understanding of scripts, so please correct me if I'm wrong! In the `docker/setup` script, `data_reset` is defined as `data_reset="${1:-false}"`. This means that `data_reset` is either the first command-line argument if present, or false otherwise. So in order to run the migration, sample data and xapian index you would have to specify `./docker/setup true`. Given the current description in the installation guide is to do this, I think we should include the command line argument.
- Adds a section on how to visit the MailHog instance.
- Minor grammar fixes.

## Why was this needed?
To solve issues with databases after Docker installation, and to provide more information in the docs.

### DIFF
--- a/docs/installing/docker.md
+++ b/docs/installing/docker.md
@@ -52,7 +52,7 @@ create your own <a href="{{ page.baseurl }}/docs/glossary/#theme" class="glossar
   </p>
 </div>
 
-### How to setup the Docker container
+### How to set up the Docker container
 
 The supplied scripts in the `./docker` directory will create you a Docker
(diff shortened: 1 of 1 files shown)


gbp (commented):
Thanks for this. Not sure how I didn't catch this before.

I will change the setup script so passing in `true` isn't necessary.

gbp:
Fixed setup script in #7965, merged other changes in 81cc65713e9884d03683e0473508543abfbe7fb5

Issue #7829, opened 2023-07-17
Expiring large batches of requests crashes due to OOM


laurentS:
After sending out a series of large batches (500-1500 requests/batch), we have seen `bundle exec rake --silent stats:update_expiring_embargo_info_request_events` get killed a few times. While I have not explored the exact memory usage in detail, it seems fairly obvious that it's an OOM error, having had similar behaviour on our custom scripts elsewhere when processing large-ish numbers of requests.

I will send in a PR with a simple fix.

Related madada ticket: https://gitlab.com/madada-team/dada-core/-/issues/200

laurentS:
This was with alaveteli `v0.42.0.0`, but the code appears to remain in use with current versions.

laurentS:
Closing this as it was fixed by the above PR.

PR #7830, opened 2023-07-17
Reduce batch size to limit memory usage during task execution #7829

laurentS:
## Relevant issue(s)

#7829 

## What does this do?

This PR reduces the batch size used when expiring embargoes, from the default 1000 to 200.

## Why was this needed?

At madada.fr, we have seen crashes in this code when large numbers of embargoed requests are expired.

## Implementation notes

Not much to add, this should have no negative side effects, outside of marginally slowing down execution as there will be 5 times more DB queries sent. As this is a background job, there should be no material impact.


gbp (approved):
Can't see any reason not to change this. Thanks.

PR #7831, opened 2023-07-20
WIP: Reduce attempts to parse attachments in specs

garethrees:
Wanted to see if this had much impact on spec run time.


garethrees:
Didn't make much of a difference and is likely to get removed soon.

PR #7834, opened 2023-07-20
Account renaming post fixups

gbp:
## Relevant issue(s)

Connected to #285 

## What does this do?

Adds some changes post deploy of the self serve account renaming

## Why was this needed?

Noticed a few missing features/issues.

## Screenshots

image


gbp:
Discussed with @garethrees we agreed to merge as is as this more closely reflects how the previous User erasing works.

PR #7835, opened 2023-07-21
Use HTTPS for `PublicBody#calculated_home_page`

garethrees:
Reasonable to expect this is the default these days!


gbp (approved):

gbp (commented):
Looks good, except for the spec failures

Issue #6434, opened 2021-08-04
Prevent 'calculated homepage' from being generated for certain domains


mdeuk:
_Originally posted by @RichardTaylor in https://github.com/mysociety/alaveteli/issues/427#issuecomment-892806347_

>On a very closely related point we should stop calculated home pages eg. googlemail.com
>
>Do we need a list of exceptions ?
>
>This could be assembled from a list of the most common domains used in request addresses, presumably after excluding .ac.uk/.nhs.uk/gov.uk then  hotmail / aol / gmail / outlook would come top and we could treat the latter specially?

We did some rough statistics on this issue on https://github.com/mysociety/whatdotheyknow-theme/issues/690 previously - it is certainly causing a data quality issue in the UK, where we have a rather surprising number of public bodies which rely on either free or ISP provided email addresses!

Naturally, we should be surprised that our public bodies are conducting official business using free(mium) email products which are unlikely to meet the standards required by [public records legislation](https://www.nationalarchives.gov.uk/information-management/manage-information/policy-process/managing-email/), but at the same time, we should ensure our software doesn't unintentionally mislead people by sending them in the wrong direction.

This data quality issue creates issues for re-users of our data, such as the use case which prompted the WDTK ticket - e.g. our data being used to populate a Wikidata dataset; and given the pervasion of non-official email addresses for public bodies is unlikely to be a UK specific 'feature', I'd suggest it creates issues for re-users of data on other Alavateli websites as well.

garethrees:
> our data being used to populate a Wikidata dataset

Linking to https://github.com/mysociety/alaveteli/issues/6535.

garethrees:
I think we should only calculate a homepage for a default value in the form, which could then be deleted from the field should it not look sensible, rather than the current system that dynamically generates it.

garethrees:
Might be resolved by a better implementation of https://github.com/mysociety/alaveteli/issues/6434#issuecomment-1036466019

Issue #6676, opened 2021-11-17
Release 0.44


gbp:
Release candidate date: Summer 2023
Planned release date: Summer 2023

- [ ] Announce release candidate

[Release candidate date](https://github.com/mysociety/alaveteli/wiki/Release-Manager's-checklist#on-the-release-candidate-date):
- [ ] Make translation hotfix for current release
- [ ] Cut release branch

[Planned release date](https://github.com/mysociety/alaveteli/wiki/Release-Manager's-checklist#on-the-release-date):
- [ ] Merge release into master, tag and make GitHub release
- [ ] Announce release

Features:
- [x] #7496
~~- [ ] #6342~~
~~- [ ] #6343~~

Prev: #6657 Next: #7976

PR #7836, opened 2023-07-21
[#6434] Exclude email providers from body home pages

garethrees:
Many smaller authorities use a request email address from a general
purpose public email provider. We don't want e.g. gmail.com being set as
the home page for the authority.

This is configurable in the theme by setting the class attribute:

    # Add to the defaults
    PublicBody.excluded_calculated_home_page_domains << %w[
      example.net
    ]

    # Clear the defaults and set your own list
    PublicBody.excluded_calculated_home_page_domains = %w[
      example.org
      example.net
    ]

Fixes https://github.com/mysociety/alaveteli/issues/6434

### DIFF
--- a/app/models/public_body/calculated_home_page.rb
+++ b/app/models/public_body/calculated_home_page.rb
@@ -1,0 +1,15 @@
+# Guess the home page based on the request email domain.
+module PublicBody::CalculatedHomePage
+  # Guess home page from the request email, or use explicit override, or nil
+  # if not known.
+  #
+  # TODO: PublicBody#calculated_home_page would be a good candidate to cache
+  # in an instance variable
+  def calculated_home_page
+    if home_page && !home_page.empty?
+      home_page[URI.regexp(%w(http https))] ? home_page : "https://#{home_page}"
+    elsif request_email_domain
+      "https://www.#{request_email_domain}"
+    end
+  end
+end
--- a/app/models/public_body/calculated_home_page.rb
+++ b/app/models/public_body/calculated_home_page.rb
@@ -25,12 +25,15 @@ module PublicBody::CalculatedHomePage
     ]
   end
 
+  def calculated_home_page
+    @calculated_home_page ||= calculated_home_page!
+  end
+
+  private
+
   # Guess home page from the request email, or use explicit override, or nil
   # if not known.
-  #
-  # TODO: PublicBody#calculated_home_page would be a good candidate to cache
-  # in an instance variable
-  def calculated_home_page
+  def calculated_home_page!
     if home_page && !home_page.empty?
       home_page[URI.regexp(%w(http https))] ? home_page : "https://#{home_page}"
     elsif request_email_domain
@@ -38,8 +41,6 @@ module PublicBody::CalculatedHomePage
       "https://www.#{request_email_domain}"
     end
   end
-
-  private
 
   def excluded_calculated_home_page_domain?(domain)
     excluded_calculated_home_page_domains.include?(domain)
--- a/app/models/public_body/calculated_home_page.rb
+++ b/app/models/public_body/calculated_home_page.rb
@@ -31,15 +31,22 @@ module PublicBody::CalculatedHomePage
 
   private
 
-  # Guess home page from the request email, or use explicit override, or nil
-  # if not known.
+  # Ensure known home page has a full URL or guess if not known.
   def calculated_home_page!
-    if home_page && !home_page.empty?
-      home_page[URI.regexp(%w(http https))] ? home_page : "https://#{home_page}"
-    elsif request_email_domain
-      return if excluded_calculated_home_page_domain?(request_email_domain)
-      "https://www.#{request_email_domain}"
-    end
+    ensure_home_page_protocol || guess_home_page
+  end
+
+  # Ensure the home page has the HTTP protocol at the start of the URL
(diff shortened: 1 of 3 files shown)


garethrees:
Might want to show appending domains for configuration too:

```
PublicBody.excluded_calculated_home_page_domains << %w[
  example.org
  example.net
]
```

HelenWDTK:
> Any others we should add as defaults for Alaveteli? Can't cover every email provider but should include the most common international ones.

gmail, outlook, aol, gmx, mail.com, iCloud, proton, yahoo. - QQ, zoho, yandex and mail.ru are big but a bit more region specific.

app/models/public_body/calculated_home_page.rb
```diff
@@ -1,5 +1,27 @@
 # Guess the home page based on the request email domain.
 module PublicBody::CalculatedHomePage
+  extend ActiveSupport::Concern
+
+  included do
+    cattr_accessor :excluded_calculated_home_page_domains, default: %w[
+      aol.com
+      btinternet.com
+      gmail.com
```
garethrees, line 9: should also add googlemail.com before merge

mdeuk:
```suggestion
      gmail.com
      googlemail.com
```

app/models/public_body/calculated_home_page.rb
```diff
@@ -0,0 +1,52 @@
...
+module PublicBody::CalculatedHomePage
+  extend ActiveSupport::Concern
+
+  included do
+    cattr_accessor :excluded_calculated_home_page_domains, default: %w[
+      aol.com
+      btinternet.com
+      gmail.com
+      gmx.com
+      hotmail.com
```
mdeuk, line 11:
```suggestion
      hotmail.com
      hotmail.co.uk
      live.co.uk
      live.com
```

Add some other variants of the service formerly known as Hotmail - we already have `outlook.com` elsewhere.

WDTK has quite a few bodies with these domains - `316`, `437`, `96`, and `17` respectively.

app/models/public_body/calculated_home_page.rb
```diff
@@ -0,0 +1,52 @@
+# Guess the home page based on the request email domain.
+module PublicBody::CalculatedHomePage
+  extend ActiveSupport::Concern
+
+  included do
+    cattr_accessor :excluded_calculated_home_page_domains, default: %w[
+      aol.com
+      btinternet.com
```
mdeuk, line 8:
```suggestion
      aol.com
      aol.co.uk
      btinternet.com
      btconnect.com
      btopenworld.com
      talk21.com
      sky.com
      ntlworld.com
      virginmedia.com
      blueyonder.co.uk
      tiscali.co.uk
      talktalk.net
```

Cumulatively, we have a sizeable number across each[^1], so, in our own way - we gotta catch them all.

[^1]: Particularly the four BT ones, but the other consumer ISPs also add a fair amount of bulk… they are a favourite for Town and Parish councils!

app/models/public_body/calculated_home_page.rb
```diff
@@ -0,0 +1,52 @@
...
+      gmail.com
+      gmx.com
+      hotmail.com
+      icloud.com
```
mdeuk, line 12:
```suggestion
      icloud.com
      me.com
      mac.com
```

app/models/public_body/calculated_home_page.rb
```diff
@@ -0,0 +1,52 @@
...
+      mail.com
+      mail.ru
+      outlook.com
+      protonmail.com
+      qq.com
+      yahoo.com
```
mdeuk, line 18:
```suggestion
      yahoo.com
      yahoo.co.uk
```

We have `294` bodies with yahoo.co.uk. Another favourite for small bodies.

mdeuk (commented):
The list of domains makes sense - but, a scan of the wdtk database suggests some additions would be helpful [^1].

It seems to make sense to put them here, rather than branch them out on wdtk-theme, as they are possibly also of use to others.

[^1]: Primarily UK ISPs, but also the uk variants of Hotmail. Sadly, we have _**loads**_ of them, primarily in small Government organisations.

garethrees:
> The list of domains makes sense - but, a scan of the wdtk database suggests some additions would be helpful
>
> It seems to make sense to put them here, rather than branch them out on wdtk-theme, as they are possibly also of use to others.

Thanks, I'll add these variants. 🙌 Might pop the `.co.uk` variants on the theme; will have a think. It's a bit of an arms race. There are [libraries](https://github.com/kikobeats/free-email-domains) that track these free mail providers, so might consider using something like that should it start getting out of hand. I'm mainly interested in the most common cases, since ones where we only have a handful can be resolved by adding the actual home page.

gbp (approved):
This all looks good to me. Happy with the suggestions or we can bump some of those over to the theme.

garethrees:
I've refined this PR so that it covers international providers and made a counterpart PR for UK domains at theme level https://github.com/mysociety/whatdotheyknow-theme/pull/1789

gbp:
@garethrees I'm going to get this merged unless there is any reason not to have this in #6676?

garethrees:
If we can we should run it in prod for a bit this morning just to check it all seems okay, but in general I don't see why not. 🚀

PR #7874, opened 2023-08-29
Allow users to request account closure

### LINKED ISSUE #5052: Allow users to close their accounts and remove their username/profile
WhatDoTheyKnow users writing to us asking to close their accounts is becoming a notable fraction of support mail. 

There may be an opportunity to reduce the administration workload by letting users themselves anonymise and close their accounts (the admin button for this operation was introduced via #3074 and https://github.com/mysociety/alaveteli/pull/4869). 

There are a number of reasons though not to do this:
* Administrators currently ask users what they are seeking to achieve and act accordingly, for example sometimes unsubscribing from unwanted tracks is all that's needed to satisfy a user.
* We cannot automatically censor correspondence reliably to remove a user-name; if users contact us we can discuss and assess their needs and adjust the action we take accordingly. 
* Some users who want to remove their name from the site change their mind after correspondence with the admin team who, where relevant, stress that making a FOI request is something positive, something to be proud of. 

See also letting users unsubscribe from all email from the system: #4603

chrismytton:
## Relevant issue(s)

Fixes #5052

## What does this do?

Adds a "Close account" link to the users profile, which when clicked takes the user to a confirmation screen explaining what happens when an account is closed. If they confirm then an account closure request is created and a confirmation email is sent to the user.

There is then a list of account closure requests in the admin which an admin can go through. That list has a link to the user's account, and a close button, which closes the account directly.

## Why was this needed?

- Allow users to request to have their accounts closed.
- To reduce the administrative workload related to "right to erasure" requests.

## Implementation notes

- Adds a `Users::CloseAccountController` with `new` and `create` actions for users to request account closure.
- Adds an `AccountClosureRequest` model to manage closure requests.
- Sends email notifications to users upon request of account closure.
- Updates admin interface to allow viewing and management of account closures.
- Adds links for account closure in both user profile and admin interface.

## Video

https://github.com/mysociety/alaveteli/assets/22996/024072a6-bfd1-4d02-930f-73cfe48c6333

## Notes to reviewer

- Should the user be logged out after they have requested account closure?
- Should the "Close account" button on the user's profile be more (or less!) prominent?
- Is it now too easy for an admin to close an account?
- Should the list of account closure requests be added to the admin homepage "todo" list?


garethrees:
Closing as this has been sat for an eternity. Will revisit in future. Sorry that we couldn't get this over the line.

PR #7890, opened 2023-09-07
Stop duplicate log entries when closing requests.

### LINKED ISSUE #4776: Multiple allow_new_responses_from updates on a single request
> On the request at:
> 
> https://www.whatdotheyknow.com/admin/requests/412220
> 
> there are 82 events in quick succession closing the request to new
> responses. They're dated around
> 
> 2018-06-22 03:46:06 +0100
> 
> and the event records say
> 
> allow_new_responses_from: authority_only => nobody
> editor: InfoRequest.stop_new_responses_on_old_requests
> 
> 
> As this looks unusual I'm flagging it in case any action needs to be
> taken or if it's indicative of a bug.
> 
> 
> Regards
> 
> Richard

- [ ] Should clean up the repeated events once this is fixed to make the admin pages easier to use.

dracos:
The been_published check joins to the event table, so this call was e.g. returning one row per event. Fixes #4776.

After merging, something like:
```
delete from info_request_events where id in (
  with e1 as materialized (
    select info_request_id,min(id) min_id,params,count(*) c
    from info_request_events
    where event_type='edit' and params->>'editor' = 'InfoRequest.stop_new_responses_on_old_requests'
    group by info_request_id,params
    having max(created_at)-min(created_at) <'1 hour'::interval
  )
  select e2.id
  from e1, info_request_events e2
  where e1.info_request_id=e2.info_request_id and e1.params=e2.params and e2.id>e1.min_id
);
```
would delete the duplicate rows (8,266,693 at current count).

### DIFF
--- a/app/models/info_request.rb
+++ b/app/models/info_request.rb
@@ -542,12 +542,13 @@ class InfoRequest < ApplicationRecord
   def self.stop_new_responses_on_old_requests
     # 'old' months since last change to request, only allow new incoming
     # messages from authority domains
-    InfoRequest
-      .been_published
-      .where(allow_new_responses_from: 'anybody')
-      .where.not(url_title: 'holding_pen')
-      .updated_before(requests_old_after_months.months.ago.to_date)
-      .find_in_batches do |batch|
+    InfoRequest.
+      been_published.
+      where(allow_new_responses_from: 'anybody').
+      where.not(url_title: 'holding_pen').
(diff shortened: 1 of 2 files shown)


gbp (approved):
This fixes the root cause which is great.

The query to delete events isn't quite correct tho, it would work in most cases but some requests can have their `allow_new_responses_from` state be rolled back to 'anybody' or 'authority_only` after the responses have been stopped. Will open an separate issue for this.

PR #7891, opened 2023-09-07
Disable manual upload on closed requests.

### LINKED ISSUE #7185: Disable "Authority: Respond to this request" links on requests where `allow_new_responses_from` is "nobody"
A [recent case](https://www.whatdotheyknow.com/admin/incoming_messages/2083668/edit) in the WhatDoTheyKnow team mailbox identified that users with a public body email address associated with their account can respond to old requests where the `allow_new_responses_from` flag has been set to `nobody`.

This isn't desirable behaviour, as although the initial intent may have been to stop old requests receiving **email** responses (which are often spam), we will have edge cases of public bodies responding via the web portal, so ought to treat them in the same way.

dracos:
Fixes #7185. I imagine it may want a different template instead of my reuse here, but hopefully is the right idea.

### DIFF
--- a/app/assets/stylesheets/responsive/_request_layout.scss
+++ b/app/assets/stylesheets/responsive/_request_layout.scss
@@ -125,6 +125,12 @@
   margin-top: 2em;
 }
 
+#request_upload_response .request-header__closed_to_correspondence {
+  @include grid-column(12);
+  margin-top: 0.5em;
+  margin-bottom: 1em;
+}
+
 
 .feed_link_toolbar {
   width: 100%;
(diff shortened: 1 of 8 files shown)


gbp:
Screenshots compared to other messages which prevent the upload form from rendering

Screenshot 2023-09-25 at 13-01-34 Make and browse Freedom of Information (FOI) requests
Screenshot 2023-09-25 at 13-01-57 Make and browse Freedom of Information (FOI) requests
Screenshot 2023-09-25 at 13-02-13 Make and browse Freedom of Information (FOI) requests

gbp (approved):
Nice. Thanks for this.

We should really extract/refactor the permissions and action filter here but the whole controller action really needs reworking and we're not going to get to that any time soon. This approach which copies the existing way of preventing uploads works for now.

I have updated the styling for the white label theme and WDTK theme so this notice renders as the existing ones do.

gbp:
New styles:

image

PR #7892, opened 2023-09-07
[#7189] Details page showing keys metadata

### LINKED ISSUE #7189: Describe request tag changes as request tag changes rather than merely an "edit" on a request's public event log
Currently request tag changes are logged on the public event log as an event_type of "edit". 

Ideally the type of edit would be shown ie. "tags edited". 

eg. 

https://www.whatdotheyknow.com/details/request/statistics_of_how_many_tickets_h

Reason for acting:

* Some users can be concerned by unexplained "edits" to requests.

dracos:
Fixes #7189 

Two suggestions, could take one or both - the first changes "edit" to "edit metadata" on the details page, which I think makes it clearer (especially when there's no edit_incoming or edit_outgoing etc) that an edit of a request is editing its metadata rather than text of messages within the request.

And then as suggested on the ticket adds a column to show the keys that have changed as part of an event.

### DIFF
--- a/app/views/request/details.html.erb
+++ b/app/views/request/details.html.erb
@@ -25,7 +25,7 @@
           :site_name => site_name, :contact_path => help_contact_path) %>
 </p>
 
-<% columns = ['id', 'event_type', 'created_at', 'described_state', 'calculated_state', 'last_described_at' ] %>
+<% columns = ['id', 'event_type', 'details', 'created_at', 'described_state', 'calculated_state', 'last_described_at' ] %>
 
 <table>
   <thead>
@@ -41,7 +41,11 @@
     <tr class="<%= cycle('odd', 'even') %>">
     <% for column in columns %>
       <td data-label="<%= column %>"><%
-        d = info_request_event.send(column)
+        if column == 'details'
+          d = info_request_event.params_diff[:new].keys().join(', ')
+        else
+          d = info_request_event.send(column)
+        end
         if column == 'event_type' and d == 'edit'
           d = 'edit metadata'
         end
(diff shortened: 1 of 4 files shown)


app/views/request/details.html.erb
```diff
@@ -41,7 +41,11 @@
     <tr class="<%= cycle('odd', 'even') %>">
     <% for column in columns %>
       <td data-label="<%= column %>"><%
-        d = info_request_event.send(column)
+        if column == 'details'
+          d = info_request_event.params_diff[:new].keys().join(', ')
```
garethrees, line 45: @gbp I don't _think_ there should be a problem with this since it's only the keys, but could you construct a query to pull out all possible values that might be displayed here? Just want any keys in the params column that _don't_ start with `old_` I think.

gbp:
```sql
=> SELECT DISTINCT key FROM (SELECT jsonb_object_keys(params) AS key FROM info_request_events) AS keys WHERE key NOT LIKE 'old_%';
             key
-----------------------------
 email
 title
 destination_request
 api
 visible
 reason
 token
 smtp_message_id
 subject
 prominence
 attachment_id
 prominence_reason
 comment
 user
 shared
 comment_if
 raw_data
 attachment
 deleted_incoming_message_id
 comments_allowed
 incoming_message
 handle_rejected_responses
 script
 tag_string
 actions
 embargo
 embargo_extension
 awaiting_description
 request
 explanation
 attention_requested
 message
 id
 described_state
 allow_new_responses_from
 outgoing_message
 user_url_name
 body
 stop_new_responses
 public_body_url_name
 questions
 deleted_outgoing_message_id
 event_created_at
 editor
```

garethrees: Yep sweet, nothing concerning there.

PR #7893, opened 2023-09-08
Add ability to ask external caches to invalidate.

dracos:
Add a cached_urls method to the Comment, FoiAttachment, InfoRequest, PublicBody and User models, containing the entries that those models wish to invalidate when an event is logged via log_event or when the censor rules change. Entries are either a fixed path to be purged or a regex beginning with '^' to be banned.

Adds a NotifyCacheJob that for each of the object's cached_urls, for each locale, for each host in VARNISH_HOSTS, makes BAN or PURGE HTTP requests to the host, to invalidate the relevant entries.

### DIFF
--- a/app/jobs/notify_cache_job.rb
+++ b/app/jobs/notify_cache_job.rb
@@ -1,0 +1,68 @@
+require 'net/http'
+
+class Net::HTTP::Purge < Net::HTTP::Get
+  METHOD = 'PURGE'
+end
+
+class Net::HTTP::Ban < Net::HTTP::Get
+  METHOD = 'BAN'
+end
+
+##
+# Job to notify a cache of URLs to be purged or banned, given an object
+# (that must have a cached_urls method).
+#
+# Examples:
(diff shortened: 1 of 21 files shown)


gbp (approved):
This is looking good to me. There are more I could do to this but for now I have committed some fixups to do things is slightly better Ruby approach but functionally this should be the same. Many thanks @dracos 

Changes:
1. remove multiple conditionals, in favour for a callback when enqueuing the job
2. move post event logging code into InfoRequestEvent after create callback, and added specs
3. refactored job, removing heavily nested code and improve success/failure logging.

dracos:
Thanks - that callback is definitely nicer :) Only query I have is can Net::HTTP hold open a connection - I had it outside the URLs loop in the hope it would use the one connection for all of them, but probably makes no difference.

gbp:
> Only query I have is can Net::HTTP hold open a connection

Ah, as it is then no it doesn't. We can close the connection manually though if we don't use a do/end block. I'll push another fixup.

PR #7894, opened 2023-09-08
Remove Nicaragua and Turkey

sagepe:
Removes a couple of sites from the gh-pages branch that sadly no longer operate.


garethrees (approved):
Thanks!

PR #7912, opened 2023-09-22
WIP: Tests for case insensitivity

### LINKED ISSUE #7902: Uppercase characters in request email divert incoming message to holding pen
Here's a redacted but real example:

```
To: Request-[id]-[idhash]@Whatdotheyknow.Com
```

* Uppercase `R` of `Request`
* Uppercase `W` of `Whatdotheyknow`
* Uppercase `C` of `.Com`

We should do a case-insensitive comparison of these parts.

alexander-griffen:
## Relevant issue(s)
Fixes #7902 
## What does this do?
Two new tests have been introduced, to test for case sensitivity, with the expectation that they would fail. However, they currently seem to pass.

### DIFF
--- a/spec/mailers/request_mailer_spec.rb
+++ b/spec/mailers/request_mailer_spec.rb
@@ -20,7 +20,43 @@ RSpec.describe RequestMailer do
       deliveries = ActionMailer::Base.deliveries
       expect(deliveries.size).to eq(1)
       mail = deliveries[0]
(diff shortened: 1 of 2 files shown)


gbp:
Not required. Issue was different to what was originally thought. These tests while useful have been superseded by ones which check the Message-ID headers. See #7913

PR #7922, opened 2023-09-28
Add Xapian Sidekiq queue

gbp:
## What does this do?

Queue `InfoRequestExpireJob` on a Xapian Sidekiq queue which has concurrency disable so only one job is processed at a time.

## Why was this needed?

Calls to `InfoRequestExpireJob` when passed a `User` would block processing of other jobs such as attachment processing and cause support mail.

## Implementation notes

The order of the queues in `config/general.yml` is important (as they aren't weighted). the worker threads will always prioritise the first queue (`default`) so attachment processing and other jobs get priority.

Because we dealing with Xapian we also want to ensure only one Xapian related task is happening at any one time (although currently the expire job, isn't actually running Xapian indexing, but having multiple expire job, for the same resource running at the same time, could cause issues too) adding the `sidekiq-limit_fetch` to configure the `limits` in `config/general.yml` allows us to do this.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -120,6 +120,7 @@ gem 'ruby-msg', '~> 1.5.0', git: 'https://github.com/mysociety/ruby-msg.git', br
 gem 'rubyzip', '~> 2.3.2'
 gem 'secure_headers', '~> 6.5.0'
 gem 'sidekiq', '~> 6.5.9'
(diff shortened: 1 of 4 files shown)


gbp:
I have tested this on our staging ENV and it seems to be working well. i.e. two queues, all `InfoRequestExpireJob` jobs are on the `xapian` queue, getting exec one at a time.

PR #7940, opened 2023-10-06
Import changelog

garethrees:
Make it easier for admins to read about new features that have been added to Alaveteli from within Alaveteli itself.

Screenshot 2023-10-06 at 16 29 01

Screenshot 2023-10-06 at 16 29 10


app/views/layouts/admin.html.erb
```diff
@@ -44,13 +44,16 @@
...
-          <span class="span4 offset8">
+          <span class="span6 offset6">
             <ul class="nav nav-pills">
               <li>
                 <a href="http://alaveteli.org/docs/running/admin_manual/">
                   Admin Manual
                 </a>
               </li>
+              <li>
+                <%= link_to 'Alaveteli Changelog', admin_changelog_index_path %>
```
garethrees, line 55: I wonder about making this less "technical" sounding?

Could use "New Features", though that kind of only makes sense for the ~week or so after upgrade.

HelenWDTK: Are there always new features, or will it sometimes be bug fixes? How about "Latest changes" or "Recent changes"?

garethrees: Yeah it varies. Sometimes a release only includes translation updates. I quite like "Latest changes" or "Recent changes" as more human friendly, but I wonder if that's too general and could mean e.g. on-site content changes like "new requests" or whatever. Could call it "Software Updates" to make it align more with operating systems, which is quite universally understood?

garethrees: Went with "Changelog" after some discussion about what we do on other mySociety services. Also added some intro description when you click through to explain what's been viewed.

gbp (commented):
Nice. This could be useful. Any though given to making it clearer which version of Alaveteli is currently installed? Just rendering the changelog could make re-users think they have the latest features.

garethrees:
> Nice. This could be useful.

🎉 

> Any though given to making it clearer which version of Alaveteli is currently installed?

Could add a quick line like (which links to `/admin/debug`):

Screenshot 2023-10-06 at 16 40 45

> Just rendering the changelog could make re-users think they have the latest features.

I was thinking that the changelog that gets rendered is only ever up to the installed version, which will have the version number as the first heading (e.g. if you have 0.41.0.0 installed, the first heading will be 0.41.0.0). Then when you click through to one of the GitHub links you'd see a newer number. Being explicit about the current version doesn't hurt though, so have pushed an extra commit (ce1db92b4).

gbp:
> I was thinking that the changelog that gets rendered is only ever up to the installed version, which will have the version number as the first heading (e.g. if you have 0.41.0.0 installed, the first heading will be 0.41.0.0)

Oh, of course, that makes sense!

gbp (approved):

PR #7951, opened 2023-10-11
[EU#124] Fix loading custom states in theme

gbp:
## Relevant issue(s)

Fixes: https://github.com/mysociety/asktheeu-theme/issues/124

## What does this do?

Fix loading custom states in theme

## Why was this needed?

Let Zeitwerk ignore files which don't match the expected file structure.

Credit to @laurentS for working out the issue in #7917

### DIFF
--- a/config/initializers/theme_loader.rb
+++ b/config/initializers/theme_loader.rb
@@ -17,6 +17,9 @@ def require_theme(theme_name)
 
   require theme_main_include
 
+  # Let Zeitwerk ignore files which don't match the expected file structure
+  Rails.autoloaders.main.ignore(theme_lib.join('customstates.rb'))
+
   Rails.configuration.paths['config/refusal_advice'].push(
     root.join('config/refusal_advice')
   )


gbp:
Intending to backport this to a 0.43.2.0 release

laurentS:
Can I suggest adding some sort of warning in the logs or something? It took me quite a while to figure out what was happening. Or maybe a note in the template file, or the method that loads custom states. I feel like silently ignoring files leads to headaches for reusers.

gbp:
> Can I suggest adding some sort of warning in the logs or something? It took me quite a while to figure out what was happening. Or maybe a note in the template file, or the method that loads custom states. I feel like silently ignoring files leads to headaches for reusers.

This change should make everything work as it did before using the Zeitwerk autoloader. I'm not sure what the warning in the the log would need to say.

Maybe you misunderstood the change here, we're not ignoring the file, rather telling Zeitwerk to ignore it and that `lib/customstates.rb` in the theme can't be autoloaded when the `Customstates` class is used. This is correct because `Customstates` doesn't exist. Ultimately this means the file needs to be manually required which happens already. 

I'm going to backport as 0.43.2, I will include a upgrade note pointing to this PR in case Alaveteli themes have other custom files which also need this change. In this case the theme author should be able to add `Rails.autoloaders.main.ignore(...)` to `lib/alavetelitheme.rb`.

laurentS:
> Maybe you misunderstood the change here,

I did indeed :sweat: thanks for clarifying!

PR #7979, opened 2023-10-25
[#7977] Update `AttachmentController#show_as_html` to do attachment masking in the background

### LINKED ISSUE #7977: Update "view as HTML" to do processing into the background - masking
Our view attachments in the browser feature, "view as HTML", is doing alot of processing inline in the web processes and this sometimes causes memory issues on the server and results in oom-kill running. 

We should applying censor rules and masks via the `FoiAttachmentMaskJob` in the background as we do for attachment downloads.

gbp:
## Relevant issue(s)

Fixes #7977

## What does this do?

This updates `AttachmentController#show_as_html` to do attachment masking in the background.

## Why was this needed?

Our "view attachments in the browser" feature, or "view as HTML", is doing a lot of processing inline in the web processes and this sometimes causes memory issues on the server and results in oom-kill running.

## Implementation notes

Uses the same logic as the `#show` action to check an attachment is masked before proceeding with the action. This means it will time outs after 5 seconds and then redirect the `AttachmentMaskController` to wait for the job to complete before rendering a view saying the attachment is now viewable.

### DIFF
--- a/app/controllers/attachment_masks_controller.rb
+++ b/app/controllers/attachment_masks_controller.rb
@@ -9,6 +9,8 @@ class AttachmentMasksController < ApplicationController
 
   def wait
     if @attachment.masked?
+      redirect_to @referer and return if refered_from_show_as_html?
+
       redirect_to done_attachment_mask_path(
         id: @attachment.to_signed_global_id,
         referer: verifier.generate(@referer)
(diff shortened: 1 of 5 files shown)


gbp:
Testing on staging and this works but I have changed the implementation slightly. Because we're not downloading files we don't need to render the `done` action at all, and can just redirect the to show as HTML action.

gbp:
This is up on staging for testing, I have added a `sleep 10` line to the job to simulate a slow job to show the redirect to the attachment mask controller.

In order to ensure attachments need to be masked, in the staging Rails console run: `FoiAttachment.update_all(masked_at: nil)`

garethrees (approved):
Reviewed on a call together. Looks great!

PR #7984, opened 2023-10-27
[#7784] Update request URLs use `url_title`

### LINKED ISSUE #7784: Make all request URLs use `url_title` instead of `id` as the identifier
When dealing with support requests we often have to figure out multiple permutations of request (and associated content) URLs.

For example, given the request `/request/foo`, we might end up with:

* `/request/foo`
* `/cy/request/foo`
* `/request/123/response/456/attach/2/bar.doc`
* `/request/123/response/456/attach/2/bar.doc.html`
* `/request/123/followups/new/456`
* `/details/request/foo`
* `/feed/request/foo`

While the locale prefix (`/cy/`) and other routing (e.g. `/feed/request/…`) is a bit more of a faff to fix, knowing that public request URLs will consistently use the `url_title` instead of `id` as the identifier is a good thing to not have to worry about.

This ticket should:

- [ ] Replace all uses of `InfoRequest#id` as a URL parameter `InfoRequest#url_title` for any public request URL
- [ ] Handle existing routes that use `InfoRequest#id` by permanently redirecting to the updated URL (i.e. we should always be able to _use_ `id` in place of the `url_title`, but it should always result in a redirect)

alexander-griffen:
## Relevant issue(s)

Fixes #7784 

## What does this do?

This changes instances of `id` in URLs to `url_title`

## Why was this needed?

This was needed to reduce possible URL permutations, to make admin tasks easier.

## Notes to reviewer

Commits b56786b...6b07a03 are refactoring to allow the change to be made in the other commits.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -279,10 +279,7 @@ class ApplicationController < ActionController::Base
   # Return logged in user
   def authenticated_user
     return unless session[:user_id]
-
-    @user ||= User.find_by(
-      id: session[:user_id], login_token: session[:user_login_token]
-    )
+    @user ||= User.authenticate_from_session(session)
   end
 
   # For CanCanCan and other libs which need a Devise-like current_user method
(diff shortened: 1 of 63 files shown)


gbp (approved):
I have tested this thoroughly both locally and on staging.

Issue #1106, opened 2013-09-20
Make decoded but unredacted version of incoming email available via admin interface


RichardTaylor:
Sometimes administrators need to access the unredacted plain text of an incoming email body for example to provide users with access to email addresses or phone numbers which have been removed by Alaveteli. 

Raw emails, which are available via the admin interface,  sometimes contain the body of the email in Base64 rather than plain text. 

I can't imagine why admins would need " Cached main body text folded " or " Cached main body text unfolded" in relation to incoming messages. Perhaps one of these could be replaced with "main body text unredacted" ?

See also #10 which is related.

kingqueen3065:
This is something we have to do pretty often for various reasons; predominantly to release contact details at the user's request, but also for other reasons. It is a bar to new administrators being able to undertake the work needed and so it would be great if this could be made easier.

garethrees:
## Context

Admins often want to view the contents of an incoming message without any masking or censor rules applied.

In some cases you can inspect the raw email, but it's often base64 encoded, which means needing to download the email and open it in a mail client. Even when the email is not base64 encoded, it's often difficult to mentally parse the information out when there are lots of headers, attachment sections and HTML body content.

To fix this we'll render the main body text of an incoming message into the admin interface, but without any redactions applied.

## UI

We can just render the unredacted body out under the attachments listing on the incoming message admin page. Ideally we'll allow folding/unfolding of quoted sections, but not essential if it adds a load of complexity.

![Doc](https://user-images.githubusercontent.com/282788/191958231-33de1959-b3b8-4bd6-ba1a-2c0bd16682d5.jpg)

## Mechanics

We'll want to use cached content to avoid fetching the raw email from Active Storage, but there's huge gotcha in that we currently cache the main body text _with masks_.

Given we apply censor rules on the fly, I don't see why we don't also do that with the text masks (which are effectively hard-coded censor rules anyway).

Of course, we'd then need to ensure that the masks get applied where we *do* show text publicly. It looks like we do call the `cached_main_body_text_*` methods in a few places, so they'll need to be updated to call an alternative method.

## Other considerations

Re-caching all incoming messages doesn't seem like a useful idea, so I think we can take the position that this will only become possible for new messages, or admins can ask for developer assistance to re-parse old messages on request. Running `apply_masks` to cached masked content shouldn't be problematic.

I don't think this poses any data retention policy issues, since we already hold and render the unredacted data in the raw email. We will want to make this change clear in the upgrade notes though.

[...]

PR #7999, opened 2023-11-07
[#1106] Admin view of unmasked incoming message

garethrees:
Admins often want to view the contents of an incoming message without any masking or censor rules applied.

In some cases you can inspect the raw email, but it's often base64 encoded, which means needing to download the email and open it in a mail client. Even when the email is not base64 encoded, it's often difficult to mentally parse the information out when there are lots of headers, attachment sections and HTML body content.

This commit adds an unmasked view of main body part attachments. This is equivalent to the main body contents of an incoming message that we show in public.

This does require "downloading" the attachment from Active Storage, but at present we use local file storage for attachments, so that's fine. Even if we moved them to S3 it wouldn't be a disaster to fetch them each time, as this isn't a frequently viewed page.

Needed to use an `!important` CSS rule to improve the visual rendering as for some reason bootstrap sets `margin-bottom: 0;` on `p` tags within a `blockquote` element (See
bootstrap-sass-2.3.2.2/vendor/assets/stylesheets/bootstrap/_type.scss).

Fixes https://github.com/mysociety/alaveteli/issues/1106.

Screenshot 2023-11-07 at 15 52 55

### DIFF
--- a/app/assets/stylesheets/admin.scss
+++ b/app/assets/stylesheets/admin.scss
@@ -145,6 +145,15 @@ body.admin {
     display: inline;
   }
 
+  blockquote .well p {
+    font-size: 14px;
+    margin-bottom: 1em !important;
+
+    &:last-child {
+      margin-bottom: 0 !important;
+    }
+  }
+
   div#user_locale_switcher {
     div.btn-group:before,
     div.btn-group:after {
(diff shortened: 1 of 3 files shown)


gbp (approved):
Nice. Added a fixup for slightly better styling - using EMs, remove the last margin and set the font size to match the rest of the content on the page.

image

Issue #1782, opened 2014-08-12
Encourage users to find new email addresses for bodies themselves in case of errors/bounces


RichardTaylor:
When users mark a response for our attention we give them a free text box to explain the problem. 

If the issue is the email address for the authority needs updating we should point them to the update authority form.

https://www.whatdotheyknow.com/change_request/new?body=bens_test_authority

We should encourage users to research a new address for themselves - eg. stressing the site is administered by volunteers and they can help us by finding the address and saying if they provide the address we might be able to re-send their request quicker too.

garethrees:
Could also include this where we show a failed delivery status.

PR #8000, opened 2023-11-09
[#1782] Help us CTA from delivery error message

garethrees:
Finding contact details can take a fair amount of work. Users are often happy to help, but the only signpost we have for this is when submitting the classification. They might not have time to source a new address there and then – and the form itself could be improved – but a reminder for them and other users viewing the requests might help distribute some of this work to the community.

Fixes https://github.com/mysociety/alaveteli/issues/1782.

**Public Request:**

Screenshot 2023-11-08 at 20 41 37

**Private Request:**

Screenshot 2023-11-09 at 09 53 55

### DIFF
--- a/app/helpers/info_request_helper.rb
+++ b/app/helpers/info_request_helper.rb
@@ -185,10 +185,14 @@ module InfoRequestHelper
       public_body_link: public_body_link(info_request.public_body))
   end
 
-  def status_text_error_message(_info_request, _opts = {})
+  def status_text_error_message(info_request, _opts = {})
+    body = info_request.public_body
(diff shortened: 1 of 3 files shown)


HelenWDTK:
Love this.

gbp (approved):
Haven't run but should be better than what we currently have. Would be good to open a ticket to link the delivery error report to the change request. Could we use the case reference email subject some how.

garethrees:
> Would be good to open a ticket to link the delivery error report to the change request.

Done! https://github.com/mysociety/alaveteli/issues/8004

PR #8011, opened 2023-11-14
[#8009] Update model invalidate cache callbacks

### LINKED ISSUE #8009: Large number of `NotifyCacheJob` queued overnight
There seems to be processes queuing `NotifyCacheJob` jobs excessively.

- On WDTK, `script/update_public_bodies_stats` runs around 4am and causes a save (`without_revision`) which triggers the `invalidate_cached_pages` callback.
- In `app/mailers/track_mailer.rb` there is a User `save!(touch: false)` which will trigger the `invalidate_cached_pages` callback too.
- others?

We already prevent these saves from causing a Xapian reindex. We could update the callback to only run is `no_xapian_reindex` is unset.

gbp:
## Relevant issue(s)

Fixes #8009

## What does this do?

Update model invalidate cache callbacks.

Only run the callbacks when `no_xapian_reindex` hasn't been set.

## Why was this needed?

This will prevent excessive `NotifyCacheJob` jobs from being created overnight when `script/update_public_bodies_stats` or when user alert tracks email is sent.

## Implementation notes

Might want to consider the naming of `no_xapian_reindex`.

### DIFF
--- a/app/models/info_request_event.rb
+++ b/app/models/info_request_event.rb
@@ -88,7 +88,7 @@ class InfoRequestEvent < ApplicationRecord
     self.event_type = "hide"
   end
   after_create :update_request, if: :response?
-  after_create :invalidate_cached_pages
+  after_update :invalidate_cached_pages, unless: :no_xapian_reindex
 
   after_commit -> { info_request.create_or_update_request_summary },
                   on: [:create]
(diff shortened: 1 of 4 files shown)


gbp:
Merged in ba0de9166abf4dc35fb5157f46b575b54a803919

Issue #7946, opened 2023-10-11
Only include active users in leaderboards


garethrees:
On `/statistics` we show the top requesters over the last 28 days. Sometimes users are in this list due to misuse and have had their account suspended. We shouldn't include these in the list.

While it's less likely for `/categorise/play`, we should also do this for the leaderboards there.

chrismytton:
@garethrees As well as banned users, do you think we should exclude users who have closed their accounts from leaderboards?

garethrees:
I think yes, to keep things simple. There's a `User.active` scope, so should just be using that to filter the existing list.

PR #8012, opened 2023-11-14
Only show active users on leaderboards

chrismytton:
## Relevant issue(s)

Fixes https://github.com/mysociety/alaveteli/issues/7946

## What does this do?

Change the `Statistics::Leaderboard` class to only consider active users (i.e. users that haven't closed their account or been banned) from appearing on leaderboards.

### DIFF
--- a/app/controllers/request_game_controller.rb
+++ b/app/controllers/request_game_controller.rb
@@ -38,7 +38,7 @@ class RequestGameController < ApplicationController
     end
 
     @league_table_28_days = RequestClassification.league_table(10,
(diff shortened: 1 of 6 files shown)


garethrees (commented):
Looks great. Good to merge as is. If it's quick and easy would be worth doing for the [request game too](https://github.com/mysociety/alaveteli/blob/develop/app/controllers/request_game_controller.rb#L40-L42) (`/categorise/play`), but not worth hours of work if it's a faff.

garethrees (approved):
⚡️ nice!

PR #8015, opened 2023-11-16
Extract FoiAttachment::Dsn & restore functionality

garethrees:


gbp (changes requested):

gbp (changes requested):
Oh. `dns.rb` is still present

> ls app/models/foi_attachment
> delivery_status_notification.rb  dsn.rb  prominence.rb

gbp (approved):

Issue #7811, opened 2023-07-04
Create generalised `Category` system


garethrees:
We currently have `PublicBodyCategory` and `PublicBodyCategoryHeading` to group `PublicBody` records based on their tags.

We now have several `Taggable` models that we’d like to be able to categorise – `InfoRequest` in the first instance, but this will also be useful for a [knowledge base](https://github.com/mysociety/alaveteli/issues/7764).

## For `InfoRequest`

We’ll add a new admin menu item: Requests → Categories.

From there we’ll see an interface much like the authority categories – a list of headings with categories, drag and droppable, buttons for new ones, and links to the edit forms of existing ones.

The only difference will be that these categories are scoped to `InfoRequest` records with the given tags, rather than authorities.

## For `PublicBody`

For public bodies, the category system should look and feel exactly the same as it does now; just powered by a new generic backend.

We’ll need to include a way to migrate the existing records to the new generic backend.

## Behind the scenes

I expect we’ll want to make `Category` a polymorphic record, includable via a concern:

```rb
class PublicBody
  include Categorisable
end

class InfoRequest
  include Categorisable
end
```

PR #8023, opened 2023-11-22
[#7811] Generalised category system

gbp:
## Relevant issue(s)

Part of #7811

## What does this do?

Adds generalised category system, replacing the old `PublicBodyCategory` models.

## Why was this needed?

So we can categorise requests.

## Implementation notes

~~There is one minor changes to the old category feature. Due to how nested sets work categories can now only have one parent/heading.~~

~~This means:~~
1. ~~I didn't add to `Category` the uniqueness validation which is on `PublicBodyCategory#category_tag`.~~

2. ~~Some categories have been duplicated, for WDTK this means 27 categories with the same title as another will be created. I think this is a good compromise given the reduced code complexity and added flexibility of being able to reuse the new categories feature for `InfoRequest` easily.~~


garethrees:
> There is one minor changes to the old category feature. Due to how nested sets work categories can now only have one parent/heading… Some categories have been duplicated, for WDTK this means 27 categories with the same title as another will be created.

Could you paste a list of these?

gbp:
> Could you paste a list of these?

```ruby
irb(main):002:0> PublicBodyCategory.all.select { _1.public_body_headings.count > 1 }.map(&:title).sort
=>
["Ambulance Services",
 "Approved regulators of legal services",
 "Areas of Outstanding Natural Beauty Conservation Boards",
 "Associations of Local Councils",
 "Dance companies",
 "Education authority",
 "Education regulators",
 "Financial regulators",
 "Groups of universities",
 "Health and Social Care Partnerships (Scotland)",
 "Health and social care regulators",
 "Health regulators",
 "Heraldic authorities",
 "Highways Operators",
 "Housing ALMOs",
 "Housing regulators",
 "Integrated Care Partnerships",
 "Integration Joint Boards",
 "Media regulators",
 "Military colleges",
 "Money laundering supervisory authorities",
 "Police regulators",
 "Telecommunications companies",
 "Theatres",
 "Transport regulators",
 "Utility regulators",
 "Veterans Advisory and Pensions Committees"]
```

spec/fixtures/categories.yml
```diff
@@ -0,0 +1,47 @@
...
+  parent_id: 1
+  lft: 1
+  rgt: 8
+
+useless_category:
+  id: 3
+  parent_id: 2
+  lft: 2
+  rgt: 3
+  category_tag: useless_agency useless
```
gbp, line 33: This matches the `spec/fixtures/public_body_categories.yml` but should it just be:

```suggestion
  category_tag: useless_agency
```

garethrees: Yeah I don't think the current system supports finding from multiple tags yet. That category would create the URL `/body/list/useless_agency%20useless` which won't find anything.

app/models/category.rb
```diff
@@ -0,0 +1,33 @@
...
+
+  translates :title, :description
+  include Translatable
+
+  has_many :tags,
+           foreign_key: :name,
+           primary_key: :category_tag,
+           class_name: 'HasTagString::HasTagStringTag'
+
+  has_many :public_bodies,
```
garethrees, line 27: Don't think this is required

lib/tasks/temp.rake
```diff
@@ -1,4 +1,39 @@
 namespace :temp do
+  desc 'Migrate PublicBodyCategory into Category model'
+  task migrate_public_body_categories: :environment do
```
garethrees, line 3: Probably want to error `if Category.any?` to avoid issues where category already exists

garethrees (commented):
Looks good as discussed on call / demo.

Main thing is to make sure that we can add categories to many Taggable things in future. Seems like we should be able to push the currently hard-coded things referencing an explicit model into a concern:

```rb
# BEFORE:
Category.public_body_tree.each do |heading|

# AFTER: 
PublicBody.categories.each do |heading|

# BEFORE: In category
def self.public_body_root
  Category.roots.find_or_create_by(title: 'PublicBody')
end

def self.public_body_tree
  tree(public_body_root)
end


# AFTER: In categorisable
def self.category_root
  Category.roots.find_or_create_by(title: self.class.name)
end

def self.categories
  Category.tree(category_root)
end
```

app/views/admin/categories/index.html.erb
```diff
@@ -0,0 +1,54 @@
...
+              <%= link_to "Save", '#', class: "btn btn-success disabled save-order", "data-heading-id" => heading.id, "data-list-id" => "#heading_#{heading.id}_category_list", 'data-endpoint' => reorder_admin_category_path(heading) %><p class="save-notice" data-unsaved-text="There are unsaved changes to the order of categories." data-success-text="Changes saved." data-error-text="There was an error saving your changes: ">Drag and drop to change the order of categories.</p>
+            </div>
+          </div>
+        </div>
+      </div>
+    <% end %>
+  </div>
+
+  <div class="form-actions save-panel">
+    <%= link_to "Save", '#', class: "btn btn-success disabled save-order", "data-list-id" => '#category_list', 'data-endpoint' => reorder_admin_category_path(Category.public_body_root) %><p class="save-notice" data-unsaved-text="There are unsaved changes to the order of category headings." data-success-text="Changes saved." data-error-text="There was an error saving your changes: ">Drag and drop to change the order of category headings.</p>
```
gbp, line 52:
```suggestion
    <%= link_to "Save", '#', class: "btn btn-success disabled save-order", "data-list-id" => '#category_list', 'data-endpoint' => reorder_admin_category_path(@root) %><p class="save-notice" data-unsaved-text="There are unsaved changes to the order of category headings." data-success-text="Changes saved." data-error-text="There was an error saving your changes: ">Drag and drop to change the order of category headings.</p>
```

app/controllers/admin/categories_controller.rb
```diff
@@ -0,0 +1,92 @@
...
+  def reorder
+    transaction = reorder_categories(params[:categories])
+    if transaction[:success]
+      head :ok
+    else
+      render plain: transaction[:error], status: :unprocessable_entity
+    end
+  end
+
+  protected
```
gbp, line 59:
```suggestion
  private
```

app/controllers/admin/categories_controller.rb
```diff
@@ -0,0 +1,92 @@
...
+      rescue ActiveRecord::RecordNotFound
+        error = "Couldn't find Category #{category_id}"
+        raise ActiveRecord::Rollback
+      end
+    end
+    { success: error.nil?, error: error }
+  end
+
+  private
+
```
gbp, line 76:
```suggestion
```

app/views/admin/categories/_form.html.erb
```diff
@@ -0,0 +1,82 @@
...
+      <% else %>
+        <%= f.fields_for(:translations, translation, child_index: translation.locale) do |t| %>
+          <%= render partial: 'locale_fields', locals: { t: t, locale: translation.locale } %>
+        <% end %>
+      <% end %>
+    <% end %>
+  </div>
+</div>
+
+<% if @category.leaf? || @category.new_record? %>
```
gbp, line 48:
```suggestion
<% if @category.parent != @root || @category.new_record? %>
```

app/views/admin/categories/edit.html.erb
```diff
@@ -0,0 +1,35 @@
...
+<hr />
+
+<div class="row">
+  <div class="span12">
+    <div class="well">
+      <%= form_for @category, url: admin_category_path(@category), method: 'delete', class: "form form-inline" do |f| %>
+        <%= f.submit "Destroy #{@category.title}", title: @category.title, class: "btn btn-danger", data: { confirm: 'Are you sure?' } %>
+
+        <span class="help-block">
+          Destroying a category does not destroy the public authorities associated with the category.
```
gbp, line 30:
```suggestion
          Destroying a category does not destroy the objects associated with the category.
```

lib/tasks/temp.rake
gbp: Went for `if Category.public_body_tree.any?` as `Category.public_body_root` uses a `find_or_create_by` so if that method is called first (for example loading the authorities list view) then `Category.any?` would be true.

gbp:
Reviewed and approved on a call together.

Issue #2523, opened 2015-06-04
Write systemd unit files for daemons


Flupsy:
Many modern distros are replacing conventional System V-style init scripts with systemd unit files.  We should write them for our daemons so that we have native support for systemd.

I've done this for some services on jessie so I can assist (or just do it).

sagepe:
Relevant: https://github.com/mysociety/alaveteli/commit/663bb16df54870fcab0e15a6af36647991cdca9e

gbp:
We support systemd in our install scripts https://github.com/mysociety/alaveteli/pull/7591 so this change should be relatively easy now.

HelenWDTK:
This issue has been automatically closed due to a lack of discussion or resolution for over 12 months.
Should we decide to revisit this issue in the future, it can be reopened.

sagepe:
I think this has been done via https://github.com/mysociety/alaveteli/pull/8024.

PR #8024, opened 2023-11-23
Convert the daemons to native systemd units

sagepe:
This replaces the old init scripts for the daemons with `systemd` unit files.

## Relevant issue(s)

https://github.com/mysociety/alaveteli/issues/2523

## What does this do?

Replaces the Alert Tracks, Send Notifications and Poll Incoming daemon sysvinit scripts with systemd units

## Why was this needed?

To modernise the daemons


gbp (approved):
Needed to enable ERB trim mode else the rake task didn't work. Have also update the xapian compact script to call `rails restart`.

Commits should be reordered and messages updated. I'll do this on merging.

gbp (approved):
Turns out when and install use `rbenv` these weren't working. This was down to how the paths were defined. I have moved this logic into the rake task instead of the service templates and fixed - IE single `PATH` env no `SET_PATH` env needed.

gbp:
This should be good to merge next week.

PR #8025, opened 2023-11-23
[#7811] Allow InfoRequest to be categorised 

### LINKED ISSUE #7811: Create generalised `Category` system

gbp:
## Relevant issue(s)

Depends on #8023 
Fixes #7811

## What does this do?
 
Add `categorisable` concern and adds this to `InfoRequest`

## Why was this needed?

So we can so groups of requests on [topic pages](https://github.com/mysociety/alaveteli/milestone/62).

## Screenshots

image

### DIFF
--- a/app/views/admin/categories/_scope.html.erb
+++ b/app/views/admin/categories/_scope.html.erb
@@ -1,0 +1,7 @@
+<ul class="nav nav-tabs">
+  <% Categorisable.models.sort_by(&:admin_title).each do |klass| %>
+    <%= tag.li class: { active: current_klass == klass } do %>
+      <%= link_to klass.admin_title.pluralize, url_for(model_type: klass.name, query: params[:query]) %>
+    <% end %>
+  <% end %>
+</ul>
--- a/app/controllers/admin/categories_controller.rb
+++ b/app/controllers/admin/categories_controller.rb
@@ -4,6 +4,7 @@
 class Admin::CategoriesController < AdminController
   include TranslatableParams
 
+  before_action :check_klass
   before_action :set_root, expect: [:destroy, :reorder]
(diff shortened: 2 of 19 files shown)


app/views/admin/categories/_scope.html.erb
```diff
@@ -0,0 +1,7 @@
+<ul class="nav nav-tabs">
+  <% Categorisable.models.sort_by(&:admin_title).each do |klass| %>
+    <li class="<%= 'active' if current_klass == klass %>">
```
garethrees, line 3: fyi I think you can do:

```erb
<%= tag.li class: { active: current_klass == klass } do %>
  # link_to…
<% end %>
```

gbp:
Reviewed and approved on a call together.

PR #7961, opened 2023-10-17
Explore in Datasette

garethrees:
Related to https://github.com/mysociety/alaveteli/issues/7027

Screenshot 2023-10-17 at 10 22 26

With custom URL:

Screenshot 2023-10-19 at 17 56 08

CORS headers:

Screenshot 2023-10-17 at 10 23 03

PR #8030, opened 2023-12-01
Add helper to add query params to a URL

gbp:
## What does this do?

Add helper to add query params to a URL

## Why was this needed?

In a couple of places we do this in different ways, this helper will allow us to do this in one line and is tested independently.

## Implementation notes

This helper will be useful in #7961

<hr>

[skip changelog]


garethrees (approved):
haven't run but lgtm

PR #8055, opened 2023-12-18
Fix redirect when category is destroyed

### LINKED ISSUE #8054: Unexpected redirect when you delete a request category
When you delete a request category you get redirected to the authority category list rather than the request category list.

gbp:
## Relevant issue(s)

Fixes #8054

## What does this do?

Fix redirect when category is destroyed

## Implementation notes

Pass the current model type to the admin category destroy action so we can redirect to the correct/current model index.

<hr>

[skip changelog]


confirmordeny:
That was fast. Thank you.

PR #8057, opened 2023-12-18
[#8056] Add support for Ruby 3.3

### LINKED ISSUE #8056: Support Ruby 3.3
Due out next week. 🎄

gbp:
## Relevant issue(s)

Fixes #8056

## What does this do?

Add support for Ruby 3.3

## Why was this needed?

Framework modernisation.

## Implementation notes

Seeing what test examples fail.


garethrees:
Could this PR also add deprecation notices to the changelog for the older 3.x versions? We're getting closer to [needing to make a release](https://github.com/mysociety/alaveteli/compare/master...6b002d5) and will be required for [Rails 8](https://github.com/mysociety/alaveteli/issues/8075) which is around the corner.

Issue #2663, opened 2015-07-15
Consider practiciality of generating alerts when Excel files containing hidden data are released


RichardTaylor:
There have been a number of cases on WhatDoTheyKnow of the accidental release of large amounts of sensitive personal information by public bodies who have not realised information was present, but hidden, in Excel documents released ( [Blog post](https://www.mysociety.org/2013/06/13/whatdotheyknow-team-urge-caution-when-using-excel-to-depersonalise-data/) )

The suggestion of a technical approach to assist in identifying such occurrences has been made.

If it's not technically possible to actually detect if hidden data is present, perhaps potentially problematic files, could be flagged. One idea is to alert administrators and the public body concerned when an Excel document over a certain size (by file size, rows, or sheets) is included as an attachment to a response sent to an Alaveteli site. The message to the public body could say something along the lines of:

> ~"are you sure you meant to send x MB of data in an Excel file in response to [request subject]. There have been instances of public bodies accidentally releasing information in response to requests. Please see our detailed guidance at ...  and contact us via xxxx if you are concerned you may have accidentally released more information than intended". 

Perhaps the ability to search for large attachments of a certain file type be generally useful anyway?

FOIMonkey:
Most breaches occur when the following happens:

- Data is not hidden in any way.
- Data is on a hidden spreadsheet.
- Data is in hidden rows/columns.
- Data is cached when the authority creates a graph/chart
- Data is cached when the authority creates a pivot table
- Data is cached when the authority imports data from an external file.

Xapian already happily looking inside xls and certain metadata is already discoverable by the site search eg pivot tables. It does not cope with xlsx files very well.

The easiest thing to look for is file size. Most incidents involve spreadsheets >1mb (though there have been one or two smaller than that). Of course there are also a lot of legitimate releases that are that size. As of 19 May there were 1151 spreadsheets of 1mb or larger on the site, most of these are large datasets or badly formatted excel files containing several million empty rows. The largest is 40.3mb.

RichardTaylor:
Perhaps a good question to ask would be if there are any additional features which @FOIMonkey can identify that would make it easier for them (and others) to find these problematic documents.

FOIMonkey:
Number one would be the ability to search via filesize.

skenaja:
For xlsx (which is a effectively zipped collection of xml files), these are some of the ways to find out if data has been hidden or cached:

[...]

PR #8060, opened 2023-12-19
Add XLS analyzer

gbp:
## Relevant issue(s)

Connected to #2663

## What does this do?

Add XLS analyzer

## Why was this needed?

Next step to detect more hidden data in spreadsheets.

## Implementation notes

Requires LibreOffice to be installed.


gbp:
Additional commits have been extract out into their own PRs this should be mergeable as is now.

gbp:
Marked as `awaiting-deploy` as we ran this in background jobs on production previously without issues.

PR #8062, opened 2023-12-20
Add basic admin notes index

garethrees:
Allow admins to view all notes in one place and navigate to any Note's Notable record.

The use case here is that I'm dealing with some support and want to add a note. I know a similar note will have been added in the past, so I want to base my note off that (for speed, consistency, checking I'm saying the right thing, etc). It's basically impossible to find the kind of note I'm thinking of without them in a list.

Ideally this will evolve to include the usual search & sort filtering, but even as-is it helps since I can just paginate to the end to see recent notes (which is likely where what I want will be).

Screenshot 2023-12-20 at 09 58 48

Screenshot paginates 1-per-page to check that it works.

### DIFF
--- a/app/controllers/admin/notes_controller.rb
+++ b/app/controllers/admin/notes_controller.rb
@@ -2,6 +2,10 @@ class Admin::NotesController < AdminController
   include Admin::TagHelper
   include TranslatableParams
 
+  def index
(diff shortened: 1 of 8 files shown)


garethrees:
Don't think this really needs review, but waiting until after the Christmas break to merge.

Issue #7996, opened 2023-11-06
Subject-based holding pen guesses show incorrect scores


garethrees:
Where the guess is based on subject the `id_score` / `idhash_score` shows as 100%, because we [set default values of `1`](https://github.com/mysociety/alaveteli/blob/0.44.0.0/app/models/guess.rb#L41-L49) if the guess doesn't have that info. This is misleading as it makes it look like subject-based requests are exact matches:

![Screenshot 2023-11-06 at 10 04 55](https://github.com/mysociety/alaveteli/assets/282788/e47bef31-d87b-46e6-a7ef-0b5c3d93ee73)

Related to https://github.com/mysociety/alaveteli/issues/7973

gbp:
> Any reason not to return `0` instead of `1`?

No, that should be fine afaik.

PR #8063, opened 2023-12-21
[#7996] Fix default Guess scores

garethrees:
The default score should be 0 - as in, no match – unless we actually have some sort of similarity on the attribute. Since these are guard clauses for when there is no attribute, we should return a value that implies there isn't a match.

Fixes https://github.com/mysociety/alaveteli/issues/7996

### DIFF
--- a/app/models/guess.rb
+++ b/app/models/guess.rb
@@ -39,13 +39,13 @@ class Guess
   end
 
   def id_score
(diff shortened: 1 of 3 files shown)


gbp (approved):
I don't think this will have any impact on the request guessing on receiving email so this looks fine to me.

PR #8073, opened 2024-01-02
Add XLS analyze job

gbp:
## Relevant issue(s)

Extracted from #8060 

## What does this do?

Adds background job to analyze old XLS attachments and hide them if hidden data is detected.

## Why was this needed?

Used to hide old potentially problematic attachments on WDTK.

## Implementation notes

Not sure we need this in develop but saving for posterity.

<hr>

[skip changelog]


gbp:
Not intending to merge this

PR #8096, opened 2024-01-12
[7786] WIP Remove locale prefix from URLs

### LINKED ISSUE #7786: Remove locale prefix from URLs and make session-based
Similar to https://github.com/mysociety/alaveteli/issues/7784, when dealing with support requests we often have to figure out multiple permutations of request (and associated content) URLs.

For example, given the request `/request/foo`, we might end up with:

* `/request/foo`
* `/cy/request/foo`
* `/request/123/response/456/attach/2/bar.doc`
* `/request/123/response/456/attach/2/bar.doc.html`
* `/request/123/followups/new/456`
* `/details/request/foo`
* `/feed/request/foo`

Consistently use the `url_title` will be handled in https://github.com/mysociety/alaveteli/issues/7784, and member routes handled in https://github.com/mysociety/alaveteli/issues/7785. This issue covers the locale prefix (`/cy/…`). I don't think there's a good reason to keep this in the URL. Instead, we can look it up via a value stored in the current session, which gets set to the default locale if empty, or populated with the currently signed in user's preference.

This ticket should:

- [ ] Replace the mechanism for setting the current locale from the URL prefix to a session value
- [ ] Ensure `User#locale` is set to the current locale on registration
- [ ] Update `User#locale` to their current preference as they switch locales
- [ ] Update the session locale to the `User#locale` on sign in
- [ ] Handle existing URLs with a redirect that sets the session value to the appropriate locale

alexander-griffen:
## Relevant issue(s)
Fixes #7786
## What does this do?
Removes the locale prefix from the URL, and instead uses the session value as the prioritised manner of setting the locale.
## Why was this needed?
Reduce the number or URL permutations per request in order to reduce admin burden
## Notes to reviewer

<hr>


gbp:
Transient spec failure, with `rspec spec/controllers/general_controller_spec.rb:117 spec/integration/localisation_spec.rb:27 --seed=62743`, fixed in https://github.com/mysociety/alaveteli/pull/8152

PR #8109, opened 2024-01-23
Treat unstemmed words as normal search terms.

dracos:
## Relevant issue(s)

https://github.com/mysociety/alaveteli/issues/4426

## What does this do?

has_normal_search_terms? was only returning true if the query had a stemmed search term in it, such as Zterm. Searching for a word with capital letters, such as NHS, translates to an unstemmed query term “nhs”. This would not match and the search would default to sorting by newest rather than relevance. Unstemmed query terms should still match as a normal search term, not a prefixed one.

## Why was this needed?

The default search results should not change their sort order based solely upon the capitalisation of the search provided.

### DIFF
--- a/doc/CHANGES.md
+++ b/doc/CHANGES.md
@@ -2,6 +2,7 @@
 
 ## Highlighted Features
 
+* Treat unstemmed words as normal search terms. (Matthew Somerville)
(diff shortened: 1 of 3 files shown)


gbp (approved):
Thanks for this. I have pushed some RSpec test examples showing the change.

PR #8120, opened 2024-02-05
[#5851] Review new framework defaults

### LINKED ISSUE #5851: Review new framework defaults

gbp:
## Relevant issue(s)

Fixes #5851

## What does this do?

Lazy review of the new framework defaults. Testing to see if the specs pass with the new defaults used.

## Why was this needed?

So we can upgrade Rails further.


gbp:
This is on deployed to staging. I'm thinking this might be good to merge but will see if there are issues when testing other features on there.

config/application.rb
```diff
@@ -31,6 +31,7 @@ class Application < Rails::Application
     #
     # config.time_zone = "Central Time (US & Canada)"
     # config.eager_load_paths << Rails.root.join("extras")
+    config.load_defaults = 7.0
```
gbp, line 34: This PR isn't doing anything - we need to call the `#load_defaults` method!

```suggestion
    config.load_defaults 7.0
```

Issue #7383, opened 2022-10-22
Not set request prominence reason to the notification's relevant text


WilliamWDTK:
It tends to read oddly.

It should be possible for the prominence reason to be determined automatically, but not by duplicating the varying part of the notification.

Alternatively, the prominence reason could be left blank, with the sample background text (can't think of the technical term right now, but the text that's not really 'there') matching what's given in the hint ("There are various reasons why we might have done this, sorry we can't be more specific here.").

I think the choice between the two will depend how much we want to have the prominence reason.

E.G.:

Invalid request:
Message
> **Your request on WhatDoTheyKnow hidden**
> Dear [NAME],
>
>Your request '[REQUEST NAME]' at [URL] has been reviewed by moderators.
>
>We consider it is not a valid FOI request, and have therefore hidden it from other users.
>
>You will still be able to view it while logged in to the site. Please reply to this email if you would like to discuss this decision further.
>
>Yours,
>
>The WhatDoTheyKnow team.

Prominence:
>We consider it is not a valid FOI request, and have therefore hidden it from other users.

It gets more confusing when the message refers to "you":
Personal correspondence:
Message
>**Your request on WhatDoTheyKnow hidden**
>Dear [NAME],
>
>Your request '[REQUEST NAME]' at [URL] has been reviewed by moderators.
>
>We do not allow requests for information via WhatDoTheyKnow about your personal circumstances. We have therefore hidden your request from other users.
>
>You will still be able to view it while logged in to the site. Please reply to this email if you would like to discuss this decision further.
>
>Yours,
>
>The WhatDoTheyKnow team.

Prominence reason:
>We do not allow requests for information via WhatDoTheyKnow about your personal circumstances. We have therefore hidden your request from other users.

WilliamWDTK:
Sorry for not knowing the relevant terminology.

Related to:
- #6746

HelenWDTK:
This issue has been automatically closed due to a lack of discussion or resolution for over 12 months.
Should we decide to revisit this issue in the future, it can be reopened.

PR #8121, opened 2024-02-05
Add variables to hidden user explanation templates

gbp:
## Relevant issue(s)

Needed to improve https://github.com/mysociety/whatdotheyknow-theme/pull/1823

## What does this do?

Add variables to hidden user explanation templates

## Why was this needed?

Allow templates to be improve the rendering of prominence reasons and/or explanations using conditionals to check these new variables so we can render different text for each field.


WilliamWDTK:
See also:

- #7383

PR #8123, opened 2024-02-09
[#8047] Add automatic parsing of emails with spreadsheets

### LINKED ISSUE #8047: Detect problems in spreadsheet attachments
Authorities making data breaches by including hidden data in Excel spreadsheets has been a [long-known](https://www.mysociety.org/2013/06/13/whatdotheyknow-team-urge-caution-when-using-excel-to-depersonalise-data/) issue. As WhatDoTheyKnow has grown we’ve seen increasing numbers of Excel attachments, and therefore Excel-based data breaches. We now know that our reporting of these to authorities and ICO has not improved behaviour. [Enough is enough](https://www.mysociety.org/2023/08/30/psni-data-breach/).

**Goal: Scan Excel files for potential problems and act appropriately if issues are detected.**

At a basic level, we want to automatically scan Excel files for known signs of hidden data and then alert appropriate people so that the problem can be handled. What this latter step looks like is TBC, so this issue focuses on the problem detection and recording of issues.

## Which files?

The key problem file type to target is XLSX, as that’s now the most commonly released.

🕳️⚠️ There are a variety of other Excel file formats, but they’re a tiny proportion of releases. The slight exception to this is XLS, which was the default but we’re seeing less each year. We could add a basic file size check here. It’s not worth jumping through hoops to add advanced checking mechanisms until we’ve got XLSX checking in operation.

## Checking for problems

At present we only create attachments the first time the IncomingMessage is viewed in a browser, which could be months after receiving the response. The reason for this is to even out processing load and not take up space extracting attachments that no one wants to see.

In the context of checking for problems, we should scan attachments as soon as possible after they’re received. That means we should “peek” inside the mail while we’re ingesting it and force a full parse of the raw email immediately if there’s a scannable file type attached:

    class IncomingMessage
      after_create :parse_raw_email!, if: -> { raw_email.contains_spreadsheet? }

    class RawEmail
      def contains_spreadsheet?
        mail.attachments.any? do
          |attachment| spreadsheet_extensions.include?(File.extname(attachment.filename)
        end
      end

🕳️⚠️ If “peeking” gets complicated and we can’t easily prevent publication prior to the analyser finishing, let’s discuss. We can probably live with it if there’s a minor chance that an attachment may be briefly publicly available.

We only want to automatically do this scanning the first time we receive a response. We may choose to force manual rescans in future – if we add enhanced checks – but what we don’t want is this automation to start firing if an attachment is recreated e.g. due to a digest mismatch, which happens from time to time as underlying libraries change. Ideally we’ll avoid any complexity by having this scanning process kick off from the creation of a RawEmail or IncomingMessage – that way we won’t have to maintain state about what’s been checked.

## What are the signs?

As well as a large file size, we have a [growing collection](https://github.com/mysociety/alaveteli/issues/2663#issuecomment-121943978) of knowledge about how to detect hidden content. We don’t need to ship with every possible check in place from the outset; we just want the key checks there and have the ability to add to them later.

gbp:
## Relevant issue(s)

Fixes #8047

## What does this do?

Add automatic parsing of emails with spreadsheets

## Why was this needed?

We want to parse these email automatically so the `Excel Analyzer` runs
and detects issues with hidden data.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -169,7 +169,7 @@ gem 'azure-storage', require: false
 gem 'google-cloud-storage', '~> 1.47', require: false
 
 # Storage content analyzers
-gem 'excel_analyzer', path: 'gems/excel_analyzer'
+gem 'excel_analyzer', path: 'gems/excel_analyzer', require: false
 
 group :test do
   gem 'fivemat', '~> 1.3.7'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -56,6 +56,8 @@ PATH
   specs:
     excel_analyzer (0.0.1)
       activestorage
+      mahoro
+      mail
       rubyXL
       rubyzip
 
--- a/config/initializers/excel_analyzer.rb
+++ b/config/initializers/excel_analyzer.rb
@@ -1,0 +1,1 @@
+require "excel_analyzer"
--- a/config/initializers/excel_analyzer.rb
+++ b/config/initializers/excel_analyzer.rb
@@ -1,1 +1,7 @@
 require "excel_analyzer"
+
+ExcelAnalyzer.on_spreadsheet_received = ->(raw_email_blob) do
+  incoming_message = IncomingMessage.joins(raw_email: :file_blob).
+    find_by(active_storage_blobs: { id: raw_email_blob })
+  incoming_message&.parse_raw_email!
+end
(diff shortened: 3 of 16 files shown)


gbp:
Tested working on staging

PR #8125, opened 2024-02-09
Update attachment body assignment

gbp:
## What does this do?

Update attachment body assignment

## Why was this needed?

When masking attachments avoid replacing the blob with a new blob, instead update which retains the blob key and existing metadata.

This means no `ActiveStorage::PurgeJob` or an additional `ActiveStorage::AnalyzeJob` for the new blob.


gbp:
Everything is working as expected on staging with this change.

PR #8130, opened 2024-02-12
8129 list attachments on info request admin page

### LINKED ISSUE #8129: List attachments on info request admin page
We have individual attachment hiding now, and it's sometimes time consuming to find the right incoming message admin link from the request admin page to take you to the attachment. It'd also be good for admins to get an overview of what (substantive) attachments are associated with a request.

HelenWDTK:
## Relevant issue(s)
Fixes #8129
## What does this do?
Adds attachment links on request admin page
## Why was this needed?
To make it easier to see attachments per request
## Screenshots
Screenshot 2024-02-12 at 21 58 22
Screenshot 2024-02-12 at 21 58 42

## Notes to reviewer

This will probably need fixing, but is a proof of concept

### DIFF
--- a/app/helpers/info_request_helper.rb
+++ b/app/helpers/info_request_helper.rb
@@ -335,4 +335,8 @@ module InfoRequestHelper
   def public_token?
     defined?(public_token) && public_token.present?
   end
+
+  def fetch_foi_attachments(info_request)
+    info_request.incoming_messages.includes(:foi_attachments).flat_map(&:foi_attachments)
+  end
 end
--- a/app/helpers/info_request_helper.rb
+++ b/app/helpers/info_request_helper.rb
@@ -337,6 +337,9 @@ module InfoRequestHelper
   end
 
   def fetch_foi_attachments(info_request)
-    info_request.incoming_messages.includes(:foi_attachments).flat_map(&:foi_attachments)
+  info_request.incoming_messages
+              .includes(:foi_attachments)
+              .flat_map(&:foi_attachments)
+              .reject(&:main_body_part?)
   end
 end
--- a/app/views/admin_request/show.html.erb
+++ b/app/views/admin_request/show.html.erb
@@ -372,6 +372,11 @@
 
 <hr>
 
+<h2> Attachments </h2>
+<%= render partial: 'admin_request/some_attachments' %>
+
+<hr>
+
 <h2>Annotations</h2>
 
 <%= render partial: 'admin_request/some_annotations' ,
--- a/app/views/admin_request/_some_attachments.html.erb
+++ b/app/views/admin_request/_some_attachments.html.erb
@@ -1,0 +1,27 @@
+<fieldset id="attachments" class="form-horizontal">
+  <% foi_attachments = fetch_foi_attachments(@info_request) %>
+  <% if foi_attachments.any? %>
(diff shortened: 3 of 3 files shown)


app/views/admin_request/show.html.erb
```diff
@@ -372,6 +372,11 @@
 
 <hr>
 
+<h2> Attachments </h2>
+<%= render partial: 'admin_request/some_attachments' %>
```
garethrees, line 376: Should be able to re-use the existing partial rather than duplicating it.

For example:

```erb
<%= render partial: 'admin_incoming_message/foi_attachments',
           locals: { foi_attachments: fetch_foi_attachments(@info_request) } %>
```

However, see above notes for tweaks to how we collect the attachments to render.

app/helpers/info_request_helper.rb
```diff
@@ -335,4 +335,11 @@ def attachment_params(attachment, options = {})
   def public_token?
     defined?(public_token) && public_token.present?
   end
+
+  def fetch_foi_attachments(info_request)
```
garethrees, line 339: At the moment this won't be very performant, and is kind of in the wrong place. Helpers are more for presentation logic than querying and computation.

I think we could replace most of this with a [`has_many :through` association](https://guides.rubyonrails.org/association_basics.html#the-has-many-through-association). That way Rails takes care of all the SQL joins and – I haven't tested this so I might have missed a gotcha – should allow us to do `@info_request.foi_attachments`.

We'd then want to filter out the main body part. That would be better as a [`scope`](https://guides.rubyonrails.org/active_record_querying.html#scopes) on `FoiAttachment`:

```rb
scope :substantive, -> { reject(&:main_body_part?) }
```

Then we could chain that on to our previous improvement and end up with `@info_request.foi_attachments.substantive`. We'd want to add a test for this as there's a little more logic involved than plugging together the Rails components.

We can then just call that directly when rendering the attachments:

```erb
<%= render partial: 'admin_incoming_message/foi_attachments',
           locals: { foi_attachments: @info_request.foi_attachments.substantive } %>
```

app/views/admin_request/show.html.erb
```diff
@@ -372,6 +372,11 @@
 
 <hr>
 
+<h2> Attachments </h2>
```
garethrees, line 375: I'd add a little help block here explaining that we're excluding the main body attachments:

```html
<h2>Attachments</h2>

<div class="span12 help-block">
  <p>Excludes main body part attachments.</p>
</div>
```

garethrees (changes requested):
Nice work! 👏 I've left some suggestions to get this to the point of being mergeable. It's pretty close, so I think should be something you can have a crack at if you're interested. If not LMK and I can clean it up to get it over the line 🐎

app/views/admin_request/show.html.erb
HelenWDTK: This causes ugliness due to the legend in the existing partial, but if we can live with that, then that's possible. 
Screenshot 2024-02-13 at 17 59 08

HelenWDTK:
This works if we can live with the double heading. I don't get the changelog stuff, so will leave that to someone else if this seems ok.

garethrees:
Closing for now; may revisit in future.

PR #8142, opened 2024-03-01
[7787] Always use /request/ prefix

### LINKED ISSUE #7787: Always use `/request/` prefix when displaying private requests
Similar to https://github.com/mysociety/alaveteli/issues/7784, when dealing with support requests we often have to figure out multiple permutations of request (and associated content) URLs.

While not a public URL, we currently generate  a different request URL for a request currently under a privacy period. For example, `/alaveteli_pro/info_requests/foo` in pro would become `/request/foo` when published.

This ticket should:

- [x] Update the routing so that private requests' canonical URL is `/request/:url_title`
- [x] Ensure the pro livery is still used on private requests
- [x] Ensure requests to the current format of URL (`/alaveteli_pro/info_requests/:url_title`) are permanently redirected to the correct URL.

alexander-griffen:
## Relevant issue(s)

Fixes #7787 

## What does this do?

This PR removes the Alaveteli Pro request path - `/alaveteli_pro/info_requests/:url_title` and ensures that the standard request path - `/request/:url_title` is followed instead.

## Why was this needed?

Simplification of admin procedures.

### DIFF
--- a/app/controllers/alaveteli_pro/classifications_controller.rb
+++ b/app/controllers/alaveteli_pro/classifications_controller.rb
@@ -10,7 +10,7 @@ class AlaveteliPro::ClassificationsController < AlaveteliPro::BaseController
     set_described_state
 
     flash[:notice] = _('Your request has been updated!')
-    redirect_to show_alaveteli_pro_request_path(
+    redirect_to show_request_path(
       url_title: @info_request.url_title
     )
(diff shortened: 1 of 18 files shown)


gbp:
Merged in 923a21ee39544149a0cd55790e79f87d84cb4055

PR #8150, opened 2024-03-05
Refactor localised admin forms

gbp:
## What does this do?

Refactor localised admin forms

## Why was this needed?

To reduce complexity and make the forms consistent with each other.

## Implementation notes

Switches admin forms with translated fields to use the `translated_form_for` form builder and `translated_fields` form helper. Means we don't have to manually build the tabbed interface for the different locales in use.

This is already in use for the `OutgoingMessage::Snippet` forms and can be re-used for `Announcement` `Category`, `PublicBody` and `Note` forms.

<hr>

[skip changelog]


garethrees:
Haven't checked thoroughly but looks good in principle!

PR #8152, opened 2024-03-06
Refactor spec configuration option assignment

gbp:
## What does this do?

Refactor spec configuration option assignment

## Why was this needed?

Some spec examples require certain configuration options to be set, this commit switches to stubbing the option instead of reassigning the option to the required value to get the test to pass.

This means we don't need to revert the value after the test example has been run back to the original value.

<hr>

[skip changelog]


garethrees:
nice ✨

Issue #8141, opened 2024-02-27
Project dataset export missing headers


garethrees:
See [user query](https://groups.google.com/a/mysociety.org/g/alaveteli/c/04Cbj3XmFDI/m/bMU7dFgJAwAJ).

`Project::Export` uses the first request in the project to [generate the CSV headers](https://github.com/mysociety/alaveteli/blob/0.44.0.0/app/models/project/export.rb#L33). However, if the request hasn't had data extracted, the headers for the dataset columns will be blank, even though other requests in the project may have had a dataset extraction.

```rb
project = Project.find(28)
# => #<Project:0x00007fc4d2992e58

Project::Export::InfoRequest.new(project, project.info_requests.first).data
# =>
# => {:request_url=>"https://www.whatdotheyknow.com/request/[REDACTED]",
# =>  :request_title=>"[REDACTED]",
# =>  :public_body_name=>"Blaenau Gwent County Borough Council",
# =>  :request_owner=>"[REDACTED]",
# =>  :latest_status_contributor=>"[REDACTED]",
# =>  :status=>"partially_successful",
# =>  :dataset_contributor=>nil}

Project::Export::InfoRequest.new(project, project.info_requests.last).data
# =>
# => {:request_url=>"https://www.whatdotheyknow.com/request/[REDACTED]",
# =>  :request_title=>"[REDACTED]",
# =>  :public_body_name=>"Wyre Forest District Council",
# =>  :request_owner=>"[REDACTED]",
# =>  :latest_status_contributor=>"[REDACTED]",
# =>  :status=>"partially_successful",
# =>  :dataset_contributor=>"[REDACTED]",
# =>  "Was this [REDACTED]"=>"1",
# =>  "Were [REDACTED]"=>"0",
# =>  "What reasons [REDACTED]"=>"[REDACTED]",
# =>  "How many years [REDACTED]"=>"[REDACTED]",
# =>  "Total number of [REDACTED]"=>"[REDACTED]",
# =>  "Total number of [REDACTED]"=>"[REDACTED]",
# =>  "How many [REDACTED]"=>"[REDACTED]",
# =>  "Was the [REDACTED]"=>"[REDACTED]"}
```

garethrees:
As a quick hack we could pick the hash with the most keys as the header:

```diff
  def to_csv
    CSV.generate do |csv|
-     header = data.first
+     header = data.sort_by { |h| h.keys.size }.last
      csv << header.keys.map(&:to_s) if header
[...]

PR #8159, opened 2024-03-11
Fix missing headers when exporting Project data

garethrees:
Project::Export uses the first request in the project to generate the CSV headers. However, if the request hasn't had data extracted, the headers for the dataset columns will be blank, even though other requests in the project may have had a dataset extraction.

This commit ensures that headers are always present by looking up the project's key set, and iterating through that to fetch the relevant submission value for that key, or otherwise assigning a nil value.

This has the benefit of ensuring that the keys/values exported are always in sync with the current project key set.

Fixes https://github.com/mysociety/alaveteli/issues/8141.

### DIFF
--- a/spec/models/project/export/info_request_spec.rb
+++ b/spec/models/project/export/info_request_spec.rb
@@ -6,6 +6,10 @@ RSpec.describe Project::Export::InfoRequest do
   include LinkToHelper
 
   let(:project) { FactoryBot.build(:project) }
+
+  let(:key_set) { FactoryBot.build(:dataset_key_set, resource: project) }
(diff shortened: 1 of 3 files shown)


spec/models/project/export/info_request_spec.rb
```diff
@@ -71,7 +75,8 @@
           request_owner: info_request.user.name,
           latest_status_contributor: contributor.name,
           status: info_request.described_state,
-          dataset_contributor: nil
+          :dataset_contributor => nil,
```
gbp, line 78:
```suggestion
          dataset_contributor: nil,
```

gbp (approved):
Nice one. This works well

PR #8197, opened 2024-04-11
[#7812] Allow requests to be browsed by tag

### LINKED ISSUE #7812: Allow requests to be browsed by tag
Create a `/list/:tag` route for `InfoRequest` records analogous to `/body/list/:tag` that lists requests with the given tag.

For an easy first version of the UI we can just modify the search term heading to include the tag that we're filtering on.

Screenshot 2023-07-04 at 16 57 27

gbp:
## Relevant issue(s)

Fixes #7812

## What does this do?

Allow requests to be browsed by tag

## Why was this needed?

So we can view tagged requests and make topic pages.

## Screenshots

image

### DIFF
--- a/app/controllers/request_controller.rb
+++ b/app/controllers/request_controller.rb
@@ -130,12 +130,21 @@ class RequestController < ApplicationController
 
     @filters = params.slice(:query, :request_date_after, :request_date_before)
     @filters[:latest_status] = params[:view]
+    @tag = @filters[:tag] = params[:tag]
 
     @results = InfoRequest.request_list(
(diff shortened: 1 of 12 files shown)


gbp:
Have forced push some updates so the existing form filters now keep the current tag.

This gives us URLs in the form of `/list/climate/successful`, `/list/climate/awaiting` etc...

garethrees:
This all looks great to me!

Would it be easy to change the heading to just the category title?

Screenshot 2024-04-17 at 11 24 49

PR #8223, opened 2024-04-30
[#8132] Update actions and pages which set "noindex", "nofollow" crawler directives

### LINKED ISSUE #8132: Reduce external search indexing of request list pages
The main things we want indexed are record pages themselves (info request pages, user pages, authority pages, etc).

Snippets of request content often appear on list pages, and create a whack-a-mole situation when unhappy users find that external search engines have indexed a list page (e.g. `/body/foo?page=12`) that contains a cached snippet of PII that we've removed from the request page itself.

We should stop indexing of:

- [ ] Request list pages (`/list/all`, `/list/successful`, etc) with a `page=` query param
- [ ] Similar requests page (`/request/:url_title/similar`)
- [ ] Body pages with a page= query param (`/body/:url_name?page=N`)
- [ ] User pages with a page= query param (`/user/:url_name?page=N`)
- [ ] User "wall" page (`/user/:url_name/wall`)

We might be able to do this via `robots.txt`, or could set via the `X-Robots-Tag` header depending on the page number:

```rb
before_action :set_no_crawl_headers, if: -> { params[:page].to_i > 1 }
```

gbp:
## Relevant issue(s)

Fixes #8132

## What does this do?

Update actions and pages which set "noindex", "nofollow" crawler directives

## Why was this needed?

> Snippets of request content often appear on list pages, and create a whack-a-mole situation when unhappy users find that external search engines have indexed a list page (e.g. /body/foo?page=12) that contains a cached snippet of PII that we've removed from the request page itself.

## Implementation notes

@garethrees are you happy with changing the number of paginated pages which are indexed? I'm concerned this might impact search ranking due to newer request pages not being indexed at all.

### DIFF
--- a/app/controllers/citations_controller.rb
+++ b/app/controllers/citations_controller.rb
@@ -5,6 +5,7 @@ class CitationsController < ApplicationController
   before_action :authenticate, except: :index
   before_action :load_resource_and_authorise, except: :index
   before_action :set_in_pro_area, except: :index
+  before_action :set_no_crawl_headers, except: :index
 
   skip_before_action :html_response, only: :index
 
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -33,6 +33,7 @@ class ApplicationController < ActionController::Base
   include FastGettext::Translation # make functions like _, n_, N_ etc available)
   include AlaveteliPro::PostRedirectHandler
   include ReadOnly
+  include RobotsHeaders
 
   # NOTE: a filter stops the chain if it redirects or renders something
   before_action :html_response
--- a/app/controllers/attachment_masks_controller.rb
+++ b/app/controllers/attachment_masks_controller.rb
@@ -32,10 +32,6 @@ class AttachmentMasksController < ApplicationController
 
   private
 
-  def set_no_crawl_headers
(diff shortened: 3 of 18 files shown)


garethrees (commented):
Worth making the dependency on `RobotsHeaders` explicit in the concerns that use it:

```diff
  module ProminenceHeaders
    extend ActiveSupport::Concern
+    include RobotsHeaders

  module PublicTokenable
    extend ActiveSupport::Concern
+    include RobotsHeaders
```

https://api.rubyonrails.org/classes/ActiveSupport/Concern.html says you can do this (in the intro section of the documentation)

garethrees:
Reassigning as discussed to make a few tweaks.

app/controllers/citations_controller.rb
```diff
@@ -5,6 +5,7 @@ class CitationsController < ApplicationController
   before_action :authenticate, except: :index
   before_action :load_resource_and_authorise, except: :index
   before_action :set_in_pro_area, except: :index
+  before_action :set_no_crawl_headers
```
garethrees: Needs `except: :index`

garethrees (commented):
1 minor change but LGTM after that

PR #8246, opened 2024-05-08
Change docker/setup to not require bash

### LINKED ISSUE #8244: Erronous shebang in docker/reset
The shebang is #!/bin/sh, but on my Ubuntu 22.04 the script does not run with bash in sh mode.

I see three solutions, which is why I do not make a pull request:

1. Change the shebang to #!/bin/bash
2. Change the script so that it runs in sh mode
3. Change the documentation to do "bash docker/reset" rather than "docker/reset" (and probably remove the shebang).

andersjl:
## Relevant issue(s)

Fixes #8244

## What does this do?

A simple rewrite of docker/setup script. No new behaviour.

## Why was this needed?

See #8244

## Notes to reviewer

Found no established way to test the project's scripts, so wrote no test.

### DIFF
--- a/docker/reset
+++ b/docker/reset
@@ -6,7 +6,7 @@ cd "$(dirname "$0")/.."
 
 docker compose down
 
-ARGS=()
+ARGS=""
 DB_VOLUME="$(basename $(pwd))_postgres"
 if docker volume ls | grep $DB_VOLUME >/dev/null; then
   while true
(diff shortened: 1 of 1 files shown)


gbp (approved):
Not tested on Ubunutu but this seems to work fine on my macOS machine. Thanks for the fixes.

gbp:
Merged 75a336b73e93945a56f100c0dfa5dfaa4000f03b

PR #8280, opened 2024-05-29
[#1636] Show AlaveteliConfiguration in admin interface

### LINKED ISSUE #1636: Add page to admin interface for displaying the configured values in the config file
Similar to what's in FMS.

garethrees:
## Relevant issue(s)

Fixes #1636

## What does this do?

Extract `AdminGeneralController#debug` to `Admin::DebugController#index`, tidy up its view, and render a sanitised version of`AlaveteliConfiguration` so that admins can view the configuration without a server ssh session.

## Why was this needed?

Helps less technical users and admins of Alavetelis hosted by mySociety.

Playing around with speaking at chatGPT after work because I was annoyed at having to ssh to figure out the `AlaveteliConfiguration.reply_very_late_after_days` value.

## Implementation notes

Weird issue with the routing – will try to resolve that before merge

## Screenshots

Screenshot 2024-05-29 at 18 48 22

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -547,7 +547,15 @@ Rails.application.routes.draw do
   end
   ####
 
+  #### Admin::Debug controller
+  namespace :admin do
+    # FIXME: For some reason the resources call is generating the route as
+    # admin_debug_index_path rather than the standard admin_debug_path.
+    # resources :debug, only: [:index]
+    get 'debug', to: 'debug#index', as: :debug
+  end
   ####
+
   #### AdminTag controller
   namespace :admin do
     resources :tags, param: :tag, only: [:index, :show],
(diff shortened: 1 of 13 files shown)


config/routes.rb
```diff
@@ -526,7 +526,15 @@ def matches?(request)
   end
   ####
 
+  #### Admin::Debug controller
+  namespace :admin do
+    # FIXME: For some reason the resources call is generating the route as
+    # admin_debug_index_path rather than the standard admin_debug_path.
+    # resources :debug, only: [:index]
```
gbp, line 533: This is to be expected when using the singlar form of a resource. Obviously we don't want route to be `/admin/debugs` tho, I would be fine with having to use `admin_debug_index_path`.

PR #8336, opened 2024-07-25
[#8251] Project dataset publishing

### LINKED ISSUE #8251: Ability to Publish Project Dataset
The first step to making Projects more public will be the ability to publish the Dataset created through the Project.

To do this we’ll basically create a web view of the Dataset CSV download.

In the future we might want to apply Embargo mechanics to Projects, but for now a boolean state is fine. We should have UI that makes the visibility state clear. For now owners should be able to toggle it at will. _Only the dataset will be public_; not the main Project interface.

If published, anyone should be able to also download the CSV version.

We can add an optional rich text description to the Dataset so that Project owners can give context to the data.

Screenshot 2024-05-10 at 17 52 08

gbp:
## Relevant issue(s)

Fixes #8251

## What does this do?

Adds project dataset view for project owners and allows them to specify a description and make the dataset public.

## Why was this needed?

Allows us to show good usage of FOI and the Pro/Projects tooling. 

## Implementation notes

Commits 48943c5...0097583 are minor changes and refactoring in preporation for the main feature in the following commits.

## Screenshots

As the project owner:
image
image

Public view:
image

Table content overflow:
https://github.com/user-attachments/assets/35ae18a6-d96a-4f98-999a-fc8127dafae7


app/views/projects/dataset/edit.html.erb
```diff
@@ -0,0 +1,46 @@
...
+      <%= form_for @project, url: project_dataset_path(@project) do |f| %>
+        <div class="form_input">
+          <%= f.label :dataset_description, _('Description'), class: 'form_label' %>
+          <%= f.rich_text_area :dataset_description %>
+        </div>
+
+        <div class="form_input">
+          <%= f.label :dataset_public, _('Visibility'), class: 'form_label' %>
+
+          <div class="radio_buttons">
```
garethrees, line 28: Since the default is "Private", think it would be best to swap the order here so that the Private radio button appears first (and is selected by default), as (at least for most alavetelis) the ordering of left to right is more natural.

At present, the default state feels a bit weird. It feels like the right-most radio button is the "option" rather than the "default".

image

I'd expect this, where the selected default appears first, with the option value second and unselected by default.

image

garethrees (commented):
Haven't reviewed the code side thoroughly as the demo looked great. Two really minor tweaks to UI and then let's get this merged.

PR #8366, opened 2024-08-16
[#8365] Add special not_requestable tag

### LINKED ISSUE #8365: Add generic `not_requestable` special tag
We have a set of special tags that when applied to a `PublicBody` modify the behaviour.

Currently only `not_apply` and `defunct` prevent requests, and display an appropriate message to users for those situations.

What we don't have is a generic way of preventing requests to an authority when those tags aren't quite appropriate.

We should add a `not_requestable` tag that displays a generic message that users can't make requests via the platform to this authority. We might want to caveat that this might not mean that they couldn't make requests via other means.

We can then use notes to elaborate on the specific reasons we've tagged the authority as such.

gbp:
## Relevant issue(s)

Fixes #8365

## What does this do?

Add special not_requestable tag

## Notes to reviewer

<hr>

### DIFF
--- a/app/helpers/public_body_helper.rb
+++ b/app/helpers/public_body_helper.rb
@@ -18,6 +18,10 @@ module PublicBodyHelper
       reasons.push _('Freedom of Information law does not apply to this authority, so you cannot make a request to it.')
     end
 
+    if public_body.not_requestable?
+      reasons.push _('We are unable to make requests to this authority.')
+    end
+
     unless public_body.has_request_email?
       # Make the authority appear requestable to encourage users to help find
       # the authority's email address
--- a/app/models/public_body.rb
+++ b/app/models/public_body.rb
@@ -68,7 +68,11 @@ class PublicBody < ApplicationRecord
 
   # Any PublicBody tagged with any of the follow tags won't be returned in the
   # batch authority search results or batch category UI
-  cattr_accessor :batch_excluded_tags, default: %w[not_apply defunct]
+  cattr_accessor :batch_excluded_tags, default: %w[
+    defunct
+    not_apply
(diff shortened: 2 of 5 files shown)


app/models/public_body.rb
```diff
@@ -364,10 +375,12 @@ def site_administration?
 
   # Can an FOI (etc.) request be made to this body?
   def is_requestable?
-    has_request_email? && !defunct? && !not_apply?
+    has_request_email? && !defunct? && !not_apply? && !not_requestable?
   end
 
-  scope :is_requestable, -> { with_request_email.not_defunct.foi_applies }
+  scope :is_requestable, -> {
+    with_request_email.not_defunct.foi_applies.requestable
```
garethrees, line 382: `is_requestable` and `requestable` scopes feel quite prone to confusion. I'd be inclined to go with:

```rb
with_request_email.not_defunct.foi_applies.without_tag('not_requestable')
```

app/helpers/public_body_helper.rb
```diff
@@ -18,6 +18,10 @@ def public_body_not_requestable_reasons(public_body)
       reasons.push _('Freedom of Information law does not apply to this authority, so you cannot make a request to it.')
     end
 
+    if public_body.not_requestable?
+      reasons.push _('We are unable to make requests to this authority.')
```
garethrees, line 22: Not 100% on the phrasing here but don't have an immediate suggestion. @HelenWDTK any thoughts?

garethrees:
Should also add to the UI help https://github.com/mysociety/alaveteli/blob/develop/app/views/admin_public_body/_form.html.erb#L17-L24

Issue #8369, opened 2024-08-29
Memory leak viewing all events in admin timeline


gbp:
```
/admin/timeline?event_type=info_request_event&start_date=1970-01-01T00%3A00%3A00Z
CONSUME MEMORY: 5547600 KB

/admin/timeline?start_date=1970-01-01T00%3A00%3A00Z
CONSUME MEMORY: 5615192 KB

/admin/timeline?event_type=info_request_event&start_date=1970-01-01T00%3A00%3A00Z
CONSUME MEMORY: 5616836 KB

/admin/timeline?start_date=1970-01-01T00%3A00%3A00Z
CONSUME MEMORY: 5637360 KB
```

PR #8370, opened 2024-08-29
[#8369] Remove "All time" filter from admin timeline

garethrees:
On WhatDoTheyKnow this loads 17 million+ events, which attempts to use over 5GB of memory.

In all but the smallest installs pagination will be required for this page to be performant, so this commit removes the filter entirely.

Fixes https://github.com/mysociety/alaveteli/issues/8369.

BEFORE

Screenshot 2024-08-29 at 16 54 26

AFTER

Screenshot 2024-08-29 at 16 54 18


gbp:
Closing in favour of #8458

PR #8383, opened 2024-09-13
Configure hash style

gbp:
Allow shorthand syntax but allow key values need to be consistent.

Good:
```
  { foo: 'foo', bar: 'bar' }
```
Good:
```
  foo = 'foo'; bar = 'bar'
  { foo:, bar: }
```
Bad:
```
  foo = 'foo'
  { foo:, bar: 'bar' }
```

<hr>

[skip changelog]


gbp:
This was getting annoying, the linter in my editor keeps saying "Omit hash value".

PR #8391, opened 2024-09-23
Layout and spacing improvements to popup banner

lucascumsille:
## Relevant issue(s)
Working on another ticket I noticed the markup for the `popup` and realised that there is a close button, however is not visible enough(opacity and size) and the clickable area is also quite small.

Screenshot 2024-09-23 at 07 28 22

## What does this do?

- Increases the size of the button
- Get rids of the opacity
- It also positions the close button vertically in the middle, specially on smaller screen it think it looks nicer.
- The content has being centred. For this type of one liner messages it looks a bit nicer and for having a different alignment than the rest of the page it pops up a bit more, but happy if we want to keep it the left aligned.

## Screenshots

https://github.com/user-attachments/assets/ee75629d-91f0-40be-baa7-57748480a2a7

## Notes to reviewer

<hr>

[skip changelog]

### DIFF
--- a/app/assets/stylesheets/responsive/_popups_layout.scss
+++ b/app/assets/stylesheets/responsive/_popups_layout.scss
@@ -9,20 +9,24 @@
 .popup__close {
   position: absolute;
   display: inline-block;;
-  top: -0.25em;
+  top: 50%;
+  transform: translateY(-50%); // positions the icon vertically in the middle of the banner.
   right: 0.9375rem; // equal to the padding of .popup minus padding
   z-index: 50;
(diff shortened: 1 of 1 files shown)


gbp:
@lucascumsille there is [this recent issue](https://github.com/mysociety/whatdotheyknow-theme/issues/1920) related to the popup if you're able to adjust the colour too that would be amazing.

lucascumsille:
thanks @gbp I'm doing it at the moment, but in the WDTK repo, I noticed the styling is in there.

gbp (approved):

PR #8397, opened 2024-09-30
Dataset select

### LINKED ISSUE #8117: Datasets: Add select and multi-select fields
Add select field to `Dataset::Key::FORMATS`

We're often getting initial question sets where the project owner wants to supply a preset list of options.

gbp:
## Relevant issue(s)

Fixes #8117

## What does this do?

Adds a project question select format

## Why was this needed?

Projects owners want to be able to provide a list of options for contributors to pick from.

## Screenshots

image
image
image
image

### DIFF
--- a/app/assets/stylesheets/responsive/_global_style.scss
+++ b/app/assets/stylesheets/responsive/_global_style.scss
@@ -191,7 +191,7 @@ form button {
   padding: 2px 6px 3px;
   transition: none;
   &:hover, &:focus {
-    background: none;
+    background-color: inherit;
     color: inherit;
   }
 }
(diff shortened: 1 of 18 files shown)


garethrees (changes requested):
Looks like some issues with adding select options:

projects-select-2

projects-select-remove-add

Seems to add multiple questions back if removing then re-adding a select:

projects-select-remove-question


Also when submitting a single option select. Looks like the related code is expecting an array rather than a string:

projects-select-submit

Guess it probably needs:

```diff
     when 'select'
-      super(new_value.to_a.reject(&:blank?).join(', '))
+      super(Array(new_value).reject(&:blank?).join(', '))
```

General workflow and publishing/download looks great.

gbp:
Above issues should be solved in the pushed fixups.

garethrees (commented):
no need to re-review if issues mentioned resolved

PR #8400, opened 2024-10-04
[#8314] Public citations view

### LINKED ISSUE #8314: Citations: Public List
# Citations: Public List

Let’s showcase the growing list of citations in public! This will help further inspire and demonstrate how ATI can be used.

## Before we start: Academic → Research

We [previously](https://github.com/mysociety/alaveteli/pull/8126) generalised the types of citation. One further improvement that we should make before making these more public is renaming “Academic” to “Research”. This encompasses a broader set of similar work, and I think will be less off-putting to less formal links that we’d still be very interested in.

https://github.com/mysociety/alaveteli/commit/84cf979e018683a114800150d235c3e6385052cb has an example of what needs to be done here.

BEFORE: 🎓 Academic
AFTER:  📚 Research

## The Citations Page

We’ll basically mirror the admin view at `/admin/citations` in a user-friendly way.

We’ll start off with a header and a bit of blurb about what citations are and why it’s useful for us to track them, and then a short description of each of the citation types.

Screenshot 2024-07-10 at 12 59 30

I’m not over the moon with “Citations” as the main name for these, so open to options about how we communicate these in general. Loop @MyfanwyNixon in when we've got the bones of this in place and start thinking of the copy we need ([the intro blog post could be useful here](https://www.mysociety.org/2020/01/14/has-your-foi-request-been-used-in-a-news-story-now-you-can-let-everyone-know/)).

We’ll then render a paginated list of citations, 50 per page, ordered by most recent first.

We’ll just show the `Citation#title` if present, falling back to `Citation#source_url`, pointing at the `InfoRequest#title` and then the `PublicBody#name`.

Not rendering the names of requesters or the user who added the Citation helps reduce the chances of PII data indexing issues.

Where a `Citation#description` is present, render it underneath the Citation in relatively small/unobtrusive text, perhaps in a `<summary>` tag.

## RSS / JSON

We should also add a way to programmatically access the feed of citations. JSON or RSS is fine – whichever makes most sense.

## Citation Title & Description

Two small additions we’ll make as part of this project is to add some building blocks for a more human friendly display of the `Citation`, rather than the potentially long and ugly `source_url`.

**`title`:** A `String` attribute that we can use to display the citation. e.g. instead of rendering the `source_url` of `https://www.mysociety.org/2020/01/14/has-your-foi-request-been-used-in-a-news-story-now-you-can-let-everyone-know/`, we can populate the `title` attribute with `“Has your FOI request been used in a news story? Now you can let everyone know.”` and render that instead.

**`description`:** A `Text` attribute that we can later use to add a summary or excerpt of what the Citation is about. We should limit the length of this to about a paragraph of text.

*We won’t expose these fields on the Citation creation form*; only in the admin interface. These will be placeholders for future work – using AI to generate a description and fetching the `source_url`’s `<title>` attribute for example. For now, we’re just getting them in place and rendering nicely if we’ve manually added them via the admin interface.

gbp:
## Relevant issue(s)

Fixes #8314

## What does this do?

1. Add public citations view
2. Adds `Citation#title` and `description` fields

## Why was this needed?

Shows impact of FOI in action.

## Screenshots

image

### DIFF
--- a/app/assets/stylesheets/responsive/_citations.scss
+++ b/app/assets/stylesheets/responsive/_citations.scss
@@ -1,0 +1,29 @@
+.citations-types {
+  border-top: 1px solid #d1d1d1;
+  border-bottom: 1px solid #d1d1d1;
+  margin-top: 2em;
+  margin-bottom: 1em;
+}
+
+ul.citations-list {
+  list-style: none;
+  padding-left: 0;
+
+  li.citations-list__citation {
+    padding: 1em 0;
+    border-bottom: 1px solid #d1d1d1;
+
+    display: flex;
+    align-items: flex-start;
+    gap: 1em;
+
+    .citation_icon {
+      flex-shrink: 0;
+    }
+
+    .citation__title {
+      flex-basis: 50%;
+      word-break: break-word;
+    }
+  }
+}
--- a/app/assets/stylesheets/responsive/_citations.scss
+++ b/app/assets/stylesheets/responsive/_citations.scss
@@ -25,5 +25,9 @@ ul.citations-list {
       flex-basis: 50%;
       word-break: break-word;
     }
+
+    .citation__description {
+      flex-basis: 50%;
+    }
   }
 }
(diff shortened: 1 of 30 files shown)


gbp:
Merged in 96773153a

PR #8405, opened 2024-10-09
Stripe API upgrade

gbp:
## What does this do?

Upgrades the Stripe API version used. 

## Why was this needed?

To hopefully help in migrating from Stripe plans to prices which in turn will help us add new prices and change existing subscriptions.

### DIFF
--- a/doc/CHANGES.md
+++ b/doc/CHANGES.md
@@ -2,6 +2,7 @@
 
 ## Highlighted Features
 
+* Upgrade Stripe API version (Graeme Porteous)
 * Drop support for Azure storage (Graeme Porteous)
 * Add basic Citation searching in admin UI (Gareth Rees)
 * Improve citations admin to allow title and description updates (Graeme
@@ -139,6 +140,10 @@
   files. Please ensure you review the changes to `config/nginx.conf.example` or
   `config/nginx-ssl.conf.example` and update your production configuration if
   needed.
+
+* _Note:_ If you have Pro pricing enabled, this release changes the Stripe API
+  version from `2017-01-27` to `2020-03-02`. No changes should be necessary to
+  your Stripe account.
 
 # 0.44.0.1
 


gbp:
https://docs.stripe.com/upgrades#2019-03-14
> There are a few changes to the [invoice object](https://stripe.com/docs/api/invoices):
> - The date property has been renamed to created.

Need to change https://github.com/mysociety/alaveteli/blob/fb4d9f5a385179f3c2fb5f1039a27c854f102140/app/views/alaveteli_pro/invoices/_invoice.html.erb#L2 and https://github.com/mysociety/alaveteli/blob/fb4d9f5a385179f3c2fb5f1039a27c854f102140/app/models/alaveteli_pro/invoice.rb#L17

Should also add RSpec to catch this.

gbp:
https://docs.stripe.com/upgrades#2018-02-05
> **Major**: Each plan object is now linked to a product object .... The plan object ... and **name** attributes have been [moved to product](https://docs.stripe.com/api#product_object-statement_descriptor) objects, and plan objects now have a nickname attribute.

Need to change https://github.com/mysociety/alaveteli/blob/fb4d9f5a385179f3c2fb5f1039a27c854f102140/app/views/alaveteli_pro/plans/show.html.erb#L32 and https://github.com/mysociety/alaveteli/blob/fb4d9f5a385179f3c2fb5f1039a27c854f102140/app/views/alaveteli_pro/subscriptions/_subscription.html.erb#L4

Should also add test examples to catch this.

gbp:
https://docs.stripe.com/upgrades#2018-07-27
> The percent_off field of coupons was changed from Integer to Float, with a precision of two decimal places.

Need to check the calculations in https://github.com/mysociety/alaveteli/blob/fb4d9f5a385179f3c2fb5f1039a27c854f102140/app/models/alaveteli_pro/subscription_with_discount.rb#L74-L80

gbp:
The above comments have been addressed. While I can write failing tests for them (stripe-ruby-mock doesn't enforce valid attributes, and it does [expand](https://docs.stripe.com/expand) Stripe resources) I have ensured they are all fixed.

app/controllers/alaveteli_pro/stripe_webhooks_controller.rb
```diff
@@ -60,10 +60,9 @@ def invoice_payment_succeeded
       subscription = Stripe::Subscription.retrieve(subscription_id)
       plan_name = subscription.plan.name
```
dracos, line 61: Should this be plan.product.name (expanding product), if you're deciding to use that everywhere, to match the other change?

dracos (changes requested):
One question about a `plan.name` that's left. There's also one more mention in app/models/webhook.rb that could be updated, but I guess you'd have to call something else to get the product name (you get the nickname in the same blob) - the webhook includes the new data as well as the old, I've just checked (e.g. https://dashboard.stripe.com/events/evt_1QA8dQLoAAr9vgdbpgEl1nZm ) so you could switch there, but might not be worth it.

app/controllers/alaveteli_pro/stripe_webhooks_controller.rb
gbp: Pushed [eb0f277](https://github.com/mysociety/alaveteli/pull/8405/commits/eb0f2770a2097d2157ed5d91146c74e4cb09e5e0) to change this. Annoyingly the Stripe mocking gem we're using doesn't "expand" resources so I've needed to stub this in some RSpec examples.

dracos (approved):

dracos:
Sorry, after reviewing the last PR (incoming soon), I have a question - the commit message here says "Bumps the API version to the latest version support by stripe-ruby-mock." - but according to https://github.com/stripe-ruby-mock/stripe-ruby-mock/blob/master/CHANGELOG.md it supports up to SDK v11 which is the most recent Stripe SDK? Just checking I hadn't missed something there. Think the only major change since the one you're upgrading to here is the change from tax_percent to tax rate IDs.

gbp:
In the `stripe-ruby-mock` readme they [specify their target API version](https://github.com/stripe-ruby-mock/stripe-ruby-mock?tab=readme-ov-file#specifications).

I tried later versions of the Stripe API version but unfortunately it resulted in so many broken specs.

The change to allow SDK v11 is just a change to relax the version requirement on the main Stripe ruby client gem. It does nothing to actually support those changes.

PR #8407, opened 2024-10-10
Stripe refactorings

gbp:
## What does this do?

Refactor Stripe wrappers classes and minor template updates.

## Why was this needed?

Help with future migration to Stripe Prices.

<hr>

[skip changelog]


app/helpers/alaveteli_pro/plan_helper.rb
```diff
@@ -3,16 +3,43 @@
...
+    elsif interval(plan) == 'year' && interval_count(plan) == 1
       _('Billed: Annually')
+    else
+      _('Billed: every {{interval}}', interval: interval(plan))
     end
   end
+
+  def billing_interval(plan)
+    if interval_count(plan) == 1
+      _('per user, per {{interval}}', interval: pluralize_interval(plan))
```
dracos, line 22: Given you've already checked interval_count, this could only display  'plan' directly without calling pluralize_interval

app/controllers/alaveteli_pro/subscriptions_controller.rb
```diff
@@ -196,36 +189,15 @@ def check_has_current_subscription
     redirect_to pro_plans_path
   end
 
-  def non_namespaced_plan_id
```
dracos, line 202: Think this could be removed in the previous commit, 691a1f4d5583d3fce4d72bb20f1e486711342652

app/models/alaveteli_pro/subscription/discount.rb
```diff
@@ -0,0 +1,73 @@
...
+  include Taxable
+
+  tax :discounted_amount
+
+  def discounted_amount
+    plan.amount - reduction
+  end
+
+  def discounted?
+    discounted_amount < plan.amount
```
dracos, line 27: This could be reduced to reduction > 0, presumably.

dracos (commented):
Looks good :) I left a few tiny comments is all.

PR #8408, opened 2024-10-11
[#7293] Update Notable#all_notes

### LINKED ISSUE #7293: Notes set against a `key` do not render when `key:value` is present
On WhatDoTheyKnow, we're considering using the new generalised notes functionality to display messaging on requests, e.g. where a generic event has occurred that effects multiple requests, or where a _right to be forgotten_ indication has arrived from a search engine.

The best way for us to do this, would be to present both a key (e.g. `rtbf`) and a value (e.g. `GDPR/SE/20220910-1`) so that we can tie our internal records together.

It **is** possible to render a note on a request if a key is presented **but** if key _and_ value are presented, a note linked to the key tag **doesn't render on a request**.

### How to reproduce

1. Create a new request in a test authority
2. If one doesn't already exist, create a new note linked to a tag (e.g. `rtbf`)
3. Add the tag to the request via /admin, and note that on the request view, the tag is now present.
4. Amend the tag on the request to key:value format, e.g. `rtbf:test`.
5. Refresh the public view of the request, and note the tag is no longer rendered.

### Expected behaviour
Note should display wherever the tag matches a key that has a note assigned.

### Actual behaviour
Adding the value to the tag removes the note.

### Reproducibility
I have tested this on WDTK requests 651067 and 896135, using note 8296. The former is embargoed, the latter is not. I used [note 8296](https://www.whatdotheyknow.com/admin/tags/rtbf?model_type=InfoRequest), setup against tag `rtbf`.

The issue is always reproducible; it is possible to reverse it by removing `:value` from the tag, but re-adding it will cause the issue to re-appear. This does not appear to be related to a cache.

### Curio
After a value is added to the tag on a request, the request still shows on /admin/tags at key level. E.g. if I tag a request in the format above, it will still show on the admin page for the `rtbf` tag (under requests).

gbp:
## Relevant issue(s)

Fixes #7293

## What does this do?

Update Notable#all_notes

## Why was this needed?

Update the concern so that records tagged with `name:value` will be associated with notes tagged as `name`.

### DIFF
--- a/spec/models/concerns/notable_and_taggable.rb
+++ b/spec/models/concerns/notable_and_taggable.rb
@@ -4,12 +4,31 @@ RSpec.shared_examples 'concerns/notable_and_taggable' do |factory_opts|
   describe '#all_notes' do
     subject { record.all_notes }
 
-    before { record.tag_string = 'foo' }
-
     let!(:tagged_note) { FactoryBot.create(:note, notable_tag: 'foo') }
     let!(:other_tagged_note) { FactoryBot.create(:note, notable_tag: 'bar') }
 
-    it { is_expected.to include tagged_note }
-    it { is_expected.to_not include other_tagged_note }
+    let!(:tagged_note_with_value) do
+      FactoryBot.create(:note, notable_tag: 'foo:1')
+    end
+
+    let!(:tagged_note_with_other_value) do
+      FactoryBot.create(:note, notable_tag: 'foo:2')
+    end
+
+    context 'with name tag' do
+      before { record.tag_string = 'foo' }
+      it { is_expected.to include tagged_note }
+      it { is_expected.to_not include tagged_note_with_value }
+      it { is_expected.to_not include other_tagged_note }
+      it { is_expected.to_not include tagged_note_with_other_value }
+    end
+
+    context 'with name:value tag' do
+      before { record.tag_string = 'foo:1' }
+      it { is_expected.to include tagged_note }
+      it { is_expected.to include tagged_note_with_value }
+      it { is_expected.to_not include other_tagged_note }
+      it { is_expected.to_not include tagged_note_with_other_value }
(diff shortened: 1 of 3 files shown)


spec/models/concerns/notable_and_taggable.rb
```diff
@@ -4,12 +4,20 @@
   describe '#all_notes' do
     subject { record.all_notes }
 
-    before { record.tag_string = 'foo' }
+    before { record.tag_string = 'foo:1' }
```
garethrees, line 7: might be worth making this an additional spec so that we're testing both `foo` and `foo:1`?

garethrees (commented):
haven't looked in detail but looks good and be good to have this working

spec/models/concerns/notable_and_taggable.rb
gbp: Yep good shout, I have added this in a fixup.

Before this PR we had:
- PublicBody tagged with "foo:bar" would be associated with a note with tagged with "foo:bar" but not "foo"
- PublicBody tagged with "foo" would be associated with a note with tagged with "foo" but not "foo:bar"

After this PR:
- PublicBody tagged with "foo:bar" would be associated with a note with tagged with "foo:bar" **or** "foo"
- PublicBody tagged with "foo" would be associated with a note with tagged with "foo" but not "foo:bar"


What do you think about that last case... should a PublicBody tagged with "foo" be associated with a note with tagged with "foo:bar".

garethrees: > After this PR:
>
> * PublicBody tagged with "foo:bar" would be associated with a note with tagged with "foo:bar" or "foo"
> * PublicBody tagged with "foo" would be associated with a note with tagged with "foo" but not "foo:bar"

This sounds like the correct behaviour to me.

PR #8460, opened 2024-11-20
Admin insights

gbp:
## Screenshots

image
image


app/models/insight.rb
```diff
@@ -0,0 +1,63 @@
...
+  def self.client
+    @client ||= Ollama.new(
+      credentials: { address: ENV['OLLAMA_URL'] },
+      options: { server_sent_events: true }
+    )
+  end
+
+  def self.models
+    client.tags.first['models'].map { _1['model'] }.sort
+  end
```
gbp, line 39: This could error when `OLLAMA_URL` isn't accessible from the web host. As it stands on myScoiety infra `OLLAMA_URL` is only available for the secondary sidekiq instance running on our dev server.

gbp: I have changed the model select field to a text field, easiest solution to this for now but admins will need to know which models are pulled into Ollama.

app/views/admin/insights/_list.html.erb
```diff
@@ -0,0 +1,34 @@
+<div class="row">
+  <% if insights.any? %>
+    <table class="table table-condensed table-hover span12">
+      <tr>
+        <th>ID</th>
+        <th>Model</th>
+        <th>Template</th>
```
gbp, line 7:
```suggestion
        <th>Temperature</th>
        <th>Template</th>
```

PR #8472, opened 2024-11-27
Stripe Prices fixes

gbp:
## What does this do?

Adds a couple of Stripe fixes for issues preventing signups.

- Change `Price.id` to `Price.to_param` so prices can be loaded
- Correct payment intent status (missing an `s`) should fix incomplete subscriptions
- Update payment failed email, don't send to users without current subscriptions - ie those who are attempting to sign up. 

## Why was this needed?

Getting payments working correctly.

<hr>

[skip changelog]

### DIFF
--- a/app/models/alaveteli_pro/price.rb
+++ b/app/models/alaveteli_pro/price.rb
@@ -4,6 +4,8 @@
 class AlaveteliPro::Price < SimpleDelegator
   include Taxable
 
+  UnknownPrice = Class.new(StandardError)
+
   tax :unit_amount
 
   def self.list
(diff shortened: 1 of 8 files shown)


app/models/alaveteli_pro/price.rb
```diff
@@ -12,15 +12,15 @@ def self.list
...
+  def self.retrieve(param)
+    id = AlaveteliConfiguration.stripe_prices.key(param)
+    new(Stripe::Price.retrieve(id))
   rescue Stripe::InvalidRequestError
     nil
   end
 
   def to_param
-    AlaveteliConfiguration.stripe_prices[id] || id
+    AlaveteliConfiguration.stripe_prices.fetch(id)
```
gbp, line 23: Switched to fetch to see if specs fail. Wondering if we have need the `||` or not

PR #8494, opened 2024-12-18
Gh pages docker

garethrees:
Add Dockerfile 

Avoid having to faff with local dependencies.

    docker-compose build
    docker-compose up
    open http://0.0.0.0:4000

[skip changelog]

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,11 @@
+FROM ruby:3.2
+
+WORKDIR /app
+COPY . .
+
+RUN gem install bundler jekyll
+RUN bundle install
+
+EXPOSE 4000
+
+CMD ["bundle", "exec", "jekyll", "serve", "--host", "0.0.0.0"]
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,9 @@
+version: '3.8'
+
+services:
+  app:
+    build:
+      context: .
+      dockerfile: Dockerfile
+    ports:
+      - 4000:4000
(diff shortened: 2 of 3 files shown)


Dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM ruby:3.2
+
+WORKDIR /app
+COPY . .
+
+RUN gem install bundler jekyll
```
gbp, line 6: No point in this as its already in the Gemfile.lock and could install a different versions than we're really using.

```suggestion
```

Dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM ruby:3.2
```
gbp, line 1: Lets keep the image and small as possible

```suggestion
FROM ruby:3.2-slim

RUN apt-get update -qq && \
    apt-get install -y \
    build-essential \
    git \
    && rm -rf /var/lib/apt/lists/*
```

Dockerfile
```diff
@@ -0,0 +1,11 @@
...
+
+WORKDIR /app
+COPY . .
+
+RUN gem install bundler jekyll
+RUN bundle install
+
+EXPOSE 4000
+
+CMD ["bundle", "exec", "jekyll", "serve", "--host", "0.0.0.0"]
```
gbp, line 11: This would be good, if it works?

```suggestion
CMD ["bundle", "exec", "jekyll", "serve", "--host", "0.0.0.0", "--livereload"]
```

docker-compose.yml
```diff
@@ -0,0 +1,9 @@
+version: '3.8'
+
```
gbp, line 2: Docker warns against including a version now. We should remove. 

```suggestion
```

gbp (changes requested):

PR #8542, opened 2025-01-20
Add InfoRequest embargo scopes

gbp:
## What does this do?

Adds scope for `InfoRequest.was_embargoed` and `InfoRequest.never_embargoed`.

## Why was this needed?

For @garethrees 

## Implementation notes

Needed to update factory to ensure set_embargo events are created.

<hr>


app/models/info_request.rb
```diff
@@ -146,6 +146,8 @@ def self.admin_title
   scope :is_searchable, Prominence::SearchableQuery.new
   scope :embargoed, Prominence::EmbargoedQuery.new
   scope :not_embargoed, Prominence::NotEmbargoedQuery.new
+  scope :was_embargoed, Prominence::WasEmbargoedQuery.new
```
garethrees, line 149: I think I'd call this `ever_embargoed` since it also includes currently embargoed. "was" implies "did have one but now hasn't" so could be a bit of a gotcha.

garethrees (approved):
Thanks!

Issue #8445, opened 2024-11-04
Design improvments for topic pages/category pages


gbp:
On WDTK, we have increased the number of [topic pages](https://www.whatdotheyknow.com/browse) to a point where the [index view](https://github.com/mysociety/alaveteli/blob/develop/app/views/request/index.html.erb) is getting quite long and needs some better UI design.

Maybe we could collapse the long lists and have an option to expand to show everything. See current index view:
![image](https://github.com/user-attachments/assets/bcc76ca8-cc3c-4c16-8ba4-49139286a47f)

<hr>

On the [individual topic page](https://www.whatdotheyknow.com/list/air_pollution) which uses the [list view](https://github.com/mysociety/alaveteli/blob/develop/app/views/request/list.html.erb) (where `@category` is set), it would also be good to improve UI.

This view is also used for the [main search result list](https://www.whatdotheyknow.com/list) so we will need to be mindful of changes that work for both or separate the view.

Given the topic page is already filtered and showing a collection of results, my guess would be the search form won't be used as much as the main view. So maybe we could move the search form into the sidebar or hide it completely until it is revealed. See current list view:

![image](https://github.com/user-attachments/assets/6c4cf61b-0b33-48ba-836a-9af6bdab3d2f)

Aside: If the "Follow" and "RSS" `feed_link` button styling could be fixed, that would be amazing.

PR #8568, opened 2025-02-06
New layout for topic page(/browse)

lucascumsille:
## Relevant issue(s)
Fixes part of: https://github.com/mysociety/alaveteli/issues/8445#top

## What does this do?
Tidy up the list views in `/browse` page. With this PR we make sure that headings with many items don't use a lot of vertical space.

## Implementation notes
I thought about making accordions items instead of cards, but I wasn't really fond of the idea because:

- Depending on how users interact with this page, there might be situations when they'll have to be doing a lot of clicking to see what is inside each category. Not sure if this would be an outliner case though.
- My hunch is that the accordion approach seems more suitable for users that already know the site and roughly know what is inside each heading, but not so much for someone browsing. 

To prevent the above we could include a checkbox that expands all the accordions in the page, and also allow them to open/close each item one by one. But not sure if this would be overcomplicating the problem.

### Desktop view

https://github.com/user-attachments/assets/09508d6e-f93f-46a4-8895-d040ca5f1140

### Mobile view

https://github.com/user-attachments/assets/4da69e5b-608e-47f4-b210-876dac15beb3

## Notes to reviewer

<hr>


garethrees:
Thanks for having a crack at this – hadn't realised it had got this far. While the current version isn't _great_, I think this introduces a few issues. Had a quick chat with @zarino and @gbp about it at the team meeting. Not super urgent though, so might just need some more noodling around in downtime.

PR #8578, opened 2025-02-11
[#2733/#8331] Prevent request classification notifications

### LINKED ISSUE #2733: Old Request Classification: Prevent Notifications & Bumping
Prevent classification of old requests bumping them in sort orders and prompting alert mails

When we're playing the categorisation game etc., and we classify a very old request, the request gets processed for notification emails and for the order of requests on authority pages (and personal pages etc.) 

This can be confusing as where users have notifications set for the request or for search terms that catch the request, they get notified in the same manner as if there had been a new response from the authority. It isn't apparent to the user why they have been notified about this old request when there hasn't been any new communication, and the email and the request page don't highlight to the user that the request classification has been updated prompting the notification.

A staff member from a public authority has also experienced the apparent sudden re-ordering of requests with very old but recently classified requests coming first before much more recent ones. 

I suggest altering Alaveteli such that classification of requests after a certain period (? 6 months) does not result in notification emails being triggered or the requests being brought to the top of authority page request lists and similar.

### LINKED ISSUE #8331: Projects: Prevent Classification Notifications
Project contributors classifying requests generates multiple "someone has updated the status of your request" emails

I set up a project and am receiving one email per request classified by a contributor, despite these being in a batch. This was unexpected. Project owners do not need to receive these, as they can see the progress of the classification from the project dashboard.

gbp:
## Relevant issue(s)

Fixes #2733
Fixes #8331

## What does this do?

1. Prevent request classification notifications for older requests (older than 6 months)
2. Prevent project classification notifications

## Why was this needed?

If a requester hasn't classified their request within 6 months then more than likely they don't care if their request has been classified.

Sometimes this years can go by before classification which can cause in the requester contacting us for support.

For projects classification, many notifications can be sent over a short period of time. These changes prevents them from being sent at all.

## Implementation notes

Build upon CanCan abilities so we can, in the future, offer fine grain permissions for users on which emails they'll receive.

### DIFF
--- a/app/controllers/concerns/classifiable.rb
+++ b/app/controllers/concerns/classifiable.rb
@@ -66,7 +66,7 @@ module Classifiable
     )
   end
 
-  def set_described_state
+  def set_described_state(default_log_params = {})
     described_state = classification_params[:described_state]
     message = classification_params[:message]
 
@@ -74,7 +74,7 @@ module Classifiable
       user_id: current_user.id,
       old_described_state: @info_request.described_state,
       described_state: described_state
-    }
+    }.merge(default_log_params)
 
     log_params[:message] = message if message
 
--- a/app/controllers/projects/classifications_controller.rb
+++ b/app/controllers/projects/classifications_controller.rb
@@ -33,7 +33,7 @@ class Projects::ClassificationsController < Projects::BaseController
     {
       user: current_user,
       info_request: @info_request,
-      resource: set_described_state
+      resource: set_described_state(project: @project)
     }
   end
 end
--- a/app/mailers/application_mailer.rb
+++ b/app/mailers/application_mailer.rb
@@ -20,6 +20,9 @@ class ApplicationMailer < ActionMailer::Base
   # Site-wide access to configuration settings
   include ConfigHelper
 
+  # Check user can receive the emails by wrapping `mail_user`.
+  prepend CheckMailerAbility
+
   # This really should be the default - otherwise you lose any information
   # about the errors, and have to do error checking on return codes.
   self.raise_delivery_errors = true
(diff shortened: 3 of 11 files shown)


garethrees (commented):
> any thoughts on the implementation approach I've used to isolate the logic using CanCan abilities?

Interesting! Novel use of abilities and quite nice that we could use them to keep a bunch of "send if/unless" type logic contained so that the code that calls the mailer method doesn't have to keep repeating it all over the shop.

On the other hand I do wonder whether it adds a bit too much indirection. My first thought was that I could imagine extracting the mailer method and all those conditionals to a `ClassificationMailer`, though I haven't thought much about the specific details here and how we'd pass some of the of the required arguments around.

Looks like `old_unclassified_updated` is only actually called [in one place](https://github.com/mysociety/alaveteli/blob/1811cc22ff58be9d4f3728cb3c78ade4291fa2b0/app/models/info_request.rb#L1033-L1036). This has got me thinking whether we could just drop this entirely.

I wonder how many requests that are [`old_unclassified`](https://github.com/mysociety/alaveteli/blob/1811cc22ff58be9d4f3728cb3c78ade4291fa2b0/app/models/info_request.rb#L507-L520) are actually less than 6 months old…

```
alaveteli(prod)> InfoRequest.is_public.where_old_unclassified.count
# => 154420
alaveteli(prod)> InfoRequest.is_public.where_old_unclassified.where(created_at: 6.months.ago..Time.zone.now).count
# => 13452
```

…less than 10%. Not sure it seems worth it for a small minority of requests that would potentially receive this mail?

We could add a condition to `where_old_unclassified` / `is_old_unclassified` to only include requests > 6 months old and just drop the mail entirely. Some requests may get classified a bit slower than before, but we're probably not talking loads.

If there is a good reason for keeping it, then I think tidying up the "should we actually send this" logic into one place would be good (so, the > 6 month check plus all the existing ones).

gbp:
After discussing with @garethrees and @HelenWDTK we've opted to remove this email. See: #8580

PR #8580, opened 2025-02-13
[#2733/#8331] Remove old_unclassified_updated email

### LINKED ISSUE #2733: Old Request Classification: Prevent Notifications & Bumping

### LINKED ISSUE #8331: Projects: Prevent Classification Notifications

gbp:
## Relevant issue(s)

Fixes #2733
Fixes #8331

## What does this do?

Remove old_unclassified_updated email

## Why was this needed?

Remove email notifications sent to requesters when others classify their requests.

These notifications were causing confusion and generating unnecessary support.


gbp:
There hasn't been any objections to removing this feature: https://groups.google.com/g/alaveteli-dev/c/2UnFt69jynw

PR #8652, opened 2025-04-10
[#8622] Add attachment locking

### LINKED ISSUE #8622: Attachment Locking
- [x] Add locked boolean column to foi_attachments table
   - Create migration script
   - Add model validations and scopes

- [x] Update IncomingMessage attachment extraction process
   - Prevent re-extraction for locked attachments
   - Add safety checks in relevant controller/service methods

- [x] Update InfoRequest cache expiry process
   - Modify InfoRequest#clear_attachment_masks! method
   - Scope to only clear unlocked attachments
   - Ensure censor rules don't affect locked attachments

gbp:
## Relevant issue(s)

Fixes #8622

## What does this do?

Add attachment locking

## Why was this needed?

Easier admin and better compliance with GDPR requests. This will open the door to allow us to delete raw emails so we can fully remove PII.

### DIFF
--- a/app/models/foi_attachment.rb
+++ b/app/models/foi_attachment.rb
@@ -50,6 +50,8 @@ class FoiAttachment < ApplicationRecord
   before_destroy :delete_cached_file!
 
   scope :binary, -> { where.not(content_type: AlaveteliTextMasker::TextMask) }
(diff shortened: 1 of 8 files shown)


gbp:
This has been deployed and tested on staging. LGTM

PR #8662, opened 2025-04-22
[#8624] Attachment locking - main body masking

### LINKED ISSUE #8624: Attachment Locking - Main Body Part Handling
- [x] Special handling for main body part attachments
   - Update caching behavior when main body part is edited/locked
   - Ensure Xapian search index is updated correctly

gbp:
## Relevant issue(s)

Fixes #8624

## What does this do?

Ensures locked attachments aren't masked when caching the main body part in the incoming message instance or when indexing attachments in Xapian.

## Why was this needed?

Easier admin and better compliance with GDPR requests. This will open the door to allow us to delete raw emails so we can fully remove PII.

### DIFF
--- a/app/assets/stylesheets/admin.scss
+++ b/app/assets/stylesheets/admin.scss
@@ -142,6 +142,11 @@ body.admin {
     border-color: #771414;
   }
 
+  i.icon-foi-attachment--locked {
+    @include prominence_icon;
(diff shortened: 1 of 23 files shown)


gbp:
This has been deployed and tested on staging. LGTM

PR #8675, opened 2025-04-29
Improved the semantic for the navigation bar

lucascumsille:
By adding the role="navigation" to the <ul> it removes the default semantic of the element so a <li> element inside an element that is not an <ul> or <ol> can be confusing to assistive devices. This commits wraps the <ul> into a <nav> element solving the issue above.

## Relevant issue(s)

Fixes part of: https://github.com/mysociety/sysadmin/issues/2036

## What does this do?

Fixes some accessibility issues across Alaveteli

## Why was this needed?

Improve UX and comply with The Illegal Content Codes of Practice

## Notes to reviewer

<hr>


gbp:
@lucascumsille can you complete this please "Have you updated the changelog? If this is not necessary, put square brackets around this: skip changelog"

zarino (approved):
This looks good to me! :+1:

@gbp are you happy to take it from here?

PR #8711, opened 2025-05-28
[#8696] Fix blank responses caused by deeply nested emails

### LINKED ISSUE #8696: Responses displayed as blank messages
We've sporadically been seeing some incoming messages being displayed as blank messages when, in actual fact, there is content in the HTML main body part.

Turns out the HTML part could be using HTML5 tags and we're currently using `Loofah.scrub_document`, which is [aliased to their HTML4 scrubber](https://github.com/flavorjones/loofah/blob/6970d9bc552b8e1b8e106329c7645104d24583c0/lib/loofah.rb#L141).

We should switch to an HTML5 scrubber. I reckon there is no reason not to switch to Rails `ActionView::Helpers::SanitizeHelper#sanitize` method; this has been [switched to HTML5 by default](https://github.com/rails/rails/blob/4077fe0ff07601ae3ea20b14815aed1f94d39bd6/actionview/lib/action_view/helpers/sanitize_helper.rb#L22-L23).

Alongside this, we are seeing `Document tree depth limit exceeded` exception coming from Nokogiri when attempting to view a main body attachment from a response which is displayed blank in the admin UI. This is due to it having a lot of nested divs. The same exception happens when switching to `sanitize`, so we should add an initializer to increase the `Nokogiri::Gumbo::DEFAULT_MAX_TREE_DEPTH` setting.

gbp:
## Relevant issue(s)

Fixes #8696

## What does this do?

Fix blank responses caused by deeply nested emails

## Why was this needed?

Show incoming responses that are rendered blank.

## Implementation notes

https://github.com/mysociety/alaveteli/pull/8711#issuecomment-4386223923


gbp:
When testing this locally I noticed the text folding of incoming responses wasn't working as I expected. Unsure if this is due to the changes here or another reason (maybe dev environment related). Needs further testing.

gbp:
Having tested again I'm not seeing the original issue I was concerned about. I think this is good to review.

spec/models/foi_attachment_spec.rb
```diff
@@ -819,21 +819,8 @@ def last_event
...
-          <meta http-equiv="Content-Type" content="text/html; charset=UTF-8">
-        </head>
-        <body>dull
-        </body>
-      </html>
-      EOF
-
       subject
-      expect(attachment.body.squish).to eq(expected)
+      expect(attachment.body.strip).to eq('dull')
```
garethrees: Haven't looked at this in detail but it appears that previously we were expecting HTML to still exist in the output – i.e. not strip all tags; just `script`. Could you elaborate on how the new stripping preserves/intentionally changes this previous behaviour?

gbp:
Looked at this a bit closer to workout exactly what in these emails is preventing injection. Turns out the HTML5 is misleading; while the fix is to switch to the HTML5 parser it isn't because of HTML5 tags rather how deep nested tags are in some the emails. 

The HTML4 parser (libxml2) has a hardcoded depth limit of 255 with no configurable option. Setting `max_tree_depth` which is a Gumbo (HTML5 parser) parameter only and is why this works.

How did I work this out. I used the following in production Rails console to find messages which are blank. The results was able to highlight messages in the issue and others like 3294869 which when inspecting manually all seem to have many nested divs - something which I suspect MS Word to HTML conversion might be to blame.

```ruby
blank_messages = IncomingMessage.where(cached_main_body_text_unfolded: "\n\n")

results = { depth_exceeded: [], errors: [], other: [] }

blank_messages.find_each do |msg|
  begin
    part = msg.get_main_body_text_part

    body = part.unmasked_body

    unless part.content_type == 'text/html'
      results[:other] << { id: msg.id, content_type: part.content_type, size: body.size }
      next
    end

    depth = 0
    max_depth = 0
    body.scan(/<\/?[a-z][^>]*>/i) do |tag|
      if tag.start_with?('</')
        depth -= 1
      else
        depth += 1
        max_depth = depth if depth > max_depth
      end
    end

    if max_depth > 400
      results[:depth_exceeded] << { id: msg.id, depth: max_depth }
    else
      results[:other] << { id: msg.id, depth: max_depth, content_type: part.content_type, size: body.size }
    end
  rescue => e
    results[:errors] << { id: msg.id, error: "#{e.class}: #{e.message}" }
  end
end
```

spec/models/foi_attachment_spec.rb
gbp: Not 100% sure why I did like this originally but I think this was a side effect from using the Rails' ActionView::Helpers::SanitizeHelper which allowed us to set `max_tree_depth` globally.

I have switch now to `Loofah.html5_document(...).scrub` now (and setting `max_tree_depth` in the config block for the document), which keeps the doctype/head.

garethrees (approved):
Ah nice sleuthing! I haven't run this but the diagnosis code and fix look reasonable to me. Let's try it!

PR #8736, opened 2025-06-19
Attachment locking - destroy raw email

gbp:
Part of https://github.com/mysociety/alaveteli/issues/8625, https://github.com/mysociety/alaveteli/issues/8803

## What does this do?

Feature: Attachment locking - add ability to destroy raw emails

## Why was this needed?

This is the final hurdle to add the ability to completely remove PII from Alaveteli this will be used for GDPR RtE requests where we would like to keep the response in an redacted form without PII.

### DIFF
--- a/doc/CHANGES.md
+++ b/doc/CHANGES.md
@@ -2,6 +2,7 @@
 
 ## Highlighted Features
 
+* Add admin ability to destroy raw emails (Graeme Porteous)
 * Add support for Debian 13 "Trixie" (Graeme Porteous)
 * Add support for Ubuntu 24.04 LTS "Nobel Numbat" (Graeme Porteous)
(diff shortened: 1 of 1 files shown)


gbp:
> Main thing decided is to change from destroying raw email database record to just erasing the active storage file record; think we want to look at adding columns to the raw email model to ensure correct attributes cached before we allow admins to start destroying

Moving the cache columns from IncomingMessage to RawEmail is proofing difficult in terms of sorting out the test examples. Everything is too tightly coupled and unpicking is too much work for this PR. It has already taken up too much time and serves no real benefit as we can just leave the cache columns where they are - I believe it is not worth doing it anymore.

This also means the work to remove the attached ActiveStorage file only and leaving the RawEmail instance isn't worth it as the instance won't have any use being just blank with no attributes.

PR #8774, opened 2025-07-31
Add classification icons to request lists in admin

garethrees:
Makes it easier to quickly identify e.g. groups of successful/refused requests.

Could do with reducing the CSS duplication; will speak to others about this to get some input.

Screenshot 2025-07-31 at 17 39 57


app/assets/stylesheets/admin.scss
```diff
@@ -352,3 +352,90 @@ pre.error {
...
+/* Status icons */
+@mixin classification-icon() {
+  background-image: image-url('status/status-sprite.png');
+  background-size: auto 14px;
+  @media (-webkit-min-device-pixel-ratio: 1.5), (min-resolution: 144dpi) {
+    background-image: image-url('status/status-sprite@2.png');
+  }
+}
+
+.classification_icon-standalone {
```
zarino: I _think_ if you just put the contents of the `classification-icon` mixin inside here, then you wouldn’t need the mixin any more, and it wouldn’t need to be replicated inside each status ruleset further down.

(Or are there cases where `classification_icon-standalone` is used on its own, with no additional status class?)

garethrees: Yeah I'm not actually sure what the `-standalone` class is meant to be doing, so could probably tidy that lot up.

The duplication I was referring to is actually that this essentially exists in the public interface – [mixin](https://github.com/mysociety/alaveteli/blob/0.45.3.0/app/assets/stylesheets/responsive/_utils.scss#L171-L178) & [individual icons](https://github.com/mysociety/alaveteli/blob/0.45.3.0/app/assets/stylesheets/responsive/_lists_style.scss#L81-L158).

I didn't have chance to try/look in to sharing files in the different namespaces (`assets/responsive` for public interface and `assets/admin` for admin) – its been a while since I've worked with all this.

Couple of issues with just copy-pasting that code to `admin.scss`:

The admin interface bootstrap adds css that targets…

```scss
.admin [class^="icon-"],
.admin [class*=" icon-"] 
```

…which took precedence over our sprite; can probably work around that though.

The main difference is that in the public version we use 22px x 22px dimensions, whereas 14x14 looks better in the admin interface to line up with the [bootstrap icons](https://getbootstrap.com/2.2.0/base-css.html#icons). This then needs all the individual icon classes to change to select the right part of the sprite, but there might be a better way, so that's why just re-writing it was easier as a proof of concept.

Would be nice to have just one place to define these (with maybe a `--small` modifier class for the admin UI).

zarino: Ahhh I see. Hmm yes, that is a bit trickier.

My instinct is that—even though they’re using the same underlying sprite—these icons are different enough (different size, different class name to avoid the `icon-` conflict) that it’s probably simplest to handle them separately. Once the repeated `@include` is removed, and the CSS is gzipped, it’ll only be a few extra bytes, and only for admin users.

It’s tempting to store the names and positions of each of the sprite’s icons in a single Sass map, that then gets looped over in both the public and admin stylesheet – but since the two spreadsheets are so separate (eg: no obvious `_variables.scss`-style shared file which they both import), this is probably more hassle than it’s worth, just for a little bit of DRYing.

PR #8785, opened 2025-08-04
Make requests sortable in admin interace

garethrees:
Extracts `Admin::Sortable` concern, using it to allow requests to be sorted in the admin interface.

Sometimes it's more useful to view by created rather than updated (e.g. "show me most recent requests made by this body").

## admin/user/show

Screenshot 2025-08-04 at 09 59 45

## admin/body/show

Screenshot 2025-08-04 at 09 59 59

## admin/requests/index

Screenshot 2025-08-04 at 10 00 16

## admin/users/index

Now uses created date when sorted by created_at options

Screenshot 2025-08-04 at 10 01 31


app/views/concerns/admin/sortable/_sort_options.html.erb
```diff
@@ -0,0 +1,11 @@
+<% if defined?(sort_options) && sort_options %>
```
gbp, line 1: Do we really need this conditional?

app/controllers/concerns/admin/sortable.rb
```diff
@@ -39,10 +39,12 @@ def set_sort_options(attrs)
     end.with_indifferent_access
   end
 
-  def set_sort_order
+  def set_sort_order(default: nil)
     @sort_order =
       if params[:sort_order].in?(@sort_options.keys)
         params[:sort_order]
+      elsif default && @sort_options.key?(default)
```
gbp: Could use a call to `#in?` like above for consistency

app/controllers/admin_request_controller.rb
```diff
@@ -5,10 +5,14 @@
 # Email: hello@mysociety.org; WWW: http://www.mysociety.org/
 
 class AdminRequestController < AdminController
+  include Admin::Sortable
+
   before_action :set_info_request, :check_info_request, only: %i[
     show edit update destroy move generate_upload_url hide
   ]
 
+  sortable default: :updated_at_desc, only: [:index]
```
gbp, line 14: Did you mean to change the sort order from `created_at desc` here?

app/views/admin_request/_info_request.html.erb
```diff
@@ -12,7 +12,12 @@
       <%= both_links(info_request.user) %>
       <%= arrow_right %>
       <%= both_links(info_request.public_body) %>,
-      <%= admin_date(info_request.updated_at, ago_only: true) %>
+
+      <% if defined?(sort_order) && sort_order =~ /created_at/ %>
+        <%= admin_date(info_request.created_at, ago: false) %>
+      <% else %>
+        <%= admin_date(info_request.updated_at, ago_only: true) %>
+      <% end %>
```
gbp, line 20: Approach here differs from `app/views/admin_user/_user.html.erb` maybe we should look to be consistence?

gbp (commented):
A few comments and questions, nothing major. Would be good to address the Rubocop comments tho.

app/views/concerns/admin/sortable/_sort_options.html.erb
garethrees: It prevents the page breaking if no sort options are defined at controller level, but where this view gets rendered from elsewhere. Might be able to drop it / work around it not sure.

app/controllers/admin_request_controller.rb
garethrees: The current sort order is updated at desc, so yeah kept it as that for now

app/views/admin_request/_info_request.html.erb
garethrees: Yeah I just worked off the existing implementation in each case – don't know if there's a good reason that they were different. Don't think the consistency is critical at this point but agree could look at in future.

PR #8792, opened 2025-08-11
Add Taggable concern to User

gbp:
## Relevant issue(s)

https://github.com/mysociety/whatdotheyknow-private/issues/397

## What does this do?

Adds the Taggable concern to User allowing users to be assigned tags.

## Why was this needed?

Adding the ability to tag users will allow admins to search for specific users or groups based on their tags.

## Screenshots

image

image

image

image


app/views/admin_user/_form.html.erb
```diff
@@ -123,3 +123,12 @@
     </div>
   </div>
 </div>
+<div class="control-group">
+  <label for="user_tag_string" class="control-label">Tags</label>
+  <div class="controls">
+    <%= text_field :admin_user, :tag_string, :class => "span4" %>
+    <div class="help-block">
+      space separated list tags
```
gbp:
```suggestion
      space separated list of tags
```

garethrees (commented):
Any specific questions about this? I don't think this needs significant review since it's juts adding an existing pattern. I'm happy if you're happy.

gbp:
> Any specific questions about this? I don't think this needs significant review since it's juts adding an existing pattern. I'm happy if you're happy.

This was more to check I haven't missed anything.

PR #8830, opened 2025-08-15
Improve /browse page and individual category pages

### LINKED ISSUE #8445: Design improvments for topic pages/category pages

zarino:
## Relevant issue(s)

Fixes #8445. Replaces #8568.

## What does this do?

Makes the `/browse` page easier to skim read…

new column layout on the browse page

Adds a client-side search to narrow down the page contents if there’s a particular term you care about…

search term highlighted on the browse page

And also improves the layout at the top of the individual category page, hiding the search form inside a `<details>` element…

search form hidden on category page

## Notes to reviewer

Didn’t know whether this is a significant enough change to qualify for the changelog, and if so, how I should word it. So I’ll leave that to @gbp / @garethrees.


garethrees (approved):
This looks great!

I was wondering whether to use the `<details>` filter when we're not showing a category, but I'm not sure. It leaves a load of empty space as we don't have the category body text / notes:

Screenshot 2025-08-15 at 17 28 18

Think just keep it as it is now as we'll be revisiting these pages in the near future as we work on the search tech.

PR #8838, opened 2025-08-19
Add `last_sign_in_at` column to users

gbp:
## What does this do?

Records last sign in time when a user logs in. Will allow admins to see if user profiles are inactive or not.

## Screenshots

image

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -166,12 +166,9 @@ class ApplicationController < ActionController::Base
     session[:user_id] = user.id
     session[:user_login_token] = user.login_token
     session[:remember_me] = remember_me
-    # Intentionally allow to fail silently so that we don't have to care whether
-    # sign in recording is enabled.
-    user.sign_ins.create(
+    user.record_sign_in(
       ip: user_ip, country: country_from_ip, user_agent: request.user_agent
     )
-    user.touch(:last_sign_in_at)
   end
 
   # Logout form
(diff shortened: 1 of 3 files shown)


app/controllers/application_controller.rb
```diff
@@ -171,6 +171,7 @@ def sign_in(user, remember_me: nil)
     user.sign_ins.create(
       ip: user_ip, country: country_from_ip, user_agent: request.user_agent
     )
+    user.touch(:last_sign_in_at)
```
garethrees: think its worth refactoring the sign_in creation and touch to `User#record_sign_in`

gbp: I don't understand. Are you talking about `User::SignIn` creation? Also `User#record_sign_in` doesn't exist?

garethrees: Yeah, so basically add:

```rb
class User
  def record_sign_in(args)
    sign_ins.create(args)
    touch(:last_sign_in_at)
  end
end
```

Then in the controller we can just call:

```diff
- user.sign_ins.create(
-   ip: user_ip, country: country_from_ip, user_agent: request.user_agent
- )
+ user.record_sign_in(ip: user_ip, country: country_from_ip, user_agent: request.user_agent)
```

Was just thinking it would keep the controller a bit cleaner, and we could add a test for the model level method to assert that it touches the record and creates/doesn't create the sign in depending on whether retention is enabled.

PR #8839, opened 2025-08-19
Add spam email format option to UserSpamScorer

gbp:
## What does this do?

Allow sites to set `UserSpamScorer.spam_email_format` with an array of regexp. These will be matched against email address to test if they are spammy or not.


garethrees (commented):
Looks good, but should add some tests

app/controllers/concerns/user_spam_check.rb
```diff
@@ -41,6 +41,7 @@ def spam_scorer_config
         email_from_suspicious_domain?: 10,
         email_from_spam_domain?: 13,
         email_from_spam_tld?: 1,
+        email_is_spam_format?: 13,
```
garethrees: think this should be a lower score on reflection – maybe 2 or 3 – given some of the formats are quite generic and could easily be used by genuine users rather than a smoking gun sign of a problem.

garethrees: same for the defaults

gbp: Have lowered the default scores. Not sure if that will mean we prevent on the formats we are looking at but we can always adjust if needed.

PR #8842, opened 2025-08-22
Restrict web reply domains

### LINKED ISSUE #427: Be more careful about who gets "authority" privileges
Currently anyone with an email address @ the same domain as the FOI address can do special things on behalf of the authority. This goes wrong for small councils that use personal email accounts instead of having their own domain - e.g. anyone @gmail.com could upload to some councils.

This will also become more of a risk if and when we open up more features to the authority.

A few ideas for fixing this
- don't apply the logic to .com addresses
- have a blacklist somewhere of domains the logic doesn't apply to
- add a field to the authority page for specifying the domain to accept with "[none]" or similar to disable, and default to the current logic and/or some of the other proposals

garethrees:
## Relevant issue(s)

Fixes #427

## What does this do?

Add ability to prevent domain-based access to response upload web UI and limit to only users registered with the main FOI address

## Why was this needed?

Some domains have a large number of users spanning multiple bodies, or some authorities use quite generic domains.

## Implementation notes

Could do with a lot of refactoring but did some basic steps to at least make this change a bit neater

## Screenshots

```rb
PublicBody.excluded_foi_officer_access_domains << 'hump.com'
```

Screenshot 2025-08-22 at 15 22 27

## Notes to reviewer

Not super happy with having `UploadResponseController` at the top level while also inheriting from `RequestController`. I did initially have it as `Request::UploadResponseController` but got some random spec errors, presumably from `Request` being an existing class. I think what we have is fine for this quick fix and can come back to refactor in future.

Not going to address the Rubocop comments where we've just moved code from one file to another – a problem for another day / automated cleanup.


app/models/public_body/foi_officer_access.rb
```diff
@@ -3,14 +3,61 @@
...
+  end
+
+  def foi_officer_domain_excluded?
+    excluded_foi_officer_access_domains.include?(request_email_domain)
+  end
 
   # Does this user have the power of FOI officer for this body?
   def is_foi_officer?(user)
+    return true if registered_with_main_foi_address?(user)
+    return false if outside_user?(user)
```
garethrees, line 36: Actually should probably swap the order of these:

```rb
    return false if outside_user?(user)
    return true if registered_with_main_foi_address?(user)
```

mdeuk:
Fixes #427

gbp:
Can't see any check been run on this. Might need a rebase to get CI to check it

PR #8849, opened 2025-08-29
Extract PublicBody::CsvImport concern

garethrees:
Simply move the methods to a concern given they're an isolated concept and take up a significant amount of line length in the main class.

Obviously there's lots of refactoring to be done here – ideally this would be a well-tested class in its own right with the concern being a simple interface, but for now just extracting the pieces out to their own file is a nice improvement.

Tested a basic CSV-based update in the browser and all still works as expected.

[skip changelog]

### DIFF
--- a/app/models/public_body.rb
+++ b/app/models/public_body.rb
@@ -26,7 +26,6 @@
 #  disclosure_log                         :text
 #
 
-require 'csv'
 require 'securerandom'
 require 'set'
 require 'confidence_intervals'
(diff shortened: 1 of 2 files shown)


app/models/public_body.rb
```diff
@@ -26,21 +26,19 @@
...
 
-require 'csv'
 require 'securerandom'
 require 'set'
 require 'confidence_intervals'
 
 class PublicBody < ApplicationRecord
   include CalculatedHomePage
   include Categorisable
+  include CsvImport
```
gbp, line 36:
```suggestion
  include PublicBody::CsvImport
```

gbp: Including the namespace is what we do in User

garethrees: hmm looks like we haven't been consistent about that, but yeah think we should do this so its nice and easy to see what's a global concern and what's just nesting some stuff under the model

garethrees: Will fix that up in a separate pr now

PR #8855, opened 2025-09-02
Don't spam-check admins

garethrees:
Admins may do things that trigger the spam scorer for testing purposes (e.g. use plus accounts). Admins should always be allowed to log in.

### DIFF
--- a/spec/controllers/users/sessions_controller_spec.rb
+++ b/spec/controllers/users/sessions_controller_spec.rb
@@ -236,7 +236,8 @@ RSpec.describe Users::SessionsController do
         FactoryBot.create(
           :user,
           email: 'spammer@example.com', name: 'Download New Person 1080p!',
-          password: 'password1234', email_confirmed: true
+          password: 'password1234', email_confirmed: true,
+          confirmed_not_spam: false
         )
       end
 
@@ -300,6 +301,20 @@ RSpec.describe Users::SessionsController do
       context 'when the user is an admin' do
         before do
           user.add_role(:admin)
+
+          allow(@controller).
+            to receive(:spam_should_be_blocked?).and_return(true)
+        end
+
+        it 'allows the signin' do
+          do_signin(user.email, 'password1234')
+          expect(session[:user_id]).to eq user.id
+        end
+      end
+
+      context 'when the user is confirmed_not_spam' do
+        before do
+          user.update(confirmed_not_spam: true)
 
           allow(@controller).
             to receive(:spam_should_be_blocked?).and_return(true)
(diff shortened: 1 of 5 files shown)


spec/controllers/users/sessions_controller_spec.rb
```diff
@@ -296,6 +297,34 @@ def do_signin(email, password)
           expect(session[:user_id]).to eq user.id
         end
       end
+
+      context 'when the user is an admin' do
+        before do
+          user.add_role(:admin)
+
```
gbp, line 304: Not massively bother but I would write these tests as:

```suggestion
        let(:user) { FactoryBot.create(:admin_user, ...) }

        before do
```

spec/controllers/users/sessions_controller_spec.rb
```diff
@@ -236,7 +236,8 @@ def do_signin(email, password)
         FactoryBot.create(
           :user,
```
gbp, line 237: Change to use the `unconfirmed_user` factory. Then changes below to set `confirmed_not_spam: false` wouldn't be needed.

spec/controllers/users/sessions_controller_spec.rb
```diff
@@ -296,6 +297,34 @@ def do_signin(email, password)
...
+        it 'allows the signin' do
+          do_signin(user.email, 'password1234')
+          expect(session[:user_id]).to eq user.id
+        end
+      end
+
+      context 'when the user is confirmed_not_spam' do
+        before do
+          user.update(confirmed_not_spam: true)
+
```
gbp, line 318:
```suggestion
        let(:user) { FactoryBot.create(:confirmed_user, ...) }

        before do
```

Although `confirmed_user` factory doesn't exist, can just use `user` factory as these are confirmed by default but maybe we should add an alias to it.

spec/controllers/users/sessions_controller_spec.rb
garethrees: Ah the reason I did this is because the existing user is set up to look spammy, so it ensures that they would get flagged as spam if they didn't have the admin role. I didn't test this, but the `:admin_user` factory might not get flagged so the test would pass but not because we'd added the explicit check around admin users.

spec/controllers/users/sessions_controller_spec.rb
garethrees: Don't think an `unconfirmed_user` would get spam checked (https://github.com/mysociety/alaveteli/blob/develop/app/controllers/users/sessions_controller.rb#L36), so would have to then set `email_confirmed` here instead. Seems a bit clearer in intent as is.

PR #8856, opened 2025-09-02
Model concern include consistency

garethrees:
Include namespace in model concern includes. This makes it obvious at a glance whether a concern is shared or just nesting some behaviour under the specific model.

Order includes by most general to most specific. This is more likely to ensure generic methods are included first so that more specific concerns that rely on them have the methods needed. This isn't guaranteed so we can deviate from this as necessary.

[skip changelog]


gbp (approved):
Nice. One minor and kinda related improvement could be move `include Rails.application.routes.url_helpers` into `LinkToHelper`?

PR #8857, opened 2025-09-03
Refactor include of url_helpers

garethrees:
This include is only required because we're including LinkToHelper in models, so include it there to reduce duplication.

[skip changelog]


garethrees:
@gbp was this what you had in mind in https://github.com/mysociety/alaveteli/pull/8856#pullrequestreview-3180207109?

garethrees:
Closing as this didn't work as seamlessly as expected.

Issue #2046, opened 2015-01-03
Inconsistent naming of norwegian translations on Transifex and git


petterreinholdtsen:
The Norwegian bokmål translation is named nb_NO, while Norwegian nynorsk is named nn.  It would make more sense if they were named nb and nn (or nb_NO and nn_NO, but I recommend against it to stay consistent with most free software translated to Norwegian).

Can Transifex and git be updated to use nb and nn, not nb_NO and nn?

garethrees:
For reference, Transifex [supports](https://www.transifex.com/languages):

- Norwegian Bokmål (nb)
- Norwegian Bokmål (Norway) (nb_NO)
- Norwegian (Norway) (no_NO)
- Norwegian Nynorsk (nn)
- Norwegian Nynorsk (Norway) (nn_NO)

garethrees:
```irc
[09:45:05]  <gareth>	 pere so just to check my understanding, the action here is to rename `locale/nb_NO` to `locale/nb`?
[09:46:45]  <pere>	 yes.  and change it in transifex too.
```

garethrees:
```irc
[10:03:27]  <pere>	 btw, on <URL: https://alaveteli-dev.nuug.no/ >, where do the string 'Norsk bokmål' come from?  I suspect it would be better to just write 'Bokmål'.
[10:06:40]  <gareth>	 pere https://github.com/mysociety/alaveteli/blob/rails-3-develop/lib/languages.rb#L114
```

petterreinholdtsen:
You might find
<URL: http://people.skolelinux.org/pere/blog/Spr_kkoder_for_POSIX_locale_i_Norge.html >
useful as background information.  It is only in Norwegian, sorry about
that.

-- 
Happy hacking
Petter Reinholdtsen

garethrees:
@petterreinholdtsen I've made you a Coordinator on the `nb` locale for Alaveteli https://www.transifex.com/organization/mysociety/team/33529/members/nb/

garethrees:
Could you confirm that the changes in this PR and on Transifex meet your needs; if so, I'll get it merged. We'll leave removing `nb_NO` until Release 0.21 (#2049).

petterreinholdtsen:
[Gareth Rees]
> Could you confirm that the changes in this PR and on Transifex meet
> your needs; if so, I'll get it merged. We'll leave removing `nb_NO`
> until Release 0.21 (#2049).

[...]

PR #8858, opened 2025-09-04
[#2046] improve rake tasks to remove user generate content

gbp:
Part of https://github.com/mysociety/sysadmin/issues/2046

## What does this do?

Improve rake tasks to remove user generate content optimising so tasks run quicker. Initially on WDTK these will still take a long time but running on a rolling basis should be much quicker now.

<hr>

[skip changelog]

### DIFF
--- a/lib/tasks/users.rake
+++ b/lib/tasks/users.rake
@@ -145,9 +145,16 @@ namespace :users do
   desc 'Purge profile content from limited users'
   task purge_limited: :environment do
     users = User.limited_profile.where(created_at: ...6.months.ago)
-    users.find_each do |user|
-      user.update!(about_me: '') if user.about_me.present?
-      user.profile_photo&.destroy
+
+    users_with_about_me = users.where.not(about_me: '')
+    users_with_profile_photos = users.joins(:profile_photo)
+    ids = users_with_about_me.ids + users_with_profile_photos.ids
+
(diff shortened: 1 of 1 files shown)


lib/tasks/users.rake
```diff
@@ -145,15 +145,50 @@ namespace :users do
...
 
   desc 'Destroy user accounts that have not created any content'
   task destroy_unused: :environment do
     users = User.unused.where(created_at: ...2.years.ago)
-    users.destroy_all
+
+    TrackThing.where(tracked_user_id: users).update_all(tracked_user_id: nil)
+
+    ActiveRecord::Base.logger.silence do
+      users.preload(
```
garethrees, line 168: Is it worth thinking about automating this list so that it's a bit more resilient to new associations, e.g something like:

```rb
User.reflect_on_all_associations.
  select { |a| [:has_many, :has_one].include?(a.macro) && !a.options[:through] }.
  map(&:name)
```

gbp: Yeah probably save us hassle in the future. There are some minor differences in the associations so will double check them first tho

garethrees: Thinking about it not sure we need to exclude `:through` option as it probably doesn't matter that it's preloaded and not used.

Issue #7620, opened 2023-03-01
Generic user content rate limiter


garethrees:
Extract from https://github.com/mysociety/alaveteli/pull/7602 to be able to apply to any `User` association records.

Here's a sketch.

```diff
diff --git a/app/models/user.rb b/app/models/user.rb
index ceb0eab68..f1653a9bc 100644
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -36,7 +36,42 @@
 #  closed_at                         :datetime
 #  login_token                       :string
 #
+class ContentRateLimiter
+  def self.from_relation(records)
+    new(records.klass, records)
+  end
+
+  def initialize(klass, records)
+    @klass = klass
+    @records = records
+  end
+
+  def exceeded_limit?
+    exceeded_daily_cap? || exceeded_creation_rate?
+  end
+
+  def exceeded_daily_cap?
+    return false unless defined?(klass::MAX_PER_DAY)

+    # OR         (created_at: 24.hours.ago..)
+    records.where(created_at: Time.zone.now.at_beginning_of_day..).size >=
+      klass::MAX_PER_DAY
+  end
+
+  def exceeded_creation_rate?
+    return false unless defined?(klass::CREATION_RATE_INTERVALS)
+
+    klass::CREATION_RATE_INTERVALS.any? do |limit, duration|
+      return false if records.size < limit
+      records.limit(limit).all? { |record| record.created_at > duration.ago }
+    end
+  end
+
+  private
+
+  def records
+    records.reorder(created_at: :desc)
+  end
+end
 class User < ApplicationRecord
   include AlaveteliFeatures::Helpers
   include AlaveteliPro::PhaseCounts
@@ -89,10 +124,17 @@ class User < ApplicationRecord
            -> { order(created_at: :desc) },
            inverse_of: :user,
            dependent: :destroy
+
   has_many :comments,
            -> { order(created_at: :desc) },
            inverse_of: :user,
-           dependent: :destroy
+           dependent: :destroy do
+             def limiter
+               ContentRateLimiter.new(proxy_association.reflection, proxy_association.target)
+            end
+          end
+
+
   has_many :public_body_change_requests,
            -> { order(created_at: :desc) },
            inverse_of: :user,
@@ -451,9 +493,7 @@ def can_make_followup?
   def can_make_comments?
     return false unless active?
     return true if is_admin? || is_pro_admin?
-
-    !exceeded_limit?(:comments) &&
[...]

PR #8859, opened 2025-09-05
Add request creation interval rate limit

garethrees:
## Relevant issue(s)

Fixes https://github.com/mysociety/whatdotheyknow-private/issues/415

## What does this do?

Prevents too many requests being created too quickly.

## Why was this needed?

We don't want to allow a user to sign up and smash through their daily limit in 5 minutes, then create a new account to do the same, and so on.

This ensures that the allocation is used sensibly. It hopefully won't affect the majority of genuine users.

## Implementation notes

Same vibe as https://github.com/mysociety/alaveteli/pull/7602. Extracted out a shared concern (cc https://github.com/mysociety/alaveteli/issues/7620) 

## Screenshots

Screenshot 2025-09-08 at 08 32 06

### DIFF
--- a/app/models/concerns/rate_limited.rb
+++ b/app/models/concerns/rate_limited.rb
@@ -1,0 +1,41 @@
+# Ability to set and check the rate at which records are created.
+#
+# Custom limits can be set per record type:
+#
+# rate_limited 5 => 30.minutes, 10 => 2.hours
+#
+# Read this as: "A maximum of N (records) every X (time)"
+module RateLimited
+  extend ActiveSupport::Concern
+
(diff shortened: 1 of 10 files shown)


app/models/concerns/rate_limited.rb
```diff
@@ -0,0 +1,43 @@
+# rate_limited 1 => 2.seconds, 2 => 5.minutes
```
gbp: Is this line meant to be here?

garethrees: oops no – was the stub of the concern doc

PR #8869, opened 2025-09-09
404 for not found AttachmentToHTML sub-files

garethrees:
When we extract PDFs to HTML we try to extract the text to HTML text, but many will also include images for parts of the PDF that can't be parsed or don't make sense to parse as text.

Parts of this can fail, leaving missing images. Every missing image in an attachment causes a redirect to the request page, meaning we see large spikes to request pages when accessing attachments with lots of missing images, effectively self-DDOSing the app.

The 303 redirect was introduced in 501b2b8. For top-level attachments this makes more sense, since occasionally the part number / filename can change (through reparsing, and since then we can now modify attachment names), so in that case it's nicer to take users back to at least the request page where the attachment originated as they may be able to find the equivalent current version.

It doesn't make sense to do this for these sub-files that are used in the to-HTML process, as these aren't intended to be accessed directly.

h/t @dracos for the diagnosis and bones of this improvement.

### DIFF
--- a/app/controllers/attachments_controller.rb
+++ b/app/controllers/attachments_controller.rb
@@ -102,6 +102,9 @@ class AttachmentsController < ApplicationController
         request.format = :html
         render_hidden('request/hidden_attachment')
       end
+    elsif params[:file_name] =~ /^foiextract.*/
+      # 404 for AttachmentToHTML sub-files that can't be found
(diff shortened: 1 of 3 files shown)


gbp (approved):
Looks like this would work. Happy to deploy as is and test.

PR #8871, opened 2025-09-09
Update site locale storage

dracos:
Switch to use a separate `locale` cookie instead of storing in the encrypted Rails session cookie. This should prevent the activating of a session if there is no session.

This will allow better caching in Varnish when the Rails session isn't in use and prevent requests from hitting the application stack.


gbp:
All seems to be working from the application side. Will deal with Varnish separately.

Issue #8870, opened 2025-09-09
Users can create requests without confirming their email address


garethrees:
I'm not sure how to replicate this, but somehow it is possible for users to create requests (or other content like Tracks) without having confirmed their email address. It doesn't look like this is a new issue, given we have records of this going back years:

```rb
User.where(email_confirmed: false).where.not(info_requests_count: 0).order(created_at: :desc).pluck(:created_at)
# =>
[2025-02-01 01:57:26.204303000 GMT +00:00,
 2024-06-14 21:45:59.251306000 BST +01:00,
 2024-05-28 15:47:09.491374000 BST +01:00,
 2023-09-16 18:08:03.218541000 BST +01:00,
 2023-01-25 06:16:09.707959000 GMT +00:00,
 2022-12-19 20:57:14.805979000 GMT +00:00,
 # …snip…
 2015-08-08 10:16:54.101664000 BST +01:00,
 2015-06-04 11:21:37.865381000 BST +01:00,
 2014-12-07 02:08:09.439255000 GMT +00:00,
 2014-08-25 20:48:06.955872000 BST +01:00,
 2013-06-18 02:27:37.343444000 BST +01:00,
 2012-03-31 20:17:21.661622000 BST +01:00]
```

I've tested the two main pathways:

* Click sign in / sign up in the nav bar; sign up; try to sign in without confirming
* Try to make a request; go through sign in / sign up; try to sign in without confirming

In both cases I can't manage to get the user signed in – it redirects the user to the "confirm your email" view.

garethrees:
Start with an unconfirmed user:

```rb
User.find_by(email: 'foobar@localhost').email_confirmed?
# => false
```

Here's the password reset flow, starting from the sign in page.

<img width="1912" height="1237" alt="Image" src="https://github.com/user-attachments/assets/15468b76-9a15-4de4-8652-bb87795ca1d4" />

Click "Forgotten your password?"

<img width="1912" height="1237" alt="Image" src="https://github.com/user-attachments/assets/c5e22fec-899f-4d92-8988-2bd5bf1bab06" />

Fill in your email and click "Send me the email".

<img width="1912" height="1237" alt="Image" src="https://github.com/user-attachments/assets/63924a96-3359-433a-86bc-8c8387c8f41a" />

Go to your inbox.

<img width="897" height="432" alt="Image" src="https://github.com/user-attachments/assets/09427750-4280-48aa-9079-e8f6fb726f42" />

Follow the link.

[...]

PR #8872, opened 2025-09-09
[#8870] Confirm users after following password reset link

garethrees:
If an unconfirmed user resets their password, they are logged in and can perform actions like creating requests while their account is still unconfirmed.

The password reset link acts as a confirmation email, so it doesn't seem like there's a reason not to confirm them in this step.

Fixes https://github.com/mysociety/alaveteli/issues/8870.

## Notes to Reviewer

@gbp can you think of any reasons _not_ to do this?

### DIFF
--- a/app/controllers/users/confirmations_controller.rb
+++ b/app/controllers/users/confirmations_controller.rb
@@ -13,9 +13,10 @@ class Users::ConfirmationsController < UserController
         clear_session_credentials
       end
 
+      confirm_user!(post_redirect.user)
+
       redirect_to SafeRedirect.new(post_redirect.uri).path
       return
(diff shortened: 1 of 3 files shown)


app/controllers/users/confirmations_controller.rb
```diff
@@ -13,6 +13,7 @@ def confirm
         clear_session_credentials
       end
 
+      @user = confirm_user!(post_redirect.user)
```
garethrees: don't need to set `@user` here actually – will remove

gbp (approved):
This looks to be working okay and I agree seems like a sensible thing to do

PR #8884, opened 2025-09-19
[#4286] Classification form icons

### LINKED ISSUE #4286: Add status icons to describe state form
screen shot 2017-10-11 at 16 37 41

This is really hard to parse and figure out what its asking.

Adding the state icons next to each option would help users find the item they're after.

garethrees:
## Relevant issue(s)

Fixes #4286

## What does this do?

Add icons to the describe state form

## Why was this needed?

This helps the user better identify which state to choose.

## Implementation notes

Helper is only used in the describe state form so safe to add the icon there

## Screenshots

Screenshot 2025-09-19 at 15 28 30

## Notes to reviewer

Needs some alignment tweaks.

<hr>

### DIFF
--- a/app/assets/stylesheets/responsive/_request_style.scss
+++ b/app/assets/stylesheets/responsive/_request_style.scss
@@ -174,7 +174,8 @@ a.cplink__button {
    padding: 1.25em 1.25em;
    .icon-standalone {
      float: left;
-     margin-right: 11px;
+     margin-right: 8px;
+     margin-top: 1px;
    }
    :last-child {
      margin-bottom: 0;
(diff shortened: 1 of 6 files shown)


lucascumsille:
My first approach was to try to handle all the alignment using flex, instead of using margins which normally leads to this kind of situations, but that would mean changing the html markup. I'll put it on this ticket on next sprint so I can give it more time

lucascumsille:
@garethrees I just added a fixup commit that should improved the vertical alignments.
let me know if there is anything else.
Screenshot 2025-10-07 at 10 29 28

Screenshot 2025-10-07 at 10 29 56

Screenshot 2025-10-07 at 10 25 21

PR #8889, opened 2025-09-22
Update `cattr_*` calls

gbp:
## What does this do?

Update `cattr_*` calls

## Why was this needed?

We noticed in https://github.com/mysociety/whatdotheyknow-theme/pull/2052 that using the defaults value option for `cattr_*` can cause issues when the defaul value itself is sourced from the class attribute. This PR fixes this and updates other `cattr_*` as now I've looked more at the documentation of the method I understand what it is actually doing and in some cases we don't need the instance methods in defined by default.

<hr>

[skip changelog]

### DIFF
--- a/app/models/public_body/calculated_home_page.rb
+++ b/app/models/public_body/calculated_home_page.rb
@@ -2,9 +2,14 @@
 module PublicBody::CalculatedHomePage
   extend ActiveSupport::Concern
 
-  included do
-    cattr_accessor :excluded_calculated_home_page_domains,
-                   default: Domains.webmail_providers
+  class_methods do
+    def excluded_calculated_home_page_domains
+      @excluded_calculated_home_page_domains ||= Domains.webmail_providers
+    end
+
+    def excluded_calculated_home_page_domains=(domains)
+      @excluded_calculated_home_page_domains = domains
+    end
   end
 
   def calculated_home_page
@@ -34,6 +39,6 @@ module PublicBody::CalculatedHomePage
   end
 
   def excluded_calculated_home_page_domain?(domain)
-    excluded_calculated_home_page_domains.include?(domain)
+    self.class.excluded_calculated_home_page_domains.include?(domain)
   end
 end
--- a/app/models/concerns/rate_limited.rb
+++ b/app/models/concerns/rate_limited.rb
@@ -23,8 +23,6 @@ module RateLimited
     end
 
     cattr_accessor :creation_rate_limits,
-                   instance_reader: false,
-                   instance_writer: false,
                    instance_accessor: false,
                    default: DEFAULT_CREATION_RATE_LIMITS
   end
--- a/app/models/note.rb
+++ b/app/models/note.rb
@@ -21,14 +21,18 @@ class Note < ApplicationRecord
   delegate :rich_body, :rich_body=, :rich_body?, to: :translation
   after_save { rich_body.save if rich_body.changed? }
 
-  cattr_accessor :default_style, default: 'original'
-  cattr_accessor :style_labels, default: {
-    '🔴 Red': 'red',
-    '🟡 Yellow': 'yellow',
-    '🟢 Green': 'green',
-    '🔵 Blue': 'blue',
-    'Original': 'original'
-  }
+  cattr_accessor :default_style,
+                 instance_accessor: false,
+                 default: 'original'
+  cattr_accessor :style_labels,
+                 instance_accessor: false,
+                 default: {
+                   '🔴 Red': 'red',
+                   '🟡 Yellow': 'yellow',
+                   '🟢 Green': 'green',
+                   '🔵 Blue': 'blue',
+                   'Original': 'original'
+                 }
 
   enum :style, Note.style_labels.values.index_by(&:itself),
                default: Note.default_style,
--- a/app/models/public_body.rb
+++ b/app/models/public_body.rb
@@ -50,7 +50,9 @@ class PublicBody < ApplicationRecord
   attr_accessor :no_xapian_reindex
 
   # Set to 0 to prevent application of the not_many_requests tag
-  cattr_accessor :not_many_public_requests_size, default: 5
+  cattr_accessor :not_many_public_requests_size,
+                 instance_writer: false,
+                 default: 5
 
   # Any PublicBody tagged with any of the follow tags won't be returned in the
   # batch authority search results or batch category UI
--- a/app/models/public_body.rb
+++ b/app/models/public_body.rb
@@ -54,7 +54,9 @@ class PublicBody < ApplicationRecord
 
   # Any PublicBody tagged with any of the follow tags won't be returned in the
   # batch authority search results or batch category UI
-  cattr_accessor :batch_excluded_tags, default: %w[not_apply defunct]
+  cattr_accessor :batch_excluded_tags,
+                 instance_accessor: false,
+                 default: %w[not_apply defunct]
 
   has_many :info_requests,
            -> { order(created_at: :desc) },
--- a/app/models/public_body/csv_import.rb
+++ b/app/models/public_body/csv_import.rb
@@ -9,7 +9,7 @@ module PublicBody::CsvImport
   included do
     # Default fields available for importing from CSV, in the format
     # [field_name, 'short description of field (basic html allowed)']
-    cattr_accessor :csv_import_fields do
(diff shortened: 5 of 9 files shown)


app/models/public_body/calculated_home_page.rb
```diff
@@ -4,7 +4,13 @@ module PublicBody::CalculatedHomePage
...
   included do
     cattr_accessor :excluded_calculated_home_page_domains,
-                   default: Domains.webmail_providers
+                   instance_accessor: false
+
+    def self.excluded_calculated_home_page_domains
+      var = :@@excluded_calculated_home_page_domains
+      value = class_variable_get(var) if class_variable_defined?(var)
+      value || Domains.webmail_providers
+    end
```
garethrees, line 12: Ah okay so if we're needing to fall back to the class method (totally fine) then maybe we should just implement our own methods rather than the `cattr_accessor` as this override is a bit confusing and the below seems to work fine:

```rb
  class_methods do
    def excluded_foi_officer_access_domains
      @excluded_foi_officer_access_domains ||= Domains.webmail_providers
    end

    def excluded_foi_officer_access_domains=(domains)
      @excluded_foi_officer_access_domains = domains
    end
  end
```

gbp: Could certainly drop `cattr_accessor` although not sure we want to use `||=` as it'll cache the value and I would expect it to be updated if `Domains.webmail_providers` changes.

The `cattr_accessor` method sets/get the `@@` class var which is why I've had to call `class_variable_get` and `class_variable_defined?` otherwise we get `NameError (uninitialized class variable @@excluded_calculated_home_page_domains in PublicBody)` errors.

garethrees: I'm not too worried about any of these (the `PublicBody` methods or `Domains`) changing after app boot, as they'll only be modified once in the theme, so I think caching is fine here.

To step through this (for my own benefit!), assuming the following change:

```diff
diff --git a/app/models/public_body/foi_officer_access.rb b/app/models/public_body/foi_officer_access.rb
index 5dc323e69..783ee9971 100644
--- a/app/models/public_body/foi_officer_access.rb
+++ b/app/models/public_body/foi_officer_access.rb
@@ -8,8 +8,18 @@ module PublicBody::FoiOfficerAccess
                    default: Domains.webmail_providers
   end

+  class_methods do
+    def excluded_foi_officer_access_domains
+      @excluded_foi_officer_access_domains ||= Domains.webmail_providers
+    end
+
+    def excluded_foi_officer_access_domains=(domains)
+      @excluded_foi_officer_access_domains = domains
+    end
+  end
+
   def foi_officer_domain_excluded?
-    excluded_foi_officer_access_domains.include?(request_email_domain)
+    self.class.excluded_foi_officer_access_domains.include?(request_email_domain)
   end

   # Does this user have the power of FOI officer for this body?
@@ -32,7 +42,7 @@ def domain_within_authority?(user)
   end

   def no_restrictions_on_domain?(user)
-    !excluded_foi_officer_access_domains.include?(user.email_domain)
+    !self.class.excluded_foi_officer_access_domains.include?(user.email_domain)
   end

   def outside_user?(user)
```

…and the following to make this testing a bit easier:

```diff
diff --git a/lib/domains.rb b/lib/domains.rb
index 25b5ebfdd..e3486a954 100644
--- a/lib/domains.rb
+++ b/lib/domains.rb
@@ -1,23 +1,7 @@
 # Helpers related to generic domains
 class Domains
   cattr_accessor :webmail_providers, default: %w[
-    aol.com
     gmail.com
-    googlemail.com
-    gmx.com
     hotmail.com
-    icloud.com
-    live.com
-    mac.com
-    mail.com
-    mail.ru
-    me.com
-    outlook.com
-    protonmail.com
-    qq.com
-    yahoo.com
-    yandex.com
-    ymail.com
-    zoho.com
   ]
 end
```

With no theme customisations, when we boot the app we get the built-in webmail providers:

```rb
Domains.webmail_providers
# => ["gmail.com", "hotmail.com"]
PublicBody.excluded_foi_officer_access_domains
# => ["gmail.com", "hotmail.com"]
```

We now customise the theme with UK-specific webmail providers:

```diff
diff --git a/lib/model_patches.rb b/lib/model_patches.rb
index a7db7ff..e28f1ef 100644
--- a/lib/model_patches.rb
+++ b/lib/model_patches.rb
@@ -32,6 +32,8 @@
                      "to government services.")
   end

+  Domains.webmail_providers += %w[hotmail.co.uk btinternet.com]
+
   InfoRequest.class_eval do
     def email_subject_request(opts = {})
       html = opts.fetch(:html, true)
```

This updates `Domains` and correctly populates `PublicBody`:

```rb
Domains.webmail_providers
# => ["gmail.com", "hotmail.com", "hotmail.co.uk", "btinternet.com"]
PublicBody.excluded_foi_officer_access_domains
# => ["gmail.com", "hotmail.com", "hotmail.co.uk", "btinternet.com"]
```

We now want to add a specific excluded foi officer access domain:

```diff
diff --git a/lib/model_patches.rb b/lib/model_patches.rb
index a7db7ff..9f7138b 100644
--- a/lib/model_patches.rb
+++ b/lib/model_patches.rb
@@ -32,6 +32,10 @@
                      "to government services.")
   end

+  Domains.webmail_providers += %w[hotmail.co.uk btinternet.com]
+
+  PublicBody.excluded_foi_officer_access_domains += %w[nhs.net]
+
   InfoRequest.class_eval do
     def email_subject_request(opts = {})
       html = opts.fetch(:html, true)
```

We now have the default and UK specific webmail providers, as well as the specific domain for foi officer access:

```rb
Domains.webmail_providers
# => ["gmail.com", "hotmail.com", "hotmail.co.uk", "btinternet.com"]
PublicBody.excluded_foi_officer_access_domains
# => ["gmail.com", "hotmail.com", "hotmail.co.uk", "btinternet.com", "nhs.net"]
```

The only gotcha is that we need to remember to modify `Domains` before anything else that uses it, e.g:

```diff
diff --git a/lib/model_patches.rb b/lib/model_patches.rb
index a7db7ff..d1e643f 100644
--- a/lib/model_patches.rb
+++ b/lib/model_patches.rb
@@ -32,6 +32,10 @@
                      "to government services.")
   end
 
+  PublicBody.excluded_foi_officer_access_domains += %w[nhs.net]
+
+  Domains.webmail_providers += %w[hotmail.co.uk btinternet.com]
+
   InfoRequest.class_eval do
     def email_subject_request(opts = {})
       html = opts.fetch(:html, true)
```

```rb
Domains.webmail_providers
# => ["gmail.com", "hotmail.com", "hotmail.co.uk", "btinternet.com"]
PublicBody.excluded_foi_officer_access_domains
# => ["gmail.com", "hotmail.com", "nhs.net"]
```

gbp: You've noticed the ordering issue... surely changing `||=` to `||` fixes that then we're all good?

garethrees: You still need to define the `webmail_providers` first even with the `||`:

Checked out 3b4a6f0e8 in core with `webmail_providers` reduced as above. Modify theme with:

```diff
diff --git a/lib/model_patches.rb b/lib/model_patches.rb
index a7db7ff..be26d98 100644
--- a/lib/model_patches.rb
+++ b/lib/model_patches.rb
@@ -114,22 +114,10 @@ def late_calculator
     unit
   ]
 
-  PublicBody.excluded_calculated_home_page_domains += %w[
-    aol.co.uk
-    blueyonder.co.uk
-    btconnect.com
-    btinternet.com
-    btopenworld.com
-    hotmail.co.uk
-    live.co.uk
-    ntlworld.com
-    sky.com
-    talk21.com
-    talktalk.net
-    tiscali.co.uk
-    virginmedia.com
-    yahoo.co.uk
-  ]
+
+  PublicBody.excluded_foi_officer_access_domains += %w[nhs.net]
+
+  Domains.webmail_providers += %w[hotmail.co.uk btinternet.com]
 
   PublicBody.class_eval do
     # Return the domain part of an email address, canonicalised and with common

```

```rb
Domains.webmail_providers
# => ["gmail.com", "hotmail.com", "hotmail.co.uk", "btinternet.com"]
PublicBody.excluded_foi_officer_access_domains
# => ["gmail.com", "hotmail.com", "nhs.net"]
```

Doesn't _really_ matter too much as we're not calling this all over the place of course, but doesn't seem any reason _not_ to cache given the above?

PR #8892, opened 2025-09-22
[#8890] Refactor user exceeded content limits handling

### LINKED ISSUE #8890: [ERROR] request#new (RuntimeError) "Unknown reason for ban"
```
A RuntimeError occurred in  request#new:

  Unknown reason for ban
  app/models/user.rb:538:in 'User#can_fail_html'


Request:

  * URL        : https://www.whatdotheyknow.com/new/north_somerset_council
  * HTTP Method: GET
  * IP address : REDACTED
  * Parameters : {"controller" => "request", "action" => "new", "url_name" => "north_somerset_council"}
  * Timestamp  : 2025-09-22 08:44:34 +0100
    * Rails root : /data/vhost/www.whatdotheyknow.com/alaveteli-2025-09-18T15-38-55

Session:

  * session id: [FILTERED]
  * data: {"session_id" => "REDACTED",
   "locale" => "en",
   "user_id" => REDACTED,
   "user_login_token" => "REDACTED",
   "remember_me" => true,
   "post_redirect_token" => "REDACTED",
   "_csrf_token" => "REDACTED"}

Backtrace:

  app/models/user.rb:538:in 'User#can_fail_html'
  app/controllers/request_controller.rb:193:in 'RequestController#new'
  app/controllers/application_controller.rb:127:in 'ApplicationController#record_memory'
```

gbp:
## Relevant issue(s)

Fixes #8890

## What does this do?

Refactor user exceeded content limits handling

## Why was this needed?

- Bug introduced in #8859
- Refactoring code for consistency and improved readability

<hr>

[skip changelog]

### DIFF
--- a/app/controllers/comments_controller.rb
+++ b/app/controllers/comments_controller.rb
@@ -13,7 +13,6 @@ class CommentsController < ApplicationController
   before_action :find_info_request
 
   before_action :reject_unless_comments_allowed
-  before_action :reject_if_user_banned
   before_action :set_in_pro_area
 
   before_action :build_comment
@@ -21,6 +20,7 @@ class CommentsController < ApplicationController
 
   before_action :reedit_comment, only: [:create]
   before_action :authenticate, only: [:create]
+  before_action :reject_if_user_banned
   before_action :check_for_spam_comment, only: [:create]
 
   before_action :validate_comment, only: [:preview, :create]
@@ -80,14 +80,13 @@ class CommentsController < ApplicationController
                 notice: _("Comments are not allowed on this request")
   end
 
-  # Banned from adding comments?
   def reject_if_user_banned
-    return if !authenticated? || authenticated_user.can_make_comments?
+    return unless authenticated?
 
-    if authenticated_user.banned?
+    if authenticated_user.suspended?
       @details = authenticated_user.can_fail_html
       render template: 'user/banned'
-    else
+    elsif authenticated_user.exceeded_comment_limits?
       render template: 'comments/rate_limited'
     end
   end
--- a/app/controllers/followups_controller.rb
+++ b/app/controllers/followups_controller.rb
@@ -4,9 +4,10 @@ class FollowupsController < ApplicationController
   before_action :set_incoming_message,
                 :set_info_request,
                 :set_last_request_data,
-                :check_can_followup,
-                :check_user_credentials,
+                :check_incoming_message_can_be_followed_up,
                 :check_request_matches_incoming_message,
+                :authenticate,
+                :check_user_can_followup,
                 :set_params,
                 :set_internal_review,
                 :set_outgoing_message,
@@ -43,7 +44,7 @@ class FollowupsController < ApplicationController
 
   private
 
-  def check_can_followup
+  def check_incoming_message_can_be_followed_up
     unless @info_request.is_followupable?(@incoming_message)
       @reason = @info_request.followup_bad_reason
       render action: 'followup_bad'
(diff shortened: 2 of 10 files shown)


app/models/user.rb
```diff
@@ -503,8 +503,11 @@ def exceeded_request_limits?
       InfoRequest.exceeded_creation_rate?(info_requests)
   end
 
-  def can_make_followup?
-    active?
+  def exceeded_follow_limits?
```
garethrees: If we keep this would rename this so that it's not confused with "following" (i.e. tracks)

```diff
- def exceeded_follow_limits?
+ def exceeded_followup_limits?
```

app/controllers/followups_controller.rb
```diff
@@ -4,8 +4,9 @@ class FollowupsController < ApplicationController
   before_action :set_incoming_message,
                 :set_info_request,
                 :set_last_request_data,
-                :check_can_followup,
-                :check_user_credentials,
+                :check_incoming_message_can_be_followed_up,
+                :check_user_can_followup,
+                :authenticate,
```
garethrees: should `:authenticate` happen before `:check_user_can_followup`?

app/controllers/followups_controller.rb
```diff
@@ -80,20 +81,24 @@ def check_responses_allowed
...
+
+  def check_user_can_followup
+    return unless authenticated?
+
+    if authenticated_user.suspended?
       @details = authenticated_user.can_fail_html
       render template: 'user/banned'
-      return
+    elsif authenticated_user.exceeded_follow_limits?
+      render template: 'comments/rate_limited'
```
garethrees: This doesn't make sense here. The template is all about annotations, not correspondence with the authority. ("Too many annotations. You have hit the rate limit on annotations. Users are ordinarily limited to 10 annotations per day.")

It's a little but of a moot point since at the moment there are no rate limits on followups, other than returning true if the user is suspended (so they'll hit the `user/banned` template first).

I think the only (theoretical) way a user could hit this is if they were not active but following up (which doesn't seem possible since you need to be active to make a request). Should this part of the conditional just be dropped entirely?

app/controllers/comments_controller.rb
```diff
@@ -65,7 +65,7 @@ def find_info_request
     @info_request = InfoRequest.find_by_url_title!(params[:url_title])
     return unless @info_request.embargo && cannot?(:read, @info_request)
 
-    raise ActiveRecord::RecordNotFound
+    render_hidden
```
garethrees: What's the reason for this change?

We explicitly 404 for private requests so that we don't give away their existence through the URL title.

app/controllers/followups_controller.rb
gbp: I think I was following the order of the before action hooks in CommentController - where `reject_if_user_banned` is before `authenticate`... I recall I was going to rename these `check_*` hooks `reject_if_*` eventually. But yes, we probably want to reorder these.

app/controllers/followups_controller.rb
gbp: As above I was basing the changes in this controller on some existing methods in the CommentController. This obviously isn't needed until/if we add followup limits.

app/controllers/followups_controller.rb
garethrees: If it works then no worries for now – just caught my eye that it might not work as expected, but I didn't dig in to it in detail

app/controllers/comments_controller.rb
gbp: I have reverted this

PR #8894, opened 2025-09-24
Refactor Users::ConfirmationsController

gbp:
## What does this do?

This change unifies actions for consistency.

All actions for the different post redirect circumstances are basically doing the same thing.

The only difference is for the 'change_password' action. If we take into account the change password controller/action which we get redirected to these actually perform the same operations just in different orders. EG before this change, the user would be sign in after password change form is submitted, now they get signed in before being redirected to the form.

Additionally the `@user` var was being used in two different contexts; the current logged in user and the post redirect's user, so this refactor clears that up also.

## Why was this needed?

Needed for https://github.com/mysociety/whatdotheyknow-theme/pull/2035

<hr>

[skip changelog]


garethrees (commented):
Looks okay to me in principle, and much less confusing 👍 

I don't think signing the user in for the password reset really opens up any compromise vectors that wouldn't have already existed, but plz check in with other devs to sense check that in case we've got experience of thinking this through on other apps.

Issue #8676, opened 2025-04-29
Projects: General Availability


garethrees:
# Projects: General Availability

We’ve had Projects running with beta users for a while, with the high-volume Scorecards collaboration now being done without any technical assistance. We’ll now make Projects a default feature for Pro users, and as a consequence Pro users of other Alaveteli installs.

We’ll focus on tightening up what we have rather than adding anything substantially new. Once we’ve resolved these issues, we’ll remove the feature flag.

## Project Setup Workflow

Projects is a new concept in Alaveteli so we’ll need to explain it. The key things we can add here are:

* A brief description of Projects in the dashboard view covering what it’s for and why it’s useful.  
* Explanatory notes on each setup step so that users understand why they’re filling in each field.  
* Ability to skip each step (“I’ll do this later”) and make it clear that steps can be edited after setup.

We’ve got a [public blog post](https://www.mysociety.org/2020/07/24/whatdotheyknow-projects/) and some [draft user guides](https://github.com/mysociety/alaveteli-project/issues/265#issuecomment-2304472421) that we can pull content from. We could also consider adding the draft user guides to the [Pro help page](http://help/pro) and linking to that from the interface if there’s more to say than inline help can fit.

## Improve Editing Workflow

After setting up a project, the title & briefing, requests, dataset questions and contributors are all editable.

When you navigate to one of these pages from the project home page and click “Update Project” you get taken to the next step in the initial setup wizard workflow. This is quite surprising in this context. When making these edits after initial setup, it would be better to show an indication of success (like a flash message or other UI element) and either stay on the current page, or redirect back to the homepage.

It’s not obvious that there are two steps to removing contributors, as described in [\#8510](https://github.com/mysociety/alaveteli/issues/8510). Given generating the invite link button doesn’t need a second step, we should make the “remove” button perform the removal and do away with the “update project” button here.

We should also make sure that there’s a clear link back to the project homepage from each of the edit steps.

## Editing Datasets

The one area where we should add enhancements is the ability for project owners to edit datasets. This will serve a dual purpose of allowing them to correct any issues from incorrect contributions, and also allow admins to use the interface to fulfil any support requests.

Each submission should be editable – both the classification and the dataset. Editing should just directly edit the submission, rather than creating a new submission. This will allow admins to edit through the same interface and not have the submission author change.  

<img width="1596" alt="Image" src="https://github.com/user-attachments/assets/bfc8caa2-2db8-460e-95bc-7042fce57c48" />

[...]

PR #8911, opened 2025-10-07
[#8676] Add projects admin UI

gbp:
## Relevant issue(s)

Part of #8676

## What does this do?

Add a basic projects admin UI with index/show/edit/destroy actions. Visible only to `pro admins` when the `projects` feature is enabled.

## Screenshots

image

image

image

image

### DIFF
--- a/app/controllers/admin/projects/submissions_controller.rb
+++ b/app/controllers/admin/projects/submissions_controller.rb
@@ -1,0 +1,18 @@
+##
+# Controller for managing Project::Submissions from the admin interface
+#
+class Admin::Projects::SubmissionsController < AdminController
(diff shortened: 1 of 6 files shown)


app/controllers/admin/projects_controller.rb
```diff
@@ -0,0 +1,65 @@
...
+
+  before_action :set_project, except: :index
+  before_action :authorise
+
+  sortable default: :created_at_desc, only: [:index]
+
+  def index
+    @query = params[:query]
+    if @query
+      projects = Project.where("lower(title) like lower('%'||?||'%')", @query)
```
garethrees, line 15: Could extract this to `Project.search`

Issue #2303, opened 2015-04-09
'View as HTML' label is meaningless to the citizen


equivalentideas:
For the link citizens use to view a document attached to a request in their browser:

![screen shot 2015-04-07 at 12 16 03 pm](https://cloud.githubusercontent.com/assets/1239550/7015983/1630e22a-dd20-11e4-87fc-5bceeee014db.png)

Why are we telling them what format it's in? Do they know what it means that it's in HTML?

Maybe a label like: 'View online' or 'View in your browser' or 'View text version' would be better?

I guess the key questions is 'what is the benefit of viewing it in HTML in your browser?' We should try and make the action label focus on the use to the citizen.

On the document view page we also say "This is an HTML version of an attachment to the Freedom of Information request". Same question applies here.

garethrees:
I quite like "View in your browser"

crowbot:
I think one of the main reasons for having the HTML view originally was to allow google to index the contents of the documents, so I guess the users benfitting were the ones who found what they were looking for from a google search.

garethrees:
> I think one of the main reasons for having the HTML view originally was to allow google to index the contents of the documents, so I guess the users benfitting were the ones who found what they were looking for from a google search.

As I understand it, I think the issue is more with the wording of the link rather than the actual use of viewing in the browser.

We could just have "Download" and "View"?

crowbot:
On 14 April 2015 at 11:05, Gareth Rees <notifications@github.com> wrote:

> I think one of the main reasons for having the HTML view originally was to
> allow google to index the contents of the documents, so I guess the users
> benfitting were the ones who found what they were looking for from a google
> search.
>
> As I understand it, I think the issue is more with the wording of the link
> rather than the actual use of viewing in the browser.
>
Yes, Luke's question was about what the benefit was to the user of having
the link - so that we can give the link a name that reflects that benefit.
My point was that the main benefit is rather indirect.

We could just have "Download" and "View"?
>
We could, but depending on people's setups, the "Download" link may also
display the document in the browser.

garethrees:
> We could, but depending on people's setups, the "Download" link may also
display the document in the browser.

Yeah that's true, and could be confusing.

equivalentideas:
We had a little exchange on the use of the doc web view in https://github.com/openaustralia/righttoknow/issues/473 :

> It would be interesting to see if there are any documented use cases, but here's an idea:

[...]

PR #8916, opened 2025-10-08
[#8873] Remove PDF attachment to HTML conversion

### LINKED ISSUE #8873: Replace View as HTML for PDFs
## Context

We currently use `pdftohtml` to convert PDFs to HTML for the “View as HTML” option for attachments.

We’ve had all sorts of problems with this over the years – parsing issues (https://github.com/mysociety/alaveteli/issues/3446), memory leaks (https://github.com/mysociety/alaveteli/issues/1827#issuecomment-54796601), creating thousands of images ([https://github.com/mysociety/alaveteli/issues/772](https://github.com/mysociety/alaveteli/issues/772)), and most recently missing imagescausing high numbers of unnecessary redirects ([https://github.com/mysociety/alaveteli/pull/8869](https://github.com/mysociety/alaveteli/pull/8869))[^1].

On top of that, the HTML conversion is average at best.

## Why do we extract to HTML?

As I understand it, there were three benefits to extracting PDFs to HTML[^2]:

1. At the time, PDF indexing by search engines was limited. Extracting to HTML helped get information releases indexed and findable by external search.  
2. At the time, browsers didn’t have PDF viewers and would just download the file. Having a way to preview “on site” provided a nicer user experience.  
3. We render the extracted HTML in a branded wrapper. This helps visitors entering the site via one of these pages understand a bit of context about where it's hosted (an FOI release on a third party platform, rather than an authority website) and ways to get back to the correspondence that resulted in its release.

I don’t *think* any OCR or similar is happening via this process, so I don’t think we’re getting any particular benefit in making scanned PDFs any more usable, but that’s something to check.

Given PDFs are now generally indexed and browsers have built in PDF viewers, the main benefit to the “View as HTML” view is the wrapper.

There may be *some* SEO benefit to having HTML according to [this article](https://www.thewebmaster.com/google-pdf-indexing/), but I doubt what we’re doing now is the ideal way in 2025\.

## Alternative approaches

We’ve considered replacing the PDF View as HTML view with PDF.js ([https://github.com/mysociety/alaveteli/issues/833](https://github.com/mysociety/alaveteli/issues/833)). This might enable extra benefits in future – potentially being able to highlight/annotate PDFs etc, but we don’t have any specific plans around that yet.

Instead, we could just render the PDF in an iframe, which uses the browser’s PDF viewer:

```html
<!-- Need to remove the #wrapper div and add some styling hacks to fill the page, but this is the gist -->
<div id="view-html-content">
  <iframe src="/request/implementation_report_on_europea_2/response/46227/attach/5/HR%202023%2083%20Redacted.pdf"></iframe>
</div>
```

Image

## Scope of this issue

1. Check whether `pdftohtml` is doing any OCR / text extraction from scanned or non-text-based PDFs that we’d lose if we removed it.  
2. Check whether the iframe approach works reasonably well in major browsers.  
3. Do some brief research into any additional benefits we’d get by using PDF.js instead of the iframe approach.

Unless any of the above throw up any reasons not to, we should replace the PDF HTML view with the `iframe` approach.

As part of this work, we should also rename the “View as HTML” link to just “View”, and swap the ordering so that its View (if the viewer is available) then Download.

Image

[^1]:  I’m not sure *any* of the images are working based on a few PDFs I’ve recently loaded in View as HTML.
[^2]:  https://github.com/mysociety/alaveteli/commit/729161fb2bdc308c2bb8d90f0a9a44f3b19c92c7

gbp:
## Relevant issue(s)

Fixes #8873
Fixes https://github.com/mysociety/alaveteli/issues/2303

## What does this do?

Remove PDF attachment to HTML conversion

## Why was this needed?

This was overly complex and caused numerous issues in user support, server infra, and developer happiness. Since this was originally implements modern browsers all support rendering PDFs inline using PDF.js or similar so isn't needed anymore.


app/controllers/attachments_controller.rb
```diff
@@ -2,7 +2,6 @@
 # Controller to serve FoiAttachment records in both raw and as HTML.
 #
 class AttachmentsController < ApplicationController
-  include FragmentCacheable
```
garethrees, line 5: Would obviously love to get rid of this but non-pdf stuff appears to be cached too. Run this:

```
cache/views/request$ find . -type f | grep -Ev "(pdf|jpg|png)"
```

Plenty of xls, doc, txt, etc entries.

I'm not sure whether they're actually _needed_, but worth checking.

gbp: Those should all be cached versions of the show as HTML view, ie the Google Doc Viewer wrapper. Not the actual content of the attachment.

PR #8942, opened 2025-10-23
Projects dataset export optimisation

gbp:
## What does this do?

Projects dataset export optimisation

## Notes to reviewer

<hr>

[skip changelog]

### DIFF
--- a/spec/models/project/export_spec.rb
+++ b/spec/models/project/export_spec.rb
@@ -2,7 +2,10 @@ require 'spec_helper'
 
 RSpec.describe Project::Export do
   let(:project) { instance_double('Project') }
+  let(:key_set) { instance_double('KeySet') }
   let(:instance) { described_class.new(project) }
+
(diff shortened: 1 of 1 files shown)


app/models/project/export/info_request.rb
```diff
@@ -69,7 +79,7 @@ def classification_is_latest_status_update?
   end
 
   def extraction_submission
-    submissions.extraction.last
+    extraction_project_submissions.last
```
gbp, line 82: This is not scoped to the project

app/models/project/export/info_request.rb
```diff
@@ -32,34 +30,46 @@ def base_data
...
+    end
   end
 
   def submissions
     project.submissions.where(info_request: id)
   end
 
   def classification_submission
-    submissions.classification.last
+    classification_project_submissions.last
```
gbp, line 67: This is not scoped to the project

PR #8995, opened 2025-11-10
Add skip CI action checkboxes

gbp:
## What does this do?

Add skip CI action checkboxes to GH pull request template

## Why was this needed?

Sometimes we know better if the changelog and/or RuboCop actions should be run or not.

<hr>

- [x] Skip changelog
- [x] Skip RuboCop


garethrees:
Love this

garethrees:
Could live with it. Other option would be to use labels, but that's much less obvious to know that they exist. I don't mind either way (including sticking with something like the current version).

Issue #8803, opened 2025-08-15
Version raw emails to persist redactions and ability to destroy original copy


garethrees:
## Persisting Redactions

We already implemented attachment locking in https://github.com/mysociety/alaveteli/issues/8622. We did discuss revisiting this to see if we can move to more of a one-way flow.

## Ability to destroy original copy of raw email

Irrespective of whether we decide to look at making destructive modifications to the raw email to persist redactions, in some cases we'll want to totally remove it entirely.

We don't want to destroy the `RawEmail` database record, as that provides key metadata (e.g. https://github.com/mysociety/alaveteli/issues/8804) that response did arrive. It's better to think of this that we want to _erase the PII_. That means destroying the underlying Active Storage file, and handling the case that the `RawEmail` now no longer has any data.

While this is more about the underlying data storage and removing the original because we've now got redacted/locked versions, we'll want to be mindful of the interaction with https://github.com/mysociety/alaveteli/issues/8806 here.

garethrees:
Closing this as the main functionality has been completed via https://github.com/mysociety/alaveteli/pull/9040.

PR #9021, opened 2025-12-03
Move raw email cache columns from incoming messages

gbp:
## Relevant issue(s)

Connected #8803 

## What does this do?

Move raw email cache columns from incoming messages

### DIFF
--- a/doc/CHANGES.md
+++ b/doc/CHANGES.md
@@ -2,6 +2,7 @@
 
 ## Highlighted Features
 
+* Move raw email cache columns from incoming messages (Graeme Porteous)
 * Improve raw email testing fixtures (Graeme Porteous)
 * Add support for Debian 13 "Trixie" (Graeme Porteous)
 * Add support for Ubuntu 24.04 LTS "Nobel Numbat" (Graeme Porteous)
@@ -12,6 +13,10 @@
 * Block various action links via robots.txt (Laurent Savaete)
 
 ## Upgrade Notes
+
+* _Recommended:_ To migrate `RawEmail` cache columns, please run:
+
+      bin/rails temp:migrate_cache_columns_to_raw_emails
 
 ### Changed Templates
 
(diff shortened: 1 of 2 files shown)


app/models/raw_email.rb
```diff
@@ -22,6 +22,8 @@
 class RawEmail < ApplicationRecord
   # deliberately don't strip_attributes, so keeps raw email properly
 
+  prepend RawEmail::CacheAttributes
```
garethrees, line 26: This file/code is missing so the spec failures aren't representative of the actual issues.

app/models/raw_email/cache_attributes.rb
```diff
@@ -0,0 +1,42 @@
...
+      from_name: MailHandler.get_from_name(mail),
+      from_email: from_email,
+      from_email_domain: PublicBody.extract_domain_from_email(from_email) || '',
+      valid_to_reply_to: ReplyToAddressValidator.valid?(from_email) &&
+        !empty_return_path? && !auto_submitted?
+    )
+  end
+
+  def should_cache_attributes?
+    file.attached? && read_attribute(:subject).blank?
```
garethrees: What happens when a valid email comes in with no subject? Would checking `from_email` be more reliable?

app/models/raw_email.rb
garethrees, line 26: OOI, why prepend rather than include?

gbp: Include caused spec failures - iirc the methods, which we dynamically define in the concern to ensure the values are cached, these weren't being used, Instead it the methods defined by ActiveRecord/Model would be called.

app/models/raw_email/cache_attributes.rb
gbp: Am going to also cache the Message-ID header and switch to that. We already use the message IDs when we look to [accept a message onto an info request correspondence](https://github.com/mysociety/alaveteli/blob/e36bca90ed6e81c4a335a289bd8121d184cfd034/app/models/info_request.rb#L929) so will improve performance there.

garethrees: Nice yeah that's better

PR #9023, opened 2025-12-04
Handle array return value for auto-submitted mail header

gfrmin:
## Summary

This PR fixes a crash in the mail handler when processing emails with multiple `auto-submitted` headers.

## Problem

The Mail gem returns an Array when multiple headers with the same name exist in an email (which is valid per RFC 5322). The `get_auto_submitted` method in `MailHandler::Backends::MailBackend` assumed a single header object and would crash with `NoMethodError: undefined method 'value' for Array` when encountering duplicate auto-submitted headers.

## Solution

- Added handling for both single header objects and arrays of headers
- When multiple headers exist, returns the first value
- Maintains backward compatibility with single header case

## Files Changed

- `lib/mail_handler/backends/mail_backend.rb` - Added array handling in `get_auto_submitted` method
- `spec/lib/mail_handler/backends/mail_backend_spec.rb` - Added test case for multiple headers

## Test Coverage

Added new test case that matches existing testing patterns:
```ruby
it 'returns the first value when multiple auto-submitted headers exist' do
  mail = Mail.new
  mail['auto-submitted'] = 'auto-replied'
  mail['auto-submitted'] = 'auto-generated'
  expect(get_auto_submitted(mail)).to eq('auto-replied')
end
```

## Impact

Prevents email processing crashes when receiving emails with duplicate auto-submitted headers from certain mail servers.

### DIFF
--- a/doc/CHANGES.md
+++ b/doc/CHANGES.md
@@ -2,6 +2,7 @@
 
 ## Highlighted Features
 
+* Handle array return value for auto-submitted mail header (gfrmin)
 * Restore logging of :email parameters (Gareth Rees)
 * Fix importing holidays from iCal feed (Gareth Rees)
 * Ability to search users by tag in admin interface (Gareth Rees)
--- a/lib/mail_handler/backends/mail_backend.rb
+++ b/lib/mail_handler/backends/mail_backend.rb
@@ -162,7 +162,10 @@ module MailHandler
       end
 
       def get_auto_submitted(mail)
-        mail['auto-submitted'] ? mail['auto-submitted'].value : nil
+        field = mail['auto-submitted']
+        return nil unless field
+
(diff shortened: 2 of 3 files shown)


garethrees:
This looks good in principle.

[RFC 5322](https://www.rfc-editor.org/rfc/rfc5322) doesn't seem to say much about this header, but [RFC 3834](https://www.rfc-editor.org/rfc/rfc3834) states:

> The maximum number of Auto-Submitted fields that may appear in a
> message header is 1.

https://gitlab.com/gitlab-org/gitlab-foss/-/issues/30617 also suggests that we should only be expecting one.

That said, I appreciate RFC 3834 may have been updated and in any case, we've got to deal with what we receive so I think this should be fine.

Issue #6931, opened 2022-04-06
Extract major domain concerns for application composability


garethrees:
The application's code quality and development flexibility could be greatly improved by making it more composable. There are quite a few useful areas that could be generalised. This issue is just a kicking off point and way to group the similar work.

There are obviously more extractions we can make across the codebase – `Classifiable`, `Trackable`, `Citable`, etc – but the list below represents the major generalisations that can be _widely shared_ and are the most important for adding new behaviour quickly.

Each of these concerns warrant their own issue, but at a high level I will articulate them here.

## `Reportable`

Instead of marking records as `attention_requested`, and having to repeat all the reporting functionality across models, we should have a `Report` that handles all the behaviour associated with a report – alerting, closing, etc. We then make content `Reportable` by including a concern and creating a report against the reported record.

## `Commentable`

Comments should be much more general purpose. There's the current in-public use, but we'd like to bring more admin discussion into the app. For example, I can imagine a `PublicBodyChangeRequest` having comments associated with it as part of internal discussion as to whether to accept the request. All of these are essentially comments, but right now a `Comment` is too tied to `InfoRequest` to be useful for anything else.

We'll also want to `include Prominenceable` to comments at some point, rather than the current "hidden" mechanism.

## `Prominenceable`

We use prominence in several places, and have made a good start at extracting shared behavior into `MessageProminence`, but right now it doesn't _feel_ easy to add prominence to other models – most notably `FoiAttachment` and `Comment`. `Prominenceable` shouldn't just be about messages.

It's probably not appropriate to back `Prominenceable` by a table in the same way as `Reportable`, but perhaps we want to take an approach similar to the `has_tag_string` or `acts_as_xapian` macros which allow the configuration of the appropriate prominence values.

## `Reindexable`

This is a tiny piece of the code, but is core to how Alaveteli handles changes to records. We've got a hotchpotch of mechanisms for this right now:

```ruby
OutgoingMessage
  after_update :xapian_reindex_after_update

IncomingMessage
  after_destroy :update_request
  after_update :update_request

Comment
  after_save :reindex_request_events

# And so on…
```

After we add `Prominenceable` to `FoiAttachment`, for example, which of these methods do we use? It would be better to just `include Reindexable`.

## `Taggable`

We already have a `has_tag_string` macro and `Taggable` concern, but the underlying code is pretty confusing, it's not a polymorphic model, and there's probably a bit of cleanup in how we assign and use tags throughout the rest of the application.

[...]

PR #9040, opened 2025-12-18
Ability to erase RawEmail

garethrees:
## Relevant issue(s)

Part of #8803
Supersedes https://github.com/mysociety/alaveteli/pull/8736.

## What does this do?

Allows admins to erase the underlying raw email data of responses.

## Why was this needed?

Destroying the underlying copy of the response email we received is the only way to entirely remove PII held within responses from the system.

## Implementation notes

- [x] ~Might want to call `parse_raw_email!` before erasing to ensure that we've cached _a_ version before total destruction. Done~ EDIT: now unnccessary.

## Screenshots

Imagine an incoming message with an attachment that's difficult to redact:

Screenshot 2025-12-19 at 15 43 18

We navigate to the attachment to replace it:

Screenshot 2026-01-27 at 11 54 23

Which adds our replacement:

Screenshot 2025-12-19 at 15 58 40

And locks the attachment:

Screenshot 2025-12-19 at 16 01 47

But we still hold the original attachment in the `RawEmail#file`. Now we can visit the raw email (via the incoming message) in order to erase the underlying raw email file that we hold:

Screenshot 2026-01-27 at 11 56 13

This preserves the `RawEmail` database record, but handles the fact it doesn't have an attached `file`.

Screenshot 2025-12-19 at 16 03 31

We also see that downloading the raw email has been disabled, and all attachments have been locked.

Screenshot 2025-12-19 at 16 04 07

On the admin page for the request we see an event history of locking the unlocked attachments and erasing the raw email:

Screenshot 2026-01-27 at 11 59 33

On the public view we still see main body content and replaced attachment:

Screenshot 2025-12-19 at 16 11 14

Screenshot 2025-12-19 at 16 11 59

## Notes to reviewer

Assuming the standard fixture data is loaded, you can create the above email with something along the lines of:

```
$ cat spec/fixtures/files/incoming-request-pdf-attachment.eml | \
  sed 's/^From: EMAIL_FROM$/From: Olive Officer <officer@example.org>/' | \
  sed 's/^To: FOI Person <EMAIL_TO>$/To: Robin Houston <foi+request-107-9ea720b7@alaveteli.10.10.10.30.xip.io>/' | \
  docker-compose run --rm -T app script/mailin
```

And then visit `/request/107` in the browser.

<hr>

### DIFF
--- a/doc/CHANGES.md
+++ b/doc/CHANGES.md
@@ -2,6 +2,7 @@
 
 ## Highlighted Features
 
+* Allow erasure of underlying raw email data (Graeme Porteous, Gareth Rees)
 * Restore logging of :email parameters (Gareth Rees)
 * Fix importing holidays from iCal feed (Gareth Rees)
 * Ability to search users by tag in admin interface (Gareth Rees)


app/models/incoming_message.rb
```diff
@@ -96,6 +96,8 @@ def parse_raw_email!(force = nil)
     raise "Incoming message id=#{id} has no raw_email" if raw_email.nil?
 
     if !force.nil? || last_parsed.nil?
+      raise "Can't parse an erased raw_email" if raw_email_erased?
```
garethrees: Should probably make this an actual error class (`RawEmailErased` ?) so that we can catch it if we're ever writing scripts that force parsing.

spec/models/foi_attachment_spec.rb
```diff
@@ -645,6 +645,49 @@ def last_event
...
+      it { is_expected.to eq(true) }
+    end
+
+    context 'when it is already locked' do
+      let(:foi_attachment) { FactoryBot.create(:body_text, :locked) }
+
+      it 'remains locked' do
+        expect(foi_attachment.reload).to be_locked
+      end
+
```
garethrees, line 892: Could also assert that it doesn't create an event:

```rb
      it 'does not log an event' do
        expect { subject }.not_to change { last_event }
      end
```

app/views/admin/foi_attachments/edit.html.erb
```diff
@@ -120,9 +120,22 @@
...
 
-  <blockquote>
-    <div class="well well-small">
-      <%= simple_format(@foi_attachment.unmasked_body) %>
+  <% if @foi_attachment.incoming_message.raw_email_erased? %>
+    <div class="row">
+      <div class="span12">
+        <div class="alert alert-warning">
+          <p>
+            Cannot show the unmasked data as the underlying raw email for this
```
garethrees: underlying ~raw email~ data

app/views/admin_raw_email/show.html.erb
```diff
@@ -9,10 +9,19 @@
...
-  <span class="muted">Download the email for an exact rendering.</span>
-</p>
+
+<% if @raw_email.erased? %>
+  <p>
+    <%= link_to 'Download', '', class: 'btn', disabled: true %>
+    <span class="muted">Erased raw emails cannot be downloaded.</span>
+  </p>
+<% else %>
+
```
garethrees: remove extra newline

app/views/admin_raw_email/show.html.erb
```diff
@@ -75,5 +75,61 @@
...
+          <p>
+            <ul>
+              <li>Destroys the underlying file representation of the response email received</li>
+              <li>Locks all attachments that were extracted from this response</li>
+            </ul>
+          </p>
+
+          <p>
+            Note that <strong>this will not erase from any backups</strong>
+            created outside of Alaveteli or from blob storage retention
```
garethrees: "blob storage retention schemes" doesn't make sense to a non-technical user; need to rephrase

app/controllers/admin/raw_emails/erasures_controller.rb
```diff
@@ -0,0 +1,33 @@
...
+  def set_raw_email
+    @raw_email = RawEmail.find(params[:raw_email_id])
+  end
+
+  def erasure_reason
+    erasure_params[:erasure_reason]
+  end
+
+  def erasure_params
+    params.require(:raw_email).permit(:erasure_reason)
```
garethrees: In a future commit I've made this required at the html form level

app/models/foi_attachment.rb
```diff
@@ -306,6 +309,31 @@ def update_and_log_event(event: {}, **params)
     )
   end
 
+  def lock(editor:, reason: nil)
```
garethrees: Not sure why rubocop thinks this is a predicate method

app/models/raw_email.rb
```diff
@@ -75,7 +77,36 @@ def data_as_text
   end
 
   def erased?
-    false
+    !file.attached? && erased_at.present?
+  end
+
+  def erase(editor:, reason: nil)
```
garethrees: Might want to call `IncomingMessage#parse_raw_email!` before erasing to ensure that we've cached a version before total destruction.

app/models/foi_attachment.rb
```diff
@@ -314,6 +314,10 @@ def lockable?
     unlocked?
   end
 
+  def unlockable?
+    !incoming_message.raw_email_erased?
```
gbp: The incoming_message association is optional?!? Maybe that is wrong... if not this could result in a NoMethodError.

Assuming it is wrong maybe we need a direct association to raw_email?

app/models/raw_email.rb
```diff
@@ -67,6 +76,39 @@ def data_as_text
...
 
+  def erased?
+    !file.attached? && erased_at.present?
+  end
+
+  def erase(editor:, reason: nil)
+    raise AlreadyErasedError if erased?
+
+    transaction do |t|
+      t.after_rollback { return false }
```
gbp, line 94: Have you checked this works?

I would assume the `return false` is superseded by the transaction block completing so the erase method with return truthful.

In the controller does the error message actually render?

app/models/raw_email.rb
gbp: Data can be nil now. A NoMethodError can be raised

app/controllers/admin/raw_emails/erasures_controller.rb
```diff
@@ -0,0 +1,34 @@
+# Allow admins to erase RawEmail records.
+class Admin::RawEmails::ErasuresController < AdminController
+  before_action :set_raw_email
+
```
gbp, line 7: Do we need a `rescue_from RawEmail::AlreadyErasedError` here?

Without it we should except exception emails, maybe add a test example for this.

app/views/admin_raw_email/show.html.erb
```diff
@@ -9,10 +9,18 @@
...
 <% end %>
 
-<p>
-  <%= link_to 'Download', admin_raw_email_path(@raw_email, format: 'eml'), class: 'btn' %>
-  <span class="muted">Download the email for an exact rendering.</span>
-</p>
+
+<% if @raw_email.erased? %>
+  <p>
+    <%= link_to 'Download', '', class: 'btn', disabled: true %>
```
gbp, line 15: Does the disable attribute work for anchor tags?

app/views/admin_raw_email/show.html.erb
```diff
@@ -9,10 +9,18 @@
...
+    <span class="muted">Erased raw emails cannot be downloaded.</span>
+  </p>
+<% else %>
+  <p>
+    <%= link_to 'Download', admin_raw_email_path(@raw_email, format: 'eml'), class: 'btn', disabled: @raw_email.erased? %>
```
gbp: The disable attribute can be dropped inside the conditional

app/models/foi_attachment.rb
```diff
@@ -306,6 +309,31 @@ def update_and_log_event(event: {}, **params)
...
 
+  def lock(editor:, reason: nil)
+    return true if locked?
+
+    update_and_log_event(
+      event: { editor: editor, reason: reason },
+      locked: true
+    )
+
+    # TODO: Do we need to expire the request here?
```
gbp: TODO

app/models/raw_email.rb
```diff
@@ -67,6 +76,39 @@ def data_as_text
...
+
+      # TODO: Use a GID here, or pass additional params?
```
gbp: TODO

app/models/foi_attachment.rb
garethrees: Yeah I don't think there are any valid cases where we'd have an orphan `FoiAttachment` – doesn't exist on WDTK either:

```rb
FoiAttachment.where(incoming_message_id: nil).count
# => 0
```

Was added in a17230914c874824b1f56cf739cc5a1bb9e48048 as part of a wider config change so I guess that was just taking the simple route and mirroring the general previous behaviour?

app/views/admin_raw_email/show.html.erb
garethrees: Yup:

Image

```html
<p>
  <a class="btn" disabled="disabled" href="">Download</a>
  <span class="muted">Erased raw emails cannot be downloaded.</span>
</p>
```

app/controllers/admin/raw_emails/erasures_controller.rb
garethrees: Yeah could do. Its prevented in the UI so quite unlikely, but easy to add so might as well do so.

app/models/raw_email.rb
garethrees: Yep, it works, although I got the flash key wrong (and need to set explicitly because of the non-standard flash key).

Here's a diff to test (assuming the default sample data):

```diff
diff --git a/app/controllers/admin/raw_emails/erasures_controller.rb b/app/controllers/admin/raw_emails/erasures_controller.rb
index 576de5494..1d7187001 100644
--- a/app/controllers/admin/raw_emails/erasures_controller.rb
+++ b/app/controllers/admin/raw_emails/erasures_controller.rb
@@ -8,7 +8,8 @@ def create
     if @raw_email.erase(editor: admin_current_user, reason: erasure_reason)
       redirect_to admin_raw_email_path(@raw_email), notice: success_message
     else
-      redirect_to admin_raw_email_path(@raw_email), alert: failure_message
+      flash[:error] = failure_message
+      redirect_to admin_raw_email_path(@raw_email)#, error: failure_message
     end
   end
 
diff --git a/app/models/foi_attachment.rb b/app/models/foi_attachment.rb
index 3b0c3d3c1..99f96994d 100644
--- a/app/models/foi_attachment.rb
+++ b/app/models/foi_attachment.rb
@@ -310,6 +310,8 @@ def update_and_log_event(event: {}, **params)
   end
 
   def lock(editor:, reason: nil)
+    return false if id == IncomingMessage.find(2).foi_attachments.first.id
+
     return true if locked?
 
     update_and_log_event(
```

Navigate to `/admin/raw_emails/2` and attempt to erase the email:

Image

I haven't tested all permutations yet (e.g. what if `log_event` fails; what if `file.purge` fails, etc) but in principle it works.

We could drop the `after_rollback` callback (so a failed transaction would return `nil`), but I thought the explicit `false` return on failure was clearer in this situation.

garethrees: > I haven't tested all permutations yet (e.g. what if `log_event` fails; what if `file.purge` fails, etc) but in principle it works.

Need to explicitly raise a rollback for some of these

app/models/raw_email.rb
garethrees: yes; erasing before attachments have been extracted / masked / processed causes breakages.

app/models/foi_attachment.rb
garethrees: Needed `optional: true` otherwise lots of tests fail (presumably because we're creating orphan `FoiAttachment` records to test incidental behaviour).

Maybe just raise an exception here if the incoming message doesn't exist as it doesn't look like we'd hit an error for this in reality.

app/models/foi_attachment.rb
garethrees: We expire the attachment in the controller: https://github.com/mysociety/alaveteli/blob/raw-email-erase/app/controllers/admin/foi_attachments_controller.rb#L16C8-L16C29

So yes.

app/models/raw_email.rb
garethrees: Should be able to do something like:

```rb
lock_all_attachments(editor: editor, reason: 'Erasing', raw_email: self)
```

This will then generate params to the effect of:

```
reason: erasing
raw_email: gid-of-raw-email
```

app/controllers/admin/raw_emails/erasures_controller.rb
garethrees: We require reasons for locking and replacement so might as well be consistent here too.

app/models/foi_attachment.rb
garethrees: In fact, the `NoMethodError` will do for now. I don't _think_ we'll see this in practice. If we do we can add extra handling.

app/models/foi_attachment.rb
gbp: I think if you return true instead of calling `locked?` at the end of the method it'll make Rubocop happy

app/models/raw_email.rb
garethrees: Need to accept generic params/args in:

* `IncomingMessage#lock_all_attachments`
* `FoiAttachment#lock`

garethrees:
```diff
diff --git a/app/models/foi_attachment.rb b/app/models/foi_attachment.rb
index 7ca1c3728..3bf9f8bb9 100644
--- a/app/models/foi_attachment.rb
+++ b/app/models/foi_attachment.rb
@@ -309,11 +309,11 @@ def update_and_log_event(event: {}, **params)
     )
   end
 
-  def lock(editor:, reason: nil)
+  def lock(editor:, reason: nil, event: {})
     return true if locked?
 
     update_and_log_event(
-      event: { editor: editor, reason: reason },
+      event: event.merge(editor: editor, reason: reason),
       locked: true
     )
diff --git a/app/models/incoming_message.rb b/app/models/incoming_message.rb
index fc884eee9..def02c711 100644
--- a/app/models/incoming_message.rb
+++ b/app/models/incoming_message.rb
@@ -505,12 +505,12 @@ def extract_attachments
     old_attachments.each(&:mark_for_destruction)
   end
 
-  def lock_all_attachments(editor:, reason: nil)
+  def lock_all_attachments(editor:, reason: nil, event: {})
     result = nil
 
     transaction do
       result = foi_attachments.all? do |attachment|
-        attachment.lock(editor: editor, reason: reason)
+        attachment.lock(editor: editor, reason: reason, event: event)
       end
 
       raise ActiveRecord::Rollback unless result
diff --git a/app/models/raw_email.rb b/app/models/raw_email.rb
index 6da74310f..7ade7c26c 100644
--- a/app/models/raw_email.rb
+++ b/app/models/raw_email.rb
@@ -88,11 +88,14 @@ def erase(editor:, reason: nil)
     transaction do |t|
       t.after_rollback { return false }
 
-      # TODO: Use a GID here, or pass additional params?
-      lock_reason = "Erasing RawEmail##{id}"
+      lock_all_params = {
+        editor: editor,
+        reason: 'RawEmail#erase',
+        event: { raw_email: self }
+      }
 
       raise ActiveRecord::Rollback unless
-        lock_all_attachments(editor: editor, reason: lock_reason)
+        lock_all_attachments(lock_all_params)
 
       raise ActiveRecord::Rollback unless
         log_event(
```

Screenshot 2026-01-07 at 15 48 08

garethrees: Works, but adds quite a bit of complexity 🤔

gbp: That was kinda how I had thought... the `editor` and `reason` arguments could also be passed in within the `event` argument hash.

I would look at our `update_and_log_event` methods and see if there is a common approach to all `*_and_log_event` methods which could be encapsulated and extracted into a concern to reduced complexity - maybe not right now but in the future it seems like it would be a good approach.

garethrees: Yeah I was also thinking about an `Eventable` concern (https://github.com/mysociety/alaveteli/issues/6931), but would need a bit of work.

I added the editor and reason as kwargs so that they're _required_ for these operations. Don't think there's a way to do do this in another way (that's more elegant)?

garethrees: Suppose it will be a bit cleaner when I make `reason` required rather than a `nil` default.

garethrees: We now log the raw email for the locking and erasure:

Screenshot 2026-01-08 at 12 23 40

app/models/raw_email.rb
```diff
@@ -62,9 +71,46 @@ def data
...
+      raise ActiveRecord::Rollback unless
+        log_event(
+          'erase_raw_email',
+          editor: editor,
+          reason: reason,
+          raw_email: self,
+          storage_key: storage_key
+        )
+
+      file.purge
```
garethrees: Note this should probably be `purge_later` since the [documentation for `purge`](https://api.rubyonrails.org/classes/ActiveStorage/Blob.html#method-i-purge) states:

> deleting the file off the service will initiate an HTTP connection to the service, which may be slow or prevented, so you should not use this method inside a transaction or in callbacks. Use [purge_later](https://api.rubyonrails.org/classes/ActiveStorage/Blob.html#method-i-purge_later) instead

config/routes.rb
```diff
@@ -699,6 +699,16 @@ def self.matches?(request)
   end
   ####
 
+  #### Admin::RawEmails::Erasure controller
```
garethrees: Erasure_s_

app/models/raw_email.rb
garethrees: @gbp what do you think about this? By this point in the transaction we're happy that the key things have succeeded (locking & log event). Assume we're confident enough that the purge job is always going to succeed?

app/models/foi_attachment.rb
```diff
@@ -306,6 +309,31 @@ def update_and_log_event(event: {}, **params)
...
 
+  def lock(expire: true, editor:, reason:, **event)
+    return true if locked?
+
+    update_and_log_event(
+      event: { **event, editor: editor, reason: reason },
+      locked: true
+    )
+
+    public_send(:expire) if expire
```
garethrees: dur can make this:

```diff
- public_send(:expire) if expire
+ self.expire if expire
```

app/controllers/admin/raw_emails/erasures_controller.rb
```diff
@@ -0,0 +1,42 @@
...
+  private
+
+  def set_raw_email
+    @raw_email = RawEmail.find(params[:raw_email_id])
+  end
+
+  def erasure_reason
+    erasure_params[:erasure_reason].presence ||
+      raise(ActionController::ParameterMissing, :erasure_reason)
```
gbp: I think this works:

```suggestion
  params.require(:raw_email).require(:erasure_reason)
```

gbp: Could drop the `erasure_params` method too then

app/models/raw_email.rb
```diff
@@ -62,9 +73,51 @@ def data
...
+      raise ActiveRecord::Rollback unless
+        log_event(
+          'erase_raw_email',
+          editor: editor,
+          reason: reason,
+          raw_email: self,
+          storage_key: storage_key
+        )
+
+      file.purge_later
```
gbp, line 112: Have we considered what happens if this background job fails?

app/controllers/admin/raw_emails/erasures_controller.rb
gbp: We're raising `UnmaskedAttachmentsError` from inside `#erase` now too. Probably should rescue from that too

app/models/foi_attachment/lockable.rb
```diff
@@ -6,14 +6,37 @@ module FoiAttachment::Lockable
   included do
     before_save :handle_locked
 
+    validate :must_be_unlockable_to_unlock, on: :update
+
     scope :locked, -> { where(locked: true) }
     scope :unlocked, -> { where(locked: false) }
   end
 
+  def lock(expire: true, editor:, reason:, **event)
```
gbp: Wouldn't a better name for this method be `lock!`?

app/models/raw_email.rb
```diff
@@ -62,9 +73,51 @@ def data
...
+
+  def erasable?
+    all_attachments_masked?
+  end
+
+  def erase(editor:, reason:)
+    raise AlreadyErasedError if erased?
+    raise UnmaskedAttachmentsError unless all_attachments_masked?
+
+    incoming_message.parse_raw_email
```
gbp: Do we still need to call this if we're ensuring attachments are masked?

app/models/raw_email.rb
gbp: Still wasn't 100% convinced this worked as expected... but from testing it does seem okay:
```
[1] pry(main)> results = []
[1] pry(main)>
[1] pry(main)> test_transaction = -> do
[1] pry(main)>   ActiveRecord::Base.transaction do |t|
[1] pry(main)>     t.after_rollback { return false }
[1] pry(main)>     results << 'inside_transaction_1'
[1] pry(main)>     raise ActiveRecord::Rollback
[1] pry(main)>     results << 'inside_transaction_2'
[1] pry(main)>   end
[1] pry(main)>   results << 'post_transaction'
[1] pry(main)>   true
[1] pry(main)> end
[1] pry(main)>
[1] pry(main)> p test_transaction.call
[1] pry(main)> p results
[1] pry(main)>
false
["inside_transaction_1"]
```

app/models/foi_attachment/lockable.rb
garethrees: For what reason are you thinking? At the moment it's using `update_and_log_event`, not `update_and_log_event!` so it's not using an `!` save/update call. I've also always [understood](https://thoughtfulapps.com/articles/ruby/about-ruby-bang-methods) that you should only add bang methods when there's a safe counterpart. Sometimes in Rails apps it is nicer to add `!` to signify the type of update under the hood, even if there's no safe counterpart, but that doesn't apply in this case.

Besides, aiming to refactor this a bit more as per [this branch](https://github.com/mysociety/alaveteli/compare/refactor-locking-replacement), so think its fine as it is for now unless there's a really compelling reason to change.

garethrees: Though in looking at this I think we need to handle the return value of `update_and_log_event` since `lock` will return true even if the update fails.

app/models/raw_email.rb
garethrees: Good point – no, we can remove this

app/models/raw_email.rb
garethrees: Do we get exception notifications for failed background jobs? I was assuming that we did (i.e. it retries for the specified time/number of retries, then emails once it becomes dead).

If not then I guess we'll want to add a task along the lines of https://guides.rubyonrails.org/active_storage_overview.html#purging-unattached-uploads

gbp: We're not notified at present. They remain in the dead queue and we periodically check this.

garethrees: I've extracted this out to its own issue (https://github.com/mysociety/alaveteli/issues/9099) since I don't think its a blocker for merge and initial release on WDTK, but obviously something we want to do something about as part of the wider project.

PR #9051, opened 2025-12-29
Fix crashes from invalid public_body_id in request form

gfrmin:
## Summary

Defend against invalid `public_body_id` parameters that cause crashes in the request form.

## Problem

Attackers sending invalid or malicious `public_body_id` values (e.g., `abc`, `99999`, or SQL injection attempts) causes:

1. ActiveRecord safely rejects the invalid value → returns `nil` for `public_body`
2. Controller/View tries to access `@info_request.public_body.is_requestable?` → crashes with `undefined method for nil:NilClass`
3. Error emails spam administrators via exception notification

## Solution

Add explicit nil check for `@info_request.public_body` in both controller paths (`new` method and `render_new_compose`) before accessing any `public_body` methods. Invalid values now redirect to frontpage instead of crashing.

## Test plan

- [x] Verify normal request form still works with valid public body
- [ ] Test that invalid `public_body_id` parameter (e.g., `abc`, `99999`) no longer causes crashes
- [ ] Confirm error notifications are no longer triggered by these attacks

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/app/controllers/request_controller.rb
+++ b/app/controllers/request_controller.rb
@@ -213,6 +213,12 @@ class RequestController < ApplicationController
                                                       outgoing_message_params)
     @outgoing_message = @info_request.outgoing_messages.first
 
+    # Redirect if public body doesn't exist (invalid public_body_id)
+    unless @info_request.public_body
+      redirect_to frontpage_url
+      return
+    end
+
     # Maybe we lost the address while they're writing it
     unless @info_request.public_body.is_requestable?
       render action: "new_#{ @info_request.public_body.not_requestable_reason }"
@@ -550,6 +556,12 @@ class RequestController < ApplicationController
     params[:info_request][:tag_string] = params[:tags] if params[:tags]
 
     @info_request = InfoRequest.new(info_request_params)
+
+    # Redirect if public body doesn't exist (invalid public_body_id)
+    unless @info_request.public_body
+      redirect_to frontpage_url
+      return
+    end
 
     params[:info_request_id] = @info_request.id
 
--- a/doc/CHANGES.md
+++ b/doc/CHANGES.md
@@ -2,6 +2,7 @@
 
 ## Highlighted Features
 
+* Reduce exception notification spam from invalid public body requests (gfrmin)
 * Fix importing holidays from iCal feed (Gareth Rees)
 * Ability to search users by tag in admin interface (Gareth Rees)
 * Add ATI Network Impacts Showcase (Lucas Cumsille Montesinos, Gareth Rees)


commonlib
gbp, line 1: We shouldn't rollback commonlib to an older version. Please keep at 24bf6d5.

app/controllers/request_controller.rb
```diff
@@ -213,6 +213,13 @@ def new
                                                       outgoing_message_params)
     @outgoing_message = @info_request.outgoing_messages.first
 
+    # Check if public body exists and is requestable
+    if @info_request.public_body.nil?
+      @info_request.errors.add(:public_body, "Please select an authority")
+      render action: 'new'
```
gbp: Is rendering `new` the best thing to do here? Without a public body perhaps we should redirect back to `/select_authority` form or the root path

app/controllers/request_controller.rb
```diff
@@ -584,6 +593,13 @@ def render_new_compose
     end
     @outgoing_message.set_signature_name(@user.name) unless @user.nil?
 
+    # Check if public body exists
+    if @info_request.public_body.nil?
+      @info_request.errors.add(:public_body, "Please select an authority")
+      render action: 'new'
```
gbp: As above, is rendering `new` the best idea?

gbp (changes requested):
Are you sure this PR is complete? In my quick testing it doesn't fix the input validation issues.

Using this cURL request as an example:
```
curl 'http://127.0.0.1:3000/new' --data 'authenticity_token=&outgoing_message%5Bbody%5D=Dear+Ministry+of+Silly+Walks%2C%0D%0A%0D%0AHow+many+staff+do+you+have%3F%0D%0A%0D%0AYours+faithfully%2C%0D%0A%0D%0ABob&info_request%5Btitle%5D=Staff+count&info_request%5Bpublic_body_id%5D=abc&info_request%5Btag_string%5D=&submitted_new_request=1&preview=0&reedit=Edit+your+request'
```

It has `public_body_id=abc` but yet it is erroring on `app/views/request/new.html.erb:35`.

Thanks for the contribution. It would help if you could ensure that commits in your PR are logically organised, with any bug fixes or corrections squashed into the original relevant commits before submission, so that each commit represents a complete, coherent change to the codebase.

gfrmin:
Thanks for the review! I've completely reworked the PR based on your feedback:

Changes made:
  - Reset the branch to develop (removes the commonlib rollback and unrelated file changes)
  - Squashed everything into a single clean commit
  - Changed approach from render action: 'new' to redirect_to frontpage_url as you suggested

The fix now:
  - Adds explicit nil check for @info_request.public_body in both controller paths (new method and render_new_compose)
  - Redirects to frontpage when public_body is nil, preventing the view from ever trying to access methods on nil

Your curl test should now redirect to the frontpage instead of crashing at line 35. The key difference is that my previous code tried to render the form with a nil public_body (which still crashed in the view), whereas now it redirects before reaching the view.

Please re-test with the updated code.

PR #9057, opened 2026-01-06
Extract `IncomingMessage#parse_raw_email`

garethrees:
Remove force option from IncomingMessage#parse_raw_email! and add the safe IncomingMessage#parse_raw_email counterpart so that we have a more conventional bang/non-bang pair of methods.

Currently retains the exception raising if there's no raw email, but this doesn't seem necessary for the non-bang version as it only matters when we actually attempt to parse the raw email.

[skip changelog]

### DIFF
--- a/app/controllers/attachments_controller.rb
+++ b/app/controllers/attachments_controller.rb
@@ -71,7 +71,7 @@ class AttachmentsController < ApplicationController
 
   def find_attachment
     @attachment = (
-      @incoming_message.parse_raw_email!
+      @incoming_message.parse_raw_email
 
       IncomingMessage.get_attachment_by_url_part_number_and_filename!(
         @incoming_message.get_attachments_for_display,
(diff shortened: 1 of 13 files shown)


gbp (approved):
Looked at this one last week. Seems okay to me although I've not doubled checked you've caught all invocations of the method... presumably as the specs pass it is okay tho.

garethrees:
> caught all invocations of the method

Yep pretty confident:

```
# Check for any calls like parse_raw_email!(true)
$ rg 'parse_raw_email!\('
# No output

# Check for any calls like parse_raw_email! true
$ rg 'parse_raw_email!\s'
app/models/incoming_message.rb
114:    parse_raw_email! if last_parsed.nil?
# No further output

# Also check the theme for any calls
gareth: ~/src/mysociety/alaveteli-themes/whatdotheyknow-theme (master $=)
$ rg parse_raw_email
# No output
```

PR #9034, opened 2025-12-12
RawEmail related cleanup

garethrees:
Part of #8803.

[skip changelog]

PR #9060, opened 2026-01-08
Sample data fixes

garethrees:
* Fix missing raw email file attachments when using sample data.

Extracted from https://github.com/mysociety/alaveteli/pull/9034. In service of https://github.com/mysociety/alaveteli/issues/8803.

[skip changelog]

### DIFF
--- a/script/load-sample-data
+++ b/script/load-sample-data
@@ -52,6 +52,19 @@ end
 
 # run everything as a single transaction - all the deletes then all the inserts
 ActiveRecord::FixtureSet.create_fixtures(fixtures_dir, fixture_files)
+
+# Copy the RawEmail fixture data into the development environment
+raw_emails_yml = Rails.root.join('spec', 'fixtures', 'raw_emails.yml')
+
+yamls = YAML.load_file(raw_emails_yml, permitted_classes: [Time])
+yamls.map { |_,v| v['id'] }.each do |raw_email_id|
+  raw_email = RawEmail.find(raw_email_id)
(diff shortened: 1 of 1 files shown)


garethrees:
Hm, yes, the above does illustrate the underlying reason that I made the PR – that the blobs are stored in `test_fixtures` service, rather than the `raw_emails`  service that new incoming mail gets stored under to (e.g. if you pipe a mail in via `script/mailin`).

```
$ dc run --rm app bin/rails runner "puts ActiveStorage::Blob.pluck(:service_name)"
test_fixtures
test_fixtures
test_fixtures
test_fixtures
test_fixtures # <<--- The fixture data blobs

$ cat spec/fixtures/files/incoming-request-pdf-attachment.eml |   sed 's/^From: EMAIL_FROM$/From: Olive Officer <officer@example.org>/' |   sed 's/^To: FOI Person <EMAIL_TO>$/To: Robin Houston <foi+request-109-89900s32@alaveteli.10.10.10.30.xip.io>/' |   docker-compose run --rm -T app script/mailin

$ dc run --rm app bin/rails runner "puts ActiveStorage::Blob.pluck(:service_name)"
test_fixtures
test_fixtures
test_fixtures
test_fixtures
test_fixtures # <<--- The fixture data blobs
raw_emails    # <<--- The newly ingested response from above blob
```

The issue was essentially that I was getting some sort of record not found error due to the app trying to find the relevant blob ID with the `service_name` of `'raw_emails'` , but that blob actually had the `service_name` value of `'test_fixtures'`.

I can't reproduce this at the moment, so I'll separate the two commits to individual PRs and close this one until I encounter the error again.

PR #9062, opened 2026-01-08
Redelivery cleanup

garethrees:
Clean up redelivery model and controller code.

In service of https://github.com/mysociety/alaveteli/issues/8803 to help simplify implementation of https://github.com/mysociety/alaveteli/pull/9040.

[skip changelog]


gbp (changes requested):
Looks good. Can we get rid of `Admin::IncomingMessages#redeliver` now?

PR #9072, opened 2026-01-14
[#8807] Add ability for attachments to be erased

### LINKED ISSUE #8807: Ability to destroy individual attachments while leaving a record that it once existed
In some cases it's too difficult / undesirable to create a redacted copy of an attachment and make a replacement – we just want to permanently remove it.

At the moment we can't do this without collateral damage – i.e. we have to destroy the entire `IncomingMessage` and other potentially useful attachments.

As an admin, we want to be able to go to the admin page for an attachment, click destroy / erase, and have only that attachment be removed. Sometimes we may want to leave an explanation about why:

Image

In the case of removing attachments, we'll need to lock all other attachments and ultimately destroy the raw email data (https://github.com/mysociety/alaveteli/issues/8803).

We should consider this in light of https://github.com/mysociety/alaveteli/issues/8805 and https://github.com/mysociety/alaveteli/issues/8806 to make sure we're consistent with UI, behaviour and code where it makes sense to be.

There may be some complexity here given the way `FoiAttachment` records are created.

gbp:
## Relevant issue(s)

Fixes #8807

## What does this do?

Add ability for attachments to be erased

## Implementation notes

Attempted to use the same pattern as #9040

### DIFF
--- a/doc/CHANGES.md
+++ b/doc/CHANGES.md
@@ -2,6 +2,7 @@
 
 ## Highlighted Features
 
+* Add ability to erase attachments (Graeme Porteous)
 * Added Web Share API option on request pages (Lucas Cumsille Montesinos)
 * Add prioritisation indication of admin summary todos (Gareth Rees)
 * Update Exim message ID matching (Graeme Porteous)


app/models/foi_attachment/lockable.rb
```diff
@@ -59,15 +59,15 @@ def lockable?
   end
 
   def unlockable?
-    !incoming_message.raw_email_erased?
+    !erased? && !incoming_message.raw_email_erased?
```
garethrees: Won't `FoiAttachment#erased?` imply that the underlying `RawEmail` is erased?

Think we could move this conditional to `FoiAttachment#erased?` if we think its generally necessary…

```rb
class FoiAttachment
  def erased?
    erased_at.present? && !incoming_message.raw_email_erased?
  end
```

…or just trust that `FoiAttachment#erased_at` being present is enough to know that the underlying `RawEmail` is gone so we can't do anything that would attempt to reparse.

app/views/admin/foi_attachments/edit.html.erb
```diff
@@ -45,6 +45,16 @@
...
+<% if @foi_attachment.erased? %>
+  <div class="row">
+    <div class="span12">
+      <div class="alert alert-warning">
+        <p>Cannot show the mask, lock or replacement options as this attachment has been erased.</p>
+      </div>
+    </div>
+  </div>
+<% else %>
+  <% # TODO: fix indent %>
```
garethrees: Probably worth extracting out these actions to partials so the conditional is nicer:

```erb
<!-- psuedocode -->
else
  render partial: 'mask'
  render partial: 'lock'
  render partial: 'replace'
  render partial: 'erase'
```

app/models/foi_attachment.rb
```diff
@@ -58,12 +59,14 @@ class FoiAttachment < ApplicationRecord
...
   scope :binary, -> { where.not(content_type: AlaveteliTextMasker::TextMask) }
+  scope :erased, -> { where.not(erased_at: nil) }
 
   delegate :expire, to: :info_request
   delegate :metadata, to: :file_blob, allow_nil: true
 
   admin_columns exclude: %i[url_part_number within_rfc822_subject hexdigest],
-                include: %i[redacted_filename display_filename metadata]
+                include: %i[redacted_filename display_filename metadata
+                            erased_at]
```
garethrees: `erased_at` is included automatically as its a database column – the explicit include is only necessary for non-attribute methods. This duplicates the value:

Image

app/models/foi_attachment.rb
```diff
@@ -272,6 +287,34 @@ def load_attachment_from_incoming_message
     )
   end
 
+  def erased?
+    erased_at.present?
+  end
+
+  def erase(editor:, reason:)
```
garethrees, line 294: This doesn't erase the underlying `RawEmail`. We'll need to do this as the intention is that when an `FoiAttachment` is erased, we no longer hold a copy of the data it contains, as noted in the [issue description](https://github.com/mysociety/alaveteli/issues/8807#issue-3324679816). At present that won't be true since we'd still hold it in the `RawEmail`.

I don't think there's a case where we'd want to erase an attachment but continue to hold the contents – we'd use hiding via the prominence options for that.

app/models/foi_attachment/lockable.rb
garethrees: `FoiAttachment#erased?` should imply that the underlying `RawEmail` is erased (see comment on `FoiAttachment#erase`), so we could then make this an `||` for a little performance gain.

app/models/foi_attachment.rb
```diff
@@ -272,6 +287,34 @@ def load_attachment_from_incoming_message
...
+
+    transaction do |t|
+      t.after_rollback { return false }
+
+      raise ActiveRecord::Rollback unless
+        log_event(
+          'erase_attachment',
+          editor: editor,
+          reason: reason,
+          raw_email: self,
```
garethrees: Mismatch of model/record: `raw_email: {gid: "gid://alaveteli/FoiAttachment/286"}`

app/views/admin/foi_attachments/edit.html.erb
garethrees: Probably worth extracting out these actions to partials at this point so the conditional is nicer:

```erb
<!-- psuedocode -->
else
  render partial: 'mask'
  render partial: 'lock'
  render partial: 'replace'
  render partial: 'erase'
```

app/views/admin/foi_attachments/edit.html.erb
```diff
@@ -148,6 +158,34 @@
...
 <% end %>
 
+<div class="row">
+  <div class="span12 danger-zone">
+    <h1 class="span12 danger-zone__header">Danger Zone</h1>
+
+    <div class="danger-zone__item">
+      <div class="span7">
+        <p>
+          <strong>Erase this Attachment</strong>
```
garethrees: Need some documentation as to what this does (and the implications, e.g. it will erase the underlying raw email which will lock all other attachments)

app/views/request/_attachments.html.erb
```diff
@@ -44,6 +44,19 @@
               <% end %>
             </p>
 
+          <% elsif a.erased? %>
+            <p class="attachment__name">
+              <%= h a.display_filename %>
```
garethrees, line 49: Erasure should purge the filename too, since this can often contain PII (e.g. `FOI-1234 Bob Smith.pdf`).

app/views/request/_attachments.html.erb
```diff
@@ -44,6 +44,19 @@
...
+            </p>
+
+            <p class="attachment__meta">
+              <em>This file has been erased.</em>
```
garethrees: gettext wrapper needed

app/views/request/erased_attachment.html.erb
```diff
@@ -0,0 +1,11 @@
+<% @title = _('Attachment has been erased') %>
+
+<h1><%= @title %></h1>
+
+<p>
+  This attachment has been removed and is no longer available.
```
garethrees: gettext wrapper needed

spec/models/foi_attachment_spec.rb
```diff
@@ -66,6 +68,18 @@
     it { is_expected.to include(unlocked_attachment) }
   end
 
+  describe '.erased' do
+    subject { described_class.erased }
+
+    let!(:erased_attachment) do
+      FactoryBot.create(:body_text, erased_at: Time.zone.now)
```
garethrees: Probably worth adding an `:erased` trait for all these.

app/views/admin/foi_attachments/edit.html.erb
garethrees: Currently erasing a main body part raises an `FoiAttachment::MissingAttachment` as it's not caught by this conditional. Should be fixed by resolving the comment on `FoiAttachment#erase`, but wanted to note to make sure we check this case once that's been resolved.

app/models/foi_attachment.rb
gbp: Oh right, thats not how I understood the issue.

Let's get this merged and then add that as a separate issue and PR.

app/views/admin/foi_attachments/edit.html.erb
gbp: Fine as it is for now

garethrees (commented):
Had a quick scan through the fixups and they look fine

RE: https://github.com/mysociety/alaveteli/pull/9072#discussion_r2817962692

Happy to merge this as an initial part of the issue (once failing specs resolved), but lets feature flag the UI / controller until it works as intended.

PR #9083, opened 2026-01-21
Separate attachment updates

garethrees:

Handling multiple behaviours (prominence, locking, replacing) in a single update method (`FoiAttachment#udpdate`) / controller action (`Admin::FoiAttachments#update`) is becoming too unwieldy.

Updating prominence, locking and replacements now all have their own form. I've also added the ability to force masking as we'll later want to prevent locking until masking has been run.

Screenshot 2026-01-21 at 16 04 19

[skip changelog]

### DIFF
--- a/app/controllers/admin/foi_attachments/masks_controller.rb
+++ b/app/controllers/admin/foi_attachments/masks_controller.rb
@@ -1,0 +1,32 @@
+# Allow admins to force the masking of FoiAttachment records.
+class Admin::FoiAttachments::MasksController < AdminController
+  before_action :set_foi_attachment, :set_incoming_message, :set_info_request
+  before_action :check_info_request
+
+  def create
+    @foi_attachment.mask_later
+    redirect_to edit_admin_foi_attachment_path(@foi_attachment),
+                notice: 'Masking queued.'
+  end
+
+  private
+
+  def set_foi_attachment
+    @foi_attachment = FoiAttachment.find(params[:foi_attachment_id])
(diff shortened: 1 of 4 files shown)


app/controllers/admin/foi_attachments/locks_controller.rb
```diff
@@ -0,0 +1,63 @@
...
+          additional censor rules.
+        TXT
+      else
+        flash[:notice] = 'Attachment locked.'
+      end
+
+      redirect_to edit_admin_foi_attachment_path(@foi_attachment)
+    else
+      flash.now[:error] = @foi_attachment.errors.full_messages.to_sentence
+      render 'admin/foi_attachments/edit'
```
gbp, line 25: Would prefer this as rendering the template results in the UI having the wrong state, ie we can see a "Unlock" button when locking failed. Same for the `#destroy` method, we see a "Lock" button when unlocking fails. I guess the same issue in the replacements controller too?

```suggestion
      flash[:error] = @foi_attachment.errors.full_messages.to_sentence
      redirect_to [:edit, :admin, @foi_attachment]
```

garethrees: Ah yep this definitely makes more sense now that we're not re-rendering a user input form

garethrees: https://github.com/mysociety/alaveteli/pull/9095

PR #9086, opened 2026-01-22
[#9085] Update locale cookie handling

### LINKED ISSUE #9085: Locale switcher not working as expected
## Problem

The Swedish site has available locales set to `en` and `sv`, with `sv` as the default locale.

After changes in version 0.46, users cannot view the `sv` locale when their browser doesn't prefer `sv` over `en` (via the `HTTP_ACCEPT_LANGUAGE` header). The site is ignoring both the user's manual locale selection and the configured default locale.

This issue doesn't occur on WDTK when switching to Welsh because WDTK's default locale is `en`, which matches most browser requests.

## Root Cause

In 0.46, we made two changes that interact to cause this issue:

1. **Improved `HTTP_ACCEPT_LANGUAGE` parsing**: We now correctly match browser requests like `en-GB` to the `en` locale
2. **Increased browser locale precedence**: The browser's locale preference now takes precedence over the site's configured default locale

As a result, when a browser requests `en` (or `en-*`), this overrides the site's `sv` default.

Additionally, we currently remove the locale cookie when it's set to the default locale. This was originally implemented for caching reasons in WDTK's Varnish configuration. However, Varnish has since been improved to ignore the locale cookie entirely, making this removal unnecessary.

## Solution

- [ ] Allow the locale cookie to be set even when the user selects the default locale

gbp:
## Relevant issue(s)

Fixes #9085

## What does this do?

Update locale cookie handling

Always persist the locale cookie as this is the only way to configure the users choice when the default locale doesn't match what their browser requests via the `HTTP_ACCEPT_LANGUAGE` header.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -91,13 +91,8 @@ class ApplicationController < ActionController::Base
 
   def store_gettext_locale
     # set the current stored locale to the requested_locale
-    locale = params[:locale].presence || AlaveteliLocalization.locale
-
-    if locale == AlaveteliLocalization.default_locale
-      cookies.delete(:locale)
-    else
-      cookies[:locale] = locale
-    end
+    locale = AlaveteliLocalization.locale
+    cookies[:locale] = locale
 
     # ensure current user locale attribute is up-to-date
     current_user.update_column(:locale, locale) if current_user
--- a/doc/CHANGES.md
+++ b/doc/CHANGES.md
@@ -2,6 +2,7 @@
 
 ## Highlighted Features
 
+* Update locale cookie handling (Graeme Porteous)
 * Restore logging of :email parameters (Gareth Rees)
 * Fix importing holidays from iCal feed (Gareth Rees)
 * Ability to search users by tag in admin interface (Gareth Rees)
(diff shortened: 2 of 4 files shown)


gbp:
@sagepe could you double check this won't impact our caching done in Varnish?

app/controllers/application_controller.rb
```diff
@@ -91,13 +91,8 @@ def set_gettext_locale
...
     # set the current stored locale to the requested_locale
-    locale = params[:locale].presence || AlaveteliLocalization.locale
-
-    if locale == AlaveteliLocalization.default_locale
-      cookies.delete(:locale)
-    else
-      cookies[:locale] = locale
-    end
+    locale = AlaveteliLocalization.locale
+    cookies[:locale] = locale
```
garethrees, line 95: OOI could we put this in the session rather than a separate cookie and have the same effect?

gbp: No! We had to move it out of the session to allow us to serve cached versions of both `en` and `cy` locales.

sagepe: And this allows us to manage locale in a cacheable way for non-logged in users.

garethrees: 👍

PR #9095, opened 2026-01-27
Redirect on attachment admin action failure

garethrees:
In all of these cases we're not handling form errors in the usual way, so rendering the edit template doesn't really work as expected and in the case of locking, can show an incorrect lock status.

Instead just redirect on both success and failure with the general error message.

Also adds a missing test for displaying the error flash on replacement failure.

See https://github.com/mysociety/alaveteli/pull/9083#pullrequestreview-3710281012

[skip changelog]

### DIFF
--- a/app/controllers/admin/foi_attachments/locks_controller.rb
+++ b/app/controllers/admin/foi_attachments/locks_controller.rb
@@ -10,20 +10,19 @@ class Admin::FoiAttachments::LocksController < AdminController
       )
       @foi_attachment.expire
 
-      if @foi_attachment.locked? && !@foi_attachment.masked?
-        flash[:notice] = <<~TXT.squish
+      flash[:notice] = if @foi_attachment.locked? && !@foi_attachment.masked?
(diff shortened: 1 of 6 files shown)


app/controllers/admin/foi_attachments/locks_controller.rb
```diff
@@ -10,20 +10,19 @@ def create
...
           additional censor rules.
         TXT
       else
-        flash[:notice] = 'Attachment locked.'
+        'Attachment locked.'
       end
-
-      redirect_to edit_admin_foi_attachment_path(@foi_attachment)
     else
       flash.now[:error] = @foi_attachment.errors.full_messages.to_sentence
```
gbp:
```suggestion
      flash[:error] = @foi_attachment.errors.full_messages.to_sentence
```

garethrees: oops missed that

gbp (approved):

PR #9102, opened 2026-01-28
Add ActionMailbox

### LINKED ISSUE #8801: Integrate ActionMailbox
The overall goal here is to do away with as much of our custom mail ingestion and move to relying on Rails to do much of the hard work for us. [Configuring the MTA](https://alaveteli.org/docs/installing/email/) is pretty complicated, so the more we can ship in the app the better.

Long after Alaveteli was created, [Action Mailbox](https://guides.rubyonrails.org/action_mailbox_basics.html) was introduced. This will allow an easier MTA configuration experience, allow us to handle the app-specific mail routing in the app, shift some of the work to background processing, and cut custom code out of the workflow.

Here's an outline of some of the key and stretch things I think we can do. We'll need to refine this as we get stuck in of course:

Key:

* Integrate Action Mailbox for the main response ingestion. We've [spiked this](https://github.com/mysociety/alaveteli/pull/8049) in the past so we know it should work in principle.
  * Ideally we'd get `RequestMailer` out of the loop entirely here.
* Once the email has been ingested, kick off parsing in the background so that we're not parsing responses in the HTTP request cycle.
* It would be good to clean up the customisations in `MailHandler` – we don't need to worry about old TMail behaviour at this point.

Stretch:

* Move some of the [other actions](https://alaveteli.org/docs/installing/email/#set-up-your-contact-email-recipient-groups) we configure in the MTA to ActionMailbox routing. The ideal is as close to zero-config in the MTA as we can get.
  * If its possible it would be great to be able to configure some of this from the app's admin interface (e.g. we often get requests to update the forwarding addresses for `team@some-site`), though this is a real nice to have.
* Use this as an opportunity to review and clean up some of the mail-related parsing in [`MailHandler`](https://github.com/mysociety/alaveteli/blob/develop/lib/mail_handler.rb), [`mail_extensions`](https://github.com/mysociety/alaveteli/blob/develop/lib/mail_handler/backends/mail_extensions.rb) and [`MailBackend`](https://github.com/mysociety/alaveteli/blob/508317c712b67b1fc9e23d3ca2ee943a02529adb/lib/mail_handler/backends/mail_backend.rb) – it's hard to be specific here but lots of this looks like it ought to be handled in the underlying [Mail library](https://github.com/mikel/mail).
  * If there are opportunities to contribute some of our customisations back to Mail then we should take them.

gbp:
## Relevant issue(s)

Fixes #8801

## What does this do?

Adds the ability to receive email via ActionMailbox.

## Why was this needed?

Removing custom mail handling code and switching to the Rails-way

### DIFF
--- a/app/models/raw_email.rb
+++ b/app/models/raw_email.rb
@@ -38,6 +38,13 @@ class RawEmail < ApplicationRecord
   delegate :lock_all_attachments, to: :incoming_message
   delegate :all_attachments_masked?, to: :incoming_message
 
+  def inbound_email
+    ActionMailbox::InboundEmail.find_by(
+      message_id: message_id,
+      message_checksum: message_checksum
+    ) if self[:message_id] && self[:message_checksum]
+  end
+
   def addresses(include_invalid: false)
     MailHandler.get_all_addresses(mail, include_invalid: include_invalid)
   end
@@ -123,6 +130,7 @@ class RawEmail < ApplicationRecord
         )
 
       file.purge_later
+      inbound_email&.destroy
       touch(:erased_at)
 
       expire(preserve_database_cache: true)
--- a/spec/models/raw_email_spec.rb
+++ b/spec/models/raw_email_spec.rb
@@ -252,6 +252,40 @@ RSpec.describe RawEmail do
     end
   end
 
+  describe '#inbound_email' do
+    let(:raw_email) { FactoryBot.create(:incoming_message).raw_email }
+
+    context 'when message_id and message_checksum are present' do
+      let!(:inbound_email) do
+        ActionMailbox::InboundEmail.create!(
+          message_id: raw_email.message_id,
+          message_checksum: raw_email.message_checksum,
+          status: :delivered
+        )
+      end
+
+      it 'returns the matching InboundEmail' do
+        expect(raw_email.inbound_email).to eq(inbound_email)
+      end
+    end
+
+    context 'when message_id or message_checksum are not set' do
+      before do
+        raw_email.update_columns(message_id: nil, message_checksum: nil)
+      end
+
+      it 'returns nil' do
+        expect(raw_email.inbound_email).to be_nil
+      end
+    end
+
+    context 'when no matching InboundEmail exists' do
+      it 'returns nil' do
+        expect(raw_email.inbound_email).to be_nil
+      end
+    end
+  end
+
   describe '#erasable?' do
     subject { raw_email.erasable? }
 
@@ -348,6 +382,17 @@ RSpec.describe RawEmail do
     it 'expires the associated info_request' do
       expect(raw_email.info_request).to receive(:expire)
       subject
+    end
+
+    it 'destroys the associated inbound_email' do
+      inbound_email = ActionMailbox::InboundEmail.create!(
+        message_id: raw_email.message_id,
+        message_checksum: raw_email.message_checksum,
+        status: :delivered
+      )
+      expect { subject }.to change {
+        ActionMailbox::InboundEmail.exists?(inbound_email.id)
+      }.from(true).to(false)
     end
 
     it { is_expected.to eq(true) }


app/mailboxes/request_mailbox.rb
```diff
@@ -0,0 +1,6 @@
+class RequestMailbox < ApplicationMailbox
+  def process
+    mail = MailHandler.mail_from_raw_email(inbound_email.source)
```
garethrees: This is now `MailHandler.mail_from_string` as of https://github.com/mysociety/alaveteli/commit/019a088de8ac0c85310caca85f138a68c7009ea1

doc/CHANGES.md
```diff
@@ -20,8 +21,20 @@
...
-  1. Postfix/Exim `./script/mailin` pipe or 2. POP poller.
-  While you can have multiple sources configured we recommend using one only.
+  1. Postfix/Exim `./script/mailin` pipe, 2. POP poller or 3. ActionMailbox
+  While you can have multiple sources configured we recommend migrating to
+  ActionMailbox as the others are depreicated and will be removed in a future
+  release.
+  See: https://github.com/mysociety/alaveteli/wiki/ActionMailbox-Migration-Guide
+
+* _Required:_ After ActionMailbox migration please backup `config/master.key`.
+  This file gets generated automaticaally and is the private encryption key for
```
garethrees: typo: `/automaticaally/automatically/`

doc/CHANGES.md
```diff
@@ -20,8 +21,20 @@
...
+  the credentials. Without it the application can't read credentials. If you
+  ever move servers or reinstall Alaveteli then you might need this file.
```
garethrees: `/might/will/`?

Might be worth linking to the upstream rails guides here (https://guides.rubyonrails.org/security.html#custom-credentials I think is the main guide but not super familiar with this)

doc/CHANGES.md
```diff
@@ -20,8 +21,20 @@
...
+
+* _Required:_ Please update your `config/storage.yml` file to include a
+  production configuration for `inbound_emails`. See
+  `config/storage.yml-example` as an example.
```
garethrees, line 38: Would it make sense to set filesystem-based as the default? I expect that will be what most start with.

spec/support/email_helpers.rb
```diff
@@ -9,7 +9,7 @@ def receive_incoming_mail(filename_or_string, **kwargs)
   )&.route
 end
 
-def get_fixture_mail(filename_or_string, email_to = nil, email_from = nil)
+def get_fixture_mail(filename_or_string, email_to: nil, email_from: nil)
```
garethrees: Could make these just `to: nil, from: nil` as its fairly obvious these are about the mail headers?

spec/support/email_helpers.rb
```diff
@@ -33,14 +33,9 @@ def load_mail_server_logs(log)
...
 def gsub_addresses(content, **kwargs)
-  content.gsub!('EMAIL_TO', kwargs[:email_to]) if kwargs[:email_to]
-  content.gsub!('EMAIL_FROM', kwargs[:email_from]) if kwargs[:email_from]
-  content.gsub!('EMAIL_CC', kwargs[:email_cc]) if kwargs[:email_cc]
-  content.gsub!('EMAIL_BCC', kwargs[:email_bcc]) if kwargs[:email_bcc]
-  if kwargs[:email_envelope_to]
-    content.gsub!('EMAIL_ENVELOPE_TO', kwargs[:email_envelope_to])
-  end
-  content
+  kwargs.slice(*%i[email_to email_from email_cc email_bcc email_envelope_to]).
```
garethrees: Similarly could remove the `email_` prefix from the kwargs to make method calls shorter.

garethrees (commented):
Nice, so much cleanup in all of this 🙌

doc/CHANGES.md
gbp: The default in `config/storage.yml-example` is to store emails on the filesystem.

gbp:
This was merged in 7de1ef9e301b44c980241188474499906b713855

Issue #8928, opened 2025-10-16
Network Impact Showcase


garethrees:
## Context

When you land on an Alaveteli site, it’s not obvious at all that it’s part of a wider network of Alavetelis and other ATI platforms.

This hasn’t been a problem historically given most of the focus was on citizen use, but as we push for broader high impact users (journalists, campaigners and researchers), it can be useful for them to know that there are other request platforms, and organisations behind the platforms that can help them navigate specific national contexts.

We’ve seen enough seedling examples of high impact users making use of Alavetelis outside of their country, and have also tracked some cross-border examples, to know there’s an improvement we could make here.

## Goals

A page built in to Alaveteli that demonstrates:

* the breadth of the global ATI network  
* exemplar cross-border use  
* an overwhelming sense of credibility from impactful national use  
* that there’s a committed network advocating for ATI rights

This page will primarily be targeted at:

* High Impact users (journalists, campaigners and researchers), who may make use of several national platforms  
* Funders, to help national sites benefit from the wider network impact in their own funding pitches

## The Network Showcase

Here’s a starting point of an idea for the kind of page I’m thinking we’ll create. It doesn’t have to be exactly like this, so very happy to riff on the idea.

We’ll start with something like the map from the [alaveteli.org deployments page](http://alaveteli.org), and a condensed list of ATI sites (e.g. 🇬🇧[WhatDoTheyKnow,](https://www.whatdotheyknow.com) 🇪🇺[AskTheEU](https://www.asktheeu.org/), …). We’ll emphasise Alavetelis, but also acknowledge the other platforms (“And our friends at 🇩🇪[FragDenStaat](https://fragdenstaat.de/), 🇺🇸[Muckrock](https://www.muckrock.com/), …). A static map is fine for now – we just want to be clear that “this is a global network”; we don’t need a precise location map, since we’ll have that via the list of sites with emoji flags anyway.

We don’t have a huge set of cross border examples, so we’ll give some high prominence to some exemplars. This is mainly to sow the seeds that ATI, and Alavetelis, can be used for cross-border work.

[...]

PR #9117, opened 2026-02-09
Added fix to console error for masonry JS component

lucascumsille:
## Relevant issue(s)

Currently every page with no masonry grid is triggering this console error:

`TypeError: null is not an object (evaluating 'grid.getAttribute')`

https://github.com/mysociety/alaveteli/issues/8928#issuecomment-3870694394

## What does this do?

Moved conditional that checks `.js-gallery-grid` and `#load-more` elements exist right after both constant are declared. This prevents the consolore log error `Cannot read properties of null (reading 'getAttribute')` in every page that don't have those constant and their children.

<hr>

[skip changelog]

### DIFF
--- a/app/assets/javascripts/ati-masonry.js
+++ b/app/assets/javascripts/ati-masonry.js
@@ -1,10 +1,13 @@
 document.addEventListener('DOMContentLoaded', function() {
 
+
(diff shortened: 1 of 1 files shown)


gbp (approved):
Thanks for fixing!

Issue #8808, opened 2025-08-15
Ability to fully anonymise a user account


garethrees:
## The Problem

When a user has made requests, closing their account creates user-level censor rules to redact their name from the associated requests.

While the link to the user’s email address gets removed, the account is not really anonymised. Admins can still view the user’s name alongside the content they’ve created.

The account closure process does not destroy material from requests. This means that even if we took an alternative approach to the user level censor rule creation (creating a censor rule on each of the user’s requests, for example), it would still be possible to visit the account page, view a request and then reveal the original name on the account.

There’s also the possibility that non-name PII included across requests could lead to account re-identification if still linked.

The Close & Anonymise feature is still behind a feature flag.

_Aside: In some cases we record internal GDPR reference numbers as part of the censor rule audit trail. This ties the record back to the email address, though I think that should be fine since the GDPR tracking can have its own retention policy._

garethrees:
## Destructive censor rules

If censor rules were made destructive, then we’d no longer need to hold them once applied for account anonymisation. In theory this would leave us with an anonymised account, still with the requests linked to it, but with no instances of the user’s name remaining in the request contents.

Censor rules are best-effort. Even if the automatically created rules are improved to account for common name variations, there’s a non-zero chance that not all instances of a user’s name will be removed, making it possible to re-identify an anonymised account that’s still linked to a request.

While low probability, this approach doesn’t solve the issue of non-name PII included across requests enabling mosaic identification of the closed account.

There’s a risk that automated destructive edits would incorrectly remove content.

This is particularly undesirable for substantive response material. The current implementation would mean “locking” response attachments, meaning further censor rules would not be applied without additional manual intervention. To make the edits destructive, the underlying raw email of responses would need to be destroyed, which increases the risk that substantive FOI releases are irreversibly destroyed where we have a legitimate interest to preserve them (and just redact to minimise data use).

## Disassociate requests from user accounts

If we assume that removing PII from request contents will never be perfect, the only way to truly anonymise a *user account* is to disassociate the account from the requests.

When an account is anonymised, instead of a user-level censor rule, the rule could be created individually per request. This would leave the account empty of requests and censor rules, meaning all PII could be fully erased from the account record.

Individual requests would be attributable to a named individual through the request level censor rule, but it would be more difficult to link multiple requests to the same author. One way it would be possible to infer this is by viewing censor rule records and matching the name and creation timestamps.

[...]

PR #9110, opened 2026-02-03
WIP extract handle_locking

garethrees:
## Relevant issue(s)

Extracted from https://github.com/mysociety/alaveteli/pull/9100 in continuation of https://github.com/mysociety/alaveteli/pull/9101 in service of https://github.com/mysociety/alaveteli/issues/8808.

## Notes to reviewer

<hr>

Have you updated the changelog? If this is not necessary, put square brackets around this: [skip changelog]

PR #9124, opened 2026-02-16
Refactor replacement

garethrees:
Refactors attachment replacement to use an action-specific method rather than a generic call to update. This helps clean up the controllers and makes the methods more composable with other actions (like masking, expiry, etc).

Similar to https://github.com/mysociety/alaveteli/pull/9101, in service of https://github.com/mysociety/alaveteli/issues/8808.

[skip changelog]

### DIFF
--- a/app/controllers/admin/foi_attachments/replacements_controller.rb
+++ b/app/controllers/admin/foi_attachments/replacements_controller.rb
@@ -2,6 +2,7 @@
 class Admin::FoiAttachments::ReplacementsController < AdminController
   before_action :set_foi_attachment, :set_incoming_message, :set_info_request
   before_action :check_info_request
+  before_action :check_clearable, only: :destroy
 
   def create
     @foi_attachment.replace(
@@ -62,5 +63,10 @@ class Admin::FoiAttachments::ReplacementsController < AdminController
 
     raise ActiveRecord::RecordNotFound
   end
-end
 
+  def check_clearable
+    return if @foi_attachment.replacement_clearable?
+
+    raise StandardError, 'Attachment not clearable'
+  end
+end
--- a/spec/controllers/admin/foi_attachments/replacements_controller_spec.rb
+++ b/spec/controllers/admin/foi_attachments/replacements_controller_spec.rb
@@ -319,6 +319,19 @@ RSpec.describe Admin::FoiAttachments::ReplacementsController do
       end
     end
 
+    context 'if the attachment is not clearable' do
+      before do
+        allow_any_instance_of(FoiAttachment).
+          to receive(:replacement_clearable?).and_return(false)
+      end
+
+      it 'raises StandardError' do
+        expect {
+          delete :destroy, params: params
+        }.to raise_error(StandardError)
+      end
+    end
+
     context 'if the request is embargoed', feature: :alaveteli_pro do
       before { info_request.create_embargo }
 


app/models/foi_attachment/replaceable.rb
```diff
@@ -14,6 +14,20 @@ module FoiAttachment::Replaceable
     before_save :handle_replacements
   end
 
+  def replace!(editor:, reason:, replacement_body: nil, replacement_file: nil, replaced_filename: nil, **event)
+    attrs = { replacement_body: replacement_body,
+              replacement_file: replacement_file,
+              replaced_filename: replaced_filename,
+              replaced_reason: reason }.compact
```
garethrees, line 37: Should explain this in the commit message.

We can either receive a replacement_body or replacement_file, and the replaced_filename is optional. We build the hash of all replacement options then compact them to remove any with a nil value.

spec/models/foi_attachment_spec.rb
```diff
@@ -1216,6 +1216,137 @@ def last_event
...
+        foi_attachment.replace!(
+          editor: editor,
+          reason: reason,
+          replacement_file: fixture_file_upload('parrot.png', 'image/png'),
+          replaced_filename: 'redacted.png'
+        )
+      end
+
+      let(:foi_attachment) { FactoryBot.create(:body_text) }
+
```
garethrees, line 1412: missed spec to check the actual content

app/views/admin/foi_attachments/edit.html.erb
```diff
@@ -135,6 +135,13 @@
   <div class="form-actions">
     <div class="input-prepend">
       <%= f.submit 'Upload replacement', class: 'btn btn-warning' %>
+      <button type="submit"
+              class="btn btn-danger"
+              name="_method"
+              value="delete"
+              <%= 'disabled' unless @foi_attachment.replaced? %>>
+        Clear replacement
```
garethrees: Can't clear a replaced body (`FoiAttachment::MissingAttachment`)

app/controllers/admin/foi_attachments/replacements_controller.rb
```diff
@@ -4,9 +4,10 @@ class Admin::FoiAttachments::ReplacementsController < AdminController
   before_action :check_info_request
 
   def create
-    if @foi_attachment.update_and_log_event(
```
garethrees, line 7: in commit message:

call out that reason is now a separate param (as its for the event not the foi attachment)

app/models/foi_attachment/replaceable.rb
```diff
@@ -14,6 +14,14 @@ module FoiAttachment::Replaceable
     before_save :handle_replacements
   end
 
+  # Note that #replace still raises on failure. This version also runs
```
garethrees, line 17: Add commit explanation:


```
Adds a safer version of replace! that handles required pre-and-post-replacment
actions.
``

File: app/models/foi_attachment/replaceable.rb [hunk 6]
```diff
@@ -36,6 +44,25 @@ def replace!(editor:, reason:, replacement_body: nil, replacement_file: nil, rep
     true
   end
 
+  def clear_replacement!(editor:, reason:, **event)
```
garethrees, line 47: Need to handle case of erased raw email somewhere (here or prevent in UI)

app/models/foi_attachment/replaceable.rb
```diff
@@ -36,6 +44,25 @@ def replace!(editor:, reason:, replacement_body: nil, replacement_file: nil, rep
...
+
+    update_and_log_event!(
+      event: { **event, editor: editor, reason: reason },
+      replaced_at: nil,
+      replaced_reason: nil,
+      masked_at: nil,
+      filename: mail_attributes[:filename]
+    )
+
+    mask_later
```
garethrees: move to `clear_replacement`

app/models/foi_attachment/replaceable.rb
```diff
@@ -22,6 +22,15 @@ def replace(...)
     expire
   end
 
+  # Note that #clear_replacement still raises on failure. This version also
+  # runs pre-and-post-clearing steps, whereas #clear_replacement! omits these
+  # for more flexibility when composing with other actions.
+  def clear_replacement(...)
+    clear_replacement!(...)
+    mask_later
```
garethrees: Check why we explicitly mask here but only expire on replace

garethrees: Ah actually we don't force masking on any replacing – only locking (which happens on replace, but via handle_locked) – can remove here and deal with rationalisation in https://github.com/mysociety/alaveteli/pull/9110

app/models/foi_attachment.rb
```diff
@@ -64,6 +64,7 @@ class FoiAttachment < ApplicationRecord
   delegate :expire, to: :info_request
   delegate :raw_email_erased?, to: :incoming_message
   delegate :metadata, to: :file_blob, allow_nil: true
+  delegate :raw_email_erased?, to: :incoming_message
```
gbp: Already exists!

```suggestion
```

app/models/foi_attachment/replaceable.rb
```diff
@@ -14,6 +14,58 @@ module FoiAttachment::Replaceable
...
+      replaced_reason: nil,
+      masked_at: nil,
+      filename: mail_attributes[:filename]
+    )
+
+    true
+  end
+
+  def replacement_clearable?
+    replaced? && !raw_email_erased?
```
gbp:
```suggestion
    replaced? && !erased?
```

app/helpers/admin/foi_attachments/replacements_helper.rb
```diff
@@ -0,0 +1,22 @@
+module Admin::FoiAttachments::ReplacementsHelper
+  def clear_replacement_button(foi_attachment)
+    title =
+      if !foi_attachment.replaced?
+        'No replacement to clear.'
+      elsif foi_attachment.raw_email_erased?
```
gbp:
```suggestion
      elsif foi_attachment.erased?
```

garethrees:
Ah good catch missed that addition in the rebase. Fixed up now so should all be good. Ignoring the remaining rubocops.

PR #9133, opened 2026-02-25
[#8808] Correctly namespace admin user controllers

garethrees:
## Relevant issue(s)

In service of #8808

## What does this do?

Improves namespacing of controllers around closing user accounts

## Why was this needed?

Working on user erasure in #8808 and having the controller correctly namespaced first makes for a nicer set of changes; did the other related ones at the same time since they were all pretty similar.

## Notes to reviewer

 [skip changelog]

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -788,41 +788,15 @@ Rails.application.routes.draw do
   ####
 
   #### Admin::Users::Suspensions controller
-  namespace :admin do
-    resources :users, only: [] do
-      scope module: :users do
-        resource :suspension, only: [:create]
-      end
-    end
-  end
-  ####
-
+  #### Admin::Users::Closures controller
+  #### Admin::Users::Anonymisations controller
   #### Admin::Users::Erasures controller
   namespace :admin do
-    resources :users, only: [] do
-      scope module: :users do
-        resource :erasure, only: [:create]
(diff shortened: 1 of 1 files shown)


config/routes.rb
```diff
@@ -787,35 +787,19 @@ def self.matches?(request)
...
+  #### Admin::Users::Anonymisations controller
+  #### Admin::Users::Erasures controller
+  namespace :admin do
+    resources :users, only: [] do
+      scope module: :users do
+        resource :suspension, only: [:create]
+        resource :closure, only: [:create]
+        resource :anonymisation, only: [:create]
+        resource :erasure, only: [:create]
+      end
```
gbp:
```suggestion
    resources :users, only: [], module: :users do
      resource :suspension, only: [:create]
      resource :closure, only: [:create]
      resource :anonymisation, only: [:create]
      resource :erasure, only: [:create]
```

or 

```suggestion
    resources :users, only: [] do
      resource :suspension, only: [:create], module: :users
      resource :closure, only: [:create], module: :users
      resource :anonymisation, only: [:create], module: :users
      resource :erasure, only: [:create], module: :users
```

garethrees: oh yeah the first option there is much nicer

PR #9138, opened 2026-02-27
[#8808] User erasure

### LINKED ISSUE #8808: Ability to fully anonymise a user account

garethrees:
Fixes https://github.com/mysociety/alaveteli/issues/8808

Supersedes https://github.com/mysociety/alaveteli/pull/9137

[skip changelog]

### DIFF
--- a/app/models/user/erasable.rb
+++ b/app/models/user/erasable.rb
@@ -1,13 +1,13 @@
 module User::Erasable
   extend ActiveSupport::Concern
 
-  def erase
-    erase!
+  def erase(...)
+    erase!(...)
   rescue ActiveRecord::RecordInvalid
     false
   end
 
-  def erase!
+  def erase!(editor:, reason:)
     raise ActiveRecord::RecordInvalid unless closed?
 
     transaction do
--- a/app/models/user/erasable.rb
+++ b/app/models/user/erasable.rb
@@ -1,0 +1,34 @@
+module User::Erasable
+  extend ActiveSupport::Concern
+
+  def erase
+    erase!
+  rescue ActiveRecord::RecordInvalid
+    false
+  end
+
+  def erase!
+    raise ActiveRecord::RecordInvalid unless closed?
+
+    sha = Digest::SHA1.hexdigest(rand.to_s)
+
+    transaction do
+      slugs.destroy_all
+      sign_ins.destroy_all
+      email_histories.destroy_all
+      profile_photo&.destroy!
+
+      outgoing_messages.update!(
+        from_name: _('[Name Removed]')
+      )
+
+      update!(
+        name: _('[Name Removed]'),
+        email: "#{sha}@invalid",
+        url_name: sha,
+        about_me: '',
+        password: MySociety::Util.generate_token
+      )
+    end
+  end
+end
--- a/app/models/user/erasable.rb
+++ b/app/models/user/erasable.rb
@@ -10,25 +10,35 @@ module User::Erasable
   def erase!
     raise ActiveRecord::RecordInvalid unless closed?
 
-    sha = Digest::SHA1.hexdigest(rand.to_s)
-
     transaction do
-      slugs.destroy_all
-      sign_ins.destroy_all
-      email_histories.destroy_all
-      profile_photo&.destroy!
+      destroy_identifying_associations
 
       outgoing_messages.update!(
         from_name: _('[Name Removed]')
       )
 
-      update!(
-        name: _('[Name Removed]'),
-        email: "#{sha}@invalid",
-        url_name: sha,
-        about_me: '',
-        password: MySociety::Util.generate_token
-      )
+      erase_account
     end
+  end
+
+  private
+
+  def destroy_identifying_associations
+    slugs.destroy_all
+    sign_ins.destroy_all
+    email_histories.destroy_all
+    profile_photo&.destroy!
+  end
+
+  def erase_account
+    sha = Digest::SHA1.hexdigest(rand.to_s)
+
+    update!(
+      name: _('[Name Removed]'),
+      email: "#{sha}@invalid",
+      url_name: sha,
+      about_me: '',
+      password: MySociety::Util.generate_token
+    )
   end
 end
--- a/app/models/user/erasable.rb
+++ b/app/models/user/erasable.rb
@@ -1,6 +1,10 @@
 module User::Erasable
   extend ActiveSupport::Concern
 
+  def erase_later(editor:, reason:)
+    User::ErasureJob.perform_later(self, editor: editor, reason: reason)
+  end
+
   def erase(...)
     erase!(...)
   rescue ActiveRecord::RecordInvalid
--- a/app/models/user/erasable.rb
+++ b/app/models/user/erasable.rb
@@ -16,11 +16,7 @@ module User::Erasable
 
     transaction do
       destroy_identifying_associations
-
-      outgoing_messages.update!(
-        from_name: _('[Name Removed]')
-      )
-
+      make_request_redactions_permanent(editor: editor, reason: reason)
       erase_account
     end
   end
@@ -32,6 +28,13 @@ module User::Erasable
     sign_ins.destroy_all
     email_histories.destroy_all
     profile_photo&.destroy!
+  end
+
+  def make_request_redactions_permanent(...)
+    outgoing_messages.update!(from_name: _('[Name Removed]'))
+
+    return unless feature_enabled?(:user_erase)
+    info_requests.find_each { |request| request.make_redactions_permanent(...) }
   end
 
   def erase_account
--- a/app/controllers/admin/users/erasures_controller.rb
+++ b/app/controllers/admin/users/erasures_controller.rb
@@ -2,16 +2,11 @@
 # Controller for erasing accounts
 #
 class Admin::Users::ErasuresController < AdminController
-  before_action :set_erased_user
+  before_action :set_erased_user, :ensure_erasable
 
   def create
-    if @erased_user.erase(editor: admin_current_user, reason: reason)
-      flash[:notice] = 'The user was erased.'
-    else
-      flash[:error] = 'Something went wrong. The user could not be erased.'
-    end
-
-    redirect_to admin_user_path(@erased_user)
+    @erased_user.erase_later(editor: admin_current_user, reason: reason)
+    redirect_to admin_user_path(@erased_user), notice: erasure_queued
   end
 
   private
@@ -20,7 +15,18 @@ class Admin::Users::ErasuresController < AdminController
     @erased_user = User.find(params[:user_id])
   end
 
+  def ensure_erasable
+    return if @erased_user.closed?
+
+    flash[:error] = 'User accounts must be closed before erasing.'
+    redirect_to admin_user_path(@erased_user)
+  end
+
   def reason
     params[:reason]
+  end
+
+  def erasure_queued
+    'Erasure has been queued.'
   end
 end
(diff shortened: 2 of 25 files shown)


garethrees:
Main TODOs:

- [x] Add `InfoRequest#make_redactions_permanent` – delegate to each outgoing/incoming message
- [x] Update `User#erase` with event params
- [x] Add `User#erase_later` and associated job
- [x] Update `Admin::Users::ErasuresController` to queue the job
- [x] Update `User#erase` with additional action to iterate over request content
- [x] ~Deal with redactions to `InfoRequest` record itself (title is not actually redacted: https://github.com/mysociety/alaveteli/issues/1881)~
- [x] ~Do something about force-masking all attachments – either in `User::ErasureJob` or in `RawEmail#erase` itself (probably former for now so that we don't need to update other erasure calls with jobs)
- [ ] ~Look at `Current.user` for making the event params less of a faff

TODO before merge:

- [x] Add feature flag
- [x] Remove checked in debug line
- [ ] Destroy any user level censor that include the name too
- [ ] Set `User#erased_at`
- [ ] Update inline documentation

Possibly related here or maybe an issue on `develop`:

- [ ] Look at why we're needing to explicitly call (`_cache_main_body_text`)

garethrees:
@gbp I think this is ready for a first pass review. The failing spec is related to 7784c08, which I believe is fixed in an alternative way in https://github.com/mysociety/alaveteli/pull/9149.

app/models/user/erasable.rb
```diff
@@ -34,6 +34,14 @@ def destroy_identifying_associations
     profile_photo&.destroy!
   end
 
+  def make_request_redactions_permanent(...)
+    return unless feature_enabled?(:user_erase)
+    puts "CALLED make_request_redactions_permanent"
```
garethrees: oops checked in a debug line

FOIMonkey:
This is cool. I tested Close → Anonymise → Erase on a user with 12 requests. I waited for the flash message to show before clicking Erase, but something went wrong. The user's name survived in the locked attachment body text of multiple requests. The raw email was still deleted, so there was no way to fix this.                                   
                                              
It seemed to be the requests that I had viewed prior to testing.

Screenshot 2026-04-08 at 13 59 53
Screenshot 2026-04-08 at 14 00 02
Screenshot 2026-04-08 at 14 00 15
Screenshot 2026-04-08 at 14 00 27

gbp (changes requested):
I don't think this is working at all... and I've gone back to the version before I rebased too and it seems broken too.

The raw emails are being deleted and causing `FoiAttachment::MissingAttachment` exceptions preventing `User::ErasureJob` from completing.

I'm getting quite confused around all of this and I think we need to sit down and go through it together.

gbp:
> The raw emails are being deleted and causing `FoiAttachment::MissingAttachment` exceptions preventing `User::ErasureJob` from completing.

Not sure this is caused by these changes, as I've noticed failed jobs in Sidekiq. I guess one for these might have been retried and purged the files from disk?

app/models/user/erasable.rb
```diff
@@ -0,0 +1,56 @@
...
+    sign_ins.destroy_all
+    email_histories.destroy_all
+    profile_photo&.destroy!
+  end
+
+  def make_request_redactions_permanent(...)
+    return unless feature_enabled?(:user_erase)
+    puts "CALLED make_request_redactions_permanent"
+
+    outgoing_messages.update!(from_name: _('[Name Removed]'))
```
garethrees, line 34: This doesn't create edit_outgoing events

gbp:
Okay so after some more testing this does appear to be working. I think the problems I was seeing was due to left over jobs from previous tests. I've updated my setup script to ensure Redis DB is reset, removing any pending Sidekiq jobs.

I've attempted to replicate what @FOIMonkey is seeing but as of yet unable to.

@FOIMonkey any chance you too had old data in your PG or Redis DB? Maybe you could share more instructions on how you tested it and got it to fail?

FOIMonkey:
> @FOIMonkey any chance you too had old data in your PG or Redis DB? 

@gbp No. Just did a fresh test with a clean DB and Sidekiq queue empty.                                                                             
                                                                                              
Here's what I did with two requests from the same user (each with a .txt attachment containing the requester's name as on the account):                                                
                                                                           
  1. Opened request A and clicked through to view the attachment file                                                                                                                   
  2. Did not open request B at all                                                            
  3. Logged into admin → Close → Anonymise → Erase                                                                                                                                      
                                                                                              
  Result:                                                                                                                                                                               
  - Request A (the one I'd viewed): the main body attachment and txt are locked. The requester's name still is visible in the main body of the incoming message, the .txt attachment is redacted when viewed as html, not redacted when downloaded.                                                                                  
  - Request B (the one I hadn't viewed): everything is correctly redacted.                                                                                                                 
                                                                                                                                                                                        
Both raw emails were deleted afterwards, so there's no way to fix the dodgy one. 

What seems to be happening is that request B wasn't masked so gets masked during the erasure step after the censor rule exists. Request A was already masked and cached from my earlier attachment view before any censor rule existed. Would the erasure step perhaps skip it somehow as being already masked?. 

My limited understanding of censor rules has been that they're done on the fly unless manual rules applied by an admin. In this case, the page load that would normally trigger their application didn't happen, so when you finally view it, it is too late.

FOIMonkey:
It doesn't look like User#anonymise! clears the masks or expires the cache - that'll be it.

FOIMonkey:
I've confirmed it's not re-masking after the censor rule has been applied.

Before:
```
Request 119 (attachment 16)
masked_at: 2026-04-14 12:19:18 UTC  locked: false  erased: false  raw_email_erased: false
body: "Hello,\n\nThis is the response to the information request from Sarah Jennings.\n\nDear Sarah Jennings,\n\nPlease find attached our response to your request. We have identified that Sarah Jennings should receive the following documents.\n\nKind regards,\nResponder\n"

Request 120 (attachment 17)
masked_at: 2026-04-14 12:20:49 UTC  locked: false  erased: false  raw_email_erased: false
body: "Hello,\n\nThis is the response to the information request from Sarah Jennings.\n\nDear Sarah Jennings,\n\nPlease find attached our response to your request. We have identified that Sarah Jennings should receive the following documents.\n\nKind regards,\nResponder\n"
```

After Erase job ran at 2026-04-14 12:23:49 UTC

```
Request 119 (attachment 16)
masked_at: 2026-04-14 12:19:18 UTC  locked: true  erased: false  raw_email_erased: true
body: "Hello,\n\nThis is the response to the information request from Sarah Jennings.\n\nDear Sarah Jennings,\n\nPlease find attached our response to your request. We have identified that Sarah Jennings should receive the following documents.\n\nKind regards,\nResponder\n"

Request 120 (attachment 17)
masked_at: 2026-04-14 12:20:49 UTC  locked: true  erased: false  raw_email_erased: true
body: "Hello,\n\nThis is the response to the information request from Sarah Jennings.\n\nDear Sarah Jennings,\n\nPlease find attached our response to your request. We have identified that Sarah Jennings should receive the following documents.\n\nKind regards,\nResponder\n"
```

garethrees:
Need to test a bit further but I think the issues around the censor rule not expiring are resolved. Was too difficult to apply as fixups so here's a summary of the main changes:

* Ensure censor rule creation always triggers request expiry (https://github.com/mysociety/alaveteli/pull/9218)
* Prevent/skip masking attempts where the raw email is erased (a2cafc21777a981ccd7337ba99fdf300d9321d56)
* Mask any unmasked attachments before making redactions permanent on incoming message (874c27448f62b7bc5d00d76c01bffee9a40de8cd) – this used to be in `User::ErasureJob` but is now pushed down closer to the point at which its needed now that we can call it without it raising (via modifications in previous bullet).

FOIMonkey:
A few thoughts. The redacted? check seems to be asking "does any mask change anything in this message." which means the default email address regex is making this return true. On a test request, it erased and locked a message with no PII of the user in it due to the request address in the reply having been redacted.
                                                                                                                      
Almost every FOI response from a public body contains an email address, so the default mask will likely make `redacted?` return true in most cases. This matters, as a subsequent rights-based request from a different person can't then be honoured without full deletion. That might be a necessary trade off where PII was present, but seems suboptimal where it wasn't.

Erase also seems to check for closed, banned and pro, but not if Anonymise has run. If you Click Close → Erase by mistake, then attachments are locked and the message deleted if another rule or the default masks are in place. The body then contains the user's full name etc locked public in a non fixable state. 

Never underestimate my ability to press the wrong button!!!

This all feels broad for an irreversible action. I thought maybe it could check user level censor rules as a sense check, but then it’s normal to add request level rules when tackling attachments/quirky things. Could check for those too, and exclude default and body level ones which are unlikely to be user related. Slight danger there is if the default has removed PII of user, eg they have included their email address, but we don’t run the default on outgoing messages anyway, so it should be obvious enough that this has happened for the admin to add a user/request level rule. 

As this works in a way where PII present that doesn't match the censor rule would be locked public if a rule matches something else,  it might also be an idea to add a note to the collapsed help text for anonymise to make clear what censor rules don’t touch or often fail on. This would help admins know the usual red flags. From memory, this would be: zip, tiff, gif, jpeg, png, bmp, docx, xlsx, xlsb, xlsm, pptx, pdf,  and some .doc/.xls/.ppt files, nested archives/embedded files, outlook safelinks urls, comments and request titles. There are possibly more if you factor in projects things and citations (unlikely these would have pii, but not impossible). 

When responding to a GDPR request, you have to communicate what data you still hold to the data subject, so it is good to know anyway.

garethrees:
> The redacted? check seems to be asking "does any mask change anything in this message." … so the default mask will likely make redacted? return true in most cases. This matters, as a subsequent rights-based request from a different person can't then be honoured without full deletion.

Yeah, I appreciate this isn't ideal and is noted in the commit message that it's a very naive approach. I wanted to get an end to end version working, but I think its worth exploring if we can do something about this. It should be possible without _too_ much effort to limit this check to only censor rules (rather than masks + censor rules).

Future requests can be honoured, but we'd have to use attachment "replacement" rather than censor rules (i.e. redact offline then reupload that version), so it would be more laborious.

> Erase also seems to check for closed, banned and pro, but not if Anonymise has run… Never underestimate my ability to press the wrong button!!!

These attributes are slightly easier to check. For the censor rule we could check `last_edit_editor`/`last_edit_comment` for `User#anonymise!`, but both would change if the rule gets edited. Could just check for any censor rules with `[name removed]` for the replacement. Might fail in edge cases where that gets modified, but should work for 99% of cases.

> This all feels broad for an irreversible action…

Yeah, its tricky. Even if we can reduce the scope of the `redacted?` check as above, I suspect we'll end up locking/erasing more raw emails than necessary. Not sure there's a better way if we want a ~ one-click "erase" button at the user level.

Thinking about it, a different approach could be to move the erasure down to the censor rule – so for each censor rule have a button that makes _this_ censor rule permanent… would mean a bit more clicking around as we add the various user/request rules, and then a final erase step on the user that just erases the remaining account details… Not sure how possible this is with what we have right now though.

> add a note to the collapsed help text

Yep this is a todo once the implementation gets finalised.

garethrees:
Back to draft while we focus on redaction improvements (https://github.com/mysociety/alaveteli/milestone/75) which will hopefully unblock some of the issues here.

PR #9149, opened 2026-03-04
[#8807] Update attachment erasing

gbp:
## Relevant issue(s)

Fixes remaining issues in #8807

## What does this do?

Update FoiAttachment#erase to also erase raw email. This was missed as an original requirement but easy to add.

Fixes attachment filenames not being erased.

<hr>

[skip changelog]

### DIFF
--- a/app/views/admin_incoming_message/_foi_attachments.html.erb
+++ b/app/views/admin_incoming_message/_foi_attachments.html.erb
@@ -24,6 +24,9 @@
               <% if attachment.replaced? %>
                 <span class="label">Replaced</span>
               <% end %>
+              <% if attachment.erased? %>
+                <span class="label">Erased</span>
+              <% end %>
             </td>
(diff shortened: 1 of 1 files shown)


garethrees (changes requested):
Nice fixed. Sorry found one more gotcha – if we erase the main body part we should show an explicit "erased" message rather than a blank message:

Image

**EDIT:** Hm also, when erasing the PDF the main body disappears:

Before erasure:

Screenshot 2026-03-06 at 11 48 00

After erasing the PDF attachment. Should still see "PDF attachment enclosed."

Screenshot 2026-03-06 at 11 48 27

When viewing the main body attachment in the admin interface a `FoiAttachment::MissingAttachment` is raised:

Screenshot 2026-03-06 at 11 50 26

Screenshot 2026-03-06 at 11 51 15

PR #9155, opened 2026-03-10
Fix incoming message cache columns

gbp:
## Relevant issue(s)

#8807 

## What does this do?

Fixes issues where:
- [FoiAttachment wouldn't be viewable if the RawEmail has been erased](https://github.com/mysociety/alaveteli/issues/8807#issuecomment-3972073205)
- the IncomingMessage cache columns could be set to nil when either the FoiAttachment or the RawEmail has been erased.

<hr>

[skip changelog]

### DIFF
--- a/app/models/raw_email.rb
+++ b/app/models/raw_email.rb
@@ -84,7 +84,7 @@ class RawEmail < ApplicationRecord
     !file.attached? && erased_at.present?
   end
 
-  def erased!
+  def ensure_not_erased!
     raise ErasedError, "email has been erased (ID=#{id})" if erased?
   end
 
@@ -93,7 +93,7 @@ class RawEmail < ApplicationRecord
   end
 
   def erase(editor:, reason:)
-    erased!
+    ensure_not_erased!
 
     raise UnmaskedAttachmentsError unless all_attachments_masked?
 
--- a/spec/models/raw_email_spec.rb
+++ b/spec/models/raw_email_spec.rb
@@ -243,7 +243,7 @@ RSpec.describe RawEmail do
     end
   end
 
-  describe '#erased!' do
+  describe '#ensure_not_erased!' do
     let(:raw_email) do
       request = FactoryBot.create(:info_request)
       message = FactoryBot.create(:incoming_message, info_request: request)
@@ -254,7 +254,7 @@ RSpec.describe RawEmail do
 
     context 'when not erased' do
       it 'returns nil' do
-        expect(raw_email.erased!).to be_nil
+        expect(raw_email.ensure_not_erased!).to be_nil
       end
     end
 
@@ -267,7 +267,7 @@ RSpec.describe RawEmail do
       end
 
       it 'raises ErasedError' do
-        expect { raw_email.erased! }.to raise_error(
+        expect { raw_email.ensure_not_erased! }.to raise_error(
           described_class::ErasedError,
           "email has been erased (ID=#{raw_email.id})"
         )


app/models/foi_attachment.rb
```diff
@@ -292,6 +288,10 @@ def erased?
     erased_at.present? || raw_email_erased?
   end
 
+  def erased!
```
garethrees: Would favour a method name with a little extra clarity for this, even at the expense of verbosity. Maybe something like:

* `raise_if_erased!`
* `must_not_be_erased!`
* `cannot_be_erased!`

spec/models/incoming_message_spec.rb
```diff
@@ -665,16 +665,33 @@ def last_event_for(attachment)
...
-      incoming_message.reload
-      expect(incoming_message.cached_main_body_text_folded).to be_nil
+    context 'when main body text is erased' do
+      before do
+        incoming_message.get_main_body_text_folded
+        allow(incoming_message.get_main_body_text_part).
+          to receive(:erased?).and_return(true)
+      end
+
+      it 'does not clear cached values' do
```
garethrees, line 721: What if we _do_ want to erase the main body part (e.g. there's PII there that we can't keep)? We don't want that preserved in caches?

garethrees: Thinking about this, I think we want the associated check on the _caching_ side rather than the _clearing_ side? I've just tried this and it looks like it works okay, though only had a quick check this morning.

```diff
diff --git a/app/models/incoming_message.rb b/app/models/incoming_message.rb
index cb971dddb..fbd85eb67 100644
--- a/app/models/incoming_message.rb
+++ b/app/models/incoming_message.rb
@@ -167,8 +167,6 @@ def apply_masks(text, content_type)
 
   # Removes anything cached about the object in the database, and saves
   def clear_in_database_caches!
-    return if get_main_body_text_part&.erased?
-
     self.cached_attachment_text_clipped = nil
     self.cached_main_body_text_unfolded = nil
     self.cached_main_body_text_folded = nil
diff --git a/app/models/incoming_message/attachments.rb b/app/models/incoming_message/attachments.rb
index 46fc44419..a606bbf2c 100644
--- a/app/models/incoming_message/attachments.rb
+++ b/app/models/incoming_message/attachments.rb
@@ -79,7 +79,7 @@ def get_attachments_for_display # rubocop:disable Naming/AccessorMethodName
     # return what user would consider attachments, i.e. not the main body
     main_part = get_main_body_text_part
     attachments = []
-    foi_attachments.each do |attachment|
+    # ideally create a nicer scope like .not.erased or .retained
+    foi_attachments.not(erased_at: nil).each do |attachment|
       attachments << attachment if attachment != main_part
     end
     attachments
diff --git a/app/models/incoming_message/main_body.rb b/app/models/incoming_message/main_body.rb
index 193a1c978..d42f0dcbd 100644
--- a/app/models/incoming_message/main_body.rb
+++ b/app/models/incoming_message/main_body.rb
@@ -150,6 +150,8 @@ def get_body_for_quoting # rubocop:disable Naming/AccessorMethodName
   # Cached as loading raw_email can be quite huge, and need this for just
   # search results
   def _cache_main_body_text
+    return nil if get_main_body_text_part&.erased?
+
     text = get_main_body_text_internal
     # Strip the uudecode parts from main text
     # - this also effectively does a .dup as well, so text mods don't alter
```

gbp: Maybe I'm not understanding what your diff is trying to do...

But not sure why you're proposing changes to `get_attachments_for_display`?

This is called from several different points in the code base including when we list attachments in the frontend. With your changes (assuming I'm correcting it correctly: `s/not/where/` ?) it would hide erased attachments from he frontend. I'm sure you said you wanted them to remain visible so users can see something has been deleted?

garethrees: Ah, yes. We'll need that check elsewhere for the attachments case.

The main point is that when we erase an attachment, we should not leave its data cached in one of the cached columns.

gbp:
Treating as one PR in #9149

Issue #8813, opened 2025-08-15
Add new search engine system


garethrees:
For efficacy, performance and operational reasons we'd like to replace Xapian.

There are a few questions we need to work through.

* What's wrong with what we've got?
* What do we want to do that we can't?
* What do we want to keep?
* What do we replace it with?
* How do we want to design the indexing?
  * Specifically, do we want to continue indexing the _events_ or only the specific records, or some other variant?

We said we'd evaluate things like Postgres vector search, Rails’ new Active Record Search, Meilisearch, etc

garethrees:
37signals have completely open-sourced their new product: https://github.com/basecamp/fizzy/

Having a quick poke around it looks like its all database search so this could be interesting to study (and be a hint for what Active Record Search might look like).

garethrees:
> To make this happen, we've also started work on Active Record Search with full-text implementations for both SQLite and MySQL (Trilogy). https://github.com/basecamp/fizzy/blob/main/app/models/search/record/sqlite.rb
> – https://x.com/dhh/status/1995906589241311411

laurentS:
This text is meant to summarise our needs and findings about this problem and potential solutions. It will be updated as we go.

# Defining the problem

Alaveteli currently uses `xapian` for its search (and plain SQL search in the site admin section, both sides are candidates for replacement).

## Limitations of the current search system

- Many users have reported unsuccessful searches, where the obvious result didn't appear, particularly when searching for a public body to send a request to. This has happened in multiple sites (and multiple languages)
- `xapian` has limited language support. When using the French stemmer, results are somewhat better than with the default English one (for French content), but not as good as what users have come to expect. Several languages that alaveteli is used in are not supported by Xapian.
- admin docs are limited, making it much harder than it should be to configure the search engine
- index updates are slow for full re-indexing, which means bigger sites are essentially unable to do this at all. Changes to indexing logic are more or less impossible because of this.
- indexing of attachments is lacking for many document types, and non-existent for things like images, scanned pdfs, etc...
- no vector/semantic search: xapian is only able to do text search. We can search for "train" (which will also return things like "trains" for languages which have a functional stemmer), but won't match on "railroad" or "railway". 

# Requirements for the new search system

## Use cases

All these use cases should be accessible via a common search interface (at the code level):

- site end users are looking for documents relating to a given topic or mentioning a specific word/phrase
- site admins search all occurrences of a user's name as part of a GDPR request
[...]

PR #9160, opened 2026-03-15
Pg search ruby experiment (WIP)

laurentS:
## Relevant issue(s)
#8813 

## What does this do?
This is not for merging! This tries various search options using postgresql.

- for reference, reindexing the full list of 51k Ma Dada public bodies (only) with xapian on my laptop took 20 minutes 5 secs.

### multisearch

- Following the pg_search README, I created a `multisearch` table for public bodies only, with their name+short name. Building the table for the 51k bodies took 92 seconds. Changing the dictionary to use french instead of the default english, same result. This does NOT build an index that postgres can use to search. It's a very naive approach, which leads to poor performance: searching for "ministere interieur" (our pathological case) returns the expected result plus other reasonable records in 500-900ms. This is extremely slow, because postgres does a full table scan and calculates tsvectors for each record on the fly (as shown by EXPLAIN). As this is O(N), doing the same on the messages tables would lead to queries taking multiple seconds for us, likely minutes for WDTK. The solution is to precalculate tsvectors and store them in a table, and build a GIN index, as suggested under [#using-tsvector-columns](https://github.com/Casecommons/pg_search?tab=readme-ov-file#using-tsvector-columns). (it's not possible to build an index directly when using the `unaccent` function, because the latter is not "immutable", so accented languages need a stored column with tsvectors in it) (see commits marked (1))

### pg_search_scope

- pg_search_scope does not work on PublicBody against :name and/or :short_name, because those are not actual columns in the corresponding db table. There might be a way to do this through the versioning and translation magic in ruby (probably not, as `pg_search` seems to expect the identifiers to be pg column names), but it seems a lot easier (and faster at execution) to me to simply write SQL queries.

- I then created a simple search scope on IncomingMessage (15k records) against cached_main_body_unfolded worked, but it is slow: a test query for "factures de repas" (invoices for meals) returned a very good first hit, followed by several half relevant ones (matching only on "invoice"). The french stemmer seems to be the one to blame, as it cuts "repas" in a way that does not make sense. Query takes 5.2s, because of a sequential scan on the table (meaning it is not usable for WDTK). Tried a few other search strings, results are half convincing. (commits marked (2)).

### GIN index, manual queries

- adding a tsvector column to `PublicBody`, and building a GIN index on it makes search extremely fast, ~1ms for 51k public bodies. I can rebuild the full search index inside postgres in under 2 seconds. (see migration in commit (3)). Search is only using `tsvector` here, so it uses the stemmer, which won't handle typos, for instance. However, as a first step, for comparison:

#### Xapian
```ruby
alaveteli(dev)> ActsAsXapian::Search.new([PublicBody], 'bergerac', limit: 10).results.map { |pb| pb[:model].name }
["Mairie - Bergerac",
 "Trésorerie de Bergerac",
 "Bureau information jeunesse - Bergerac",
 "Sous-préfecture de Bergerac",
 "Bureau de douane (viticulture) - Bergerac",
 "Greffe des associations - Dordogne (Bergerac)",
 "Service de la publicité foncière (SPF) - Bergerac",
 "Chambre de commerce et d'industrie (CCI) - Dordogne - Bergerac",
 "Centre d\u0092information et d\u0092orientation (CIO) - Bergerac",
 "Maison de justice et du droit  de Bergerac"]
```
xapian query runtime: 0.2-3ms, plus pg queries: total ~1.5ms

#### pg_search_scope (no index used)

```ruby
alaveteli(dev)> PublicBody.pg_search_with_scope("bergerac").map { |pb| pb.name}
=>
["Centre d'information sur les droits des femmes et des familles (CIDFF) - Dordogne - Bergerac 24037-01",
 "Plateforme d'accompagnement et de répit pour les aidants de personnes âgées - Bergerac",
 "Bureau de douane (viticulture) - Bergerac",
 "Chambre de commerce et d'industrie (CCI) - Dordogne - Bergerac",
 "Bureau information jeunesse - Bergerac",
 "Centre d\u0092information et d\u0092orientation (CIO) - Bergerac",
 "Service de la publicité foncière (SPF) - Bergerac",
 "Mairie - Bergerac",
 "Mission locale pour l'insertion professionnelle et sociale des jeunes (16-25 ans) - Bergerac",
 "Maison de justice et du droit  de Bergerac",
 "Centre de protection maternelle et infantile (PMI) - Bergerac",
 "Greffe des associations - Dordogne (Bergerac)",
 "Service des impôts des entreprises du centre des finances publiques de Bergerac",
 "Sous-préfecture de Bergerac",
 "Trésorerie de Bergerac"]
```
Ranking of results isn't the best. I'd expect the Mairie to come first, but we don't give pg enough info to guess this. Query runtime 49ms. Too slow.

#### using GIN index

```ruby
alaveteli(dev)> PublicBody.search_with_pg("bergerac", "french").map {|pb| pb.name}
=>
["Centre d'information sur les droits des femmes et des familles (CIDFF) - Dordogne - Bergerac 24037-01",
 "Plateforme d'accompagnement et de répit pour les aidants de personnes âgées - Bergerac",
 "Bureau de douane (viticulture) - Bergerac",
 "Chambre de commerce et d'industrie (CCI) - Dordogne - Bergerac",
 "Bureau information jeunesse - Bergerac",
 "Centre d\u0092information et d\u0092orientation (CIO) - Bergerac",
 "Service de la publicité foncière (SPF) - Bergerac",
 "Mairie - Bergerac",
 "Mission locale pour l'insertion professionnelle et sociale des jeunes (16-25 ans) - Bergerac",
 "Maison de justice et du droit  de Bergerac",
 "Centre de protection maternelle et infantile (PMI) - Bergerac",
 "Greffe des associations - Dordogne (Bergerac)",
 "Service des impôts des entreprises du centre des finances publiques de Bergerac",
 "Sous-préfecture de Bergerac",
 "Trésorerie de Bergerac"]
```
runtime: 1.2ms + a set of pg selects to fetch translations issued by rails, which we can probably optimise like for xapian. This query uses the GIN index, so it should scale nicely.

This is not typo-tolerant yet, but we can add a `pg_trgm` query for that. More to come.

- Side note: I don't think that semantic search for public bodies makes much sense, as I expect most users know the name of the entity they're looking for (modulo spelling, typos, acronyms, etc... for which tsvectors are much more adapted than semantic search, I think)

### DIFF
--- a/app/models/public_body.rb
+++ b/app/models/public_body.rb
@@ -158,6 +158,47 @@ class PublicBody < ApplicationRecord
   # This does not work, because name and short_name are not actual columns
   # on the public_bodies table.
   # pg_search_scope :search_by_name, against: [:name, :short_name]
+ 
+  # use a pg_search_scope built on a tsvector column that is updated by
+  # a pg trigger, as shown in docs
+  pg_search_scope :pg_search_with_scope,
+                # pg_search requires a colum from public_bodies here, or it fails
+                # to build a valid SQL query, so I used something that shouldn't interfere too much
+                # with the search terms, but this is not a viable option here
+                against: :last_edit_comment,
+                using: {
+                  tsearch: {
+                    dictionary: 'french',
+                    tsvector_column: 'name_tsv'
+                  },
+                  trigram: {} # trigram does not use tsvectors
+                }
+
+  def self.search_with_pg(query, locale= AlaveteliLocalization.locale)
+    # sub-ms search, uses the GIN(tsvector) index created in migration.
+    # pg_search does not seem to be able to produce this by itself.
+    # susceptible to injection, not for production use!
+    search_terms = query.split(' ').map{|q| "to_tsquery('#{locale}', unaccent(COALESCE('#{q}', '')))"}.join(' && ')
+
+    sql = <<-SQL
+      SELECT
+        "public_bodies".* FROM "public_bodies" INNER JOIN (
+        SELECT
+          "public_bodies"."id" AS pg_search_id,
+          (ts_rank(("public_bodies"."name_tsv"), (#{search_terms})), 0) AS rank
+        FROM "public_bodies"
+        WHERE
+          (("public_bodies"."name_tsv") @@ (#{search_terms}))) AS pg_search_subreq
+      ON
+        "public_bodies"."id" = pg_search_subreq.pg_search_id
+      ORDER BY
+        pg_search_subreq.rank DESC,
+        "public_bodies"."id" ASC
+    SQL
+
+    PublicBody.find_by_sql(sql)
+  end
+
 
   non_versioned_columns << 'created_at' << 'updated_at' << 'first_letter' << 'api_key'
   non_versioned_columns << 'info_requests_count' << 'info_requests_successful_count'
--- a/db/migrate/20260315143113_add_tsvector_column_to_public_body.rb
+++ b/db/migrate/20260315143113_add_tsvector_column_to_public_body.rb
@@ -1,0 +1,43 @@
+class AddTsvectorColumnToPublicBody < ActiveRecord::Migration[8.0]
+  def up
+    add_column(:public_bodies, :name_tsv, :tsvector)
+    # this index is used by search_with_pg, but not pg_search_with_scope
+    add_index(:public_bodies, :name_tsv, using: :gin)
+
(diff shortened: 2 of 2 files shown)


laurentS:
Closing this PR as it is not meant to be merged, but keeping the code and write up for later reference.

PR #9163, opened 2026-03-16
Use FoiAttachment#display_filename when downloading

garethrees:
Ensures correct redactions are applied.

Regressions introduced in https://github.com/mysociety/alaveteli/commit/dec1b6dc06ba66c49b819a18ffb0b0c8bcb6290e (0.46.0.0) and https://github.com/mysociety/alaveteli/commit/870f8f774d8880489f678e374c38a8cb13d3290f (0.38.0.0).

### DIFF
--- a/app/controllers/attachments_controller.rb
+++ b/app/controllers/attachments_controller.rb
@@ -26,7 +26,7 @@ class AttachmentsController < ApplicationController
     else
       send_data(
         @attachment.body,
-        filename: @attachment.filename,
+        filename: @attachment.display_filename,
         type: content_type
       )
     end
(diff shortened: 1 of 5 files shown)


app/controllers/attachments_controller.rb
```diff
@@ -26,7 +26,7 @@ def show
     else
       send_data(
         @attachment.body,
-        filename: @attachment.filename,
+        filename: @attachment.display_filename,
```
garethrees, line 29: Or could use `redacted_filename` here. `display_filename` calls `redacted_filename` and then does some URL escaping stuff. Not sure which is more appropriate.

garethrees: We use `display_filename` in https://github.com/mysociety/alaveteli/blob/0.46.0.0/app/controllers/request_controller.rb#L499-L503

PR #9184, opened 2026-03-26
Add CSP script header.

dracos:
[Initial try to see what happens]

Noting https://github.com/github/secure_headers/issues/511 if a gem was to include its own script tags.


gbp:
Went with https://github.com/mysociety/alaveteli/pull/9187

PR #9187, opened 2026-03-26
Add CSP with nonce-based script protection

gbp:
- Configure CSP initializer with strict_dynamic for scripts
- Update views to use `javascript_tag`/`javascript_include_tag` with `nonce: true` instead of raw `<script>` tags.

Scripts from external sources are automatically trusted via "strict_dynamic" when loaded by nonced scripts.

### DIFF
--- a/config/initializers/content_security_policy.rb
+++ b/config/initializers/content_security_policy.rb
@@ -4,22 +4,23 @@
 # See the Securing Rails Applications Guide for more information:
 # https://guides.rubyonrails.org/security.html#content-security-policy-header
 
-# Rails.application.configure do
-#   config.content_security_policy do |policy|
-#     policy.default_src :self, :https
-#     policy.font_src    :self, :https, :data
-#     policy.img_src     :self, :https, :data
-#     policy.object_src  :none
-#     policy.script_src  :self, :https
-#     policy.style_src   :self, :https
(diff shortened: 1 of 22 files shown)


config/initializers/content_security_policy.rb
```diff
@@ -4,22 +4,23 @@
...
-#   # config.content_security_policy_report_only = true
-# end
+Rails.application.configure do
+  config.content_security_policy do |policy|
+    policy.default_src :self, :https
+    policy.font_src    :self, :https, :data
+    policy.img_src     :self, :https, :data
+    policy.object_src  :none
+    policy.script_src  :self, :strict_dynamic
+    policy.style_src   :self, :https, :unsafe_inline
```
gbp, line 14: Probably want to leave as the default

```suggestion
    policy.style_src   :self, :https
```

gbp: Actually I'm going to allow `unsafe_inline`. We have around 23 inline styles:
`clear:both`, `display:none`, `text-align:center`, `margin:...`, `overflow:hidden`, `visibility:hidden`, `width: X%"` (dynamic). Plus any in themes.

We could work to remove all these but allowing inline styles is easier for now. Can always change this in the future.

PR #9188, opened 2026-03-26
Restrict visibility of hidden and requester-only requests in projects

garethrees:
Contributors shouldn't see hidden or requester_only requests.

The tests for this are a bit confusing and not currently passing.

### DIFF
--- a/spec/models/project_spec.rb
+++ b/spec/models/project_spec.rb
@@ -170,6 +170,12 @@ RSpec.describe Project, type: :model, feature: :projects do
       it 'excludes non-project batch requests' do
         is_expected.not_to include(*other_batch.info_requests)
       end
+
+      it 'excludes hidden requests' do
+        hidden = FactoryBot.create(:info_request, :hidden)
+        project.requests << hidden
+        is_expected.not_to include(hidden)
+      end
     end
 
     it 'has one key set' do
--- a/app/controllers/projects/classifies_controller.rb
+++ b/app/controllers/projects/classifies_controller.rb
@@ -77,10 +77,17 @@ class Projects::ClassifiesController < Projects::BaseController
   end
 
   def load_info_request_from_queue
-    @info_request = (
-      @queue = Project::Queue.classifiable(@project, session)
-      @queue.next
-    )
+    @queue = Project::Queue.classifiable(@project, session)
+    @info_request = next_readable_request
+  end
+
+  def next_readable_request
+    request = @queue.next
+    while request && cannot?(:read, request)
+      @queue.skip(request)
+      request = @queue.next
+    end
+    request
   end
 
   def load_info_request_from_url_title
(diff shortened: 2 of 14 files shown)


app/models/project/info_request_extension.rb
```diff
@@ -7,11 +7,11 @@ module InfoRequestExtension
     EXTRACTABLE_STATES = %w(successful partially_successful)
 
     def classifiable
-      where(awaiting_description: true)
+      where(awaiting_description: true, prominence: %[normal backpage])
```
garethrees: Can't use `is_public` scope as that excludes private requests, which can be added to projects.

garethrees: Would be more ideal to exclude fully `hidden` requests in the main `Project#info_requests` method – that way we don't have to add filtering every time we enumerate them.

`requester_only` is a bit more complicated; ideally the requester will still be able to see them, but not contributors. I think that means checking at controller level, and then if the user can't read that request, skipping to the next one.

gbp:
```suggestion
      where(awaiting_description: true, prominence: %w[normal backpage])
```

app/models/project/info_request_extension.rb
```diff
@@ -31,6 +31,7 @@ def extractable
     def extracted
       joins(:extraction_project_submissions).
         where(project_submissions: { project: project }).
+        where(prominence: %w[normal backpage])
```
gbp:
```suggestion
        where(prominence: %w[normal backpage]).
```

spec/models/project_spec.rb
```diff
@@ -325,10 +339,15 @@
...
     let(:classifiable_request) { FactoryBot.create(:awaiting_description) }
     let(:classified_request) { FactoryBot.create(:successful_request) }
+    let(:hidden_request) { FactoryBot.create(:successful_request, :hidden) }
 
     let(:project) do
       project = FactoryBot.create(:project)
-      project.requests << [classifiable_request, classified_request]
+      project.requests << [
+        classifiable_request,
+        non_classifiable_request,
```
gbp:
```suggestion
        classified_request,
```

app/models/project/info_request_extension.rb
```diff
@@ -7,11 +7,11 @@ module InfoRequestExtension
     EXTRACTABLE_STATES = %w(successful partially_successful)
 
     def classifiable
-      where(awaiting_description: true)
+      where(awaiting_description: true, prominence: %[normal backpage])
     end
 
     def classified
-      where(awaiting_description: false)
+      where(awaiting_description: false, prominence: %[normal backpage])
```
gbp:
```suggestion
      where(awaiting_description: false, prominence: %w[normal backpage])
```

gbp (dismissed):
Few changes needed to get the specs passing

garethrees:
@gbp please do take over this PR – feel free to change authorship on the original commit if you want

gbp:
> @gbp please do take over this PR – feel free to change authorship on the original commit if you want

Yes still working on it. Just getting the specs running.

gbp:
@garethrees I think this is good to go now. If you're happy feel free to merge and deploy.

PR #9202, opened 2026-04-06
[experiment] Semantic search in postgres

laurentS:
This PR is not for merging!

It is a POC for hybrid search built directly in postgres (hybrid meaning full-text search + semantic search results combined).

It uses reciprocal ranked fusion (RRF) for merging results, which seems to be the most commonly used solution (without resorting to another neural network derivative).

On ~75k documents (email bodies) in French, FTS results are good, although ranking is difficult to assess. It is very likely that something like BM25 would improve results there.

Embedding the text from an email takes about 70-300ms. This depends on the run, more than the email content, apparently. I have not been able to identify what slows down the process, other than guessing that the process probably starts swapping at some point, as RAM usage is very high (40+GB). Running a fresh process for each batch of ~1k docs is likely to be faster than creating much larger batches, even when accounting for the time it takes to load the language model. If embedding a large backlog of docs, it's most certainly worth renting a beefy VPS with tons of RAM for a few hours.

The model I used here is limited to 512 words, but it seems to bail out much earlier, so the code artificially uses a 1600 char limit which seemed to work more often than not (not a very useful metric :wink:) 

Overall impressions:
- indexing content is reasonably fast, and even with large backlogs, we should be able to write the code so that FTS is indexed first, then other slower search data can be built in the background
- with the right code interface, we should be able to evolve the search query over time, or even keep multiple versions that users can choose from (my guess is different languages will require different settings, as the stemmer, embedder and other language specific tools will vary substantially in quality)
- with 100k records, search returns in less than 10ms. Not sure about scaling, but since all queries in the db are hitting indexes and not doing full table scans, I believe this should scale pretty nicely
- as mentioned before, because this is SQL in postgresql, we have all the fancy optimisation tooling available. Coupled with db migrations, I am convinced we can iterate over time towards a good solution, even if the first version is not much better than what we currently have.
- Harder things: using pg means one less external dependency to worry about, but we would still need pg extensions (pgvector...) but this is still probably easier to handle than something like elasticSearch
- obviously, having all the knobs exposed means we can fine tune, but also spend a ton of time on it :snail: 

The main open question in my mind is whether postgres FTS can function with a language that is not part of the 29 snowball stemmer langs that come with stock pg installs.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -186,6 +186,11 @@ gem 'excel_analyzer', path: 'gems/excel_analyzer', require: false
 # AI insights
 gem "ollama-ai", "~> 1.3.0"
 
+# search experiment
+gem "numo-narray", :git => "https://github.com/ruby-numo/numo-narray.git"
+gem "rb_sys"  # required to build transformers-rb (see mkmf error msg)?
+gem "transformers-rb"
+
 group :test do
   gem 'fivemat', '~> 1.3.7'
   gem 'webmock', '~> 3.26.1'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -23,6 +23,12 @@ GIT
     strip_attributes (1.14.1)
       activemodel (>= 3.0, < 9.0)
 
+GIT
+  remote: https://github.com/ruby-numo/numo-narray.git
+  revision: 95c05257349b954c027c2284834408736244662a
+  specs:
+    numo-narray (0.9.2.1)
+
 PATH
   remote: gems/alaveteli_features
   specs:
@@ -474,6 +480,9 @@ GEM
       zeitwerk (~> 2.6)
     rainbow (3.1.1)
     rake (13.3.1)
+    rake-compiler-dock (1.11.0)
+    rb_sys (0.9.124)
+      rake-compiler-dock (= 1.11.0)
     rdoc (7.2.0)
       erb
       psych (>= 4.0.0)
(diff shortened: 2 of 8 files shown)


laurentS:
Further notes about semantic search:
- batching the embedding generation step speeds up the process substantially. From about 50ms/record, we dropped to 10-20ms/record when embedding in batches of 50 or 100 records (although this was tested on a dump of HN comments, which tend to be shorter than emails, we only used the first 1600 chars of each item, so the actual length is somewhat mitigated by this).
- 20ms/record is about 5.5hrs per million records.
- the process seems very parallelisable, so when running the initial embedding generation, boosting the CPU of the machine running the process would speed up ~linearly.
- RAM use is high, 20-30GB, but not substantially different when batching. However, even a single record require multiple GB of RAM to embed.
- error rate is very high with [`transformers-ruby`](https://github.com/ankane/transformers-ruby), in the 20-30% range, meaning that many records were not actually embedded (the reason is unclear as the lib just errors out with "TODO"). This was tested on the `sentence-transformers/multi-qa-MiniLM-L6-cos-v1` model. It is possible that other models fare better.
- the hybrid search test done in this PR shows that the merging of each search result needs a bit of tuning, as FTS results seem to rank higher than semantic ones.
- Semantic search in English seems to work better than in French, which makes sense as the model was largely trained on English content. Fine tuning might be needed to improve results.
- On the search side, with ~1M embeddings, search still returns 10-30ms most of the time. Indexes seem to be used in all queries.
- with ~1M embeddings, the HNSW index weighs 1.87GB. The `search_documents` table weighs 8.5GB, with ~1M embeddings (fixed size per record) and 5.9M `tsvector`s (size varies with length of raw text).

laurentS:
Closing this as code for later reference.

PR #9208, opened 2026-04-09
WIP: Add AlaveteliTextMasker.pdftk_memory_limit

garethrees:
Allows customisation at theme level if needed

    Rails.application.config.after_initialize do
      AlaveteliTextMasker.pdftk_memory_limit = 2_147_483_648 # 2gb
    end

[skip changelog]

### DIFF
--- a/lib/alaveteli_text_masker.rb
+++ b/lib/alaveteli_text_masker.rb
@@ -4,6 +4,8 @@ require 'config_helper'
 module AlaveteliTextMasker
   include ConfigHelper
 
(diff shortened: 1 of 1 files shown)


garethrees:
The specs are failing because the default value (512mb) I've given is not enough memory; they pass when given 8gb.

The output is misleading – it looks like the memory limit is not being set correctly, with the opts hash being passed to the external command – this isn't the case, its just a formatting issue with the output where it interpolates the full method argument list into a string, with the underlying ExternalCommand correctly handling the cli args/opts hash.

PR #9209, opened 2026-04-09
[#8821] Add CensorRule#case_sensitive

### LINKED ISSUE #8302: Make censor be case insensitive
Firstname Surname, firstname surname, FIRSTNAME SURNAME and Firstname SURNAME should ideally be treated the same. 

This was a problem that I came across as part of dealing with an RtO/RtE where the standard C&A rule missed some versions of the full name in plain text correspondence due to capitalisation. Users often capitalise their name in an inconsistent way.

### LINKED ISSUE #8821: Ability to create template rules to reduce brittleness of rules that apply to multiple records

garethrees:
## Relevant issue(s)

Part of #8821
Fixes https://github.com/mysociety/alaveteli/issues/8302

## What does this do?

Allows censor rules to be case insensitive, whether using a plain text or regexp version.

## Why was this needed?

Annoying to have to create multiple rules for common variants where the only difference is case.

## Implementation notes

I was originally thinking we'd expose several regexp options, but since multiline is set by default the only other one that felt like it made sense was case insensitivity. As such I just went with a boolean column rather than storing the regexp opts as a string or bitmask, which makes everything a bit easier to understand. Not massively difficult to change though.

I opted to make case insensitivity apply to "non-regexp" rules so that non-technical admins would feel more confident in using the feature. It does convert the plain text to a regexp under the hood, but think this should be fine.

### Text Rule

BEFORE

Screenshot 2026-04-13 at 15 00 39

RULE

Note lower-case in name

Screenshot 2026-04-13 at 15 01 14

AFTER

Screenshot 2026-04-13 at 15 01 21

### Regexp Rule

BEFORE

Screenshot 2026-04-13 at 15 02 26

RULE

Note mixed case in name

Screenshot 2026-04-13 at 15 02 44

AFTER

Screenshot 2026-04-13 at 15 02 50

### Binary

BEFORE

Screenshot 2026-04-13 at 15 00 05

RULE

Note lowercase "fs"

Screenshot 2026-04-13 at 14 59 32

AFTER

Screenshot 2026-04-13 at 14 59 44

## Notes to reviewer

<hr>

### DIFF
--- a/app/controllers/admin_censor_rule_controller.rb
+++ b/app/controllers/admin_censor_rule_controller.rb
@@ -87,7 +87,8 @@ class AdminCensorRuleController < AdminController
   def censor_rule_params
     if params[:censor_rule]
       params.require(:censor_rule).
-        permit(:regexp, :text, :replacement, :last_edit_comment, :last_edit_editor)
+        permit(:regexp, :case_sensitive, :text, :replacement,
+               :last_edit_comment, :last_edit_editor)
     else
       {}
     end
--- a/app/models/censor_rule.rb
+++ b/app/models/censor_rule.rb
@@ -13,6 +13,7 @@
 #  created_at        :datetime         not null
 #  updated_at        :datetime         not null
 #  regexp            :boolean          default(FALSE), not null
+#  case_sensitive    :boolean          default(TRUE), not null
 #
 
 # models/censor_rule.rb:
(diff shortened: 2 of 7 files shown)


gbp (changes requested):
As mentioned review to https://github.com/mysociety/alaveteli/pull/9224 there are some merge conflicts

PR #9217, opened 2026-04-13
Allow AlaveteliTextMasker masks to be customised

### LINKED ISSUE #8822: Ensure redaction system is adaptable to international contexts

garethrees:
Part of https://github.com/mysociety/alaveteli/issues/8822.

Expose an API to add to, remove from or modify the default text masks to
better suit site specific contexts.

### DIFF
--- a/doc/CHANGES.md
+++ b/doc/CHANGES.md
@@ -2,6 +2,7 @@
 
 ## Highlighted Features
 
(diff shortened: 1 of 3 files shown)


gbp (commented):
Would ideally make these censor rules but this interface will do for now

PR #9220, opened 2026-04-16
[#8821] Regexp `User#anonymise!` rule

### LINKED ISSUE #8821: Ability to create template rules to reduce brittleness of rules that apply to multiple records

garethrees:
## Relevant issue(s)

Requires https://github.com/mysociety/alaveteli/pull/9209
Part of #8821

## What does this do?

Creates a regexp rule to handle common name variants during user anonymisation

## Why was this needed?

Just using the account name alone usually doesn't capture the variety of names used during correspondence.

## Screenshots

BEFORE

Screenshot 2026-04-16 at 11 44 32

RULE CREATION

Screenshot 2026-04-16 at 11 40 00

AFTER

Screenshot 2026-04-16 at 11 44 09

## Notes to reviewer

<hr>

### DIFF
--- a/app/models/user/anonymisable/name_pattern.rb
+++ b/app/models/user/anonymisable/name_pattern.rb
@@ -1,0 +1,76 @@
+# Split a name into parts to construct a pattern suitable for a Regexp
+# CensorRule#text
+class User::Anonymisable::NamePattern
+  DEFAULT_HONORIFICS = 'Mr|Mrs|Miss|Ms|Mx'
+  DEFAULT_PATTERN = /
+    \b
+    (?:
+      (?:%{honorifics})\.? \s+ (?:%{firstname}\s+)? %{surname}
+      |
+      %{firstname} (?: \s+ %{surname} | \s+ %{last_initial}\.? )?
+      |
+      %{first_initial}\.? \s* %{surname}
+      |
+      %{surname} ,\s+ (?: %{firstname} | %{first_initial}\.? )
+    )
+    (?!\w)
+  /x
+
+  cattr_accessor :honorifics, default: DEFAULT_HONORIFICS
+  cattr_accessor :pattern, default: DEFAULT_PATTERN
+
+  def initialize(name)
+    @name = name
+  end
+
+  def to_censor_rule_text
+    format(pattern_string, substitutions)
+  end
+
+  # Hash with Regexp-escaped values
+  # honorifics is a Regexp fragment so left unescaped
+  def substitutions
+    to_h.
+      except(:honorifics).
+      transform_values { |v| Regexp.escape(v) }.
+      merge(honorifics: honorifics)
+  end
+
+  def to_h
+    { firstname: firstname, surname: surname,
+      first_initial: first_initial, last_initial: last_initial,
+      honorifics: honorifics }
+  end
+
+  def firstname
+    parts.first
+  end
+
+  def surname
+    parts.last
+  end
+
+  def first_initial
+    parts.first.mb_chars.first.to_s
+  end
+
+  def last_initial
+    parts.last.mb_chars.first.to_s
+  end
+
+  protected
+
+  attr_reader :name
+
+  private
+
+  def pattern_string
+    return pattern unless pattern.is_a?(Regexp)
+    return pattern.source unless (pattern.options & Regexp::EXTENDED).nonzero?
+    "(?x:#{pattern.source})"
+  end
+
+  def parts
+    @parts ||= @name.strip.split(/\s+/)
+  end
+end
--- a/spec/models/user/anonymisable/name_pattern_spec.rb
+++ b/spec/models/user/anonymisable/name_pattern_spec.rb
@@ -1,0 +1,206 @@
+require 'spec_helper'
+
+RSpec.describe User::Anonymisable::NamePattern do
+  describe '.honorifics' do
+    subject { described_class.honorifics }
+    it { is_expected.to eq('Mr|Mrs|Miss|Ms|Mx') }
+  end
+
+  describe '.honorifics=' do
+    before { described_class.honorifics = 'Dr' }
+    after { described_class.honorifics = described_class::DEFAULT_HONORIFICS }
+
+    subject { described_class.honorifics }
+
+    it { is_expected.to eq('Dr') }
+  end
+
+  describe '.pattern' do
+    subject { described_class.pattern }
+
+    it { is_expected.to eq(described_class::DEFAULT_PATTERN) }
+
+    describe 'default matches common name variants' do
+      common_name_variants = [
+        'Bob Smith',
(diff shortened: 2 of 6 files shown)


app/models/user/anonymisable/name_pattern.rb
```diff
@@ -0,0 +1,58 @@
+# Split a name into parts to construct a pattern suitable for a Regexp
+# CensorRule#text
+class User::Anonymisable::NamePattern
+  DEFAULT_HONORIFICS = 'Mr|Mrs|Miss|Ms|Mx'
+  DEFAULT_PATTERN = '\b(?:(?:%{honorifics})\.?\s+(?:%{firstname}\s+)?%{surname}|%{firstname}(?:\s+%{surname}|\s+%{last_initial}\.?)?|%{first_initial}\.?\s*%{surname}|%{surname},\s+(?:%{firstname}|%{first_initial}\.?))(?!\w)'
```
gbp: Does this have to be a string? I would make a regex and use multiline extended mode to make this more readable.

Or if it does need to be a string use a heredoc

garethrees: Its input into `CensorRule#text` – might be able to set this as regexp and then convert back to text, for the `CensorRule to convert back to regexp again… didn't take that approach initially to avoid any gremlins through the conversion process.

garethrees: Added in 52eb801 but haven't had loads of time to test so if you could have a look at that a bit more thoroughly that would be great.

spec/models/user/anonymisable/name_pattern_spec.rb
```diff
@@ -0,0 +1,196 @@
...
+  describe '#last_initial' do
+    subject { described_class.new(name).last_initial }
+
+    context 'with a simple name' do
+      let(:name) { 'John Smith' }
+      it { is_expected.to eq('S') }
+    end
+
+    context 'with a multibyte first character in the surname' do
+      let(:name) { 'John Ångström' }
```
laurentS, line 143: Random edge case: does this work with a single letter family name? eg: https://en.wikipedia.org/wiki/C%C3%A9dric_O

spec/models/user/anonymisable/name_pattern_spec.rb
```diff
@@ -0,0 +1,196 @@
...
+        example.run
+        described_class.pattern = described_class::DEFAULT_PATTERN
+      end
+
+      it { is_expected.to eq('Bob B S Smith') }
+    end
+
+    context 'with custom honorifics' do
+      around do |example|
+        described_class.honorifics = 'Monsieur|M|Madame|Mme'
```
laurentS, line 198: I have not tested, but `M.` is a common abbreviation in French for "Monsieur". See for instance https://madada.fr/request/couts_de_remediation_des_inciden (first line of the response). This comment is about checking whether such an alternative would work in the pattern above

spec/models/user/anonymisable/name_pattern_spec.rb
```diff
@@ -0,0 +1,196 @@
...
+        'Bob Smith',
+        'Bob',
+        'B Smith',
+        'B. Smith',
+        'Bob S',
+        'Smith, Bob',
+        'Smith, B',
+        'Mr Smith',
+        'Mr. Smith',
+        'Mr B Smith',
```
laurentS, line 34: Would `Mr B. Smith` and variants make sense in this list (with a `.` after first name)? Or maybe it's a French thing?

spec/models/user/anonymisable/name_pattern_spec.rb
laurentS: I now see that it appears to be auto-added in the permutations listed above, if so, please ignore this comment.

laurentS (commented):
I have not tested the code, but reading through it I added a few edge cases that I'm familiar with, which could be worth testing for to confirm the regex generated will work.

spec/models/user/anonymisable/name_pattern_spec.rb
garethrees: Yeah that should work okay, providing you set the French honorifics in the theme as above, as "M" is not a usual English honorific and so not included in the default list.

spec/models/user/anonymisable/name_pattern_spec.rb
garethrees: Yep that's covered. I only added a test for `'B. Smith'` though (L28), so I've added more variants in d8c3f5405.

spec/models/user/anonymisable/name_pattern_spec.rb
garethrees: Yup! cac879349

app/models/user/anonymisable/name_pattern.rb
```diff
@@ -0,0 +1,76 @@
...
+  protected
+
+  attr_reader :name
+
+  private
+
+  def pattern_string
+    return pattern unless pattern.is_a?(Regexp)
+    return pattern.source unless (pattern.options & Regexp::EXTENDED).nonzero?
+    "(?x:#{pattern.source.squish.gsub(' ', '')})"
```
gbp: This changes the pattern source which is fine for the default but it could be customised and includes literal spaces.

gbp (approved):
This is looking alright and seems to work well.

Do we need to consider any middle names as these are silently dropped?

app/models/user/anonymisable/name_pattern.rb
garethrees: Hm, yeah. I did this because otherwise the censor rule is quite ugly / difficult to edit in the UI without it:

Screenshot 2026-05-18 at 15 06 56

All the matching tests pass if we _don't_ do this for the default, so could just live with the ugly rule in the UI as I expect no one's going to edit it anyway, unless you have a better idea?

gbp: Yes. Not too concerned about this but thought I would flag anyway. Would a quick fix be to make the input a text area?

garethrees:
> Do we need to consider any middle names

Vast majority of users have firstname surname, so I think will just merge this for now as its a huge step up in most cases. The distribution may differ in other countries though, so might come back to add an improvement at a later date if we've got time after other issues have been resolved.

```rb
User.active.group("array_length(string_to_array(trim(name), ' '), 1)").count.sort.to_h
# => {1 => 32623,
# =>  2 => 165652,
# =>  3 => 12568,
# =>  4 => 1638,
# =>  5 => 336,
# => …
```

app/models/user/anonymisable/name_pattern.rb
garethrees: Will just remove for now and see how we go; easy to change to textarea later if we feel we really need it

PR #9224, opened 2026-04-17
[#8822] Add CensorRule#diacritics_insensitive

### LINKED ISSUE #8822: Ensure redaction system is adaptable to international contexts

garethrees:
## Relevant issue(s)

Part of #8822 
Requires https://github.com/mysociety/alaveteli/pull/9209/

## What does this do?

Add CensorRule#diacritics_insensitive

## Why was this needed?

Annoying to have to enter multiple censor rules / create a regexp for common accented words.

## Implementation notes

Limitations:

* Only works when `CensorRule#regexp` is `false`. I did look at removing this constraint but it adds complexity and probably isn't a very typical case.
* Only handles single-letter diacritics (e.g. É -> E). Does not handle characters that would get mapped to multiple Ascii characters (e.g. æ -> ae).

## Screenshots

Screenshot 2026-04-17 at 17 15 37

Screenshot 2026-04-17 at 17 15 55

Screenshot 2026-04-17 at 17 16 00

### DIFF
--- a/app/models/censor_rule.rb
+++ b/app/models/censor_rule.rb
@@ -14,6 +14,7 @@
 #  updated_at        :datetime         not null
 #  regexp            :boolean          default(FALSE), not null
 #  case_sensitive    :boolean          default(TRUE), not null
+#  ignore_diacritics :boolean          default(FALSE), not null
 #
 
 # models/censor_rule.rb:
@@ -41,7 +42,11 @@ class CensorRule < ApplicationRecord
              inverse_of: :censor_rules,
              optional: true
 
-  validate :require_valid_regexp, if: -> { regexp? || !case_sensitive? }
+  validate :require_valid_regexp,
+           if: -> { regexp? || !case_sensitive? || ignore_diacritics? }
+
+  validate :regexp_ignore_diacritics,
+           if: -> { regexp? && ignore_diacritics? }
 
   validates_presence_of :text,
                         :replacement,
@@ -122,7 +127,7 @@ class CensorRule < ApplicationRecord
   end
 
   def to_replace(encoding)
-    if regexp? || !case_sensitive?
+    if regexp? || !case_sensitive? || ignore_diacritics?
       make_regexp(encoding)
     else
       encoded_text(encoding)
@@ -134,8 +139,14 @@ class CensorRule < ApplicationRecord
   end
 
   def make_regexp(encoding)
-    pattern = encoded_text(encoding)
-    pattern = Regexp.escape(pattern) unless regexp?
+    pattern =
+      if ignore_diacritics?
+        diacritic_expander.expand(encoded_text(encoding))
+      else
+        encoded_text(encoding)
+      end
+
+    pattern = Regexp.escape(pattern) unless regexp? || ignore_diacritics?
 
     ::Warning.with_raised_warnings do
       Regexp.new(pattern, regexp_options)
@@ -149,5 +160,14 @@ class CensorRule < ApplicationRecord
     options |= Regexp::IGNORECASE unless case_sensitive?
     options |= Regexp::MULTILINE if regexp?
     options
+  end
+
+  def diacritic_expander
+    DiacriticExpander.new(case_sensitive: case_sensitive?)
+  end
+
+  def regexp_ignore_diacritics
+    msg = 'Cannot use regexp and ignore diacritics option together'
+    errors.add(:text, msg)
   end
 end
--- a/doc/CHANGES.md
+++ b/doc/CHANGES.md
@@ -2,6 +2,7 @@
 
 ## Highlighted Features
 
+* Allow censor rules to ignore diacritics (Gareth Rees)
 * Allow censor rules to be case insensitive (Gareth Rees)
 * Add Content Security Policy with nonce-based script protection (Graeme
   Porteous)
@@ -114,6 +115,12 @@
   it by running:
 
     bin/rails runner "AlaveteliFeatures.backend.enable(:censor_rule_case_sensitive)"
+
+* _Optional:_ Censor rules can now ignore diacritics. This is disabled by
+  default while we beta test it before full release. Before then you can enable
+  it by running:
+
+    bin/rails runner "AlaveteliFeatures.backend.enable(:censor_rule_ignore_diacritics)"
 
 ### Changed Templates
 
--- a/spec/models/censor_rule_spec.rb
+++ b/spec/models/censor_rule_spec.rb
@@ -14,6 +14,7 @@
 #  updated_at        :datetime         not null
 #  regexp            :boolean          default(FALSE), not null
 #  case_sensitive    :boolean          default(TRUE), not null
+#  ignore_diacritics :boolean          default(FALSE), not null
 #
 
 require 'spec_helper'
@@ -98,6 +99,25 @@ RSpec.describe CensorRule do
       end
     end
 
+    context 'when sensitive to both diacritics and case' do
+      let(:rule) do
+        FactoryBot.build(
+          :censor_rule,
+          ignore_diacritics: false,
+          case_sensitive: true,
+          text: 'ecole'
+        )
+      end
+
+      it 'only matches the given text' do
+        expect(rule.apply_to_text('ecole text')).to eq('[REDACTED] text')
+        expect(rule.apply_to_text('école text')).to eq('école text')
+        expect(rule.apply_to_text('Ecole text')).to eq('Ecole text')
+        expect(rule.apply_to_text('ECOLE text')).to eq('ECOLE text')
+        expect(rule.apply_to_text('ÉCOLE text')).to eq('ÉCOLE text')
+      end
+    end
+
     context 'when case_sensitive is false with regexp rule' do
       let(:rule) do
         FactoryBot.build(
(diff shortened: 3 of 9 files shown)


garethrees:
@laurentS fancy taking a look at this one since it's more relevant to your install?

app/models/censor_rule/diacritic_folder.rb
```diff
@@ -0,0 +1,66 @@
...
+    's' => '[sśš]',      'S' => '[SŚŠ]',
+    'u' => '[uùúûüūů]',  'U' => '[UÙÚÛÜŪŮ]',
+    'y' => '[yýÿ]',      'Y' => '[YÝŸ]',
+    'z' => '[zźżž]',     'Z' => '[ZŹŻŽ]'
+  }.freeze
+
+  cattr_accessor :character_map, default: DEFAULT_CHARACTER_MAP
+
+  def initialize(case_insensitive: false)
+    @case_insensitive = case_insensitive
```
laurentS: It might be a personal preference, but in my mind `case_insensitive = false` is a double negative. I find `case_sensitive = true` to be easier to process.

spec/models/censor_rule_spec.rb
```diff
@@ -79,6 +81,72 @@
...
+      end
+
+      it 'does not match the opposite case' do
+        expect(rule.apply_to_text('Ecole text')).to eq('Ecole text')
+        expect(rule.apply_to_text('ECOLE text')).to eq('ECOLE text')
+        expect(rule.apply_to_text('ÉCOLE text')).to eq('ÉCOLE text')
+      end
+    end
+
+    context 'when diacritics_insensitive and case_insensitive are both true' do
```
laurentS: maybe worth adding a text block for when diacritics_insensitive and case_insensitive are both false, to confirm the folder does not match too much?

spec/models/censor_rule_spec.rb
```diff
@@ -79,6 +81,72 @@
...
+      let(:rule) do
+        FactoryBot.build(
+          :censor_rule,
+          :diacritics_insensitive,
+          :case_insensitive,
+          text: 'ecole'
```
laurentS, line 169: Also worth testing with a word that has more than a single diacritic, and not just in first position. `Maçã` (apple in portuguese) could do

app/models/censor_rule/diacritic_folder.rb
```diff
@@ -0,0 +1,66 @@
...
+    'e' => '[eèéêëēě]',  'E' => '[EÈÉÊËĒĚ]',
+    'i' => '[iìíîïī]',   'I' => '[IÌÍÎÏĪ]',
+    'l' => '[lłĺļľ]',    'L' => '[LŁĹĻĽ]',
+    'n' => '[nñńň]',     'N' => '[NÑŃŇ]',
+    'o' => '[oòóôõöøō]', 'O' => '[OÒÓÔÕÖØŌ]',
+    'r' => '[rŕř]',      'R' => '[RŔŘ]',
+    's' => '[sśš]',      'S' => '[SŚŠ]',
+    'u' => '[uùúûüūů]',  'U' => '[UÙÚÛÜŪŮ]',
+    'y' => '[yýÿ]',      'Y' => '[YÝŸ]',
+    'z' => '[zźżž]',     'Z' => '[ZŹŻŽ]'
```
laurentS:
```suggestion
    'y' => '[yýÿ]',      'Y' => '[YÝŸ]',
    'z' => '[zźżž]',     'Z' => '[ZŹŻŽ]',
    'œ' => '(œ|oe)',     'Œ' => '(Œ|OE)',
    'æ' => '(æ|ae)',     'Æ' => '(Æ|AE)'
```
I think your comment about this not working with multi-character replacements is incorrect.
With this change, I could do:
```ruby
=> cr = CensorRule.new(text: 'œuf', diacritics_insensitive: true, replacement: '[REDACTED]')
=> cr.apply_to_text("un œuf, des oeufs")
"un [REDACTED], des [REDACTED]s"
```
This is actually pretty useful for French, as there are a couple of very common words which *should* be spelled with `œ` (egg, heart) but most often aren't.

laurentS (commented):
This seems to work the way I'd expect it. See my comments about expanding its use case.
I think that for certain use cases, like `ö` that is sometimes spellt `oe` in German, this would not work well, but in such situations, a plain regex is probably a better solution.
It'd be useful to have users of languages that use a lot of diacritics (I'm thinking Czech for instance) test this out.

laurentS:
If forgot to mention, I played a bit with the censor rule on a 3kB long string (a paragraph from https://fr.wikipedia.org/wiki/%C5%92uf_(aliment) where I censored "œuf". It works as expected. If I multiply the string 1000 times to get a 3MB text, it still censors it pretty much instantly. However, if we need to instantiate a censor rule many times over, particularly when it's case insensitive, things get substantially slower (censoring the 3kb text 1000 times took about 0.5 seconds). I'm thinking of bulk search indexing scenarios where this might slow things down. Not a problem, but I'm leaving a trace here, just in case.

app/models/censor_rule/diacritic_folder.rb
garethrees: Oh nice! I didn't think about mapping the characters in that way.

garethrees: This does require the diacritic to be used in the censor rule text.

```rb
      # Success! ✅
      it 'handles multi-letter diacritics' do
        rule.text = 'œuf'
        expect(rule.apply_to_text('Un œuf, des oeufs')).
          to eq('Un [REDACTED], des [REDACTED]s')
      end

      # Failure ❌
      it 'handles multi-letter diacritics' do
        rule.text = 'oeuf'
        expect(rule.apply_to_text('Un œuf, des oeufs')).
          to eq('Un [REDACTED], des [REDACTED]s')
      end
```

Bit of a gotcha, but still probably better than not including?

garethrees:
Pushed commits to address notes above and also feature flagged so that we can enable on a per-site basis as we roll out.

laurentS:
looks good to me!

doc/CHANGES.md
```diff
@@ -65,6 +67,7 @@
       AtiNetworkController.showcase_enabled = false
     end
 
+<<<<<<< HEAD
```
gbp:
```suggestion
```

doc/CHANGES.md
```diff
@@ -86,6 +89,19 @@
       # Change only a default mask's replacement
       AlaveteliTextMasker.replace_mask(:mobile_number, replacement: '[cell number]')
     end
+=======
+* _Optional:_ Censor rules can now be made case insensitive. This is disabled by
+  default while we beta test it before full release. Before then you can enable
+  it by running:
+
+    bin/rails runner "AlaveteliFeatures.backend.enable(:censor_rule_case_sensitive)"
+>>>>>>> b8e52bc38 (Add CensorRule#case_sensitive)
```
gbp:
```suggestion
```

doc/CHANGES.md
```diff
@@ -86,6 +89,19 @@
       # Change only a default mask's replacement
       AlaveteliTextMasker.replace_mask(:mobile_number, replacement: '[cell number]')
     end
+=======
```
gbp:
```suggestion

```

app/models/censor_rule/diacritic_folder.rb
```diff
@@ -0,0 +1,68 @@
...
+
+  private
+
+  def fold_character(char)
+    map = character_map_case_insensitive || character_map
+
+    map[char] ||
+      map[ActiveSupport::Inflector.transliterate(char)] ||
+      char
+      #Regexp.escape(char)
```
gbp: I think it make sense to escape:

```suggestion
      Regexp.escape(char)
```

app/models/censor_rule/diacritic_folder.rb
```diff
@@ -0,0 +1,68 @@
+class CensorRule::DiacriticFolder
```
gbp: Not sure about the name of this. Doesn't it expand a character in to a character set for regexp

app/models/censor_rule.rb
```diff
@@ -2,17 +2,19 @@
...
+#  public_body_id         :integer
+#  text                   :text             not null
+#  replacement            :text             not null
+#  last_edit_editor       :string           not null
+#  last_edit_comment      :text             not null
+#  created_at             :datetime         not null
+#  updated_at             :datetime         not null
+#  regexp                 :boolean          default(FALSE), not null
+#  case_sensitive         :boolean          default(TRUE), not null
+#  ignore_diacritics      :boolean          default(FALSE), not null
```
gbp: When I migrate the DB I get changes here

gbp (changes requested):
Looks to work well. Some minor changes need to get this merged

app/models/censor_rule/diacritic_folder.rb
garethrees: Oops meant to remove the commented line. With escaping we get failures. We want the char as-is as we're constructing a string which then gets escaped fully in `CensorRule#make_regexp`.

```
Failures:

  1) CensorRule::DiacriticFolder#fold with case-insensitive mode ignores non-mapped characters the same as case-sensitive
     Failure/Error: expect(folder.fold('!@*')).to eq('!@*')

       expected: "!@*"
            got: "!@\\*"

       (compared using ==)
     # ./spec/models/censor_rule/diacritic_folder_spec.rb:99:in 'block (4 levels) in <top (required)>'

  2) CensorRule::DiacriticFolder#fold with default case-sensitive mode handles strings with spaces and punctuation
     Failure/Error: expect(folder.fold('café au lait')).to eq(expected)

       expected: "[cçćč][aàáâãäåā]f[eèéêëēě] [aàáâãäåā][uùúûüūů] [lłĺļľ][aàáâãäåā][iìíîïī]t"
            got: "[cçćč][aàáâãäåā]f[eèéêëēě]\\ [aàáâãäåā][uùúûüūů]\\ [lłĺļľ][aàáâãäåā][iìíîïī]t"

       (compared using ==)
     # ./spec/models/censor_rule/diacritic_folder_spec.rb:71:in 'block (4 levels) in <top (required)>'

  3) CensorRule::DiacriticFolder#fold with default case-sensitive mode ignores non-mapped characters
     Failure/Error: expect(folder.fold('!@*')).to eq('!@*')

       expected: "!@*"
            got: "!@\\*"

       (compared using ==)
     # ./spec/models/censor_rule/diacritic_folder_spec.rb:44:in 'block (4 levels) in <top (required)>'

Finished in 1.73 seconds (files took 2.78 seconds to load)
82 examples, 3 failures

Failed examples:

rspec ./spec/models/censor_rule/diacritic_folder_spec.rb:97 # CensorRule::DiacriticFolder#fold with case-insensitive mode ignores non-mapped characters the same as case-sensitive
rspec ./spec/models/censor_rule/diacritic_folder_spec.rb:65 # CensorRule::DiacriticFolder#fold with default case-sensitive mode handles strings with spaces and punctuation
rspec ./spec/models/censor_rule/diacritic_folder_spec.rb:42 # CensorRule::DiacriticFolder#fold with default case-sensitive mode ignores non-mapped characters

Randomized with seed 46847
```

app/models/censor_rule/diacritic_folder.rb
garethrees: Hm yeah that's true. I was thinking of it in the sense of it treats them as equivalent, but [looks like](https://docs.opensearch.org/latest/analyzers/token-filters/asciifolding/) folding in this context is mapping the diacritic to the non-diacritic version; as you say, we're doing the reverse here (even though the intent is similar).

PR #9234, opened 2026-04-28
Unify MIME detection replacing Mahoro with Marcel

gbp:
## What does this do?

Unify MIME detection replacing Mahoro with Marcel, which is already used by Active Storage internals. This eliminates the dual-library inconsistency where files could be classified differently by Mahoro (libmagic) vs Marcel.

Marcel returns clean MIME types without the charset suffixes and junk output that required workaround parsing with Mahoro.

One test expectation updated: an email attachment part with no filename that Mahoro detected as text/plain is now application/octet-stream, which is a more conservative and accurate result for unidentifiable content.


lib/alaveteli_file_types.rb
```diff
@@ -14,21 +14,24 @@ class AlaveteliFileTypes
     "xlsx" => 'application/vnd.openxmlformats-officedocument.spreadsheetml.sheet',
     "ppt" => 'application/vnd.ms-powerpoint',
     "pptx" => 'application/vnd.openxmlformats-officedocument.presentationml.presentation',
+    "eml" => 'message/rfc822',
```
garethrees, line 17: Should we add a `.email` alias as per the modified spec?

garethrees (commented):
had a quick look through and looks good in principle. happy if you are

Issue #6480, opened 2021-08-17
Attachments presented as TXT due to unexpected Content-Type


garethrees:
A [query](https://groups.google.com/a/mysociety.org/g/alaveteli-dev-private/c/n1bYBk-1xhI/m/ANx5y-o0AQAJ) came in recently about [a request](https://www.whatdotheyknow.com/request/collections_of_artwork_gifts_sil#incoming-1851657) where the authority had attached a CSV but Alaveteli is presenting it as `.csv.txt`.

Looking at the [raw email](https://www.whatdotheyknow.com/admin/raw_emails/1851505) they have indeed sent a .csv file, but (presumably) their mail client has incorrectly identified it as "application/octet-stream" rather than “text/csv”.

```
Content-Type: application/octet-stream; name="FOI-346070201 Collections.csv"
Content-Description: FOI-346070201 Collections.csv
Content-Disposition: attachment; filename="FOI-346070201 Collections.csv";
	size=17601763; creation-date="Tue, 10 Aug 2021 10:02:37 GMT";
	modification-date="Tue, 10 Aug 2021 10:09:52 GMT"
Content-Transfer-Encoding: base64
```

It looks like we attempt to [add an extension if it doesn’t match the content type](https://github.com/mysociety/alaveteli/blob/0.39.1.3/app/models/foi_attachment.rb#L240-L245), so what’s happening is that we’re identifying this as text/plain, storing that for the content_type, and as such, appending a .txt extension. I’m not actually sure how we’re getting "text/plain" here.

```sql
foi=> SELECT content_type, filename FROM foi_attachments WHERE id=3303691;
 content_type |             filename
--------------+-----------------------------------
 text/plain   | FOI-346070201 Collections.csv.txt
```

I don't think this is a _bug_ because we're doing exactly what we intended. I guess Alaveteli could be more intelligent with this kind of input and, in certain cases, prefer the file extension over the Content-Type header.

I’m not sure this is a huge problem since the CSV is still downloadable, but I guess a little confusing to users who aren’t particularly familiar with CSV.

garethrees:
Hmm so strangely on first attempt I couldn't replicate this in my development environment.

![Screenshot 2021-08-17 at 09 24 36](https://user-images.githubusercontent.com/282788/129691283-9dd7c08b-de50-4055-9762-c47f025b557c.png)

```sql
alaveteli_development=> SELECT content_type, filename FROM foi_attachments WHERE id=34;
  content_type   |           filename
[...]

Issue #8820, opened 2025-08-15
Improve text extraction from currently unindexed document types


garethrees:

garethrees:
A couple of notes from our meet up:

We might be able to leave this to the search index if it has document extraction capability; or could explore active storage variants as a way of storing the to-text extraction.

We already have libreoffice installed, and it looks promising for docx conversion if we continue to handle in the app: https://stackoverflow.com/a/12246397

```
libreoffice --headless --convert-to "txt:Text (encoded):UTF8" mydocument.doc
```

We currently do this extraction in `MailHandler` – the code around this should be extracted to something more directly related to the attachment.

laurentS:
# Findings so far

See [this spreadsheet](https://docs.google.com/spreadsheets/d/1GolE9-VV7rOQ01vBAv9pUZXPDNgBZYFM2qoaO4zNdp0/edit?gid=700018818#gid=700018818) for details.

I ran a few tools locally on a handful of documents taken from WDTK/Ma Dada. The full notebook is [here](https://codeberg.org/madada/content_extraction_evaluation).

Key take aways below:

## docling

- 🔴 very CPU/memory hungry. Extraction is 1-5 order of magnitude slower compared to the other tools below. Memory usage up to 100x higher than other tools as well.
- 🤷 while docling does better on some docs than competition, it does not seem to outperform them so much that it makes sense to spend all the extra compute
- 🔴 docling does not support old doc/xls/ppt files, which WDTK has quite a lot of
- 🐍 it is built with and for python, so interfacing from ruby is less trivial (although we are likely to run these tools over an HTTP API so this is less of a concern)
- at this point, I would discard `docling` as I don't see the 10x benefit is its output, thinking of the huge extra resource cost it comes with.

## tika

- 🦕 has been around for 20 years, was designed pre-AI craziness, so it focuses more on pulling out text, rather than trying to make sense of structure.
- 🧑‍🤝‍🧑 tika has a wide community, and is reused in a variety of major search engines (ElasticSearch, SOLR and more)
- 🟢 it supports all types of documents we need
- 🟠 it does not seem to split certain documents per page/sheet. Spreadsheets are extracted as a single string, with no clear marker between sheets (later tests seem to partially contradict this). PDFs are extracted as a single text string with character offsets of each page in a separate metadata entry, which allows figuring out where a piece of text appears, or splitting out content per page.
- 🟠 docs are "old school", aka not the easiest to navigate
[...]

PR #9235, opened 2026-04-28
[#8703/#8917] Fix attachment content type detection

### LINKED ISSUE #3604: External Command: Error from command "wvText…"
Lots of errors from `wvText` after kicking off reindexing all events in https://github.com/mysociety/whatdotheyknow-theme/issues/344.

Is it broken for old documents, or is it actually completely broken but we haven't noticed because we're infrequently running it?

- [ ] Mark https://github.com/mysociety/whatdotheyknow-theme/issues/344 as "new" once fixed.

### LINKED ISSUE #8703: Poorly encoded attachments rendering as text files
We’re seeing quite an uptick, on WDTK, of user support issues that are being caused by poorly encoded emails. A common factor seems to be a particular CRM system [^1].

The issue here is that **some** attachments, irrespective of file type, are being encoded with mime-type `text/plain`.

As an example [^2], one had the following:

```
Content-Type: text/plain; name="weee tonnages.xlsx"
Content-Description: weee tonnages.xlsx
Content-Disposition: attachment; filename="weee tonnages.xlsx"; size=18076;
	creation-date="Fri, 09 May 2025 15:26:39 GMT";
	modification-date="Fri, 09 May 2025 15:26:39 GMT"
Content-Transfer-Encoding: base64
```

Clearly, we should respect what is declared, by default, but would a secondary sense check help, so that we aren’t serving files with incorrect extensions[^3]?

Unfortunately, replacing the attachment using the new tooling does not rectify the issue.

[^1]: Verint / Lagan products, often on Exchange mailboxes - it’s unclear where the encoding is happening.

[^2]: there are several examples in the mailbox, search for Verint

[^3]: the files are fine if renamed, but that isn’t very accessible, not easily done on mobile devices

### LINKED ISSUE #8917: Attachment filenames lost when given only in content-type
The filename is being extracted by the attachment processing code - I think `part_filename` in `expand_and_normalize_parts` holds it at that point, but then the content type header is overwritten by the `normalise_content_type(get_content_type(part))` line which strips the filename, so then it is lost and the attachment is stored as `attachment.pdf`.

e.g.

```
> _mail = FoiAttachment.find(2095949).incoming_message.raw_email.mail!
> _mail.parts[1].filename
=> "CAS-183685_redacted.pdf"
> _mail.parts[1].content_type = MailHandler.normalise_content_type(MailHandler.get_content_type(_mail.parts[1]))
=> "application/pdf"
> _mail.parts[1].filename
=> nil
```

I think this can be fixed by changing `expand_and_normalize_parts` to not remove other parts of the content-type header or similar.

gbp:
## Relevant issue(s)

Fixes #8703
Fixes #8917
May fix #3604 (which would mean an improvement for #8820)

## What does this do?

Always use Marcel to detect MIME types from actual file content rather
than trusting the declared content type. Attachments can arrive with
incorrect types (e.g. xlsx files declared as text/plain).

Setting part.content_type strips the name parameter from the header,
losing filenames only specified in Content-Type and not in
Content-Disposition. Preserve the filename by re-adding the name
parameter after normalisation.

## Why was this needed?

Fixes issues with receive attachments from some authorities.


garethrees:
Also fixes https://github.com/mysociety/alaveteli/issues/6480?

Issue #9204, opened 2026-04-07
Remove commonlib


garethrees:
Placeholder.

Main todos:

* Inline ruby code to `./lib`
* Review other uses – probably install scripts; assume these can be inlined too
* Review any interaction with our deployment system; add wrappers if necessary
* Update documentation

laurentS:
If this means getting rid of the git submodules, I can only upvote the idea 👍

garethrees:
Yes!

PR #9236, opened 2026-04-28
Merge gh-pages under doc/manual

gbp:
## What does this do?

Moves GitHub pages to doc/manual.

## Why was this needed?

GitHub pages are forgotten and out-of-date. Having the source documentations along side the code will allow us to make changes while implementing features and changes.

[skip changelog]


garethrees:
I'm in favour of bringing the documentation closer to the source code, but not in this way. This isn't a top priority right now so sorry for the scrappy notes:

* Adding a submodule adds advanced git knowledge that I'd like to remove (i.e https://github.com/mysociety/alaveteli/issues/9204)
* There's a difference between the marketing site, code documentation, and the admin manual. I think each of these deserves a different approach:
  * Marketing site in its own repo (`mysociety/alaveteli.org` – a github pages site but based on the master branch)
  * Technical documentation: directly in the repo (maybe also with a generated site like https://api.rubyonrails.org/)
  * Admin manual: should be accessible within the app at `/admin/manual/*`

Hoping we can come to this in the near future as its included in a pitch we're waiting to hear back on.

gbp:
I disagree with closing this.

The distinction between marketing, technical docs, and the admin manual makes a lot of sense, and I agree they likely need different approaches long-term.

My main concern isn’t so much how we structure things eventually, but that right now the docs are effectively out of the development workflow and getting neglected as a result. Keeping them on `gh-pages` makes them easy to miss and harder to update, which is why they’ve drifted and not representing the current state of Alaveteli.

Even if the end goal is to split things out as you described, I still think there’s value in bringing the current docs in the same branch as the code also lowers the barrier for contributors to update them, rather than treating documentation as a separate task that can be forgotten and just to get them back into regular visibility and iteration.

We could treat that as a temporary step until we implement the more structured approach.

Also worth noting, other mySociety have this same approach:
https://github.com/mysociety/fixmystreet/tree/master/docs

garethrees:
Yeah ISWYM, easier to update the docs in the same PR at least.

I closed it as this work is included in a pitch we're waiting to hear back on and since we weren't agreed on approach didn't want this to detract from current projects.

What impact does this have on existing deployments wrt the submodule and reconfiguring the gh-pages setup? If its a zero/minimal-effort change then more open to just merging as is.

PR #9238, opened 2026-04-29
Replaced Twitter and Facebook with web share API inside `request-sent-message` 

### LINKED ISSUE #9165: Replace social share buttons with generic option
A continuation of https://github.com/mysociety/alaveteli/issues/8589.

When a request is created the user gets a share menu:

Image

We'll replace the buttons there with a larger version of the one implemented in https://github.com/mysociety/alaveteli/pull/9147. Same mechanics – OS share menu when supported with a fallback to copying the request URL if not.

Image

This partial lives in [`app/views/request/_request_sent.html.erb`](https://github.com/mysociety/alaveteli/blob/develop/app/views/request/_request_sent.html.erb).

There's also a [batch counterpart](https://github.com/mysociety/alaveteli/blob/develop/app/views/info_request_batch/_batch_sent.html.erb), which I don't think should be used any more, but might as well just copy the code from the individual request partial to here too.

Note that we'll also need to update our theme override with the same:

https://github.com/mysociety/whatdotheyknow-theme/blob/master/lib/views/request/_request_sent.html.erb#L19-L48

lucascumsille:
## Relevant issue(s)
Fixes: https://github.com/mysociety/alaveteli/issues/9165

## What does this do?

- Replaces Twitter and Facebook share with Web share API button
- Got rid of some extra unnecessary spacing inside the `request-sent-message` container
- Added share icon svg
- Added `button-share` styling

## Screenshots

https://github.com/user-attachments/assets/df0f5a7a-4d70-4e62-a0ad-585e63294f50

https://github.com/user-attachments/assets/c17b4c1d-0534-4bf5-8801-0a82590cbdb9

## Notes to reviewer

<hr>

[skip changelog]


gbp (approved):
Looks good

Issue #8814, opened 2025-08-15
Add search system to admin interfaces scoped to appropriate records


garethrees:
## Context

We have a variety of search / filtering options for various admin lists of records, all with a custom implementation for each record type (i.e. custom view, custom handling in controller, custom filtering/sorting, and sometimes a `.search` method on the class). Some records (e.g. InfoRequestBatch, Project, Notes, Categories, etc) don't have a search implementation.

![Image](https://github.com/user-attachments/assets/d038733a-a06b-4e9b-be05-904911b92b73)

## Limitations

This search is usually limited to some basic attributes of the record, not the full record contents – for example, we wouldn't be able to search for public bodies with "example.com" in their home page attribute, or filter requests by their classification status.

We also can't search/filter by content in associated records. This is most common where we want to find InfoRequests containing a term, but can't do so via the admin search since it doesn't search a request's outgoing or incoming messages. Another common example is wanting to search all records associated with a User (i.e. their user record attributes + requests, comments, tracks, etc).

There are also other admin-specific records that we want to be able to search – MailServerLogs, version history for versioned models, all attributes of InfoRequestEvent, and so on.

## Global Search

Ultimately admins should be able to search _all recorded information_ without developer intervention. That's the big goal.

This will likely mean we'll want a global search option to search all record types and intermingle them in the list of results:

<img width="635" height="597" alt="Image" src="https://github.com/user-attachments/assets/6418d3f9-6a5e-4d3a-867c-289eb94ee639" />

This feels like the trickiest part to implement well given the mass of data and record types that could potentially be returned, so this probably deserves extracting out to its own issue to think through the design.

Imagine we want to search for a User's name who's made very heavy site use – it may appear in a variety of outgoing messages, incoming messages, foi attachments, comments, public body change requests, event logs, etc.

This is _probably_ out of scope for this issue, but it might be worth considering a way to save a search and its results, and allow the admin to mark results as relevant/not relevant so that they can process the results bit by bit, and not end up with 100s of open tabs and trying to remember which results you want to act on and which were false positives:

<img width="581" height="452" alt="Image" src="https://github.com/user-attachments/assets/08beeeb8-3ebf-42cf-b4db-56abfa07e7a3" />

## Model Search

We'll want to keep our pattern of having a search bar on the index pages for records, where the search is scoped to just those models (and potentially their associations).

[...]

Issue #8812, opened 2025-08-15
Common search interface


garethrees:

garethrees:
@gbp I think we can consider this done? Can you link up any PRs that relate to this (either directly or preparation work). Thanks!

PR #9239, opened 2026-04-29
Setup search with postgresql

laurentS:
## Relevant issue(s)

#8812 #8813 #8814

## What does this do?

This PR sets the base for a new search system built into postgresql. It does not touch any of the existing search logic. It can be merged and deployed without impacting live deployments. In particular, no particular upgrades or extra packages are required.

## Why was this needed?

See issues above, particularly #8813 and related experiment PRs.

## Implementation notes

This lays the ground for a unified search interface done entirely in postgresql. This PR focuses mostly on full-text search (FTS), but adding in semantic search and other variants should be straightforward. It should hopefully not modify the API, so that the search system itself can be iterated on (or even replaced entirely) without affecting the rest of the codebase.

All search records are stored in a `SearchDocument` model which has a `searchable_doc` polymorphic relation to all searchable models. This allows writing code like is described in #8814, to search at various levels:

```ruby
# global search through all models
SearchDocument.search("something")

# search through public bodies
PublicBody.search("something else").where(...)

# search items related to a single object, optionally including connected
# items, like a user's requests (this part is not implemented yet)
User.find(123).search("Bob User", include_linked_items: true)
```

To separate admin accessible data, such as users' email addresses, or logs, from publicly visible content, the search index is split between `:index` and `:admin_index`. This allows calls like `SearchDocument.search("some user' name", admin_mode: true)` to return documents that should not be made public, while not impacting the regular search (for site visitors) in terms of speed.

To make a model searchable:
```ruby
class SomeModel < ApplicationRecord
  include Searchable

  searchable {
    index: {
      "name": "A",
      ".get_some_complex_attribute_in_ruby": "B"
    },
    admin_index: {...}
  }
```
where each index lists the table columns or methods (if prefixed with a `.`) to grab from the object. The `A|B|C|D` letters are the respective weight to assign to each (A being highest).

This way, anything that we can express in ruby can be indexed in search. Grabbing a table column is orders of magnitude faster when indexing, as we save a round trip to ruby.
It is theoretically possible to build postgresql stored functions to compute values inside the database and pass them to the indexing method like:
```ruby
searchable {
  index: {
    "calculate_stuff(name)": "A"
  }
}
```
however this would be a bit more complex for upgrading the code (a job would need to run when rails starts to update functions in the db). If certain fields become too slow to index, we might explore such options.

### Database internals

As all search documents will end up in the same table, partitioning in postgresql seemed like a helpful solution. This way, rails sees the db as having a single table, so that all SQL queries can be written like `SELECT * FROM search_documents WHERE searchable_doc_type='PublicBody'...`, and postgres will know to direct this query at the corresponding partition (ie. table+indices).
There is no need to manage a whole set of search document types.

The main benefit is with performance, as model specific searches only need to hit a single partition+index, meaning less load on the server, and faster results for users.

The code to handle the partitioning is surprisingly short: a few lines in the migration, and the primary key on `search_documents` must be named something else than `id`. Everything else is regular rails.

## Notes to reviewer

- [ ] hard CS problem number 2: naming things. (`search` is currently named `newsearch` to prevent collision with existing methods) Any suggestions around renaming things to make it more obvious what they do?
- [ ] Thoughts on code organisation are more than welcome. I'm not fluent enough in rails, so there is certainly a better way to locate methods, etc...
- [ ] I've tried to put ample comments in the code to serve as reference, but let me know if anything is unclear.
- [ ] It's worth thinking about which models and what fields need to be indexed. Changing fields in the future is easy (just change the definition under `searchable` and `reindex_all`), and adding more models to the search requires a migration to add the corresponding `search_documents_partition`. But for GDPR purposes, we probably want to include all possible PII locations.
- [ ] Please run `PublicBody.reindex_all` (or other models) on whatever dataset you have handy. The method can be interrupted at any time, as objects are indexed sequentially and written to db as they go. This permits running manual checks on a subset of objects.

## Scope limitations

In order for this PR to not go on endlessly, I've artificially drawn a line around the current scope. In particular, the following additions will come in subsequent PRs:
- any kind of UI (we intend to start with admin-facing search)
- searching a single object and its "dependents" (eg. everything a user has produced/participated in)
- more advanced search, such as facets, and the current search options such as "status:...", which should be straightforward to implement, as they will map onto ORM methods quite nicely
- work specifically on ranking results. This PR has mostly focused on making sure items get picked up by search.
- more test cases: I would like to collect common problems from site admins to add to the current test cases to ensure we don't break those. At this point, I'd consider an unsuccessful search as a bug (within reasonable limits). If you find some immediate obvious failures here, we can add to this PR.


app/models/incoming_message/main_body.rb
```diff
@@ -34,6 +34,20 @@ def get_main_body_text_internal # rubocop:disable Naming/AccessorMethodName
...
   end
 
+  # fetches the raw email and extracts text without applying masks nor
+  # censor rules.
+  def get_main_body_text_uncensored_for_indexing
+    parse_raw_email
+    main_part = get_main_body_text_part
+    MailHandler.get_attachment_text_one_file(
+      main_part.content_type, main_part.unmasked_body, main_part.charset
+    ).squeeze(' ')
```
garethrees: Now that https://github.com/mysociety/alaveteli/pull/9257 is merged this will be:

```rb
AttachmentToText.
  from_part(main_part, main_part.unmasked_body).
  to_text.
  squeeze(' ')
```

app/models/foi_attachment.rb
```diff
@@ -75,6 +91,63 @@ def delete_cached_file!
...
+  def is_indexable
+    # do not index images/videos as we can't extract text from them
+    # at this point
+    !(content_type.start_with?("image/") || content_type.start_with?("video/"))
+  end
+
+  def unredacted_content_for_indexing
+    MailHandler.get_attachment_text_one_file(
+      content_type, unmasked_body, charset
+    )
```
garethrees: Now that https://github.com/mysociety/alaveteli/pull/9257 is merged this will be:

```rb
AttachmentToText.
  from_string(unmasked_body, content_type: content_type).
  to_text
```

app/models/foi_attachment.rb
```diff
@@ -75,6 +91,63 @@ def delete_cached_file!
...
+           else
+             ''
+           end
+      [acc[0] + el, current]
+    end
+    out[0]
+  end
+
+  #
+  def redacted_content_for_indexing
```
garethrees: Think this method can be removed entirely now that https://github.com/mysociety/alaveteli/pull/9257 is merged and use `FoiAttachment#body_to_text` instead.

garethrees (commented):
Added a couple of notes on some changes needed now that https://github.com/mysociety/alaveteli/pull/9257 is merged.

I don't _think_ we need to pass the charset around but LMK if you think that causes any issues and we can add some additional handling.

app/models/incoming_message/main_body.rb
laurentS: This and below comments applied in the fixup commits about to be pushed.

garethrees: Great! Feel free to squash them once you're happy they work before requesting review

app/models/outgoing_message/snippet.rb
```diff
@@ -5,7 +5,7 @@
 #  id                          :bigint           not null, primary key
 #  created_at                  :datetime         not null
 #  updated_at                  :datetime         not null
-#  outgoing_message_snippet_id :bigint           not null
+#  outgoing_message_snippet_id :integer          not null
```
garethrees, line 8: why the change to integer here?

app/models/foi_attachment.rb
```diff
@@ -75,6 +91,56 @@ def delete_cached_file!
     file.purge_later if file.attached?
   end
 
+  def is_indexable?
+    # do not index images/videos as we can't extract text from them
+    # at this point
+    !(content_type.start_with?("image/") || content_type.start_with?("video/"))
```
garethrees: Appreciate this might be a total edge case but what if the filename contains PII?

app/models/outgoing_message.rb
```diff
@@ -83,6 +84,19 @@ class OutgoingMessage < ApplicationRecord
   scope :followup, -> { where(message_type: 'followup') }
   scope :is_searchable, -> { where(prominence: 'normal') }
 
+  searchable(
+    index: {
+      prominence_reason: "D",
+      ".get_text_for_indexing": "A",
+      # TODO: if the user's name is censored, from_name below still surfaces their
+      # request. should this move to admin_index? or should the name be censored too?
```
garethrees: `safe_from_name` is the one for public consumption

app/models/search_document.rb
```diff
@@ -0,0 +1,253 @@
...
+#  raw_admin_content   :text
+#  section_ref         :text
+#  language            :text
+#  content_tsv         :tsvector
+#  admin_content_tsv   :tsvector
+#  created_at          :datetime         not null
+#  updated_at          :datetime         not null
+#
+class SearchDocument < ApplicationRecord
+  belongs_to :searchable_doc, polymorphic: true
```
garethrees: Would call this `searchable` (with the column names becoming `searchable_type`, `searchable_id`).

This would be good to resolve before merge as it would be more of a pain to change afterwards.

app/models/concerns/searchable.rb
```diff
@@ -0,0 +1,258 @@
+# Define search methods common to all searchable models
```
garethrees, line 1: Think I'd put this in `app/concerns/searchable.rb` unless there's a good reason not to?

db/migrate/20260422143826_add_search_document.rb
```diff
@@ -0,0 +1,145 @@
...
+      force: :cascade,
+      options: "PARTITION BY LIST (searchable_doc_type)"
+    ) do |t|
+      t.bigserial(:sd_id)
+      t.belongs_to(:searchable_doc, polymorphic: true)
+      # plain text content of the related document.
+      # This might come from a text extraction tool (for attachments),
+      # but might also be a direct copy of admin_index fields to allow
+      # admins to do an exact text search.
+      t.text(:raw_content)
```
garethrees, line 41: Sorry I know we've discussed this (probably multiple times!), but could you remind me why we're storing both `raw_content` and `content_tsv` on `SearchDocument` records?

I know that the latter is the search-optimised version (stemmed, etc), but I can't remember why we aren't just transforming the raw content from the associated `searchable` record at index time and only persisting that output.

I remember we were at some point discussing that a postgres trigger/stored function thing could automatically do …something… when the `SearchDocument` is modified. Is it this? If so, I assume this approach wouldn't work (or would be much more complicated) if the raw content wasn't in the same table? If I _have_ remembered this correctly, where does that automated transformation happen?

garethrees (commented):
Couple of initial comments/questions after having a read through this afternoon.

app/models/outgoing_message/snippet.rb
laurentS: I have no idea, I missed this. Looking at the code, I don't even understand why this line is there at all. The column does not exist in the db, so I don't know what this refers to.

garethrees: Hm yeah it only seems to be the schema annotation rather than any actual migration change. Does look like that there was a change at some point in the past https://github.com/mysociety/alaveteli/commit/5c2cbdb0c58302c2c65a21264b9f4ce17831380c, so maybe its a hangover from that. No big deal – we'll just remove these before merge

app/models/concerns/searchable.rb
laurentS: No good reason. I put it there because I saw xapian code was under `/lib/`, but I now see that concerns seem to be the right place for shared behaviour between models. File moved.

app/models/search_document.rb
laurentS: works for me! I renamed it.

app/models/foi_attachment.rb
laurentS: Good point, corrected!

db/migrate/20260422143826_add_search_document.rb
laurentS: It's a good question. I pushed a little doc (under `doc/SEARCH.md`) that hopefully explains this. I think it's better to have it there than scattered around comments in github. Feel free to point out anything unclear in there.

PR #9247, opened 2026-05-05
Fix missing render post sending batch

gbp:
The 'batch_sent' partial isn't rendered after a batch is been created and the user redirected to `InfoRequestBatchController#show`. This is due to the flash variable not being set.


garethrees (commented):
Not sure to what extent the omission was intentional – I wonder if it was due to the presence of ["share" messaging](https://github.com/mysociety/alaveteli/blob/develop/app/views/info_request_batch/_batch_sent.html.erb#L21-L30), which doesn't make sense in the context of a private batch.

Could wrap that in a conditional though, as it's helpful that the partial explains that there's a lag to getting them sent.

PR #9249, opened 2026-05-07
Cache request count on frontpage

chrismytton:
The `InfoRequest.is_searchable.count` call takes ~600ms on the WDTK DB, so caching the result for an hour should speed up the frontpage load time a bit.

This does mean that the request count on the homepage will no longer be "realtime", but seems like a reasonable trade-off for the speed gain.

### DIFF
--- a/doc/CHANGES.md
+++ b/doc/CHANGES.md
@@ -2,6 +2,7 @@
 
 ## Highlighted Features
 
+* Cache total requests count on the front page for 1 hour (Chris Mytton)
 * Fix email change confirmation which wasn't bound to confirmed target address
   (Graeme Porteous)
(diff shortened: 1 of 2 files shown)


doc/CHANGES.md
```diff
@@ -38,6 +38,7 @@
 * Drop support for Ruby 3.3 (Graeme Porteous)
 * Drop support for Ruby 3.2 (Graeme Porteous)
 * Block various action links via robots.txt (Laurent Savaete)
+* Cache total requests count on the front page for 1 hour (Chris Mytton)
```
gbp: FYI, we put new entries to the changelog at the top of the list.

gbp (approved):
This looks fine. I have no issue with this not being realtime.

PR #9250, opened 2026-05-07
Cache recent requests on frontpage

chrismytton:
The call to `InfoRequest.recent_requests` takes 2-3 seconds on WDTK, so caching it should speed up the frontpage load times.

Note that only the InfoRequestEvent IDs are stored in the cache, then the actual events are retrieved from the database each time so we don't cache stale objects. This also allows us to filter out any requests that may have been hidden since the ID was cached.

### DIFF
--- a/app/controllers/general_controller.rb
+++ b/app/controllers/general_controller.rb
@@ -17,7 +17,7 @@ class GeneralController < ApplicationController
     medium_cache
     @locale = AlaveteliLocalization.locale
     successful_query = InfoRequestEvent.make_query_from_params( latest_status: ['successful'] )
-    @request_events, @request_events_all_successful = InfoRequest.recent_requests
+    @request_events = InfoRequest.recent_requests
     @track_thing = TrackThing.create_track_for_search_query(successful_query)
     @number_of_requests = Rails.cache.fetch(
       'frontpage/info_request_count', expires_in: 1.hour
--- a/app/controllers/general_controller.rb
+++ b/app/controllers/general_controller.rb
@@ -17,7 +17,7 @@ class GeneralController < ApplicationController
     medium_cache
     @locale = AlaveteliLocalization.locale
     successful_query = InfoRequestEvent.make_query_from_params( latest_status: ['successful'] )
-    @request_events = InfoRequest.recent_requests
+    @request_events = recent_request_events
     @track_thing = TrackThing.create_track_for_search_query(successful_query)
     @number_of_requests = Rails.cache.fetch(
       'frontpage/info_request_count', expires_in: 1.hour
@@ -195,5 +195,20 @@ class GeneralController < ApplicationController
     if feature_enabled?(:alaveteli_pro) && current_user && current_user.is_pro?
       redirect_to alaveteli_pro_dashboard_path
     end
+  end
+
+  def recent_request_events
+    ids = Rails.cache.fetch(
+      'frontpage/recent_requests', expires_in: 10.minutes, skip_nil: true
+    ) do
+      events = InfoRequest.recent_requests
+      events.empty? ? nil : events.map(&:id)
+    end || []
+
+    InfoRequestEvent.where(id: ids).
+      joins(:info_request).merge(InfoRequest.is_searchable).
+      includes(info_request: :public_body).
+      order(created_at: :desc).
+      load
   end
 end
--- a/app/controllers/general_controller.rb
+++ b/app/controllers/general_controller.rb
@@ -17,7 +17,8 @@ class GeneralController < ApplicationController
     medium_cache
     @locale = AlaveteliLocalization.locale
     successful_query = InfoRequestEvent.make_query_from_params( latest_status: ['successful'] )
-    @request_events, @request_events_all_successful = recent_request_events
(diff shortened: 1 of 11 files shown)


app/controllers/general_controller.rb
```diff
@@ -188,4 +188,20 @@ def redirect_pros_to_dashboard
       redirect_to alaveteli_pro_dashboard_path
     end
   end
+
+  def recent_request_events
+    ids, all_successful = Rails.cache.fetch(
+      'frontpage/recent_requests', expires_in: 10.minutes, skip_nil: true
+    ) do
+      events, all_successful = InfoRequest.recent_requests
+      events.empty? ? nil : [events.map(&:id), all_successful]
```
gbp:
```suggestion
      events.empty? ? nil : [events.ids, all_successful]
```

spec/controllers/general_controller_spec.rb
```diff
@@ -196,6 +196,39 @@
...
+      visible_event = FactoryBot.create(:info_request_event)
+      hidden_event = FactoryBot.create(:info_request_event)
+      hidden_event.info_request.update!(prominence: 'hidden')
+
+      allow(InfoRequest).to receive(:recent_requests).
+        and_return([[visible_event, hidden_event], true])
+
+      get :frontpage
+
+      assigned_ids = assigns[:request_events].map(&:id)
```
gbp:
```suggestion
      assigned_ids = assigns[:request_events].ids
```

chrismytton:
Just force-pushing to move the changelog entry to the top of the list (and rebase onto latest `develop`).

app/controllers/general_controller.rb
gbp: Why return `nil` maybe we can be more explicit.

```suggestion
      return nil if events.empty? # skip cache
      
      [events.ids, all_successful]
```

gbp (commented):
What an annoying method to cache.  This seems to work there are a few things to consider/improve.

Also wondering if we really care about `@request_events_all_successful`/`all_successful` boolean. Would be simpler if we didn't have that.

chrismytton:
> Also wondering if we really care about `@request_events_all_successful`/`all_successful` boolean. Would be simpler if we didn't have that.

I've pushed a fixup commit to drop the `all_successful` boolean from the cached path so `InfoRequest.recent_requests` now just returns the events array, and the cache stores only event IDs.

`@request_events_all_successful` is derived from the rehydrated events in the controller, so the template heading/copy still switches, but we no longer have to carry the `all_successful` boolean through the model -> cache -> view layers.

As a bonus it's actually more accurate than the old way. Previously all_successful only meant "the first Xapian query alone filled all 5 slots", now it reflects whether the requests actually shown are successful/partially-successful, after hydration and checks.

app/controllers/general_controller.rb
chrismytton: Unfortunately `InfoRequest.recent_requests` returns a plain Array, not an AR relation, so there's no `#ids` method.

chrismytton: I think that would return from the outer method, rather than just returning from the block. Guess we could do `next nil`, but now that I've dropped the all_successful parameter I think its ok as a ternery.

app/controllers/general_controller.rb
```diff
@@ -186,4 +187,27 @@ def redirect_pros_to_dashboard
...
     end
   end
+
+  def recent_request_events
+    ids = Rails.cache.fetch(
+      'frontpage/recent_requests', expires_in: 10.minutes, skip_nil: true
+    ) do
+      events = InfoRequest.recent_requests
+      events.empty? ? nil : events.map(&:id)
+    end || []
```
laurentS, line 206: I stumbled on this PR just now. I feel like there's another solution to your problem: The code in `InfoRequest.recent_requests` is calling xapian for something which is really SQL (there's no actual search, it's just looking at timestamps). Thinking that we're also working on getting rid of xapian, maybe this would be helpful instead:
```ruby
InfoRequestEvent.where(event_type: 'response').
				order(created_at: :desc).
				limit(5).
				joins(:info_request).
				merge(InfoRequest.is_searchable).
				where(calculated_state: [:successful, :partially_successful]).
       			includes(info_request: :public_body).
       			order(created_at: :desc).
      			to_a
```
It skips the "sent" event type, because outside of "newborn" sites, that code path is reasonably never going to be called.

Also, this query isn't incredibly fast but it can be made a lot faster with the right index (and be cached anyways, given that it's the landing page, I feel like it makes sense).

laurentS (commented):
Sorry for barging in uninvited on this PR while looking for something else. I hope the comment below is useful 🤞

app/controllers/general_controller.rb
gbp: Agree, and that's where this is heading. Request listings will be part of the PostgreSQL search work.

I'm keeping this PR to the caching so it can land on its own. Once the query is SQL and indexed we can look again at whether the cache is necessary.

gbp:
Heya @chrismytton I pushed 7a074328ad to split out the all successful flag from the caching. Once this is green I'll merge

PR #9255, opened 2026-05-12
[#7966] Fix indexing of multiple filetype values

### LINKED ISSUE #7966: Searching by filetype term only works when incoming message contains a single attachment
At the time of writing we have 11 attachments where we've identified the content type as mp4:

```rb
videos = FoiAttachment.where(content_type: 'video/mp4')
videos.count
# => 11
```

An [advanced search for `filetype:mp4`](https://www.whatdotheyknow.com/search/filetype:mp4/all) returns only one request.

Screenshot 2023-10-19 at 11 58 05

I suspect this is because the filetypes of all attachments belonging to the message are concatenated to be [indexed as a _term_](https://github.com/mysociety/alaveteli/blob/5945bbfb6a55e5a5b4a728de6ab7af87b2c5d8bd/app/models/info_request_event.rb#L133). Presumably Xapian only finds exact matches for terms, whereas in _texts_ it looks for any occurance of a given search string?

```rb
e = videos.last.incoming_message.info_request_events.first
e.send(:filetype)
# => "pdf MP4 jpg"
```

Would need to check this, but the single result findable through advanced search above has only a single attachment:

Screenshot 2023-10-19 at 11 59 00

garethrees:
## Relevant issue(s)

Fixes #7966 
Part of #8820

## What does this do?

Prior to this commit the filetypes value was a space-separated string, meaning that IncomingMessages with multiple attachments could not be searched using the `filetype:` prefix. This is due to a search of `filetype:pdf` not being an exact match of "pdf csv html" or whatever was indexed.

We must supply an array of values to the acts_as_xapian term so that Xapian can match against any of the filetypes present.

## Why was this needed?

Can only search attachments by filetype if only a single attachment type exists on a message.

## Implementation notes

Note that this will only correctly index messages received from this point on. Older messages will need to be reindexed manually.

## Notes to reviewer

Need to figure out a reindexing strategy (if we want to bother, given search engine updates) for the changelog. In most cases it would probably be okay to just query for affected events and mark them for reindexing, but WDTK scale is obviously more challenging. Open to suggestion.

```rb
# WDTK:
InfoRequestEvent.
  where(event_type: 'response').
  joins(incoming_message: :foi_attachments).
  where.not(foi_attachments: { url_part_number: 1 }). # Exclude main body part
  group('info_request_events.id').
  having('COUNT(DISTINCT foi_attachments.content_type) > 1'). # Problem occurs only when multiple attachment filetypes are present
  count.
  size
# => 994253
```

## Screenshots

Screenshot 2026-05-12 at 16 55 17

Screenshot 2026-05-12 at 16 55 23

### DIFF
--- a/app/models/incoming_message/attachments.rb
+++ b/app/models/incoming_message/attachments.rb
@@ -121,11 +121,12 @@ module IncomingMessage::Attachments
     cached_attachment_text_clipped
   end
 
-  # Returns space separated list of file extensions of attachments to this
-  # message. Defaults to the normal extension for known mime type, otherwise
-  # uses other extensions.
+  # Returns an Array of file extensions of attachments to this message.
+  # Defaults to the normal extension for known mime type, otherwise uses other
+  # extensions.
   def get_present_file_extensions # rubocop:disable Naming/AccessorMethodName
     ret = {}
+
     get_attachments_for_display.each do |attachment|
       ext = AlaveteliFileTypes.mimetype_to_extension(attachment.content_type)
       if ext.nil? && !attachment.filename.nil?
@@ -133,7 +134,8 @@ module IncomingMessage::Attachments
       end
       ret[ext] = 1 unless ext.nil?
     end
-    ret.keys.join(" ")
+
+    ret.keys
   end
 
   def locked?
--- a/app/models/info_request_event.rb
+++ b/app/models/info_request_event.rb
@@ -612,7 +612,7 @@ class InfoRequestEvent < ApplicationRecord
 
       incoming_message.get_present_file_extensions
     else
-      ''
+      []
     end
   end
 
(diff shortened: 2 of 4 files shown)


garethrees:
Decided on call not to bother with reindexing older stuff for now; will deal with when we move to postgres search.

gbp (approved):
Nice fix

gbp:
Had a go at writing a script to update "T" terms

```ruby
require 'xapian'

ActsAsXapian.prepare_environment

db = Xapian::WritableDatabase.new(ActsAsXapian.db_path, Xapian::DB_OPEN)
db.allterms('T').each do |term_item|
  term = term_item.term
  value = term[1..]
  next unless value.include?(' ')

  db.postlist(term).each do |posting|
    doc = db.document(posting.docid)
    doc.remove_term(term)
    value.split(' ').each { |v| doc.add_term("T#{v}") }
    db.replace_document(posting.docid, doc)
  end
end

db.commit
db.close
```

Issue #9241, opened 2026-04-30
Authenticator 2FA


garethrees:
# Alaveteli: Authenticator 2FA

## Goals

Critical

* Replace the current HOTP implementation with a more standard authenticator app-based TOTP workflow  
* Make it possible to technically enforce 2FA as a requirement for accounts with admin roles

Desirable

* Require 2FA on sign in when it is enabled for an account

## Context

Alaveteli has an opt-in, counter-based [2FA implementation](https://github.com/mysociety/alaveteli/pull/2877). When enabled, users must supply a one-time passcode (OTP) in addition to the normal email confirmation token when changing their password. No other actions, including sign in, require an OTP.

The current implementation was added at a time where smartphones/authenticator apps were not as prevalent, especially in the Global South where our efforts were focused at the time. SMS-based tokens were ruled out due to infrastructure cost and complexity. In 2026, this approach has a confusing user experience compared to the majority of implementations on other sites.

Users are required to store their OTP to use in the event of a password reset. They must then store a new OTP after it has been used. This has caught several users out, as there are no backup/recovery codes. If a user loses their passcode they can regenerate it while logged in, but if they are unable to log in they will be unable to access their account. There have been a few occasions where it's been difficult for administrators to [verify a user’s identity](https://wdtkwiki.mysociety.org/wiki/Resetting_user_passwords) to restore their access after the loss of an OTP.

WhatDoTheyKnow site administrators are [required by policy](https://wdtkwiki.mysociety.org/wiki/Password_policy) to have 2FA enabled, though this is not enforced by technical means. Only 497 of WhatDoTheyKnow’s 215,000+ users have 2FA enabled. There are a handful of user accounts with 2FA enabled across the international installs that we host.

## Current Implementation

2FA is controlled by a single [configuration option](https://alaveteli.org/docs/customising/config/#enable_two_factor_auth):

```
# config/general.yml
ENABLE_TWO_FACTOR_AUTH: false   # set to true to enable
```

When the flag is `false` (the default), all 2FA-related routes return 404 and the feature is completely inert regardless of what is stored in the database.

### Algorithm

Alaveteli uses **counter-based HOTP** (HMAC-based One-Time Password) via the `active_model_otp` gem (backed by `rotp`). Unlike time-based TOTP, a counter-based code does not expire after 30 seconds – it is valid until it is used. Each successful verification increments the counter and generates a new code, so every code is single-use.

### Storage

Three columns are added to the `users` table:

[...]

PR #9256, opened 2026-05-13
Add 2FA admin section

chrismytton:
Gives admins an overview of how many users have no 2FA vs HOTP vs TOTP, and also gives a list of HOTP users with how recently they've signed in. Should be useful to monitor the HOTP -> TOTP migration.

Part of https://github.com/mysociety/alaveteli/issues/9241

image

### DIFF
--- a/app/controllers/admin/two_factor_controller.rb
+++ b/app/controllers/admin/two_factor_controller.rb
@@ -1,0 +1,20 @@
+# Admin-only summary of 2FA enrolment state across the user base.
+class Admin::TwoFactorController < AdminController
+  def show
+    @all_counts = counts_for(User)
+    @admin_counts = counts_for(User.with_role(:admin))
+    @hotp_users = User.with_hotp.
+      order(User.arel_table[:last_sign_in_at].desc.nulls_last)
+  end
+
+  private
+
+  def counts_for(scope)
+    {
+      total: scope.count,
+      none: scope.without_two_factor.count,
+      hotp: scope.with_hotp.count,
+      totp: scope.with_totp.count
+    }
+  end
+end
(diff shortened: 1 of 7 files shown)


app/controllers/admin/two_factor_controller.rb
```diff
@@ -0,0 +1,20 @@
+# Admin-only summary of 2FA enrolment state across the user base.
+class Admin::TwoFactorController < AdminController
+  def show
+    @all_counts = counts_for(User)
+    @admin_counts = counts_for(User.with_role(:admin))
+    @hotp_users = User.with_hotp.
+      order(Arel.sql('last_sign_in_at DESC NULLS LAST'))
```
gbp: I guess you're trying to avoid passing raw SQL into `order`, if we're going to use Arel then we could go all the way...

I would either:

```suggestion
      order(User.arel_table[:last_sign_in_at].desc.nulls_last)
```

or accept the fact we sometimes need raw SQL

```suggestion
      order('last_sign_in_at DESC NULLS LAST')
```

gbp (approved):
Nice. Should be useful for migration. One minor comment.

app/views/admin_general/_admin_navbar.html.erb
```diff
@@ -50,6 +50,7 @@
               <li><%= link_to 'Users', admin_users_path %></li>
               <li><%= link_to 'Tracks', admin_tracks_path %></li>
               <li><%= link_to 'Tags', admin_tags_path(model_type: 'User') %></li>
+              <li><%= link_to 'Two factor', admin_two_factor_path %></li>
```
garethrees, line 71: I'd add a [divider](https://getbootstrap.com/2.2.0/components.html#dropdowns) (`<li class="divider"></li>`) above this as its more of a secondary utility rather than core functionality for that record type.

Image

chrismytton:
Have folded the fixups into the original commit.

chrismytton:

Issue #1528, opened 2014-05-22
Improve/fix HTML rendering of tables


TomSteinberg:
e.g  http://www.whatdotheyknow.com/request/it_support_services_1295#incoming-258044
 http://www.whatdotheyknow.com/request/it_support_services_347#incoming-258014
 http://www.whatdotheyknow.com/request/it_support_services_1236#incoming-257000

taken from #18 by @hsenag

hsenag:
Another example: https://www.whatdotheyknow.com/request/mental_health_services_4#incoming-601920

The original email has a nice looking HTML table.

tomchance:
I would really welcome this fix/improvement. I submit a fair few FOI requests where I end up with incomprehensible tables. For example:

https://www.whatdotheyknow.com/request/homeless_due_to_end_of_private_t_19#incoming-713565

As a result I usually write a reply asking that they email me the contents directly, which means others can't access the data if they happen across the request on the web site.

RichardTaylor:
Another example:

https://www.whatdotheyknow.com/request/number_children_home_educated_wi

What WhatDoTheyKnow shows:

> Summary of Elective Home Education Cases

>Academic Year

>2014/15 2015/16 2016/17

>SEN All From From All From From All From From
>status Cases mainstream Special Cases mainstream Special Cases mainstream Special
>Statement 10 <10 < 5 11 6 <5 9 5 <5
>(1)
>EHCP (2) 6 <5 < 5 6 6 0 8 7 0
>SEN
>Support 67 63 0 81 77 0 102 99 0
>(3)
>All SEN 83 72 < 5 98 89 <5 119 111 <5
>(4, 5)

Table in HTML:

<img width="702" alt="screen shot 2017-07-24 at 13 56 24" src="https://user-images.githubusercontent.com/547695/28524680-6b86f568-7079-11e7-8ef0-b51adddaf08a.png">

In this case we offered the user an image of the table and may upload the HTML version of the response and link it from an annotation:

https://twitter.com/WhatDoTheyKnow/status/889469208383418368

RichardTaylor:
Adding another example where a table provided in HTML format wasn't legible
https://www.whatdotheyknow.com/request/housing_register_8?unfold=1#incoming-1135657
in that case we put the tables in a Google spreadsheet and linked to it. 

Also noting #4003 is a related, broader, ticket for preserving all HTML formatting in response emails.

RichardTaylor:
Another example:
[...]

PR #9257, opened 2026-05-14
[#8820] Improve search text extraction from Microsoft Office attachments

### LINKED ISSUE #8820: Improve text extraction from currently unindexed document types

garethrees:
## Relevant issue(s)

Fixes #8820

## What does this do?

* Refactor search text extraction to `AttachmentToText`, making it easier for themes to customise content-type specific extraction processes
* Improve / add missing search text extraction from Microsoft Office attachments
* Improve search text extraction of RTF attachments
* Add search text extraction of CSV attachments

## Why was this needed?

* Helps us find PII in attachments to better respond to GDPR requests
* Improve user experience in helping them find relevant information

## Implementation notes

* Existing attachments (via their parent incoming message) will need to be reindexed to benefit.
* Would be nice in future to move the content-type specific methods into classes (e.g. `AttachmentToText::Adapters::PDF`) and have a more configurable mapping between content-type and extractor, but this will do for now.

## Notes to reviewer

* Have manually tested with a sample doc, docx, xls, xlsx, ppt, pptx, csv and rtf and all have extracted cleanly / better than prior to this PR.
- [ ] Will need packages updated in puppet before rollout to hosted sites, including WDTK


app/models/foi_attachment/content_type.rb
```diff
@@ -0,0 +1,48 @@
...
+  CONTENT_TYPE_NAMES = {
+    # Plain Text
+    "text/plain" => 'Text file',
+    'application/rtf' => 'RTF file',
+
+    # Binary Documents
+    'application/pdf' => 'PDF file',
+
+    # Images
+    'image/tiff' => 'TIFF image',
```
laurentS, line 15: Any reason not to include other image types? (from our db)
Also, since these names are displayed, it would be nice if they could be translated.

```suggestion
    'image/tiff' => _('TIFF image'),
    'image/bmp' => _('BMP image'),
    'image/gif' => _('GIF image'),
    'image/jpeg' => _('JPEG image'),
    'image/jpg' => _('JPEG image'),
    'image/png' => _('PNG image'),
    'image/svg+xml' => _('SVG image'),
    'image/tiff' => _('TIFF image'),
    'image/webp' => _('WebP image'),
    'image/x-png' => _('PNG image'),
```

app/models/foi_attachment/content_type.rb
```diff
@@ -0,0 +1,48 @@
...
+    # Word Processing
+    'application/vnd.ms-word' => 'Word document',
+    'application/vnd.openxmlformats-officedocument.wordprocessingml.document' => 'Word document',
+
+    # Presentation
+    'application/vnd.ms-powerpoint' => 'PowerPoint presentation',
+    'application/vnd.openxmlformats-officedocument.presentationml.presentation' => 'PowerPoint presentation',
+
+    # Spreadsheet
+    'application/vnd.ms-excel' => 'Excel spreadsheet',
```
laurentS, line 26: I also find the following weird types, but they obviously refer to the same thing:

```suggestion
    'application/vnd.ms-excel' => 'Excel spreadsheet',
    'application/vnd.ms-excel.sheet.binary.macroenabled.12',
    'application/vnd.ms-excel.sheet.macroenabled.12',
```

app/models/foi_attachment/content_type.rb
```diff
@@ -0,0 +1,48 @@
...
+    # Images
+    'image/tiff' => 'TIFF image',
+
+    # Word Processing
+    'application/vnd.ms-word' => 'Word document',
+    'application/vnd.openxmlformats-officedocument.wordprocessingml.document' => 'Word document',
+
+    # Presentation
+    'application/vnd.ms-powerpoint' => 'PowerPoint presentation',
+    'application/vnd.openxmlformats-officedocument.presentationml.presentation' => 'PowerPoint presentation',
```
laurentS, line 23:
```suggestion
    'application/vnd.ms-powerpoint' => _('PowerPoint presentation'),
    'application/vnd.openxmlformats-officedocument.presentationml.presentation' => _('PowerPoint presentation'),
    'application/vnd.oasis.opendocument.presentation' _('OpenDocument presentation'),
```

app/models/foi_attachment/content_type.rb
```diff
@@ -0,0 +1,48 @@
...
+
+    # Spreadsheet
+    'application/vnd.ms-excel' => 'Excel spreadsheet',
+    'application/vnd.openxmlformats-officedocument.spreadsheetml.sheet' => 'Excel spreadsheet'
```
laurentS, line 27:
```suggestion
    'application/vnd.ms-excel' => _('Excel spreadsheet'),
    'application/vnd.openxmlformats-officedocument.spreadsheetml.sheet' => _('Excel spreadsheet'),
    'application/vnd.oasis.opendocument.spreadsheet' => _('OpenDocument spreadsheet')
```

app/models/foi_attachment/content_type.rb
```diff
@@ -0,0 +1,48 @@
...
+    'application/rtf' => 'RTF file',
+
+    # Binary Documents
+    'application/pdf' => 'PDF file',
+
+    # Images
+    'image/tiff' => 'TIFF image',
+
+    # Word Processing
+    'application/vnd.ms-word' => 'Word document',
```
laurentS, line 18:
```suggestion
    'application/vnd.ms-word' => _('Word document'),
    'application/vnd.ms-word.document.macroenabled.12' => _('Word document'),
```
(we have this weird type in our db a few times, not sure where it comes from)

app/models/foi_attachment/content_type.rb
```diff
@@ -0,0 +1,48 @@
...
+    'application/vnd.openxmlformats-officedocument.wordprocessingml.document' => 'Word document',
```
laurentS, line 19:
```suggestion
    'application/vnd.openxmlformats-officedocument.wordprocessingml.document' => _('Word document'),
    'application/vnd.oasis.opendocument.text' => _('OpenDocument text'),
```

app/models/foi_attachment/content_type.rb
```diff
@@ -0,0 +1,48 @@
...
+  extend ActiveSupport::Concern
+
+  # rubocop:disable Layout/LineLength
+  CONTENT_TYPE_NAMES = {
+    # Plain Text
+    "text/plain" => 'Text file',
+    'application/rtf' => 'RTF file',
+
+    # Binary Documents
+    'application/pdf' => 'PDF file',
```
laurentS, line 12:
```suggestion
    'application/pdf' => _('PDF file'),
    
    # email delivery status files
    'message/delivery-status' => _('Email delivery status message'),
```
(we have thousands of those, and they tend to be confusing to many users, having a bit of a clearer name on them cannot hurt.

app/models/incoming_message/main_body.rb
```diff
@@ -188,14 +188,13 @@ def _convert_part_body_to_text(part)
     else
       # whatever kind of attachment it is, get the UTF-8 encoded text
       text = part.body_as_text.string
+
       if part.content_type == 'text/html'
         # e.g. http://www.whatdotheyknow.com/request/35/response/177
         # TODO: This is a bit of a hack as it is calling a
         # convert to text routine.  Could instead call a
         # sanitize HTML one.
```
laurentS, line 196: is this comment still relevant?

app/models/foi_attachment.rb
```diff
@@ -250,24 +224,23 @@ def update_display_size!
   end
 
   # Whether this type has a "View as HTML"
-  def has_body_as_html?
+  def body_to_html?
```
laurentS: I find this method name confusing, my brain goes "this is a conversion to html, but it looks like it returns a bool 🤔". `can_be_viewed_as_html?` ?

lib/attachment_to_text.rb
```diff
@@ -0,0 +1,250 @@
+class AttachmentToText
+  WORD_DOCS = %w[
+    application/vnd.ms-word
+    application/vnd.openxmlformats-officedocument.wordprocessingml.document
```
laurentS:
```suggestion
    application/vnd.ms-word
    application/vnd.openxmlformats-officedocument.wordprocessingml.document
    application/vnd.ms-word.document.macroenabled.12
    application/vnd.oasis.opendocument.text
```

I think we can include the ODF formats here, since we're using libreoffice now.

lib/attachment_to_text.rb
```diff
@@ -0,0 +1,250 @@
...
+  ]
+
+  POWERPOINT_DOCS = %w[
+    application/vnd.ms-powerpoint
+    application/vnd.openxmlformats-officedocument.presentationml.presentation
+  ]
```
laurentS:
```suggestion
  PRESENTATION_DOCS = %w[
    application/vnd.ms-powerpoint
    application/vnd.openxmlformats-officedocument.presentationml.presentation
    application/vnd.oasis.opendocument.presentation
  ]
```

Let's get rid of proprietary names 🙂

lib/attachment_to_text.rb
```diff
@@ -0,0 +1,250 @@
...
+
+  EXCEL_DOCS = %w[
+    application/vnd.ms-excel
+    application/vnd.openxmlformats-officedocument.spreadsheetml.sheet
+  ]
```
laurentS:
```suggestion
  SPREADSHEET_DOCS = %w[
    application/vnd.ms-excel
    application/vnd.ms-excel.sheet.macroenabled.12
    application/vnd.ms-excel.sheet.binary.macroenabled.12
    application/vnd.openxmlformats-officedocument.spreadsheetml.sheet
    application/vnd.oasis.opendocument.spreadsheet
  ]
```

lib/attachment_to_text.rb
```diff
@@ -0,0 +1,250 @@
...
+    # NOTE: re. charset: TMail always tries to convert email bodies
+    # to UTF8 by default, so normally it should already be that.
+    # TODO: - tell all these command line tools to return utf-8
+    case content_type
+    when 'text/plain'      then extract_plain(body)
+    when 'text/html'       then extract_html(body)
+    when 'application/pdf' then extract_pdf(body)
+    when *WORD_DOCS        then extract_ms_word(body)
+    when *EXCEL_DOCS       then extract_ms_excel(body)
+    when *POWERPOINT_DOCS  then extract_ms_powerpoint(body)
```
laurentS:
```suggestion
    when *SPREADSHEET_DOCS   then extract_ms_excel(body)
    when *PRESENTATION_DOCS  then extract_ms_powerpoint(body)
```

lib/attachment_to_text.rb
```diff
@@ -0,0 +1,250 @@
...
+      # instead, which doesn't do that.
+      AlaveteliExternalCommand.run(
+        'elinks',
+        '-eval', "set document.codepage.assume = \"#{charset}\"",
+        '-eval', 'set document.codepage.force_assumed = 1',
+        '-dump-charset', 'utf-8',
+        '-force-html', '-dump',
+        file.path,
+        binary_output: false,
+        timeout: 1200,
```
laurentS: Do we really need 20 minutes to extract content from a single file? it seems extremely long, even for a timeout. I would have expected something on the order of 10-20 seconds, maybe up to a minute for heavier formats?

lib/attachment_to_text.rb
```diff
@@ -0,0 +1,250 @@
...
+  # empty token 11 = unused here; token 11 is only for CSV import ("remove spaces").
+  # -1 in token 12 = export all sheets to separate files like sample-Sheet1.csv, sample-Sheet2.csv.
+  def extract_ms_excel(body)
+    csv_filters = '(StarCalc):44,34,76,1,,0,false,true,false,false,,-1'
+
+    with_tempfile(body) do |file|
+      in_tempdir do
+        AlaveteliExternalCommand.run(
+          'libreoffice', '--headless',
+          '--convert-to', "csv:Text - txt - csv #{csv_filters}",
```
laurentS, line 230: does this mean that a (somewhat theoretical) sparse file with a single word per line scattered over several columns would end up looking like a long list of commas? For the purpose of search indexing, this is not ideal. Postgres will remove those commas in the full-text index (though not in the raw_content column), but if we're not using this for anything else, it might be useful to `squeeze(',')` or something similar, to avoid storing unsearchable text.

spec/factories/foi_attachments.rb
```diff
@@ -63,12 +63,36 @@
       filename { 'attachment.txt' }
```
laurentS, line 63: There's a typo in the filename of this factory, an `a` is missing.

lib/attachment_to_text.rb
```diff
@@ -0,0 +1,250 @@
...
+    convert_string_to_utf8(text, 'UTF-8').string
+  end
+
+  protected
+
+  attr_reader :attachment
+
+  private
+
+  def extract_text(body, content_type)
```
laurentS, line 82: To help avoid confusion, can you rename the method or add a comment to clarify whether the text obtained is redacted or not?

spec/fixtures/files/interesting.csv
```diff
@@ -1,3 +1,3 @@
-a,b
+foo,bar
```
laurentS, line 1:
```suggestion
foo,bar
café,"maçã vermelha"
```
Let's actually verify that utf-8 and quoting do work :)

laurentS (commented):
Apologies for the long list of comments 😅 
I added some strange `content_type` strings found in our db. I don't know how many variations MS created. The ones I checked are probably from misuse of office progs, with extensions like `.docm` and such.

app/models/foi_attachment/content_type.rb
garethrees: I just moved what existed already to make the code a bit easier to work with. Will pick this up after this PR is merged.

garethrees: Ticketed in https://github.com/mysociety/alaveteli/issues/9279

lib/attachment_to_text.rb
garethrees: Think its useful to separate out MS from open stuff, but agree we could include the openoffice ones as above and then merge the two arrays together

lib/attachment_to_text.rb
garethrees: Hm yeah I just inlined the existing default without giving it that much thought: https://github.com/mysociety/alaveteli/blob/0.46.0.0/lib/mail_handler.rb#L77-L79

Feels like this could be reduced significantly

lib/attachment_to_text.rb
garethrees: Do you mean a sheet like this?

Image

Yes, that ends up as:

```
cat,,,,
dog,,,,
,giraffe,,,
shark,,,,
,,,,bear
,,,,
,,bird,,
```

This isn't used for anything else at the moment so I don't think there'd be a problem with `squeeze`.

spec/factories/foi_attachments.rb
garethrees: https://github.com/mysociety/alaveteli/pull/9280

spec/fixtures/files/interesting.csv
garethrees: 9a4f4f3a5

app/models/incoming_message/main_body.rb
garethrees: Yeah – we're still extracting text from a`text/html` part to then convert the extracted text back to HTML. The comment is saying that instead of doing this we should just sanitise the HTML we received (which would probably help with e.g. preserving tables in the response (https://github.com/mysociety/alaveteli/issues/1528)), but out of scope for now.

app/models/foi_attachment.rb
garethrees: I guess this might be a ruby-ism. Instinctively I see the `?` suffix and _[know](https://rubystyle.guide/#bool-methods-qmark)_ I'm going to get a Boolean back. We [avoid auxiliary verbs](https://rubystyle.guide/#bool-methods-prefix) like `can_` for predicate methods.

I named it `body_to_html?` to complement the actual conversion method (`body_to_html`).  I read this as "can I run `body_to_html` successfully on this record?" or "does this record have the ability to convert the body to html?"

The use case is basically:

```erb
<!-- some psuedocode here -->
<!-- only show the "view" link if we can convert the record to html -->
<% if @record.body_to_html? %>
  <%= link_to 'View', view_attachment_path(@record) %>
```

I suppose `viewable_as_html?` (EDIT: or maybe `html_viewer?`) would be the most appropriate name at the call site, at the expense of more clearly linking its relationship with `body_to_html`.

lib/attachment_to_text.rb
garethrees: Hm yeah does look like all we need to do is add the content types and the existing extraction method should work. I'll add another commit / later PR to add additional formats.

lib/attachment_to_text.rb
garethrees: It was intentional at the time e7b1cc6d8da1ebf98b5d966ffa67682b24f3a483 but can't remember what reason a very long value was chosen.

lib/attachment_to_text.rb
garethrees: Added to main class doc instead

garethrees:
Okay, considering this done. Just need to get packages updated on our internal infrastructure before I merge.

PR #9276, opened 2026-05-20
Authenticator 2fa enrolment flow

chrismytton:
## Relevant issue(s)

Part of the [Authenticator 2FA (#9241)](https://github.com/mysociety/alaveteli/issues/9241) work.

Follows on from https://github.com/mysociety/alaveteli/pull/9254 and https://github.com/mysociety/alaveteli/pull/9256

## What does this do?

Adds a TOTP enrolment flow allowing users to scan a QR code, input a TOTP code and then have 2FA activated on their account. The same flow can be used for HOTP users to migrate to TOTP.

## Implementation notes

This is just the enrolment flow currently. The gate checks (sign-in, disabling 2fa etc) will be done in a future PR.

### New 2FA setup

https://github.com/user-attachments/assets/beb31458-5366-44a8-be30-fbb38e09304c

### HOTP migration

https://github.com/user-attachments/assets/c46006b8-8ce9-4170-80f3-6515fa8834b9

## Notes to reviewer

There were a couple of things I wasn't sure of the best approach for, so any feedback welcome:

~~- `enrol_totp!` - this is a bang method because it calls `save!`, but could also arguably be `enrol_totp?` since it returns a boolean. Perhaps I should stick with plain `enrol_totp`? Maybe it's fine as-is.~~
~~- `set_provisioning_uri` feels a bit awkward, not sure if there's a better or more Rails-y way to set these instance variables?~~

I managed to solve both of these issues by introducing an `OtpEnrolment` model which handles the provisioning URI and validating the initial OTP code. I've rewritten the commit and force pushed, but you can see the previous version with the two issues above in https://github.com/mysociety/alaveteli/pull/9276/changes/286fad4cc50f30a6ab11c38c17a50cfa39008b6a.


chrismytton:
Realised that the password change gate doesn't work for TOTP users.

Seems that during model validation the gem's TOTP flow writes a timestamp to the DB for replay protection (fine) but that then triggers validations to run _again_, and this time the timestamp has been updated, so replay protection fails.

I'm just working on a fix, as ideally both HOTP and TOTP users should be able to go through the same password change flow.

app/views/one_time_passwords/show.html.erb
```diff
@@ -3,8 +3,65 @@
...
+    <p>
+      <%= _('Two factor authentication is active on your account. Sign-ins ' \
+            'and password changes will now use the entry saved in your ' \
+            'authenticator app.') %>
+    </p>
+
+    <ul class="one-time-password-actions not-for-print">
+      <li class="one-time-password-actions__action-item action-item">
+        <% confirmation_msg = _('Are you sure you want to disable two ' \
+                                'factor authentication?') %>
```
garethrees: Think we should emphasise the impact here:

"Are you sure you want to disable two factor authentication? Your account will be less secure."

app/views/one_time_passwords/show.html.erb
```diff
@@ -3,8 +3,65 @@
...
+                  'shred any printed copy.') %>
+          </strong>
+          <%= _("Don't delete the new authenticator app entry you just " \
+                "set up, that's what you will use from now on.") %>
+        </p>
+      </div>
+    <% end %>
+
+    <p>
+      <%= _('Two factor authentication is active on your account. Sign-ins ' \
```
garethrees: Maybe minor phrasing tweak here?

Two factor authentication is active on your account. You'll need to enter the code from your authenticator app for sensitive actions like sign-ins and password changes.

garethrees (commented):
Behaviour looks good – nice upgrade flow! Haven't reviewed the code, will leave that to @gbp.

On submitting an incorrect OTP from authenticator I lose the field to try again:

Image

Think we might want to improve the "about 2FA" guidance a bit, but we can get the main code working and merged and then iterate on that later.

app/models/user/one_time_password.rb
```diff
@@ -50,6 +50,17 @@ def enable_otp
...
   end
 
+  # Persist `secret` as the user's otp_secret_key and flip the user to TOTP.
+  # No verification, so callers (e.g. OtpEnrolment) are responsible for proving
+  # the user holds the secret before invoking this.
+  def enable_totp(secret:)
+    self.otp_secret_key = secret
+    self.otp_counter = nil
+    self.otp_enabled = true
+    self.otp_enabled_at = Time.current
```
garethrees: Think this should be `Time.zone.current` (see e.g. https://github.com/mysociety/alaveteli/pull/3670)

app/views/one_time_passwords/_code_form.html.erb
```diff
@@ -0,0 +1,17 @@
...
+    <%= f.text_field :otp_code,
+                     autocomplete: 'one-time-code',
+                     autofocus: true,
+                     inputmode: 'numeric',
+                     pattern: '\d{6}',
+                     size: 6 %>
+  </p>
+
+  <div class="form_button">
+    <%= f.submit local_assigns.fetch(:submit_label, _('Verify')) %>
```
garethrees, line 15: nice trick!

app/models/user/one_time_password.rb
```diff
@@ -30,6 +30,30 @@ module User::OneTimePassword
...
+      verified_at = totp.verify(code,
+                                drift_behind: options[:drift] || 0,
+                                after: otp_last_used_at)
+      if verified_at
+        self.otp_last_used_at = verified_at
+        update_column(:otp_last_used_at, verified_at) if persisted?
+      end
+      verified_at
+    end
+    private :authenticate_totp
```
garethrees: Why not add this method underneath the existing `private` keyword?

chrismytton:
> On submitting an incorrect OTP from authenticator I lose the field to try again

I think that issue was fixed in https://github.com/mysociety/alaveteli/pull/9272, so hopefully rebasing onto latest develop will fix it here as well.

app/controllers/one_time_passwords_controller.rb
```diff
@@ -5,19 +5,33 @@
...
   end
 
   def create
-    @user.enable_otp
+    @enrolment = OtpEnrolment.new(
+      user: @user,
+      secret: session[:pending_otp_secret_key],
+      otp_code: enrolment_params[:otp_code].to_s
+    )
+    was_hotp = @user.hotp?
```
gbp: Can we use one of the `*_previously_was` methods?

app/views/one_time_passwords/new.html.erb
```diff
@@ -0,0 +1,35 @@
...
+
+  <li>
+    <p>
+      <%= _('Enter the 6-digit code from your authenticator app to ' \
+            'finish setting up two factor authentication:') %>
+    </p>
+
+    <%= render 'code_form',
+                form_url: one_time_password_path,
+                submit_label: _('Enable two factor authentication') %>
```
gbp, line 33: Why are we passing these variables through to the form over hardcoding them?

spec/controllers/one_time_passwords_controller_spec.rb
```diff
@@ -6,7 +6,80 @@
...
+      expect(response.body).to include(session[:pending_otp_secret_key])
+    end
+
+    it 'reuses an existing candidate secret across refreshes' do
+      sign_in user
+      get :new
+      first_secret = session[:pending_otp_secret_key]
+      get :new
+
+      expect(session[:pending_otp_secret_key]).to eq(first_secret)
```
gbp: We can save a page load by assigning `pending_otp_secret_key` to a known value and verifying it doesn't change

spec/controllers/one_time_passwords_controller_spec.rb
```diff
@@ -6,7 +6,80 @@
...
+      user.reload
+      expect(user.otp_secret_key).to eq(original_secret)
+      expect(user.otp_counter).to eq(original_counter)
+      expect(user.otp_enabled).to eq(original_enabled)
+    end
+
+    context 'when the user is currently HOTP-enabled' do
+      it 'does not touch the persisted HOTP secret or counter' do
+        user = FactoryBot.create(:user, :enable_hotp)
+        user.save!
```
gbp: Is `save!` after creating with FactoryBot needed?

app/models/user/one_time_password.rb
gbp: Not sure about re-implementing this method from the gem. I think we could handled this a bit differently, how about something along the lines of...

```ruby
def verify_otp_code
  return if @authenticating_otp
  ...
end

def authenticate_otp(code, options = {})
  @authenticating_otp = true
  super
ensure
  @authenticating_otp = false
end
```

gbp (changes requested):
Nice work. I've left a few code changes and suggestions, let me know what you think on these.

app/models/user/one_time_password.rb
chrismytton: I think `Time.current` is [timezone aware](https://api.rubyonrails.org/classes/Time.html#method-c-current), but have changed this to `Time.zone.now`, as that seems to be the most-used approach in the codebase. 73d489ec8a407c7cf427661d2e421b49a9b1f13f

app/models/user/one_time_password.rb
chrismytton: Yeah that would have been more logical. Have removed this method following @gbp's feedback, so this statement has gone now.

app/controllers/one_time_passwords_controller.rb
chrismytton: Oooh, didn't know about these! Have updated it to use `@user.otp_counter_previously_was.present?`. 91e61e655e16090c6520cbe3504a46afa75895ce

app/views/one_time_passwords/new.html.erb
chrismytton: It's because this code form partial is going to be reused for different things (disabling TOTP, sign in etc), and those will go to different places and have differing button text (see https://github.com/mysociety/alaveteli/pull/9306). But was probably a bit premature to parameterise them in this pull request!

spec/controllers/one_time_passwords_controller_spec.rb
chrismytton: Good shout! Done in 1f601ddbca3f70413fac89b243940d9cec5d41e6

spec/controllers/one_time_passwords_controller_spec.rb
chrismytton: 🤦 Fixed in 1f601ddbca3f70413fac89b243940d9cec5d41e6

app/models/user/one_time_password.rb
chrismytton: Yeah, that's much nicer, have switched to that. `authenticate_otp` now sets an `@authenticating_otp` flag around `super` and `verify_otp_code` returns early while it's set, so the gem's TOTP verification runs unmodified and the re-entrant validation is skipped. 73d489ec8

app/views/one_time_passwords/new.html.erb
gbp: Cool. That is fine. I thought that could be the case but wanted to check just in case I was missing something.

gbp (approved):

chrismytton:

PR #9277, opened 2026-05-20
Remove user name from confirm_login email

garethrees:
If spam users register for the service with…

* A spam `User#name` (e.g. "buy my warez")
* A geniune `User#email` of a major email provider

…the "confirm your account" email is sent to the email address. Major email providers see that we're sending emails that include spam terms and may downrank the domain.

We're not losing much by removing the user's name in this case.

We also use this template for password change confirmation, but again the name is not that critical there.

It is a nice sense check for the user as they'd notice if the name was not their own, but I think the domain reputation issue is more critical.

Mentioned by @laurentS as a thing that actually happened on their site.

### DIFF
--- a/app/views/user_mailer/confirm_login.text.erb
+++ b/app/views/user_mailer/confirm_login.text.erb
@@ -1,5 +1,3 @@
-<%= raw @name %>,
-
 <%= _('Please click on the link below to confirm your email ' \
       'address.') %> <%= raw @reasons[:email] %>
 
--- a/doc/CHANGES.md
+++ b/doc/CHANGES.md
@@ -2,6 +2,8 @@
 
 ## Highlighted Features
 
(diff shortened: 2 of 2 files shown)


laurentS:
I am wondering if it would help to also add an unsubscribe link to the email.
- https://support.google.com/mail/answer/15263077?hl=en says we don't need to in the confirmation email
- https://postmarkapp.com/blog/list-unsubscribe-header neither confirms nor denies it.

garethrees:
> I am wondering if it would help to also add an unsubscribe link to the email.

Its not a mailing list so there's no real subscription to unsubscribe from. It would be good to have better control over transactional emails (https://github.com/mysociety/alaveteli/issues/4603), but that wouldn't really apply in this case as the account isn't even active yet. And we're not focused on that can of worms just yet 😅

PR #9278, opened 2026-05-21
Isolate Xapian from tests

gbp:
## What does this do?

Isolates Xapian search index access from the test suite. Specs that need a real Xapian index are tagged with `:xapian` — run them with `rspec --tag xapian` or without `rspec --tag '~xapian'`. All other specs use lightweight stubs instead of hitting the Xapian database.

## Why was this needed?

The test suite was tightly coupled to a live Xapian index. Every spec run rebuilt the index, slowing things down and making it easy for unrelated tests to accidentally depend on Xapian. Separating the two allows for faster test runs and makes it explicit which specs actually exercise search.

<hr>

[skip changelog]


garethrees:
Had a read through (commit by commit, but at a fairly general level – not line by line) and it all looks like a nice improvement.

Am I right in thinking that the xapian-tagged specs aren't run by default? Looks like that would explain the nice CI speedup, but wondering whether we should create a job that runs only these specs in parallel?

gbp:
> Am I right in thinking that the xapian-tagged specs aren't run by default?

No - all tests are run by default. 

> Looks like that would explain the nice CI speedup, but wondering whether we should create a job that runs only these specs in parallel?

I don't think there has been much speed improvement. You're probably seeing the GH actions run, these can vary quite a bit.

We should look at parallel tests as it should improve speed a fair bit. I imagine there is a lot of work still involved but reducing our dependency on Xapian will help support that.

spec/models/info_request_spec.rb
```diff
@@ -3787,141 +3787,74 @@ def create_old_unclassified_holding_pen
...
+      expect(events).to match_array([successful_event, sent_event])
+      expect(all_successful).to be false
     end
 
-    it 'coalesces duplicate requests' do
-      request_events, request_events_all_successful = InfoRequest.recent_requests
-      expect(request_events.map(&:info_request).select { |x|x.url_title =~ /^spam/ }.length).to eq(1)
+    it 'sets all_successful flag for five or more successful responses' do
+      events = Array.new(5) do
+        FactoryBot.build(:info_request_event, event_type: 'response')
```
garethrees, line 3815: Can use [`build_list`](https://thoughtbot.github.io/factory_bot/building-or-creating-multiple-records/summary.html#building-or-creating-multiple-records) for creating multiple records.

garethrees (approved):
Looks great!

PR #9290, opened 2026-05-26
Allows displaying information by Theme below text area in followup

robvandijk:
## What does this do?
It allows a Theme to display information below the text area in `followup`

## Why was this needed?
We need to inform the user about the option to pass a telephone number for contact information

## Implementation notes
We discussed this today by email. My initial thought was to extract a part of the view into a partial, and then to make adjustments to that smaller partial (smaller templates are better for maintenance purposes). Then I realized that here (and in other upcoming cases) I just want to add information, not to change any existing parts. Therefore I created this PR using a kind of plug-in mechanism for themes to be able to add info at certain (important) locations.
If you don't like this approach I can change this PR to the earlier discussed solution by extracting smaller partials (although that will require some more maintenance in the future).

<hr>


robvandijk:
BTW I have to add info at more locations. I will add these to this PR, once I've heard from you which solution you prefer (checking if partials supplied by themes exist, or splitting views into smaller partials to be overridden by themes).

garethrees:
Our usual way of providing theme hooks is a blank partial ([example](https://github.com/mysociety/alaveteli/blob/develop/app/views/general/_frontpage_extra.html.erb)), though I'm not adverse to adding hooks like this, as it avoids a blank template in core. If we went down this approach it would be nice to add a helper so we could do something like:

```erb
<%= render_theme_partial 'request/name_of_partial' %>
```

In this case though its not obvious what the proposed template is meant to do. Its name ("`theme_info_for_requester`") is pretty generic and is in the wrong folder for this view (should be in views/followups rather than views/request). I'd call this more like `followups/form/followup_message_help`.

app/views/alaveteli_pro/followups/_preview_notes.html.erb
```diff
@@ -1,4 +1,5 @@
 <ul>
     <li><%= _('Check you haven\'t included any <strong>personal information</strong>.') %></li>
     <li><%= _('Your message will be displayed publicly on this website once you publish your request.') %></li>
+    <%= render_theme_partial 'request/additional_advice_preview' %>
```
garethrees: I think this partial is small enough that it could just be entirely overridden rather than needing a theme partial hook.

app/views/alaveteli_pro/info_request_batches/_form.html.erb
```diff
@@ -29,6 +29,8 @@
       <%= f.text_area :body, :rows => 20, :cols => 60 %>
     </p>
 
+    <%= render_theme_partial 'request/below_request' %>
```
garethrees: This should be prefixed `alaveteli_pro/info_request_batches`. I don't think we should assume that the contents of these will definitely be the same across e.g. request, pro request, batch request, etc.

I'd actually nest this under `form`, too, to make it super clear its part of the form (rather than e.g. a view/preview step.

Overall I'd make this `alaveteli_pro/info_request_batches/form/below_request_body_field`.

app/views/alaveteli_pro/info_requests/_form.html.erb
```diff
@@ -32,6 +32,8 @@
       </p>
     <% end %>
 
+    <%= render_theme_partial 'request/below_request' %>
```
garethrees: Similar to the comment above, I'd make this `alaveteli_pro/info_requests/form/below_request_body_field`.

app/views/followups/_preview_notes.html.erb
```diff
@@ -1,4 +1,5 @@
 <ul>
     <li><%= _('Check you haven\'t included any <strong>personal information</strong>.') %></li>
     <li><%= _('Your message will appear in <strong>search engines</strong>') %></li>
+    <%= render_theme_partial 'request/additional_advice_preview' %>
```
garethrees: I think this partial is small enough that it could just be entirely overridden rather than needing a theme partial hook.

app/views/followups/preview.html.erb
```diff
@@ -44,6 +44,7 @@
               'to it, will be displayed publicly on this website.') %>
       </p>
     <% end %>
+    <%= render_theme_partial 'followups/below_followup_message' %>
```
garethrees: I'd call this `followups/preview/below_followup_message`

app/views/request/new.html.erb
```diff
@@ -246,6 +246,8 @@
             <% end %>
           </li>
 
+          <%= render_theme_partial 'request/additional_advice_new' %>
```
garethrees: Think it would be better to extract the advice section into its own partial (like `app/views/followups/_preview_notes.html.erb` and `app/views/alaveteli_pro/followups/_preview_notes.html.erb`) which would then be small enough to just be entirely overridden.

app/views/request/preview.html.erb
```diff
@@ -21,6 +21,7 @@
           <li><%= _('<a href="{{url}}">Do you want to request private ' \
                       'information about yourself?</a>',
                     :url => help_requesting_path(:anchor => "data_protection").html_safe) %></li>
+          <%= render_theme_partial 'request/additional_advice_preview' %>
```
garethrees: Think it would be better to extract the advice section into its own partial `app/views/request/_preview_notes.html.erb` (like `app/views/followups/_preview_notes.html.erb` and `app/views/alaveteli_pro/followups/_preview_notes.html.erb`) which would then be small enough to just be entirely overridden.

app/views/request/preview.html.erb
```diff
@@ -55,6 +56,8 @@
         </div>
       <% end %>
 
+      <%= render_theme_partial 'request/below_request' %>
```
garethrees: Would call this `request/preview/below_request`

robvandijk:
> Our usual way of providing theme hooks is a blank partial ([example](https://github.com/mysociety/alaveteli/blob/develop/app/views/general/_frontpage_extra.html.erb)), though I'm not adverse to adding hooks like this, as it avoids a blank template in core. If we went down this approach it would be nice to add a helper so we could do something like:
> 
> ```
> <%= render_theme_partial 'request/name_of_partial' %>
> ```
> 
> In this case though its not obvious what the proposed template is meant to do. Its name ("`theme_info_for_requester`") is pretty generic and is in the wrong folder for this view (should be in views/followups rather than views/request). I'd call this more like `followups/form/followup_message_help`.

It was a quick example of the direction I was thinking of, and far from perfect. Thanks for the suggestions. I added helper `render_theme_partial` and included it in various locations. For the names of the partials I decided to stay as neutral as possible and refrained from using terms such as `help`: other Themes may want to use such a partial for different reasons. So I decided to stick to the location where `render_theme_partial` is used (e.g. `below_...`).

I have also created a PR for the Woo-knop theme, see https://github.com/mysociety/wob-knop-theme/pull/47. It depends on the  `render_theme_partial` helper introduced in this PR.

app/views/request/new.html.erb
garethrees: I'd make it like this:

* request/new.html.erb
  * request/new/_advice.html.erb
    * request/new/advice/_before_you_start.html.erb
    * request/new/advice/_writing_your_request.html.erb
    * request/new/advice/_signing_your_request.html.erb

That way  you can easily override a specific section, or you could override the whole `_advice` partial, adding your custom items and/or pulling in whichever of the sub-partials you want

app/views/followups/_followup.html.erb
```diff
@@ -132,6 +132,7 @@
       <p>
         <%= o.text_area :body, rows: 15, cols: 55 %>
       </p>
+      <%= render_theme_partial 'followups/below_followup_message' %>
```
garethrees: Similar to the comment above, I'd make this `followups/form/below_request_body_field`.

In this case we haven't extracted the form to a partial, so the namespacing doesn't quite match reality, but I think we ought to extract the form (which can be done later) so might as well get this piece right now.

app/views/request/_form.html.erb
```diff
@@ -23,6 +23,8 @@
 
 <%= render partial: 'new_message_text_length' %>
 
+<%= render_theme_partial 'request/below_request' %>
```
garethrees: I'd call this `request/form/below_request_body_field`.

app/views/user/_signup.html.erb
```diff
@@ -41,6 +41,8 @@
       <%= password_field 'user_signup', 'password_confirmation', { :size => 15, :tabindex => 40, :autocomplete => "off" } %>
     </p>
 
+    <%= render_theme_partial 'user/bottom_signup', last_tabindex: 40 %>
```
garethrees: Would call this `user/signup/before_submit`

garethrees (changes requested):
Thanks for the updates. I've made a couple of naming suggestions with a view to what would make it easier for longer term maintenance, and I think a couple of the `render_theme_partial` calls could be removed where the main partial is small enough to just override in its entirity.

After these updates I think this would be good to merge, so once you've made them just squash it all in to a single commit.

robvandijk:
> Thanks for the updates. I've made a couple of naming suggestions with a view to what would make it easier for longer term maintenance, and I think a couple of the `render_theme_partial` calls could be removed where the main partial is small enough to just override in its entirity.
> 
> After these updates I think this would be good to merge, so once you've made them just squash it all in to a single commit.

Thanks Gareth for the quick response. Will get back to this after my holidays. Will process the suggested naming changes. I'm wondering about removing some of the `render_theme_partial`s and instead overwrite partials though. The method of overwriting partials, however small, is likely to cause more future maintenance than using the helper. Perhaps we can have a call to discuss this more?

garethrees:
> I'm wondering about removing some of the `render_theme_partial`s and instead overwrite partials though. The method of overwriting partials, however small, is likely to cause more future maintenance than using the helper.

I think there's a tradeoff between maintenance of core vs maintenance of the theme here. I think we want to balance this – and I completely appreciate its not a good balance right now – but more `render_theme_partial`s would mean more maintenance on our side to ensure they don't disappear underneath theme maintainers if/when we need to rename/remove them.

I also want to be a bit judicious about where they're included, especially as its a new idea that may or may not work out in the long run. If taken way too far we could end up with them before, in the middle of and after nearly every template!

Having come back and looked at this today, I think it does feel like even some of the ones I didn't initially suggest to be overrides take the idea a bit far – `request/form/below_request_body_field` for example is _very_ specific and lots of includes like this could get out of hand.

Happy to jump on a call to hash out some heuristics for when an override/include hook is most appropriate.

robvandijk:
Processed most of the proposed changes, except for the `below_request` and `below_followup` ones - let's discuss in the call how to deal with those.

Issue #8823, opened 2025-08-15
Admin UI showing which where redactions are applied and redactions apply to what


garethrees:
At the moment we've got a very basic understanding of where redactions are being applied to requests. In the admin interface we show the applicable rules for the request:

<img width="1012" height="342" alt="Image" src="https://github.com/user-attachments/assets/be0b9495-8693-4deb-b2e9-3d56af7b39b2" />

What this doesn't show is:

* Text masks
* Whether the rule actually removed anything
* Where it removed things if it did (i.e which outgoing/incoming messages, attachments, etc)

A good level up would be to include these in the listing of rules. Ideally we'd be able to quickly see which rules have been applied, and then dig in to where. This is just an idea for the UI for that:

<img width="785" height="293" alt="Image" src="https://github.com/user-attachments/assets/42d8fdbd-5a7e-418f-8854-f070116c3d4d" />

Here, we can quickly see that the only redactions are being made are by `CensorRule#2`. We can click on that to see which underlying records are being affected, and then easily click to navigate to those records.

I don't think we necessarily need to get any more specific about where replacements are happening – once you've identified the record in most cases it should be obvious enough, or easy to find with browser or site search.

We obviously might rationalise the TextMask and CensorRule systems, but I've included them in this list as it's what we have right now and knowing if a mask is doing something is quite useful.

As a stretch goal, it would also be useful to list redactions that are actually making a redaction on the lower level content under the request – i.e on the show/edit pages for outgoing message, incoming message, etc.

PR #9302, opened 2026-06-02
[#8823] Make `CensorRule` polymorphic & refactor

garethrees:
## Relevant issue(s)

Prep for https://github.com/mysociety/alaveteli/issues/8823

## What does this do?

* Use polymorphic association for censor rules
* Extract concerns
* Various cleanup refactorings

## Why was this needed?

* Tidies up existing code to use built-in framework
* Will reduce implementation complexity for #8823

## Notes to reviewer

Left a couple of inline review comments on specific things I was thinking about.

<hr>

[skip changelog]

### DIFF
--- a/app/models/censor_rule.rb
+++ b/app/models/censor_rule.rb
@@ -3,9 +3,6 @@
 # Table name: censor_rules
 #
 #  id                :integer          not null, primary key
-#  info_request_id   :integer
-#  user_id           :integer
-#  public_body_id    :integer
 #  text              :text             not null
 #  replacement       :text             not null
 #  last_edit_editor  :string           not null
@@ -15,6 +12,8 @@
 #  regexp            :boolean          default(FALSE), not null
 #  case_sensitive    :boolean          default(TRUE), not null
 #  ignore_diacritics :boolean          default(FALSE), not null
+#  censorable_type   :string
+#  censorable_id     :bigint
 #
 
 # models/censor_rule.rb:
@@ -32,14 +31,8 @@ class CensorRule < ApplicationRecord
     _('[extraneous and potentially defamatory material removed]')
   ].freeze
 
-  belongs_to :info_request,
-             inverse_of: :censor_rules,
-             optional: true
-  belongs_to :user,
-             inverse_of: :censor_rules,
-             optional: true
-  belongs_to :public_body,
-             inverse_of: :censor_rules,
+  belongs_to :censorable,
+             polymorphic: true,
              optional: true
 
   validate :require_valid_regexp,
@@ -53,11 +46,10 @@ class CensorRule < ApplicationRecord
                         :last_edit_comment,
                         :last_edit_editor
 
-  scope :global, -> {
-    where(info_request_id: nil,
-          user_id: nil,
-          public_body_id: nil)
-  }
+  scope :info_request, -> { where(censorable_type: 'InfoRequest') }
+  scope :public_body, -> { where(censorable_type: 'PublicBody') }
+  scope :user, -> { where(censorable_type: 'User') }
+  scope :global, -> { where(censorable_id: nil, censorable_type: nil) }
 
   cattr_accessor :canned_replacements,
                  instance_writer: false,
@@ -80,38 +72,32 @@ class CensorRule < ApplicationRecord
   end
 
   def is_global?
-    info_request_id.nil? && user_id.nil? && public_body_id.nil?
+    censorable_id.nil? && censorable_type.nil?
   end
 
   def expire_requests
-    if info_request
-      InfoRequest::ExpireJob.perform_later(info_request)
-      NotifyCacheJob.perform_later(info_request)
-    elsif user
-      InfoRequest::ExpireJob.perform_later(user, :info_requests)
-    elsif public_body
-      InfoRequest::ExpireJob.perform_later(public_body, :info_requests)
-    else # global rule
+    case censorable
+    when InfoRequest
+      InfoRequest::ExpireJob.perform_later(censorable)
+      NotifyCacheJob.perform_later(censorable)
+    when User, PublicBody
+      InfoRequest::ExpireJob.perform_later(censorable, :info_requests)
+    else
       InfoRequest::ExpireJob.perform_later(InfoRequest, :all)
     end
   end
 
   def censorable_requests
-    if info_request
+    case censorable
+    when InfoRequest
       # Prefer a chainable query instead of wrapping in Array for similar API
       # between CensorRule types
-      InfoRequest.where(id: info_request_id)
-    elsif user
-      user.info_requests
-    elsif public_body
-      public_body.info_requests
+      InfoRequest.where(id: censorable_id)
+    when User, PublicBody
+      censorable.info_requests
     else
       InfoRequest.unscoped
     end
-  end
-
-  def censorable
-    info_request || user || public_body || nil
   end
 
   private
--- a/app/models/censor_rule.rb
+++ b/app/models/censor_rule.rb
@@ -23,39 +23,16 @@
 # Email: hello@mysociety.org; WWW: http://www.mysociety.org/
 
 class CensorRule < ApplicationRecord
-  DEFAULT_CANNED_REPLACEMENTS = [
-    _('[Personally Identifiable Information removed]'),
-    _('[name removed]'),
-    _('[extraneous material removed]'),
-    _('[potentially defamatory material removed]'),
-    _('[extraneous and potentially defamatory material removed]')
-  ].freeze
-
-  belongs_to :censorable,
-             polymorphic: true,
-             optional: true
-
-  validate :require_valid_regexp,
-           if: -> { regexp? || !case_sensitive? || ignore_diacritics? }
-
-  validate :regexp_ignore_diacritics,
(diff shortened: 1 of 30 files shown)


db/migrate/20260601000000_add_polymorphic_to_censor_rules.rb
```diff
@@ -0,0 +1,82 @@
+class AddPolymorphicToCensorRules < ActiveRecord::Migration[8.0]
+  def up
+    add_column :censor_rules, :censorable_type, :string
+    add_column :censor_rules, :censorable_id, :bigint
```
garethrees: Not sure whether there's a better way to do this? Usually in a `change`/`create_table` migration we'd do…

```rb
t.belongs_to :imageable, polymorphic: true
```

…which also creates the indexes (last line of `#up`), but not sure if there's an equivalent for explicit up/down migrations.

db/migrate/20260601000000_add_polymorphic_to_censor_rules.rb
```diff
@@ -0,0 +1,82 @@
...
+      WHERE censorable_type = 'User'
+    SQL
+
+    execute <<~SQL
+      UPDATE censor_rules
+      SET public_body_id = censorable_id
+      WHERE censorable_type = 'PublicBody'
+    SQL
+
+    add_index :censor_rules, :info_request_id
```
garethrees: These are from `db/migrate/091_add_censor_rules_indices.rb`

app/models/censor_rule.rb
```diff
@@ -108,36 +62,7 @@ def encoded_text(encoding)
...
-
-  def diacritic_expander
-    DiacriticExpander.new(case_sensitive: case_sensitive?)
-  end
-
-  def regexp_ignore_diacritics
-    msg = 'Cannot use regexp and ignore diacritics option together'
-    errors.add(:text, msg)
+  def single_char_regexp
+    ::Regexp.new('.'.force_encoding('ASCII-8BIT'))
```
garethrees, line 64: Need to prefix with `::` as `CensorRule::Regexp` clashes, which I'm not super keen on. Could call it `CensorRule::Regex` but that's a little confusing too. On balance I went with the `::` prefix as its only an issue in this one place, but it is a bit of a gotcha.

db/migrate/20260601000000_add_polymorphic_to_censor_rules.rb
gbp: Are you looking for `add_reference :censor_rules, :censorable, polymorphic: true, index: true` ?

spec/controllers/admin/censor_rules_controller_spec.rb
```diff
@@ -833,7 +818,7 @@ def create_censor_rule
                        }
 
           expect(response).to redirect_to(
-            admin_body_path(assigns[:censor_rule].public_body)
+                            admin_body_path(public_body)
```
gbp:
```suggestion
            admin_body_path(public_body)
```

app/models/censor_rule/expiry.rb
```diff
@@ -0,0 +1,13 @@
...
+module CensorRule::Expiry
+  extend ActiveSupport::Concern
+
+  included do
+    after_commit :expire_requests
+  end
+
+  def expire_requests
+    censorable&.expire_requests ||
+      InfoRequest::ExpireJob.perform_later(InfoRequest, :all)
```
gbp: I think `#expire_requests` can return false, the last line enqueues NotifyCacheJob which has an around_enqueue callback to prevent queuing jobs when Varnish isn't configured.

This could result in global expiry of InfoRequest.all 😨.

spec/models/censor_rule_spec.rb
```diff
@@ -406,50 +406,21 @@
...
-    context 'with a public body rule' do
-      let(:body) { FactoryBot.create(:public_body) }
-      let!(:rule) { FactoryBot.create(:censor_rule, censorable: body) }
-
-      it 'expires the requests' do
-        expect { subject }.to have_enqueued_job(job).with(body, :info_requests)
+      it 'expires the requests via the censorable' do
+        expect(rule.censorable).to receive(:expire_requests)
+        subject
       end
```
gbp, line 429: Should have test to check `...to_not have_enqueued_job(InfoRequest::ExpireJob).with(InfoRequest, :all)`

app/models/censor_rule/polymorphic.rb
```diff
@@ -0,0 +1,23 @@
...
+module CensorRule::Polymorphic
+  extend ActiveSupport::Concern
+
+  included do
+    belongs_to :censorable, polymorphic: true, optional: true
+
+    scope :info_request, -> { where(censorable_type: 'InfoRequest') }
+    scope :public_body, -> { where(censorable_type: 'PublicBody') }
+    scope :user, -> { where(censorable_type: 'User') }
+    scope :global, -> { where(censorable_id: nil, censorable_type: nil) }
```
gbp, line 13:
```suggestion
    scope :global, -> { where(censorable: nil) }
```

app/models/censor_rule/polymorphic.rb
```diff
@@ -0,0 +1,23 @@
...
+  end
+
+  def global?
+    censorable_id.nil? && censorable_type.nil?
```
gbp, line 17: Probably enough just to check the ID

app/models/censor_rule/polymorphic.rb
```diff
@@ -0,0 +1,23 @@
...
+  end
+
+  def censorable_requests
+    censorable&.info_requests || InfoRequest.unscoped
```
gbp, line 21: Not called from anywhere. Do we still need this?

app/models/info_request.rb
```diff
@@ -1817,6 +1818,19 @@ def cached_urls
     ]
   end
 
+  # Return only this request in a chainable relation. This is mainly for
+  # duck-typing with other censorable relationships.
+  def info_requests
```
gbp, line 1823: Only called from censorable_requests. Do we need this?

db/migrate/20260601000000_add_polymorphic_to_censor_rules.rb
```diff
@@ -0,0 +1,82 @@
...
+      FOREIGN KEY (user_id)
+      REFERENCES users(id)
+    SQL
+
+    execute <<~SQL
+      ALTER TABLE censor_rules
+      ADD CONSTRAINT fk_censor_rules_public_body
+      FOREIGN KEY (public_body_id)
+      REFERENCES public_bodies(id)
+    SQL
```
gbp: Couldn't we use add_foreign_key here?

gbp (changes requested):
Nice refactor. Bug to fix and a few minor suggestions.

app/models/censor_rule/expiry.rb
garethrees: Ah! Okay easy fix is:

```rb
if censorable
  censorable.expire_requests
else
  # expire all
end
```

Or, could just add `true` to the end of `expire_requests`, since a `false` return isn't quite right – we know the `ExpireJob` will have been queued; returning `false` gives the indication that that hasn't happened:

```rb
  def expire_requests
    InfoRequest::ExpireJob.perform_later(self)
    NotifyCacheJob.perform_later(self)
    true
  end
```

garethrees: The former still feels a bit safer though

db/migrate/20260601000000_add_polymorphic_to_censor_rules.rb
garethrees: Ah perfect thanks!

app/models/censor_rule/polymorphic.rb
garethrees: Was added in f5ca2e07647860489d8ec51190d4dbc4aa7d7742 for temp rake task. Might be removable but will leave that to another PR as it might be useful for issues in https://github.com/mysociety/alaveteli/milestone/75

app/models/censor_rule/polymorphic.rb
garethrees: Not in the case of unsaved records:

```
From: /alaveteli/spec/models/censor_rule_spec.rb:95 :

     90:     end
     91:
     92:     context 'with a censorable' do
     93:       let(:rule) { FactoryBot.build(:user_censor_rule) }
     94:       it do
 =>  95:         binding.pry
     96:         is_expected.to eq(false)
     97:       end
     98:     end
     99:   end
    100:

[1] pry(#<RSpec::ExampleGroups::CensorRule::Global_2::WithACensorable>)> rule
=> #<CensorRule:0x0000ffff7c7b5788
 id: nil,
 text: "some text to redact",
 replacement: "[REDACTED]",
 last_edit_editor: "FactoryBot",
 last_edit_comment: "Modified by rspec",
 created_at: nil,
 updated_at: nil,
 regexp: false,
 case_sensitive: true,
 ignore_diacritics: false,
 censorable_type: "User",          <<<<<<<<<<<< user isn't saved so we get the correct type but no id yet
 censorable_id: nil>
```

I expect this isn't a real issue in production but no big deal checking both given its only occasionally called

gbp:
Need a corresponding PR for WDTK-theme

PR #9305, opened 2026-06-04
[#8812] Add common search interface

gbp:
## Relevant issue(s)

Follows on from the test isolation work in #9278
Work towards #8812

## What does this do?

Introduces a Search module that provides a backend-agnostic search interface.

Controllers, models, mailers, views, and rake tasks now interact with search through Search.search, Search.typeahead, Search.context, and Search.mark_needs_index instead of calling ActsAsXapian or TypeaheadSearch directly.

## Why was this needed?

The application was tightly coupled to Xapian through direct ActsAsXapian::Search, ActsAsXapian::Similar, and TypeaheadSearch calls scattered across controllers, models, mailers, and rake tasks. This made it impossible to swap the search backend without touching dozens of files.

The Search module establishes a clear boundary so a future backend (e.g. PostgreSQL full-text search) can be introduced by implementing Search::Backend without changing any application code.


app/search/backend.rb
```diff
@@ -0,0 +1,36 @@
...
+
+    # Required: find requests similar to the given one
+    def similar_requests(info_request)
+      raise NotImplementedError, 'Subclasses must implement #similar_requests'
+    end
+
+    # Optional: queue a record for search index update.
+    # Backends where the search index is the database itself (e.g.
+    # PostgreSQL) can leave this as a no-op.
+    def mark_needs_index(record)
```
garethrees: I'd have called this simply `reindex` (or `reindex_later`), but don't mind if we polish up naming later.

app/models/comment.rb
```diff
@@ -109,7 +109,7 @@ def hidden?
   end
 
   def reindex_request_events
-    info_request_events.find_each(&:xapian_mark_needs_index)
+    info_request_events.find_each { |e| Search.mark_needs_index(e) }
```
garethrees: Is the plan to return this to a method on a searchable record once we have the new search backend?

i.e. end up with:

```rb
info_request_events.find_each(&:mark_needs_index)
```

app/search/backend.rb
```diff
@@ -0,0 +1,36 @@
...
+      raise NotImplementedError, 'Subclasses must implement #search'
+    end
+
+    # Required: typeahead/autocomplete search returning a searchable object
+    def typeahead(query, model:, **options)
+      raise NotImplementedError, 'Subclasses must implement #typeahead'
+    end
+
+    # Required: find requests similar to the given one
+    def similar_requests(info_request)
```
garethrees: This seems quite specific to `InfoRequest` so not sure it really belongs in a model-agnostic search interface? I could see `similar` being a useful generic method. Don't mind if its a useful transitional step that we can remove later though.

app/search/context/info_request.rb
```diff
@@ -0,0 +1,27 @@
...
+    #
+    class InfoRequest
+      attr_reader :info_request
+
+      def initialize(info_request)
+        @info_request = info_request
+      end
+
+      def similar_requests
+        Search.backend.similar_requests(info_request)
```
garethrees: Why go via the backend here vs other cases where we'd have more like `Search.similar_requests`()?

app/search/results.rb
```diff
@@ -0,0 +1,128 @@
...
+      @words_to_highlight
+    end
+
+    # Enumerable interface - delegate to items
+    def each(&block)
+      items.each(&block)
+    end
+
+    # Collection query methods
+    def any?
```
garethrees: I thought you only had to define `#each` and then the other collection operations from the mixin get included automatically?

app/models/search.rb
```diff
@@ -0,0 +1,49 @@
...
+#   Search.search('freedom of information', models: [InfoRequestEvent])
+#
+# @example Typeahead search
+#   Search.typeahead('geraldine', model: PublicBody)
+#
+# @example Context-based search
+#   Search.context(info_request: request).similar_requests
+#
+module Search
+  def self.backend
```
garethrees, line 18: Could use [`mattr_accessor`](https://api.rubyonrails.org/classes/Module.html#method-i-mattr_accessor) for these?

app/models/info_request.rb
```diff
@@ -609,58 +609,6 @@ def self.stop_new_responses_on_old_requests
       end
   end
 
-  def self.request_list(filters, page, per_page, max_results)
-    query = InfoRequestEvent.make_query_from_params(filters)
```
garethrees, line 613: OOI why not keep these but defined like:

```rb
def self.request_list()
  Search.request_list()
end

def self.recent_requests()
  Search.recent_requests()
end
```

app/search/search.rb
```diff
@@ -46,4 +46,73 @@ def self.mark_needs_index(record)
   def self.queued_jobs_count
     backend.queued_jobs_count
   end
+
+  # Search for info request events and return a paginated list
+  def self.request_list(filters, page, per_page, max_results)
```
garethrees: These methods seem quite specific to `InfoRequest` rather than the general search interface.

doc/SEARCH.md
```diff
@@ -0,0 +1,381 @@
...
+### Full-text search
+
+```ruby
+# Returns a searchable object; call .search to execute
+results = Search.search('freedom of information',
+                        models: [InfoRequestEvent],
+                        sort_by: 'created_at',
+                        sort_ascending: false,
+                        collapse_by: 'request_collapse').
+            search(page: 1, per_page: 25)
```
garethrees: `Search.search().search()` seems a bit clunky.

Could look at renaming the first one, though tbf none of the alternatives I can think of are _that_ great:

* `Search.full_text` 
* `Search.fts`
* `Search.text`
* `Search.full`

Could go with query (`Search.query(…).search(…)`), but I guess we may want to reserve `query` for later use (e.g. parsing out facet parameters)?

Perhaps the second `search` method in that chain should be called `results`, since it returns a `Search::Results`:

```rb
# foo = search || typeahead
results = Search.foo().results()

results.records        # [{model:, percent:, ...}, ...]
results.total_estimate # estimated total matches
```

doc/SEARCH.md
```diff
@@ -0,0 +1,381 @@
...
+
+```ruby
+results = build_search_results(
+  items: [event],
+  total: 1,
+  page: 1,
+  per_page: 25,
+  spelling_correction: nil,
+  words_to_highlight: ['council'],
+  has_normal_search_terms: true
```
garethrees, line 318: Remind me what `has_normal_search_terms` is for? Just wondering if this is something we could drop

config/application.rb
```diff
@@ -104,6 +104,14 @@ class Application < Rails::Application
     config.autoload_paths << "#{Rails.root}/app/controllers/concerns"
     config.autoload_paths << "#{Rails.root}/app/models/concerns"
 
+    # Configure app/search as a namespaced directory so that files
+    # within it define constants under the Search namespace
+    # (e.g. app/search/backend.rb -> Search::Backend).
+    module ::Search; end
+    Rails.autoloaders.main.push_dir(
+      "#{Rails.root}/app/search", namespace: ::Search
```
garethrees: I don't feel massively strongly either way but could just put this in `app/models` as per https://github.com/basecamp/fizzy/blob/main/app/models/search.rb. What's your thinking behind `app/search`?

garethrees (commented):
I think in general all this is looking great and gets Xapian out of the way.

I think my main query is that it looks like the direction of travel is to add what looks like model-specific search methods onto the `Search` class/module (e.g. `Search.similar_requests`. We only have request-related stuff right now but I could see this growing and getting quite complex. I can see us wanting to extract the similar stuff as that's quite big, but I'd have put that in something like `InfoRequest::Similar` and then called out to the search interface when needed. Open to thoughts/suggestions on this and happy to go with less than ideal now if it helps make progress on the main goal of swapping out the search backend, but just a thought while I was reviewing.

app/search/backend.rb
gbp: Agreed. Will rename.

app/models/comment.rb
gbp: Yes, that's the intent. Records could have a backend-agnostic hook so this will become `info_request_events.find_each(&:reindex_later)`.

app/search/context/info_request.rb
gbp: Good catch — this is just inconsistent, will update

app/search/results.rb
gbp: Yes some of these aren't strictly needed as Enumerable adds them, I think we can drop `any?`, `count`, and `first`

app/models/search.rb
gbp: The reader provides a default value - using mattr would resolve the default on application boot. We will be switching backends in the tests so think this might be better. 

Will leave as is for now and these will likely change once we add a configuration option.

doc/SEARCH.md
gbp: It's still used in GeneralController to pick the default sort — a query with only prefix terms (e.g. status:successful) is "structured" so we default to newest-first, otherwise relevant-first.

So we can't drop it yet, though I can see it only being used for Xapian backend - we may want to generalise when the PG backend lands.

config/application.rb
gbp: No strong feelings about this. Would thinking search is going to be quite complex and giving it its own namespace would be good.

app/search/search.rb
gbp: Yeah I guess these could be moved the InfoRequest search context.

app/models/info_request.rb
gbp: Yes we can do this. And it makes more sense if we move the search context as we do for similar requests.

app/search/backend.rb
gbp: Yes it is quite specific to InfoRequest. I will remove from here and from the search context go straight to the SimilarRequests class.

We could make a `similar` method for other classes but trying not to add new functionality in this PR so lets do this later if needed.

doc/SEARCH.md
gbp: Agreed the double search is awkward.

I like renaming the terminal call to .results since it returns Search::Results.
```Search.search(...).results(page:, per_page:)```

I'd keep query reserved for later facet parsing as you say

gbp:
@garethrees think these changes address all the feedback. Sorry its not in fixups - was way to confusing to get everything in the correct commits.

Plus as discussed with @laurentS I've added a chainable scope so we can combine search with ActiveRecord queries.

garethrees:
> think these changes address all the feedback. Sorry its not in fixups - was way to confusing to get everything in the correct commits.

Great! No problem about the fixups. I don't think these need additional review as it was all fairly minor feedback. I think we should try to combine this and https://github.com/mysociety/alaveteli/pull/9239 as we discussed and then have another review together at that point as the general direction of this interface LGTM 🚀

config/application.rb
gbp: In the end I've moved to app/models/search - was seeing some strange auto reloading issues in development.

gbp:
@garethrees this is on staging and appears to be working as expected with the Xapian adaptor.

garethrees:
Great! I don't think additional review is required as all of the previous feedback was pretty minor.

PR #9306, opened 2026-06-05
2FA gate checks for sign-in and disabling

chrismytton:
## Relevant issue(s)

Part of the work to enable two-factor authentication #9241 

Builds on top of the enrolment work in #9276 

## What does this do?

Asks TOTP users for their current OTP code when they're signing in or trying to disable 2FA.

## Why was this needed?

This completes the OTP flow end-to-end. So a user can enrol for TOTP and then when they sign in or try to disable TOTP they will be prompted to enter their current OTP code.

## Implementation notes

Introduces a `PendingTwoFactorSignIn` class which is responsible for managing the session keys for a pending sign-in from a TOTP user. Encapsulates the user and the post redirect for the current sign-in, and makes it easy to clear all the relevant sessions keys at once.

## Screenshots

https://github.com/user-attachments/assets/aa593614-111f-41d7-a597-fa1504da3fb3

## Notes to reviewer

The last commit fixes the modal sign-in link. As the commit message says there was a redirect happening, which meant that the modal wasn't using the modal layout. Using the proper path helper seems to fix it so that the modal appears correctly when clicking the nav link when `FORCE_REGISTRATION_ON_NEW_REQUEST` is `true`.

### DIFF
--- a/app/models/pending_two_factor_sign_in.rb
+++ b/app/models/pending_two_factor_sign_in.rb
@@ -1,0 +1,52 @@
+# Wraps the session state for a sign-in that has passed the password check but
+# still needs a second factor.
+class PendingTwoFactorSignIn
+  TTL = 5.minutes
+
+  def initialize(session)
+    @session = session
+  end
+
+  def start(user:, remember_me:, post_redirect:)
+    @session[:pending_2fa_user_id] = user.id
+    @session[:pending_2fa_started_at] = Time.zone.now.to_i
+    @session[:pending_2fa_remember_me] = remember_me
+    @session[:pending_2fa_post_redirect_token] = post_redirect.token
+  end
+
+  def user_id
+    @session[:pending_2fa_user_id]
+  end
+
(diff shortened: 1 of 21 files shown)


app/models/pending_two_factor_sign_in.rb
```diff
@@ -0,0 +1,52 @@
+# Wraps the session state for a sign-in that has passed the password check but
+# still needs a second factor.
+class PendingTwoFactorSignIn
```
gbp, line 3: Maybe not one for now. If I were to implement this I would look at using `Rails.application.message_verifier`.

It could create a single message based on 'user', 'remember_me', and 'post_redirect' which we could store in the session instead of the 4 different keys - plus it automatically offers message expiry when the TTL elapses.

gbp (approved):
Nice one. I think this is good to go. One comment but not a blocker.

chrismytton:

PR #9307, opened 2026-06-05
Remove sensitive user admin columns

garethrees:
No good reason to display sensitive security information that is only used in technical processes within the app – none of these values are useful for human admins.

[skip changelog]


garethrees:
Will revive after https://github.com/mysociety/alaveteli/issues/9241 is complete.

FOIMonkey:
The last version of the 2fa changes I saw (yesterday) exposed the list of back-up codes in the admin view in plain text. It might be an idea to exclude those too if they're still there when it's merged.

garethrees:
Thanks yeah I'd clocked that too, will update this once all that's merged

PR #9317, opened 2026-06-09
Enforce two factor authentication for admins

chrismytton:
## Relevant issue(s)

Part of #9241 

Builds on top of #9306 

## What does this do?

Add an opt-in `AdminController.require_two_factor_auth` class attribute (default false).

When enabled, admins without TOTP enabled are redirected to /profile/two_factor to enable it.

It checks totp?, so HOTP and no-2FA admins are pushed to enrol, while external auth, the emergency user, and installs without two factor auth are exempt.

## Why was this needed?

Currently admin HOTP enforcement is only by convention rather than at a technical level. With this change, it can now be enforced at a technical level, forcing admins to do TOTP enrolment.

## Implementation notes

Follows a similar pattern to the `AtiNetworkController.showcase_enabled` class attribute.

## Screenshots

image

## Notes to reviewer

I've tried to handle the emergency user and external auth correctly, but I'm not overly familiar with those bits of code, so any feedback on that bit is particularly welcome.

### DIFF
--- a/app/controllers/admin_controller.rb
+++ b/app/controllers/admin_controller.rb
@@ -6,7 +6,13 @@
 
 class AdminController < ApplicationController
   layout "admin"
+
+  # Set `AdminController.require_two_factor_auth = true` in your theme to
+  # require admins to have authenticator-app (TOTP) two factor enabled.
+  class_attribute :require_two_factor_auth, default: false
+
   before_action :authenticate
+  before_action :enforce_two_factor_auth, if: :require_two_factor_auth?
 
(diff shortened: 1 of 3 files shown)


gbp (approved):
I'm happy with this.

Just need a changelog entry, maybe include an upgrade note pointing out the theme option for `AdminController.require_two_factor_auth`.

One minor addition which would be good is to update the Annie All Roles user fixtures to have a known otp_secret_key so we don't have to enable 2FA if working in the admin UI while developing Alaveteli.

chrismytton:

Issue #8824, opened 2025-08-15
Ability for censor rules to be applied on a per-role basis


garethrees:
At the moment text masks and censor rules are always applied on the request thread, irrespective of who you are (visitor, requester, admin, etc).

The most common issue this causes is that the requester can't see redacted email addresses (https://github.com/mysociety/alaveteli/issues/10). This often creates admin work to inspect the raw email, extract the masked content, and send it to the requester. The text masks exist to prevent harvesting / nuisance use of email addresses and phone numbers, so in almost all cases these will be fine to show to the requester (as if they'd made their request directly), though we may still want to mask WDTK-specific addresses / URLs.

I've recently taken a look at this and proposed a quick fix / workaround solution: https://github.com/mysociety/alaveteli/issues/10#issuecomment-3133109131

There have also been some cases where we'd have liked to have made a censor rule act for only certain user roles (e.g. only redact for non-requesters – https://github.com/mysociety/alaveteli/issues/5412). This part is definitely more of a "nice to have if its easy" rather than a critical feature.

It might be that if we end up consolidating the censor rules and text masks into a single system, we may be able to implement both the above changes in a way that gets us the ability to set something akin to prominence on any redaction rule for "free", avoiding the need to have yet another way of displaying content as discussed in https://github.com/mysociety/alaveteli/issues/10#issuecomment-3133109131.

I could imagine adding `prominence` to `CensorRule` and `TextMask`[^1] with the following options / behaviours:

* `hidden`: rule hides the content for all users; the original is only available to admins via the admin UI
* `requester_only`: rule hides content for anyone but the requester / admins
* `backpage`: rule hides the content for non-logged in users, but visible to any other user

[^1]: I know we don't have a `TextMask` class / instances of it, but that doesn't feel like it would be too difficult to add

PR #9319, opened 2026-06-12
[#8823] Redactable

garethrees:
## Relevant issue(s)

* Prep for https://github.com/mysociety/alaveteli/issues/8823
* Prep for https://github.com/mysociety/alaveteli/issues/8824
* Requires https://github.com/mysociety/alaveteli/pull/9302

## What does this do?

Adds explicit interfaces for getting redacted/unredacted versions of content

## Why was this needed?

* Make it easier to get different versions of content
* Reduce number of methods / method overrides that currently do the above
* Protects against information leaks by forcing developers to explicitly permit unredacted access to data

## Implementation notes

Here's how this will work:

```rb
class User < ApplicationRecord
  include Redactable
  redactable :from_name
end

user.censor_rules.create!(
  text: 'Bob Smith',
  replacement: '[name removed]',
  last_edit_editor: 'x',
  last_edit_comment: 'x'
)

msg = user.info_requests.first.outgoing_messages.first

msg.unredacted.from_name
# => "Bob Smith"

msg.redacted.from_name
# => "[name removed]"

msg.unredacted_access = false # the default
msg.from_name
# => Redactable::UnredactedAccessError (Redactable::UnredactedAccessError)

msg.unredacted_access = true
msg.from_name
# => "Bob Smith"
```

## Notes to reviewer

While this does force explicit unredacted permission, it still requires us to remember to add attributes to `redactable`. There's an alternative version of this where we by default assume attributes will hold PII unless explicitly told they won't, but that feels like it would get a bit _too_ annoying?

At the moment this applies both censor rules and text masks. We may want to split these up so that we can do e.g. (`record.censored.unmasked.foo`).

<hr>

### DIFF
--- a/app/controllers/admin_controller.rb
+++ b/app/controllers/admin_controller.rb
@@ -7,6 +7,7 @@
 class AdminController < ApplicationController
   layout "admin"
   before_action :authenticate
+  before_action { Redactable::Current.unredacted_access = true }
 
   # action to take if expecting an authenticity token and one isn't received
   def handle_unverified_request
--- a/app/models/concerns/redactable.rb
+++ b/app/models/concerns/redactable.rb
@@ -1,0 +1,24 @@
+module Redactable
+  extend ActiveSupport::Concern
+
+  included do
+    class_attribute :redactable_attrs, default: []
+
+    delegate :apply_masks, to: :info_request
+  end
+
+  class_methods do
+    # Allow classes to set attributes or methods that may contain personal data
+    # and so should be put through the redaction pipeline to check for any
+    # applicable redactions.
+    def redactable(*attrs)
+      self.redactable_attrs = attrs
+    end
+  end
+
+  def redacted
+    @redacted ||= Redacted.new(self)
+  end
+
+  end
+end
--- a/app/models/concerns/redactable.rb
+++ b/app/models/concerns/redactable.rb
@@ -1,10 +1,14 @@
 module Redactable
   extend ActiveSupport::Concern
 
+  UnredactedAccessError = Class.new(StandardError)
+
   included do
     class_attribute :redactable_attrs, default: []
 
     delegate :apply_masks, to: :info_request
+
+    attr_writer :unredacted_access
   end
 
   class_methods do
@@ -13,6 +17,19 @@ module Redactable
     # applicable redactions.
     def redactable(*attrs)
       self.redactable_attrs = attrs
+
+      # Raise an error if a redactable attribute is accessed without first
+      # explicitly allowing access by setting unredacted_access. We need to
+      # prepend a module so that this override sits above any defined methods
+      # in the including class.
+      prepend(Module.new do
+        attrs.each do |attr|
+          define_method(attr) do
+            raise Redactable::UnredactedAccessError unless unredacted_access
+            super()
+          end
+        end
+      end)
     end
   end
 
@@ -20,5 +37,11 @@ module Redactable
     @redacted ||= Redacted.new(self)
   end
 
+  def unredacted
+    @unredacted ||= Unredacted.new(self)
+  end
+
+  def unredacted_access
+    @unredacted_access || false
   end
 end
--- a/app/models/concerns/redactable.rb
+++ b/app/models/concerns/redactable.rb
@@ -42,6 +42,6 @@ module Redactable
   end
 
   def unredacted_access
-    @unredacted_access || false
+    @unredacted_access || Redactable::Current.unredacted_access || false
   end
 end
--- a/app/models/foi_attachment.rb
+++ b/app/models/foi_attachment.rb
@@ -36,6 +36,7 @@ class FoiAttachment < ApplicationRecord
   include LinkToHelper
 
   include MessageProminence
+  include Redactable
 
   include ContentType
   include Erasable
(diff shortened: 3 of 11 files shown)


garethrees:
This isn't directly tackling #8823/#8824, but in order to make it possible to show the different types of content, and do a comparison to see if its been redacted, it would be nice if we had a more uniform way of getting at the raw/redacted versions of that content.

This isn't ready for full review yet as I haven't updated all the call sites to use the new interface, but would be good to get some feedback on general approach/direction @gbp @laurentS.

gbp:
> This isn't ready for full review yet as I haven't updated all the call sites to use the new interface, but would be good to get some feedback on general approach/direction @gbp @laurentS.

Not sure about the `unredacted_access` flag. I think I would go with:
```ruby
msg.unredacted.from_name
# => "Bob Smith"

msg.from_name
# => "[name removed]"
```

So always return the redacted content for redactable fields.

garethrees:
> So always return the redacted content for redactable fields.

I was thinking that doing so would make admin form fields a bit complicated.

```erb
<!--
  This would render the *redacted* text into the form field; 
  Submitting the form would then persist the redacted text in the database
-->
<%= f.text_field :from_name %>
```

We'd have to remember to set the field value to the unredacted value all the time.

gbp:
> We'd have to remember to set the field value to the unredacted value all the time.

Nah, We can do this in the form_for:

```
form_for @user.unredacted do |f| ...
```

garethrees:
Oh, yeah, that's nice.

laurentS:
> > This isn't ready for full review yet as I haven't updated all the call sites to use the new interface, but would be good to get some feedback on general approach/direction @gbp @laurentS.
> 
> Not sure about the `unredacted_access` flag. I think I would go with:
> 
> ```ruby
> msg.unredacted.from_name
> # => "Bob Smith"
> 
> msg.from_name
> # => "[name removed]"
> ```
> 
> So always return the redacted content for redactable fields.

I'd vote for a mechanism that "fails safe", so agree with always returning the redacted content unless you explicitly ask for the unredacted version at the point of use. Like @gbp, I feel like the blanket `unredacted_access` is a footgun because it requires "closing the access" at some point, and it's easy to forget.

Thinking of how the search indexing works, how does the `unredacted` flag play with [this line](https://github.com/mysociety/alaveteli/pull/9239/changes#diff-f4d8b2c64074719cf7a334049cde9b596233b44b9108cc8084a7298236e65633R51) (pure ruby question, I can't seem to figure out the answer). Would we put something like `.unredacted.from_name` in the `searchable.admin_index` definition?

A more general question is about separation of concerns. I'm wondering if this belongs at the model level or presentation, but also can't think of a good way to do it at the presentation level without duplicating code everywhere.

laurentS:
One more comment: we use a modified variant of the request rendering template when sending the appeal PDF. How would we easily override all the `InfoRequest.show` templates to disable all redacting? Would I need to rewrite all of them? (not a major problem, just trying to think of use cases)

garethrees:
> I'd vote for a mechanism that "fails safe", so agree with always returning the redacted content unless you explicitly ask for the unredacted version at the point of use.

Yeah, I don't disagree and "redacted by default" was my first thought when I started thinking about this. I'm just not sure how nicely it will play with Rails' expectations when we're overloading attribute accessors in this way.

> I feel like the blanket `unredacted_access` is a footgun because it requires "closing the access" at some point, and it's easy to forget.

Sort of. In most cases one would use `.redacted.attr` or `.unredacted.attr` rather than setting the global flag. The use in the admin interface is session-based so no closure necessary. If we needed to do something else we'd add a block version that auto-closes access (analogous to [file access](https://rubystyle.guide/#auto-release-resources)).

> Thinking of how the search indexing works, how does the `unredacted` flag play with [this line](https://github.com/mysociety/alaveteli/pull/9239/changes#diff-f4d8b2c64074719cf7a334049cde9b596233b44b9108cc8084a7298236e65633R51) (pure ruby question, I can't seem to figure out the answer). Would we put something like `.unredacted.from_name` in the `searchable.admin_index` definition?

Yeah not sure entirely, but since its essentially just calling a ruby method I'd be surprised if there's a huge challenge here.

> A more general question is about separation of concerns. I'm wondering if this belongs at the model level or presentation, but also can't think of a good way to do it at the presentation level without duplicating code everywhere.

Yeah, similarly I also started off thinking about this in terms of a view presenter, but we use the redacted content in non-view places (search index, zip exports, etc). It basically _is_ a presenter though, just with a nice model-level way of generating the correct presentation object, so I think even if we only used it in views this would be the thing to do.

> One more comment: we use a modified variant of the request rendering template when sending the appeal PDF. How would we easily override all the `InfoRequest.show` templates to disable all redacting? Would I need to rewrite all of them? (not a major problem, just trying to think of use cases)

At the controller level you'd just do something like:

```rb
  def show
    @info_request = InfoRequest.find(params[:id]).unredacted
  end
```

Then in the views you'd call the attributes as normal on the wrapped object which behaves exactly like an `InfoRequest`; just with the specialised behaviour for certain methods. See [this post](https://ttb.software/2026/02/23/ruby-delegation-forwardable-vs-simpledelegator-vs-rails-delegate/) for a nice explanation of the various delegation types.

PR #9325, opened 2026-06-17
Authenticator 2fa backup codes

chrismytton:
## Relevant issue(s)

Part of #9241 

Follows on from #9317 

## What does this do?

Adds backup codes for 2FA which are issues when the user first enrols, and can be regenerated manually by the user. Includes a management page where the user can see when they last generated backup codes and how many backup codes they have remaining.

## Implementation notes

I originally implemented this using bcrypt, as that's the approach one of the open PRs on the active_model_otp repo took. But after thinking about it bcrypt has the following weaknesses:

- bcrypt is slow by design, but that means generating 12 codes and bcrypting them all takes ~3 seconds on my laptop, not terrible but not ideal. Same cost when checking backup codes as well.
- Because they're 6 digit codes it's a small keyspace to brute-force if someone gets hold of a DB dump.
- The otp_secret_key was still in plain text, so if someone got hold of a DB dump they wouldn't need to brute-force the backup codes because they could derive valid keys from this.

So given all that I switched the implementation to use Active Record Encryption, which encrypts the database columns using some new configuration variables. This makes it almost instant to generate or check a backup code, and makes a database dump far less useful to an attacker.

As a bonus encrypted attributes also show up as `"[FILTERED]"` in the console when e.g. inspecting a user, so should make it harder to accidentally leak these credentials.

## Screenshots

https://github.com/user-attachments/assets/470729fb-ba63-47f3-9ca0-e31c7ad505b4

[skip changelog]

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -73,6 +73,11 @@ module Alaveteli
     # Make Active Record use UTC-base instead of local time
     config.active_record.default_timezone = :utc
 
+    # otp_secret_key predates encryption, so an install holds plaintext secrets
+    # until its data migration encrypts them in place. Reading them has to keep
+    # working until that migration has run.
+    config.active_record.encryption.support_unencrypted_data = true
+
     # Disable the IP spoofing warning. This is triggered by a conflict between
     # the CLIENT_IP and X_FORWARDED_FOR headers. If you're using the example
     # nginx config, that should be setting X_FORWARDED_FOR which is used in
--- a/app/assets/javascripts/snippets.js
+++ b/app/assets/javascripts/snippets.js
@@ -53,4 +53,10 @@
     e.clearSelection();
   });
 
+  // Inline onclick handlers are blocked by the CSP, so bind print triggers here.
+  $(document).on('click', '[data-print]', function(e) {
+    e.preventDefault();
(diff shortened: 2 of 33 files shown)


config/application.rb
```diff
@@ -73,6 +73,21 @@ class Application < Rails::Application
     # Make Active Record use UTC-base instead of local time
     config.active_record.default_timezone = :utc
 
+    # Keys for Active Record Encryption, used to encrypt sensitive columns such
+    # as the users' otp_secret_key and otp_backup_codes. Alaveteli drives
+    # secrets through AlaveteliConfiguration rather than Rails credentials.
```
gbp: We have started using Rails.credentials with the recent ActionMailbox work. Am wondering if we should use it here too. Worth having a think about

gbp (changes requested):
Seems to work well. One minor comment, and looks like the CSS alignment is a bit broken other than that this looks good.

Image

garethrees:
Just to be explicit, I'm not concerned about the CSS alignment at this point – fine on chrome and firefox so looks like a safari-specific issue.

chrismytton:

PR #9337, opened 2026-06-26
Clarify attachment erasure flash error

garethrees:
Clarify attachment erasure error flash 

The most likely case for erasure to fail is if there are unmasked
attachments relating to the parent incoming message, so make that clear
in the message.

Retain the "request technical assistance" fallback for other cases.

Also add email fixture with multiple attachments to make testing easier.

[skip changelog]


gbp:
I had a go at improving this too.

I have two mutually exclusive commits on https://github.com/mysociety/alaveteli/pull/9342

One improves the error message to the admin user, the other moves the erasing into a job and ensures sibling attachments are all masked first which is probably better - welcome any thoughts you might have on these approaches

garethrees:
Superseded by https://github.com/mysociety/alaveteli/pull/9342

PR #9342, opened 2026-06-29
Update attachment erasing

gbp:
## What does this do?

Update attachment erasing

## Why was this needed?

Erasing a FoiAttachment cascades into erasing the shared RawEmail,
which can only be purged once every sibling attachment is masked or
erased. Previously an unmasked sibling made the erase fail.

Adds `FoiAttachment::EraseJob` and `FoiAttachment#earse_later` so the
work can be enqueue and run in the background.

Updates `FoiAttachment#erase` to masks its siblings, so the RawEmail
becomes erasable.

As masking can be slow, running it in the job keeps it out of the
request cycle.

[skip changelog]

### DIFF
--- a/app/jobs/foi_attachment/erase_job.rb
+++ b/app/jobs/foi_attachment/erase_job.rb
@@ -1,0 +1,16 @@
+##
+# Job to erase a FoiAttachment.
+#
+# Example:
+#   FoiAttachment::EraseJob.perform_later(
+#     FoiAttachment.first, editor: User.first, reason: 'GDPR request'
+#   )
+#
+class FoiAttachment::EraseJob < ApplicationJob
+  queue_as :default
+  unique :until_and_while_executing, on_conflict: :log
+
+  def perform(attachment, editor:, reason:)
+    attachment.erase(editor: editor, reason: reason)
+  end
+end
--- a/spec/jobs/foi_attachment/erase_job_spec.rb
+++ b/spec/jobs/foi_attachment/erase_job_spec.rb
@@ -1,0 +1,34 @@
+require 'spec_helper'
+
+RSpec.describe FoiAttachment::EraseJob, type: :job do
+  let(:info_request) { FactoryBot.create(:info_request_with_incoming) }
+  let(:incoming_message) { info_request.incoming_messages.first }
+  let(:attachment) do
+    FactoryBot.create(:body_text, incoming_message: incoming_message)
+  end
+  let(:editor) { FactoryBot.create(:admin_user) }
+  let(:reason) { 'GDPR request' }
+
+  def perform
+    described_class.new.perform(attachment, editor: editor, reason: reason)
+  end
+
+  it 'erases the attachment' do
+    expect(attachment).to receive(:erase).with(editor: editor, reason: reason)
+    perform
+  end
+
+  context 'with an unmasked sibling attachment' do
+    let!(:sibling) do
+      FactoryBot.create(
+        :body_text, :unmasked, incoming_message: incoming_message
+      )
+    end
+
+    it 'masks the sibling and erases the attachment' do
+      perform
+      expect(sibling.reload).to be_masked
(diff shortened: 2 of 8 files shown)


app/jobs/foi_attachment/erase_job.rb
```diff
@@ -0,0 +1,30 @@
...
+  unique :until_and_while_executing, on_conflict: :log
+
+  def perform(attachment, editor:, reason:)
+    mask_siblings(attachment)
+    attachment.erase(editor: editor, reason: reason)
+  end
+
+  private
+
+  def mask_siblings(attachment)
```
garethrees: I think this logic should be at model layer (`FoiAttachment::Maskable#mask_siblings`).

app/jobs/foi_attachment/erase_job.rb
```diff
@@ -0,0 +1,30 @@
+##
+# Job to erase a FoiAttachment. Sibling attachments sharing the same RawEmail
```
garethrees: This would be the ideal – you click the erase button and the app does whatever it needs to in order to perform the action.

Are you thinking that in the controller we'd just queue the erasure, with a "erasure queued" message, rather than the success/failure message that currently exists?

app/jobs/foi_attachment/erase_job.rb
```diff
@@ -0,0 +1,30 @@
...
+#   FoiAttachment::EraseJob.perform_later(
+#     FoiAttachment.first, editor: User.first, reason: 'GDPR request'
+#   )
+#
+class FoiAttachment::EraseJob < ApplicationJob
+  queue_as :default
+  unique :until_and_while_executing, on_conflict: :log
+
+  def perform(attachment, editor:, reason:)
+    mask_siblings(attachment)
```
garethrees: I think this logic should be at model layer (`FoiAttachment::Maskable#mask_siblings`) and called from within `FoiAttachment::Erasable#erase`; then add an `FoiAttachment::Erasable#erase_later` method for calling in the controller.

app/jobs/foi_attachment/erase_job.rb
```diff
@@ -0,0 +1,30 @@
...
+  end
+
+  private
+
+  def mask_siblings(attachment)
+    attachment.incoming_message.foi_attachments.each do |sibling|
+      next if sibling == attachment
+      next if sibling.masked? || sibling.erased?
+
+      FoiAttachment::MaskJob.perform_now(sibling)
```
garethrees: Can just call `sibling.mask` as that's synchronous as of 0649d41f30310b936ae67291097f6b93d7772481,  with `mask_later` being introduced in fee480bbd28e94d2a8692584571b49a0486e408a.

garethrees (commented):
Both better approaches than https://github.com/mysociety/alaveteli/pull/9337.

If we can handle masking without admins needing to do so that would be much better; couple of inline notes on that.

spec/jobs/foi_attachment/erase_job_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+  let(:editor) { FactoryBot.create(:admin_user) }
+  let(:reason) { 'GDPR request' }
+
+  def perform
+    described_class.new.perform(attachment, editor: editor, reason: reason)
+  end
+
+  it 'erases the attachment' do
+    expect(attachment).to receive(:erase).with(editor: editor, reason: reason)
+    perform
```
garethrees, line 18: better to assert `expect(attachment.reload).to be_erased`? either is fine though

garethrees (commented):
nice, miles better ✨

spec/jobs/foi_attachment/erase_job_spec.rb
gbp: Nah. Would be testing what FoiAttachment#erase does then. Thats already covered by model tests. All we really care about is if erase is called.

PR #9343, opened 2026-06-29
Add 2FA fixture data

garethrees:
## Relevant issue(s)

Part of https://github.com/mysociety/alaveteli/issues/9241.

Created as a new PR since it requires https://github.com/mysociety/alaveteli/pull/9317.

## What does this do?

Adds known 2FA fixture data for Annie all roles.

## Why was this needed?

Addresses review comment in https://github.com/mysociety/alaveteli/pull/9317#pullrequestreview-4499079210.

We frequently use Annie for development, so it would be annoying to have
to enrol the user to 2FA all the time. This adds a known secret key and
other 2FA data so we can keep the same authenticator configuration
across database resets.

## Implementation notes

You can generate the QR code to add it to your authenticator app by running this in a rails console:

```rb
puts RQRCode::QRCode.new(
  ROTP::TOTP.new('W5LLD4AHA7JAINQIQV7NGSEBU3CQGO3D').provisioning_uri
).as_ansi
```

You can also copy/paste the key in the "Enter a setup key" workflow of your app.

## Notes to reviewer

Haven't checked this with encrypted attributes enabled, so LMK if there's something different to do there.

<hr>

### DIFF
--- a/spec/controllers/admin/two_factor_controller_spec.rb
+++ b/spec/controllers/admin/two_factor_controller_spec.rb
@@ -9,6 +9,13 @@ RSpec.describe Admin::TwoFactorController do
     let!(:hotp_admin) { FactoryBot.create(:admin_user, :enable_hotp) }
     let!(:totp_admin) { FactoryBot.create(:admin_user, :enable_totp) }
 
+    # HACK: Exclude fixture data for clarity
+    before do
+      annie = users(:annie_all_roles_user)
+      annie.disable_otp
+      annie.save!
+    end
+
     it 'returns a successful response' do
       get :show
       expect(response).to be_successful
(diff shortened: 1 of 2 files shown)


chrismytton (approved):
Looks good, though I wonder if we need to document how to get the QR code for this secret, so devs can add it to their authenticator app?

This code generates a QR code in the Rails console:

```ruby
puts RQRCode::QRCode.new(
  ROTP::TOTP.new('W5LLD4AHA7JAINQIQV7NGSEBU3CQGO3D').provisioning_uri
).as_ansi
```

So perhaps we could just document that somewhere?

garethrees:
Ah I just added it by copy/pasting the secret key. I've updated the PR main body description and linked to the PR from the fixture file, which will do for now.

chrismytton:

PR #9344, opened 2026-06-29
Require 2FA at the email confirmation link for TOTP users

chrismytton:
## Relevant issue(s)

Part of https://github.com/mysociety/alaveteli/issues/9241

Follows on from https://github.com/mysociety/alaveteli/pull/9325

## What does this do?

Ensures the user is prompted for a TOTP code when signing in via the signup form's already registered flow.

## Why was this needed?

The /c/:email_token confirmation link (Users::ConfirmationsController#confirm) signed the user in directly, with no second-factor check. For a TOTP user this bypassed two factor entirely, clicking the link sent by the signup / already-registered flow granted a full session without the authenticator code.

Instead we route the confirmation sign-in for TOTP users through the existing 2FA challenges controller.

## Implementation notes

I did consider making a bigger change to ensure `sign_in` always checked two-factor, to prevent anything else like this from being introduced in the future, but that started to turn into quite a complex change, so backed out and just went for this simpler version that addresses the immediate issue.

[skip changelog]

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -170,8 +170,12 @@ class ApplicationController < ActionController::Base
   end
 
   # Establish the session and send the user on to where they were heading.
-  def complete_sign_in(user, post_redirect:, remember_me:)
+  # `circumstance` is re-set after sign_in because sign_in clears it (see
+  # clear_session_credentials). A sign-in resumed after a 2FA challenge, e.g.
+  # change_email confirmation, still needs it downstream.
+  def complete_sign_in(user, post_redirect:, remember_me:, circumstance: nil)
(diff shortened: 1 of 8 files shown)


gbp (approved):
LGTM. Verified the bypass replicates against the parent commit and the fix plus surrounding authentication test examples are green.

One follow-up thought (not for this PR): gating 2FA per sign-in entry point fails open (this bug was one forgotten path), and there are other `sign_in` callers. A global `before_action` that redirects any TOTP user whose session isn't marked 2FA-complete to the challenge would fail closed instead. I'd add on top of this, not a replacement. Happy to do that in a separate PR.

chrismytton:

Issue #9354, opened 2026-07-08
Duplicate attachments


garethrees:
On some incoming messages an attachment that is present once in the original raw email is extracted twice.

```rb
FoiAttachment.where(incoming_message_id: 3473179).pluck(:filename)
# => ["attachment.html", "FOI 1195 transactions under £250.xlsx", "attachment.html", "FOI 1195 transactions under £250.xlsx"]
```

In the raw email, `FOI 1195 transactions under £250.xlsx` is only attached a single time.

This is confusing for users, but particularly impactful for admins since spreadsheet files are scanned independently by analyzer.

This results in two notifications to the group, with extra work needed to identify those that we've already notified the public body about and link them up as duplicates.

garethrees:
<details>
<summary>🤖 Quick cause analysis suggest duplicate parsing due to async mask jobs; suggested fix is to lock the row while doing so…</summary>
Root cause

For incoming message 3473179, both duplicate pairs come from IncomingMessage#parse_raw_email! running twice concurrently on
the same message, each run inserting a full fresh set of FoiAttachment rows.

Here's the chain (app/models/incoming_message.rb:94, app/models/incoming_message/attachments.rb:179,
app/jobs/foi_attachment/mask_job.rb):

1. When an attachment's stored file/body can't be found, FoiAttachment#mail_attributes raises FoiAttachment::MissingError
(app/models/foi_attachment.rb:280/284).
2. FoiAttachment::MaskJob#perform rescues that and calls incoming_message.parse_raw_email! unconditionally to rebuild all
attachments from the raw email (app/jobs/foi_attachment/mask_job.rb:22, confirmed by
spec/jobs/foi_attachment/mask_job_spec.rb:29 — "parses raw email again").
3. Masking/locking flows (e.g. lock_all_attachments in app/models/incoming_message/attachments.rb:145, via handle_locked →
mask_later in app/models/foi_attachment/lockable.rb:99) enqueue one MaskJob per attachment, not one per message.
4. MaskJob is deduplicated with unique :until_and_while_executing (from the activejob-uniqueness gem), but the uniqueness key
is per-attachment argument — nothing stops two MaskJobs for two different attachments of the same incoming message from
executing at the same time.
5. If both attachments on that message hit MissingError around the same time (e.g. after a storage hiccup/migration), two job
workers each call parse_raw_email! on the same IncomingMessage concurrently. parse_raw_email! only wraps its work in a plain
[...]

PR #9359, opened 2026-07-10
Add a unique index to foi_attachments

laurentS:
## Relevant issue(s)

#9354 

## What does this do?

This prevents the insertion of duplicate attachments (with the same hexdigest) for a given message.

## Why was this needed?

Duplicate FoiAttachment records (for a same actual file) cause duplication of notification and other annoyances for users and admins.

## Implementation notes

The unique index in the database is the ultimate preventer, but it will not prevent the situation from arising (in such a case, we would get an error from postgres, instead of silently duplicating records).

The modified strategy on the job uniqueness *should* prevent 2 concurrent jobs from being queued up for a single incoming message, which we believe is the cause for this problem.

As detailed in the ticket, I do not believe that taking a lock is sufficient to prevent this bug from arising.

## Notes to reviewer

I have not been able to write a test that actually validates this PR, because it seems to be caused by concurrency and a fair bit of randomness. If you have any idea how to do this, please let me know.

### DIFF
--- a/app/jobs/foi_attachment/mask_job.rb
+++ b/app/jobs/foi_attachment/mask_job.rb
@@ -7,12 +7,17 @@
 #
 class FoiAttachment::MaskJob < ApplicationJob
   queue_as :default
-  unique :until_and_while_executing, on_conflict: :log
+  unique :until_executed, on_conflict: :log
 
   attr_reader :attachment
 
   delegate :incoming_message, to: :attachment
   delegate :info_request, to: :incoming_message
+
+  # use the IncomingMessage.id as uniqueness key when queuing up jobs
+  def lock_key_arguments
+    arguments.first.incoming_message.id
+  end
 
   def perform(attachment)
     @attachment = attachment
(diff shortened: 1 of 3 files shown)


garethrees:
@gbp can you take an initial look at this (and the investigation notes https://github.com/mysociety/alaveteli/issues/9354#issuecomment-4935402334) in terms of general approach and weigh in on any questions / comments.

db/migrate/20260709171753_add_unique_index_to_foi_attachments.rb
```diff
@@ -0,0 +1,7 @@
+class AddUniqueIndexToFoiAttachments < ActiveRecord::Migration[8.0]
+  def change
+    add_index :foi_attachments, [:incoming_message_id, :hexdigest], name: 'index_foi_attachments_uniqueness', unique: true
```
gbp, line 3: I think it is perfectly valid for an incoming message to have attachments with the same content more than once. Plus were we to dedup like this we would need to tidy up production DBs before this migration else it will fail.

app/jobs/foi_attachment/mask_job.rb
```diff
@@ -7,13 +7,18 @@
...
+  unique :until_executed, on_conflict: :log
 
   attr_reader :attachment
 
   delegate :incoming_message, to: :attachment
   delegate :info_request, to: :incoming_message
 
+  # use the IncomingMessage.id as uniqueness key when queuing up jobs
+  def lock_key_arguments
+    arguments.first.incoming_message.id
```
gbp, line 19:
```suggestion
    [arguments.first.incoming_message.id]
```

gbp (changes requested):
I've taken these changes on and think I got something working:
https://github.com/mysociety/alaveteli/pull/9385

gbp:
Closing in favour of https://github.com/mysociety/alaveteli/pull/9385

PR #9385, opened 2026-07-13
Fix duplicate attachments from concurrent parsing

### LINKED ISSUE #9354: Duplicate attachments

gbp:
## Relevant issue(s)

Fixes #9354 

## What does this do?

This prevents the insertion of duplicate attachments (with the same hexdigest) for a given message.

## Why was this needed?

Duplicate FoiAttachment records (for a same actual file) cause duplication of notification and other annoyances for users and admins.

## Implementation notes

The row-level lock taken in `IncomingMessage#parse_raw_email!` is the ultimate preventer. Two processes re-parsing the same message serialize on the message row: the second blocks until the first commits, then finds the freshly committed attachments and updates them in place instead of inserting a duplicate set. Unlike a unique index, nothing errors and no job fails - the losing parser just does no-op updates.

A unique index on `(incoming_message_id, hexdigest)` was considered and rejected: the same content can legitimately appear more than once on a message (Outlook permits duplicate attachments; admin file replacement can equalise content), existing production data already contains duplicates that would block the migration, and it would have turned the race into a `RecordNotUnique` error rather than a clean result.

The modified uniqueness strategy on `FoiAttachment::MaskJob` (one job per incoming message, `until_executed`) removes the most common source of concurrent re-parsing - two mask jobs for sibling attachments both rescuing `FoiAttachment::MissingError`. The row lock covers the remaining paths, EG `MaskJob.perform_now` invoked on demand from `FoiAttachment#body`, which bypasses queue-level deduplication entirely.

### DIFF
--- a/app/jobs/foi_attachment/mask_job.rb
+++ b/app/jobs/foi_attachment/mask_job.rb
@@ -7,7 +7,7 @@
 #
 class FoiAttachment::MaskJob < ApplicationJob
   queue_as :default
-  unique :until_and_while_executing, on_conflict: :log
+  unique :until_executed, on_conflict: :log
 
   attr_reader :attachment
 
--- a/app/models/incoming_message.rb
+++ b/app/models/incoming_message.rb
@@ -100,6 +100,11 @@ class IncomingMessage < ApplicationRecord
     raw_email_ensure_not_erased!
 
     ActiveRecord::Base.transaction do
+      # Lock the row to serialize concurrent re-parsing, otherwise two
+      # processes can each insert a full set of attachments. Locking a
+      # fresh copy avoids reloading self mid-parse.
+      self.class.lock.find(id) if persisted?
+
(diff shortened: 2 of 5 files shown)


laurentS (approved):
LGTM!

PR #9386, opened 2026-07-13
Change `Note#default_style` to "blue"

garethrees:
Change `Note#default_style` to "blue"

"original" is deprecated in favour of ActionText-based notes.

[skip changelog]


garethrees:
these specs all pass locally so not sure what's going on here.

garethrees:
Closing for now; will revive when we've got a bit more headspace

PR #9388, opened 2026-07-15
Fix bug when erasing attachments with replacements

gbp:
## What does this do?

- Fix bug when erasing attachments with replacements
- Adds logging when erasure fails so the reason can be debugged - both for FoiAttachment and RawEmail

## Why was this needed?

Couldn't perform admin on WDTK due to this issue

<hr>

[skip changelog]


garethrees (commented):
This allows erasure, but doesn't retain the replacement.

## Start with a replaced attachment

Screenshot 2026-07-15 at 10 46 03

Screenshot 2026-07-15 at 10 46 19

## Erase the attachment

Screenshot 2026-07-15 at 10 46 30

Screenshot 2026-07-15 at 10 46 48

Gets erased successfully – yay!

## But the replacement is gone

Screenshot 2026-07-15 at 10 47 02

The aim is for us to be able to make redactions in file-appropriate software outside of Alaveteli and upload as a replacement to continue to publish the remainder of the useful information, while being able to fully erase the originals that contain the data that we've redatced.

gbp:
> This allows erasure, but doesn't retain the replacement.

After testing again I'm still seeing the replacement attachment retained. Must be something else going on here.

garethrees:
I've uploaded a replacement here https://staging.alaveteli.org/admin/attachments/4225839/edit but now instead of it erasing the original and replacement, I just get "Could not erase this attachment. Request technical assistance." Both main body part and attachment are locked and masked.

gbp:
> I've uploaded a replacement here [staging.alaveteli.org/admin/attachments/4225839/edit](https://staging.alaveteli.org/admin/attachments/4225839/edit) but now instead of it erasing the original and replacement, I just get "Could not erase this attachment. Request technical assistance." Both main body part and attachment are locked and masked.

Oh sorry, I think the staging deployed error its currently on c5d806ea2a7e373b43ff3649c3b98ac2e54af427. I'll get this live now.

gbp:
Deploy was failing due to this: https://github.com/mysociety/whatdotheyknow-theme/commit/796aa20932cf24f10be5db8989af0aaae09b5204

garethrees:
Ah annoying thanks for fixing. Do we even need [`lib/censor_rules.rb`](https://github.com/mysociety/whatdotheyknow-theme/blob/master/lib/censor_rules.rb)? I can't imagine we're going to remove those rules/rebuild that table from scratch at this point!

garethrees:
Yeah looks like I'm getting the same on staging (`FoiAttachment#4225839`).

gbp:
Trying to workout what is wrong here. It seems to work:

https://github.com/user-attachments/assets/cc60f917-ea4a-4cfe-a4c2-aaefb9157058

PR #9390, opened 2026-07-16
PG search - admin user search

gbp:
## What does this do?

Moves the admin user search off `User.search` and onto the search backend interface, backed by a new PostgreSQL adapter that queries the `search_documents` admin index. The frontend is unaffected and stays on Xapian.

To support this it adds the minimum interface plumbing:

* a PostgreSQL `Search::Backend` adapter implementing the scoped-search part of the contract (`search_scope` + inline `reindex_later`)
* a `SEARCH_INDEX_BACKENDS` config listing which backends are kept indexed on every write (default `[xapian]`, unchanged), so a site can keep both indexes fresh while only `SEARCH_BACKEND` answers queries
* an optional `backend:` argument on the query methods to force a named backend for a single call, which the admin search uses to pin `:postgresql` regardless of `SEARCH_BACKEND`
* a `case_sensitive:` option for exact-mode substring search (`LIKE`/`ILIKE`), so the converted admin search keeps its case-insensitive behaviour.

## Why was this needed?

The admin user search is the first end-to-end consumer of the PostgreSQL
search index. It only needs the admin index, which Xapian doesn't have, so
routing it through the configured backend would break it under Xapian, and
there is no reason for it ever to use Xapian.

Converting it exercises the index, the adapter and the interface override with real traffic while the public search stays untouched.

<hr>

[skip changelog] Cover in pg-search-base


garethrees:
I don't want to hold this up so don't treat this as an issue to fix immediately, but the commit message says:

> results are capped at the search_scope limit of 1000 matches

Are we able to expand this? I've just had a look at a search for `@gmail.com` on WDTK and we have 700+ pages of results at 100 per pages, so far in excess of the 1000 match limit now imposed.

PR #9400, opened 2026-07-22
[#9399] Fix attachment name lost on content-type reset

### LINKED ISSUE #9399: Attachments all extracted with "attachment" as file basename
See this message: https://www.whatdotheyknow.com/request/parkland_and_nature_reserves_gra#incoming-3490291

If you open the raw email each attachment has a specific name.

May be related to https://github.com/mysociety/alaveteli/pull/9385 since one of the attachments triggered an analyzer scan.

gbp:
## Relevant issue(s)

Fixes #9399

## What does this do?

Fix attachment name lost on content-type reset

## Why was this needed?

Some attachment's filename appear only in the Content-Type `name`
parameter with a bare `Content-Disposition: attachment`.

Reassigning a part's content_type during normalisation rebuilds the
header and drops that parameter, so the filename was lost and falls back
to `attachment.<ext>`.

Re-attach the original name after normalisation when the part would
otherwise have no filename, restoring the sender's name.


garethrees (approved):
nice lgtm thanks for fixing so quickly

PR #9410, opened 2026-07-27
[#8696] Fix crash masking deeply nested HTML emails

### LINKED ISSUE #8696: Responses displayed as blank messages

gbp:
## Relevant issue(s)

Prior changes #8711
Fixes #8696

## What does this do?

Fix crash masking deeply nested HTML emails

## Why was this needed?

HTML nested deeper than the parser's tree depth limit raised ArgumentError, crashing the mask job. Emails can wrap content in thousands of empty <div> tags, exceeding the limit of 800.

Collapse long runs of the repeated wrapper before parsing so the document fits within the limit and its content is preserved as HTML.

Documents still too deep to parse fall back to escaped plain text, which cannot carry active markup.

<hr>

[skip changelog] already added in #8711


garethrees (commented):
Haven't run this but fix lgtm in principle so happy if you are. LMK if there's anything specific you wanted a second opinion on though.

PR #9411, opened 2026-07-27
[#8543] Stop retrying attachment masking after timeout

### LINKED ISSUE #8543: [ERROR] Sidekiq - Regexp::TimeoutError: regexp match timeout
Since Rails 8.0 we've seen an increase in attachment masking failures. This has been reported by users.

It seems to be due to https://github.com/rails/rails/pull/53490

We should increase the timeout in our Sidekiq jobs.

gbp:
## Relevant issue(s)

Fixes #8543

## What does this do?

Stop retrying attachment masking after timeout

## Why was this needed?

Masking large binary attachments can raise Regexp::TimeoutError. The failure is deterministic, so background retries and every inline fetch re-ran the same expensive regexps, burning CPU/memory and spamming exception notifications.

FoiAttachment#mask now records masking_failed_at and stops instead of raising, acting as a circuit breaker: a failed attachment is not re-masked, and FoiAttachment#body raises a fast MaskingError rather than re-running the masking.

The mask job notifies once, on the first failure, then exits cleanly so it is not retried. Clear masking_failed_at to allow another attempt.

## Screenshots

Screenshot 2026-07-28 at 10 39 48
Screenshot 2026-07-28 at 10 40 35

### DIFF
--- a/app/controllers/attachment_masks_controller.rb
+++ b/app/controllers/attachment_masks_controller.rb
@@ -16,6 +16,9 @@ class AttachmentMasksController < ApplicationController
         referer: verifier.generate(@referer)
       )
 
+    elsif @attachment.masking_failed?
+      redirect_to @referer
+
     else
       @attachment.mask_later
     end
--- a/app/controllers/attachments_controller.rb
+++ b/app/controllers/attachments_controller.rb
@@ -132,15 +132,22 @@ class AttachmentsController < ApplicationController
   def ensure_masked
     if @attachment.masked?
       yield
+    elsif @attachment.masking_failed?
+      render_attachment_masking_failed
     else
       @attachment.mask_later
 
       Timeout.timeout(5) do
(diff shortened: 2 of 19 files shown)


gbp:
@garethrees what do you reckon of this approach? Will want to update the frontend with relevant message too

garethrees:
Yep seems sensible to me.

garethrees (commented):
Looks good in principle. When we get these, what are our options for getting the information to the user?

I assume in most cases we'll be able to download raw email, manually extract attachment and send it to them, and if we can do that then we can also upload to file server and annotate?

What about using replacement? This feels like a reasonable use of that feature, but looks like it wouldn't work in this case. I don't think we should hold up merging since this fixes some technical issues, but something to consider how we document/handle this. Probably want to ticket it for now unless there's a really simple solution.

Issue #9334, opened 2026-06-23
Options to speed up reindexing


laurentS:
Reindexing objects in the postgres-backed search engine can involve 3 different levels of work:
- fully inside postgres: this is fastest, everything happens inside postgres, but is limited in what it can do (or rather it can do everything ruby can, but writing the code is a lot more involved). Updating a record usually takes less than a millisecond.
- ruby: gives more flexbility, at the expense of speed: typically each method called on an object will issue at least 1 SQL query. The round trip to the db means a slowdown by an order of magnitude at least (plus the fact that postgres could in theory apply a same function to an entire set of records in one go, where ruby will likely pull records one by one). We are in the 10 ms ballpark here for each record.
- the slowest updates are those that require pulling content (and possibly extract text from files) from external storage (ie. S3). This easily runs into 100+ms.

For sites with several tens of thousands of requests+responses+attachments, a full reindex can take multiple hours if not days.

Discourse uses postgresql search like we do in #9239, and they use a version column on their index tables to help with online index upgrades.

Our biggest reindexing job would be for the messages and attachments tables. There is no obvious way to speed up the initial indexing, as we have to pull content from storage, etc...

But index upgrades for messages and attachments can most likely be run as database jobs with no service interruptions:
- the `raw_content` is already in the db. If we modify the tsvector column, we can run a one-off SQL query to update the `content_tsv` column from the `raw_content` one (and likewise for admin columns). Even with a million records, this will finish in minutes, not days. As postgres would just update records, it does not need to take any exclusive table-level locks, so this can happen online. If the deployment happens off-peak-hours, few people will notice the disruption. This is akin to a database migration.
- Once that format upgrade is done, if the `raw_content` needs recomputing for any reason, it can be done as a long-running background job. The only impact being that search results will be somewhat suboptimal while this happens. This can prioritise records in any relevant way (most recently updated, most requested...) to minimise "downtime" (ie. non-optimal search results).

The above is only really important for messages/attachments. For public bodies, a full index update happens in about 10 minutes for 50k records (the largest list we know of). Other models are either not publicly exposed by search, and/or much smaller in volume, so should not cause user-visible downtime.

garethrees:
Do we know how long reindexing the request content takes?

On our dev server it did seem to take a number of days, but @gbp thought that this may have been down to the machine running out of memory and swapping, and suggested that the actual working time may have been much more reasonable had we had enough resources.

Have you tried it on Ma Dada data?

[...]

PR #9415, opened 2026-07-28
WIP: Make mail server logs searchable

laurentS:
## Relevant issue(s)

#8814

## What does this do?
This makes `MailServerLog`s searchable via the new postgres based search system.

## Implementation notes

See https://github.com/mysociety/alaveteli/issues/9334#issuecomment-4864895335 about speeding up this specific model's indexing, for the reasoning behind overriding the base `reindex_all` method.

Trimming log lines before indexing them will probably help with index size. The email log lines contain a lot of "trivial" text that is very unlikely to be searched (timestamps, server name which appears on every line, "postfix", etc...).
This would probably require a separate SQL function for each MTA.

## Notes to reviewer

<hr>

### DIFF
--- a/app/models/mail_server_log.rb
+++ b/app/models/mail_server_log.rb
@@ -18,6 +18,7 @@
 # Email: hello@mysociety.org; WWW: http://www.mysociety.org/
 
 class MailServerLog < ApplicationRecord
+  include Searchable
   # `serialize` needs to be called before all other ActiveRecord code.
   # See http://stackoverflow.com/a/15610692/387558
   serialize :delivery_status, coder: DeliveryStatusSerializer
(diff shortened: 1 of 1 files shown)


laurentS:
We might need to get rid of the overridden `reindex_all` method here if we merge the more generic one in #9439

garethrees:
> We might need to get rid of the overridden `reindex_all` method here if we merge the more generic one in #9439

Not sure if this is necessarily the best place for this comment, but just at a quick glance I wonder if we could have a setup where we have the generic `reindex_all`, but can then call a per-model custom implementation if defined:

```rb
class SomeRecord
  include Searchable
  # Uses the default reindex_all method
end

class AnotherRecord
  include Searchable

  def self.reindex_all
    # custom implementation
  end
```

laurentS:
As I didn't see any embargo related scopes or such for `MailServerLog`, I've skipped any similar filters in the admin search page. This means that as it is now, any admin would be able to find any log line, whether it relates to an embargoed request or not.
My hunch is that we should add a filter based on the admin user's permissions, but let me know if you disagree.

PR #9418, opened 2026-07-29
Add searchable to Citation

gbp:
## Relevant issue(s)

Alterative to #9412

## What does this do?

Add searchable to Citation

Declare Citation searchable on its source_url, allowing instances to be
searched through the new SearchDocument and the new interface.


garethrees:
Shall we merge 52869ae on to https://github.com/mysociety/alaveteli/pull/9412 to reduce these down to 1 PR?

Don't think it's worth updating the changelog model per model as it's going to get quite big.

PR #9420, opened 2026-07-29
Add tagging widget to admin edit forms

### LINKED ISSUE #8237: Make the tag edit box on the body admin page bigger
Replaces #7801, deleted in error

On 30 June 2023 @HelenWDTK wrote:

When a body has a lot of tags, it is hard to scroll through these to edit them, as the text field is quite small.

https://github.com/mysociety/alaveteli/assets/120410992/254a841f-3b5c-40fc-b7b6-79f029cda2c5

It would be good to make this box bigger, or to display the tags in some other way which makes them easier to
remove
***
On 3 November 2023 @garethrees wrote:
Swap to 'textarea'. Same width, but make it ~5 lines tall.

laurentS:
## Relevant issue(s)

Fixes #8237 

## What does this do?

This PR adds a much nicer way to edit tags on `Taggable` models (bodies, requests, messages...).
It relies on a basic API endpoint that supplies commonly used tags, and a few javascript functions that help with reusing them.
Usage should be pretty intuitive.

## Why was this needed?

See related issue. Mostly, this helps reduce typing and therefore typos, to avoid having tons of mistyped variants of a single tag. It also helps with editing objects that carry many tags.

## Implementation notes

See my comment on the issue. It would be good to test this with a large number of tags to see how the code behaves.

The js code relies on the browser's dropdown API, so visual aspect is likely going to be different between browsers/OSes.

For users with no javascript enabled, the form will simply use the current behaviour.

## Notes to reviewer

<hr>


gbp:
@laurentS just an FIY, not saying you need to change this right now. But we're trying not to introduce more jQuery based JS. Instead we've started migrating over to Stimulus controllers like app/javascript/admin/controllers/insight_form_controller.js

laurentS:
> @laurentS just an FIY, not saying you need to change this right now. But we're trying not to introduce more jQuery based JS. Instead we've started migrating over to Stimulus controllers like app/javascript/admin/controllers/insight_form_controller.js

Had a quick chat with @garethrees about this, most of the code here is copy/pasted from my previous work on MaDada's theme. Porting this to Stimulus is going to be a lot more work.

app/controllers/api_controller.rb
```diff
@@ -266,6 +267,22 @@ def body_request_events
     end
   end
 
+  # list available tags with their usage count.
+  # Used by tagging widget in admin pages.
+  def list_tags
```
garethrees: I think I'd put this `Admin::TagsController#index` for now. While tags are technically exposed in public (via certain json endpoints), its not that obvious that this is the case, and I don't think others would expect the whole list of tags to be able to be read without any authentication.

app/controllers/api_controller.rb
```diff
@@ -266,6 +267,22 @@ def body_request_events
...
+    query = <<-SQL
+      select
+        concat_ws(':',  name, value) as t,
+        count(*) as c
+      from has_tag_string_tags
+      group by (name, value)
+      having count(*) > 1
+      order by c desc;
+    SQL
```
garethrees: Also super minor nitpick but I'd capitalise the sql and indent it one level less:

```rb
query = <<-SQL
SELECT
  concat_ws(':',  name, value) AS t,
  COUNT(*) AS c
FROM has_tag_string_tags
GROUP BY (name, value)
HAVING COUNT (*) > 1
ORDER BY c DESC;
SQL
```

garethrees (commented):
Yeah this looks great!

Haven't reviewed the Javascript side but in terms of behaviour it all feels good. @gbp feel free to take a look – appreciate the desire to move to Stimulus but this is just porting in some existing theme customisations for a quick improvement so would prefer merge and improve later in this case.

In terms of performance running the query seems pretty fast on WDTK production console:

```rb
puts Benchmark.measure { ActiveRecord::Base.connection.execute(query).to_json }
# =>  0.012637   0.000000   0.012637 (  0.250566)
```

Removing the `count > 1` limit was still pretty fast, but did hit 1s+ on the initial run so I guess may start to feel slow in practice?

```rb
puts Benchmark.measure { ActiveRecord::Base.connection.execute(query_all_records).to_json }
# =>  0.639996   0.053966   0.693962 (  1.000513)

puts Benchmark.measure { ActiveRecord::Base.connection.execute(query_all_records).to_json }
# =>  0.339825   0.031224   0.371049 (  0.632706)
```

app/controllers/api_controller.rb
laurentS: Makes sense, I didn't really think about the fact that tags could potentially hold internal info.
I've moved the code, but `index` is already defined, and returns a different list of tags as HTML. So I made a differently named endpoint, with an obvious name to not overcomplicate that function. Let me know if you have a nicer way to do it.

laurentS:
> Removing the `count > 1` limit was still pretty fast, but did hit 1s+ on the initial run so I guess may start to feel slow in practice?

How many items does it return with the `count > 1` on (or how big is the JSON you get)? (for Ma Dada, I get ~310 items, under 10kB, which is no problem for JS to filter through) I'm wondering if we should limit to something like 10_000, just in case? Given this is admin facing, if it gets sluggish, we'll likely hear from users?

garethrees:
```rb
query = <<~SQL
SELECT
  concat_ws(':',  name, value) AS t,
  COUNT(*) AS c
FROM has_tag_string_tags
GROUP BY (name, value)
HAVING COUNT (*) > 1
ORDER BY c DESC;
SQL

JSON.parse(ActiveRecord::Base.connection.execute(query).to_json).size
# => 3869

query_all_records = <<~SQL
SELECT
  concat_ws(':',  name, value) AS t,
  COUNT(*) AS c
FROM has_tag_string_tags
GROUP BY (name, value)
ORDER BY c DESC;
SQL

JSON.parse(ActiveRecord::Base.connection.execute(query_all_records).to_json).size
# => 85259
```

From our side I think we'd be happy to drop the value side from the auto-complete, as I think it's pretty rare that any one body would have the same value (we tend to use them to link up to external dataset IDs and the like):

```rb
all_records_name_only = <<~SQL
SELECT
  name AS t,
  COUNT(*) AS c
FROM has_tag_string_tags
GROUP BY (name)
ORDER BY c DESC;
SQL

JSON.parse(ActiveRecord::Base.connection.execute(all_records_name_only).to_json).size
# => 2699
```

laurentS:
> From our side I think we'd be happy to drop the value side from the auto-complete, as I think it's pretty rare that any one body would have the same value (we tend to use them to link up to external dataset IDs and the like):

I wonder if we could do a more advanced filtering so that we keep `name:value` if there are fewer than, say, 10-20 values for `name`, otherwise we just keep `name:`. What happens if you do `having count(*) > 2`? The gap between 3869 and 2699 isn't that massive.

We use both types on Ma Dada. We have very common ones with 2 and 20 values (where the 20 values are reused hundreds of times, so losing them isn't ideal). I suppose we could do `name:value` -> `name_value`, but I'm not sure what the implications are, I'd need to look into it.

garethrees:
Maybe lets just ship the version with no `HAVING COUNT` filter and see how it feels on WDTK? If its fine there, then all good we can have all the tags. We can figure out reducing results if its slow enough to be annoying.

garethrees:
Decided on the call to merge as is

PR #9423, opened 2026-07-30
Add PostgreSQL search weights/exclude options

gbp:
## What does this do?

Add PostgreSQL search weights/exclude options

## Why was this needed?

Makes it possible to:
1. add tsvector label weighting per search
2. exclude fields from search - ie User#email_bounce_message

## Implementation notes

Changes the raw_content/raw_admin_content database columns to keyed JSONB columns so we can exclude the relevant fields based on the exclude option passed in to the search.

## Notes to reviewer

I had help from Claude to get the PG queries correct.
I believe the performance impact of this change is minor.

### DIFF
--- a/app/controllers/admin_user_controller.rb
+++ b/app/controllers/admin_user_controller.rb
@@ -42,7 +42,8 @@ class AdminUserController < AdminController
         backend: :postgresql,
         admin_mode: true,
         exact_mode: true,
-        case_sensitive: false
+        case_sensitive: false,
+        except: [:email_bounce_message]
       )
     end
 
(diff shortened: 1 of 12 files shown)


laurentS:
As a first thought, what use cases do you foresee for this, besides the exclusion of `email_bounce_message`? (we have zero values for this in Ma Dada, I don't know if we've missed something in our config?)

My thinking being that the added complexity of search may not be worth it if this is the only use case. I'm also wondering how likely it is that users would tune weights at query-time, we would have to provide ready-made options in the UI.

I will play a bit with your code, fill in the db with some values and run some `explain`s in postgres to understand the performance impact this would have.

As an alternative, building a specific index for that one column (bounce messages), or even just a SQL query for it, might be easier. But I can see how this could become a pattern of "just this one more query" throughout the codebase :thinking:

garethrees:
@laurentS will run an A/B test and review the explain's with @gbp.

PR #9435, opened 2026-08-05
Add admin search for public bodies

laurentS:
## Relevant issue(s)

Relates to #8814 

## What does this do?

This sets up a simple UI in /admin/bodies to allow admins to use either the legacy search or the new postgres search engine to find public bodies.

## Why was this needed?

We need to test the new search engine on actual data with actual use cases. Making it available on live sites without breaking existing workflows seems like a good way to do this.

## Implementation notes

I have included the few tests we had from previous commits to ensure a base level of consistency in results. We need a lot more!

To allow admin search through all the history of edits, I had to add an extra method. Naming improvements welcome :)

## Notes to reviewer

With Ma Dada data, searching for `mairie` (cityhall) leads to 35k+ results under the "exact tag match" section. As there is no pagination on that part of the page, the server has to issue something like 100k db queries to fetch translations+tags, then sends 100+MB to the browser. The client then struggles to render the result. I would suggest we remove the tag based search from this page, and rely on /admin/tags for that purpose, as it handles pagination and all (and we don't actually do partial tag matches anyway, so there is no additional use case handled here).

<hr>

### DIFF
--- a/app/models/public_body.rb
+++ b/app/models/public_body.rb
@@ -129,6 +129,20 @@ class PublicBody < ApplicationRecord
   translates :name, :short_name, :request_email, :url_name, :first_letter,
              :publication_scheme, :disclosure_log
 
+  searchable index: {
+               ".name": "A",
+               ".short_name": "A",
+               "home_page": "B",
+               ".notes_for_search": "C",
+               ".disclosure_log": "D",
+               ".publication_scheme": "D",
+             },
+             admin_index: {
+               ".request_email": "A",
+               "last_edit_editor": "D",
+               "last_edit_comment": "D",
+             }
+
   # Cannot be grouped at top as it depends on the `translates` macro
   include Translatable
 
@@ -215,6 +229,13 @@ class PublicBody < ApplicationRecord
     def editor
       User.find_by(url_name: last_edit_editor)
     end
+  end
+
+  # return only the notes attached to the body itself, excluding
+  # notes brought it by tags, as they pollute search results because
+  # they tend to be non-specific.
+  def notes_for_search
+    concrete_notes.map(&:body).join(' ')
   end
 
   # Public: Search for Public Bodies whose name, short_name, request_email or
--- a/app/controllers/admin_public_body_controller.rb
+++ b/app/controllers/admin_public_body_controller.rb
@@ -246,55 +246,67 @@ class AdminPublicBodyController < AdminController
       @page = nil if @page == ""
       @search_engine = params[:search_engine] || 'new'
 
-      if @search_engine == 'legacy' then
-        query = if @query
-          query_str = <<-EOF.strip_heredoc
-          (lower(public_body_translations.name)
-           LIKE lower('%'||?||'%')
-           OR lower(public_body_translations.short_name)
-           LIKE lower('%'||?||'%')
-           OR lower(public_body_translations.request_email)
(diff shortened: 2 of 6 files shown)


app/models/public_body.rb
```diff
@@ -129,6 +129,17 @@ def self.admin_title
...
 
+  searchable index: {
+               ".name": "A",
+               ".short_name": "A",
+               "home_page": "B",
+               ".notes_as_string": "C",
+               ".tag_string": "D"
+             },
+             admin_index: {
+               ".searchable_admin_field_versions_as_string": "A",
```
garethrees: Instead of all the versions, index the current values for:

```
request_email: A
last_edit_editor: D
last_edit_comment: D
```

For past versions we'll make `PublicBody::Version` itself searchable. We'll need to add a bit more UI to handle that, but we can do that in a separate pull request.

app/models/public_body.rb
```diff
@@ -129,6 +129,17 @@ def self.admin_title
   translates :name, :short_name, :request_email, :url_name, :first_letter,
              :publication_scheme, :disclosure_log
 
+  searchable index: {
+               ".name": "A",
+               ".short_name": "A",
+               "home_page": "B",
+               ".notes_as_string": "C",
+               ".tag_string": "D"
```
garethrees: Should add:

* disclosure_log:  C or D
* publication_scheme: C or D

Happy if we need to remove tag string as per the PR description

app/models/public_body.rb
```diff
@@ -129,6 +129,17 @@ def self.admin_title
   translates :name, :short_name, :request_email, :url_name, :first_letter,
              :publication_scheme, :disclosure_log
 
+  searchable index: {
+               ".name": "A",
+               ".short_name": "A",
+               "home_page": "B",
+               ".notes_as_string": "C",
```
garethrees: Should limit this to only notes on the body directly.

garethrees (commented):
Review as discussed on our catchup

laurentS:
Skipping changelog as it won't really provide meaningful info. Might add a line once we've added all models to search

garethrees:
> Skipping changelog as it won't really provide meaningful info. Might add a line once we've added all models to search

Yeah +1

laurentS:
The tests should be fixed (they pass locally).
I've also added an improvement to prevent N+1 queries when rendering the page. For a full list of 100 authorities, this drops the number of SQL queries from 208 to 11!

laurentS:
Ok, tests finally green. Sorry for the back and forth.

Just as a side note, in the tests, under "stop containers", there is a bunch of `FATAL` errors and such. It's probably nothing material, given that it's at the very end, but maybe worth a quick check to verify it's nothing that's going to blow up down the line.

gbp (approved):
@laurentS I've made some fixups. Mostly minor the main thing I changed was to switch from `newsearch` to `search_scope` with admin mode enabled so we can search request emails addresses on this form.

I think `newsearch` should be dropped from the codebase unless you have any reason not to favour `search_scope` ?

laurentS:
> I think `newsearch` should be dropped from the codebase unless you have any reason not to favour `search_scope` ?

No reason other than I forgot you had made these changes. Happy to drop the `newsearch` method indeed, I can't see a reason to keep it now.

PR #9436, opened 2026-08-05
Add merge driver for changelog conflicts

gbp:
Merging branches into develop almost always conflicts in doc/CHANGES.md, because every branch adds its entries to the top of the same section.

The conflicts are easily solved: both sides want to keep their entries, and the newest ones belong at the top.

Add script/merge-changelog, which resolves those conflicts by keeping both sides of each conflict block, placing "theirs" (the branch being merged) before "ours" (the target branch). It runs as a git merge driver, and .gitattributes associates it with doc/CHANGES.md.

The driver first attempts a standard three-way merge and only strips markers when that fails. If any markers survive it exits non-zero, so genuine conflicts still stop the merge for manual resolution.

Registering the driver is a one-time step per clone:

```
git config merge.changelog.name "Changelog merge driver"
git config merge.changelog.driver "script/merge-changelog %A %O %B"
```

<hr>

[skip changelog]

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+doc/CHANGES.md merge=changelog
--- a/script/merge-changelog
+++ b/script/merge-changelog
@@ -1,0 +1,80 @@
+#!/usr/bin/env bash
+##
+# Custom merge driver for doc/CHANGES.md
+#
+# Resolves conflicts by combining both sides of each conflict block,
+# placing "theirs" (the being-merged branch) entries before "ours"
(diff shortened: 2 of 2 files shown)


garethrees (approved):
Well TIL. Magic! 🪄

Haven't set it up or run but happy if you are. Will set it up and report any problems as new issues.

garethrees:
cc/ @laurentS

laurentS:
Neat! will this break git if I checkout an older branch that does not have the script?

PR #9437, opened 2026-08-06
Add drift tolerance to one time password checks

gbp:
Codes were verified with no drift, so one was rejected the instant its 30 seconds window ended.

RFC 6238 recommends comparing against past timestamps within the transmission delay, allowing at most one step.

Default the drift to 15 seconds - half a step. so we accepts the code that has just expired for the first 15 seconds of the next step.

Nothing becomes replayable: authenticate_totp still passes otp_last_used_at as `after`, so ROTP drops any step at or before the one already used, and OtpRateLimit still caps repeat attempts.

This fixes flaky tests where codes rolled over into the next window.

### DIFF
--- a/spec/models/user_spec.rb
+++ b/spec/models/user_spec.rb
@@ -1112,6 +1112,46 @@ RSpec.describe User do
     end
   end
 
+  describe 'TOTP drift tolerance' do
+    let(:user) { FactoryBot.create(:user, :enable_totp) }
+    let(:period_start) { Time.utc(2026, 8, 5, 10, 0, 0) }
+    let(:period_end) { period_start + 30.seconds }
+    let(:window_closes) { period_end + User::OneTimePassword::DRIFT }
+
+    # Read at the end of a step, submitted after it has ended
+    def expiring_code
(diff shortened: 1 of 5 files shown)


spec/models/user_spec.rb
```diff
@@ -1112,6 +1112,48 @@ def create_user(options = {})
...
+    let(:period_start) { Time.utc(2026, 8, 5, 10, 0, 0) }
+    let(:period_end) { period_start + 30.seconds }
+    let(:window_closes) { period_end + User::OneTimePassword::DRIFT }
+
+    # Read at the end of a step, submitted after it has ended
+    def expiring_code
+      code = nil
+      travel_to(period_end - 1.second) { code = user.otp_code }
+      code
+    end
```
chrismytton, line 1124: Looks like `#travel_to` returns its block's value, so this can be simplified a bit.

```suggestion
    def expiring_code
      travel_to(period_end - 1.second) { user.otp_code }
    end
```

chrismytton (approved):
Looks good to me! Just one suggestion for a potential code simplification, but no blockers.

PR #9439, opened 2026-08-06
Add admin search for info requests

laurentS:
## Relevant issue(s)

#8814 

## What does this do?

This PR marks `InfoRequest` as searchable with the new search engine.

## Implementation notes

This is mostly adapted from the previous PR for `PublicBody`.
I took the opportunity to optimise the view a bit as well. For a full page of requests, the rendering time is divided by 2-4 locally, and number of requests to db drop from ~416 to ~20.

This is probably one of the models where initial indexing time is going to matter a bit, at least for WDTK. If it runs at roughly the same speed as my local test, you're looking at ~6.5h of indexing time. I will try to implement a more generic version of the `reindex_all` I wrote in https://github.com/mysociety/alaveteli/pull/9415/changes, which can be used by the various models that don't need any ruby attributes in their `SearchDocument`s.

## Notes to reviewer

<hr>

### DIFF
--- a/app/controllers/admin_request_controller.rb
+++ b/app/controllers/admin_request_controller.rb
@@ -15,20 +15,14 @@ class AdminRequestController < AdminController
 
   def index
     @query = params[:query]
-    if @query
-      info_requests = InfoRequest.where(["lower(title) like lower('%'||?||'%')", @query])
+    @query = nil if @query == ''
+    @search_engine = params[:search_engine] || 'new'
+
+    if @search_engine == 'legacy'
+      index_legacy
     else
-      info_requests = InfoRequest
+      index_new
     end
-
-    if cannot? :admin, AlaveteliPro::Embargo
-      info_requests = info_requests.not_embargoed
-    end
-
-    @info_requests =
-      info_requests.
-      order(sort_query).
-      paginate(page: params[:page], per_page: 100)
   end
 
   def show
@@ -210,6 +204,44 @@ class AdminRequestController < AdminController
   end
 
   private
+
+  def index_legacy
+    if @query
+      info_requests = InfoRequest.where(["lower(title) like lower('%'||?||'%')", @query])
+    else
+      info_requests = InfoRequest
+    end
+
+    if cannot? :admin, AlaveteliPro::Embargo
+      info_requests = info_requests.not_embargoed
+    end
+
+    @info_requests =
+      info_requests.
+      order(sort_query).
+      paginate(page: params[:page], per_page: 100)
+  end
+
+  def index_new
+    info_requests =
+      if @query
+        InfoRequest.search_scope(@query,
+                                 backend: :postgresql,
+                                 admin_mode: true)
+      else
+        InfoRequest
+      end
+
+    if cannot? :admin, AlaveteliPro::Embargo
+      info_requests = info_requests.not_embargoed
+    end
+
+    @info_requests =
+      info_requests.
+      includes(:embargo, :user, public_body: :translations).
+      order(sort_query).
+      paginate(page: params[:page], per_page: 100)
+  end
 
   def info_request_params
     if params[:info_request]
--- a/app/models/concerns/searchable.rb
+++ b/app/models/concerns/searchable.rb
@@ -48,6 +48,10 @@ module Searchable
     @@locale_to_language_map[locale]
   end
 
+  def self.partition_table_name(model)
+    "search_documents_#{model.downcase.gsub('::', '_')}"
+  end
+
   # TODO: rename to `search`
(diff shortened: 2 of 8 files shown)


app/models/info_request.rb
```diff
@@ -212,6 +213,16 @@ class << self
   before_destroy :expire
   after_destroy :notify_associations, :update_counter_cache
 
+  searchable index: {
+               "title": "A",
+               "url_title": "B",
```
garethrees: We often receive user support reports where the authority just gives us the request email address to go on.

My usual approach is to extract out the `request-1234` part, then paste it into the URL bar, replacing the hyphen with a slash (`.com/request/1234` or `.com/admin/request/1234`). 

In an ideal world we'd be able to paste the full request address in the search and get to the right address. Would adding `id` and/or `idhash` here enable that? If its not a straight shot then we can defer this specific use case for later implementation.

laurentS: no it's easy, we can pass any valid SQL construct in that hash, so we could do (tested locally):
```ruby
searchable admin_index: {
  "concat('#{AlaveteliConfiguration.site_name}/request/', url_title)": "B",
...
```
as an alternative, we could preprocess the query in the admin form to identify the `https://mysite.com/request/url_name` pattern and actually search for the `url_name` part (and likewise for the id, but maybe that's too much trickery?). At that point, I suppose we could also recognise the request email pattern, and go straight to the matching request, I feel like it'd be a nice time-saver.

### pros & cons
- smaller index size with the latter
- might get some weird side effects, particularly with `id` where it could possibly match things like phone numbers or other refs, particularly when searching through docs

laurentS:
Added another commit to do a much faster `reindex_all`. For my test dataset, 53k requests, it reindexes the entire set in 1.5 seconds, VS 3min 5 sec with the default `reindex_all`. So you should get a full reindex in under a minute for WDTK, assuming similar hardware.

The method is generic enough to also take care of `MailServerLog` and any model that only indexes SQL constructs.

app/models/info_request.rb
garethrees: Yeah okay maybe let's park this then and keep the search index for genuinely free text content – probably go with the intercept pattern but can build that separately as it's not a core search feature.

app/models/concerns/searchable.rb
```diff
@@ -254,14 +254,96 @@ def newsearch(query,
...
+      end
+    end
+
+    # alternative implementation of reindex_all that works for models
+    # whose searchable.index and searchable.admin_index only contain
+    # SQL column names, and no ruby attributes. In that case, it is
+    # possible to send a single request to the db to generate the entire
+    # set of search_documents.
+    def reindex_all_inside_db
+      def partition_table_name(model)
```
gbp: These 3 nested method definitions within `reindex_all_inside_db` can leak on to all model classes. Should move to module level private helpers.

app/models/concerns/searchable.rb
```diff
@@ -254,14 +254,96 @@ def newsearch(query,
...
+          #{content_tsv_query(:index, language)},
+          #{content_tsv_query(:admin_index, language)},
+          CURRENT_TIMESTAMP,
+          CURRENT_TIMESTAMP
+          FROM #{table_name}
+      SQL
+
+      Rails.logger.debug(query)
+      ActiveRecord::Base.connection.exec_query(
+        "TRUNCATE #{partition_table_name(model_name.name)}"
```
gbp: This looks risky - dropping the whole partition what if the insert fails? We should wrap both in a transaction

app/controllers/admin_request_controller.rb
```diff
@@ -15,8 +15,16 @@ class AdminRequestController < AdminController
...
+    @search_engine = params[:search_engine]
+
     if @query
-      info_requests = InfoRequest.where(["lower(title) like lower('%'||?||'%')", @query])
+      if @search_engine == 'legacy'
+        info_requests = InfoRequest.where(
+          ["lower(title) like lower('%'||?||'%')", @query]
+        )
+      else
+        info_requests = InfoRequest.newsearch(@query, limit: 10000)
```
gbp: Thought we were going to search scope with admin mode. In fact there are a load of other fixups I made on PR #9435 which should be brought over here too.

I'll push some fixups

gbp (changes requested):

app/models/info_request.rb
```diff
@@ -212,6 +213,16 @@ class << self
...
+               "prominence_reason": "D"
```
gbp: Going to move this to the admin index unless there is a good reason to be public?

gbp (approved):
I've rebased the fixups and squashed some commits together. LGTM

Issue #9294, opened 2026-06-01
Correspondence header rounded corner CSS problem


garethrees:
In some cases there's a little white section on the corners:

<img width="774" height="74" alt="Image" src="https://github.com/user-attachments/assets/16dcaee3-8cb8-4d00-97e2-d93b8cbedbac" />

It looks like

* It only happens on the last message on the thread
* It only happens in the pro interface

<img width="100" height="374" alt="Image" src="https://github.com/user-attachments/assets/35159b87-0521-4e0a-81e1-64f3310cbf10" />

FOIMonkey:
This has been about for a bit. It does impact the outgoing message too sometimes eg this on an embargoed request.

<img width="59" height="58" alt="Image" src="https://github.com/user-attachments/assets/ba7c7747-5a8a-4670-acef-4bb6262f36c1" />

I've managed to pin it down to .correspondence--current. You'll see it if you follow a link to a specific message rather than to a request. Pro adds a fancy shadow that breaks the fix for this in theme. If you switch that off in dev tools it fixes it.

PR #9452, opened 2026-08-11
Removed background color from `.correspondence`

lucascumsille:
## Relevant issue(s)

Fixes part of: https://github.com/mysociety/alaveteli/issues/9294

## What does this do?
Instead the white background is being applied to the correspondence text and event actions. This prevents the odd white corners on top of the correspondence when it has rounded corners.

## Why was this needed?
When `.correspondence` has round corners there is an odd sharp edge coming from the white background from the element. Normally you would apply `overflow: hidden` to prevent it from happening, but we are using `position: sticky`, so we can't use overflow hidden.

## Implementation notes
This PR precedes this [WDTK one](https://github.com/mysociety/whatdotheyknow-theme/pull/2142)

## Screenshots
It's a two part PR. [Screenshots here](https://github.com/mysociety/whatdotheyknow-theme/pull/2142)

## Notes to reviewer

<hr>

[skip changelog]

### DIFF
--- a/app/assets/stylesheets/responsive/_request_style.scss
+++ b/app/assets/stylesheets/responsive/_request_style.scss
@@ -122,6 +122,11 @@ a.track-request-action {
   }
 }
 
+.correspondence_text,
+.event_actions {
+  background-color: #fff;
+}
+
 .correspondence__footer {
   background-color: #f4f4f4;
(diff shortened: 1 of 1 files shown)


garethrees (approved):
We reviewed on a call. Looks good!

Issue #7217, opened 2022-08-11
Remove note column from CSV `PublicBody` import


gbp:
With notes being in a separate table instead on the `PublicBody` instance it will be tricky to keep this feature but having notes associated by tags probably removed the need for it anyway.

garethrees:
Might be able to handle initial note creation for new bodies; finding and updating existing notes seems like it would be too complex to be of use

PR #9457, opened 2026-08-11
Update docs/running/ pages for Alaveteli 0.46.7.0

benrfairless:
## Relevant issue(s)

N/A

## What does this do?

Fixes verifiably stale content in the English `docs/running/` pages of the documentation site (gh-pages branch):

- **admin_manual.md**
  - Removes the `notes` column from the public authority CSV import documentation and example CSV. Notes moved to a separate `Note` model and the field was removed from `PublicBody` CSV import in 0.42.0.0 (commit `2a2c1da84`, which fixed mysociety/alaveteli#7217); an unrecognised column now fails the import, so the docs were actively misleading.
  - Rewrites the "Batch requests" section: the legacy batch request interface for regular users was removed in 0.41.1.0 and the `ALLOW_BATCH_REQUESTS` config option no longer exists. Batch requests are now an Alaveteli Professional feature (capped by `PRO_BATCH_AUTHORITY_LIMIT`, default 500). The admin **Can make batch requests** checkbox remains but now only exempts a user from daily rate limits (`User#exceeded_limit?`).
  - Fixes the admin request URL example: `/admin/requests/118` rather than the old `/admin/request/show/118` (per `resources :requests` under the `/admin` scope in `config/routes.rb`).
  - Matches current admin UI text: "New public authority", "Import from CSV file", "Destroy message", "New censor rule" (the "(for this request only)" suffix no longer exists), and the singular "Put misdelivered responses with the right request" to-do label.
  - Fixes a `title_url` → `url_title` typo in the holding pen instructions.
- **server.md**
  - Replaces `COOKIE_STORE_SESSION_SECRET` with `SECRET_KEY_BASE`, and `RECAPTCHA_PUBLIC_KEY`/`RECAPTCHA_PRIVATE_KEY` with `RECAPTCHA_SITE_KEY`/`RECAPTCHA_SECRET_KEY`, matching `config/general.yml-example`.
  - Recommends Nginx + Puma rather than Nginx + Thin: Alaveteli bundles Puma (`puma ~> 7.1.0` in the 0.46.7.0 Gemfile) and Thin is no longer a dependency.
  - Rewords the "Deployments" section to recommend deploying from a specific release tag and running the post-deploy script (linking to the upgrading page) instead of linking to the Capistrano-based deploy docs, which are being retired.
  - Points the PostgreSQL backup docs link at `/docs/current/` instead of the long-EOL 8.4 docs.
- **upgrading.md**
  - Removes the "Using Capistrano (recommended)" section. The standard upgrade path is now checking out the release tag and running `script/rails-post-deploy`; the former "Manual upgrade" steps become the main "How to upgrade" section.
  - Updates the "current release" version example from `0.19.0.6` to `0.46.7.0` (and the old `0.18` tag example).
- **requests.md**
  - "Allow new responses from..." has no `normal` value; the value that re-opens a request to responses is `anybody` (per the `validates_inclusion_of :allow_new_responses_from` list in `app/models/info_request.rb`).
- **holding_pen.md**
  - Fixes a broken glossary link for the spam address list, which pointed at `{{ page.baseurl }}/#spam-address-list` instead of `{{ page.baseurl }}/docs/glossary/#spam-address-list`.

## Why was this needed?

These pages contain instructions that no longer match the current Alaveteli release (0.46.7.0). The CSV import `notes` column in particular would cause an import failure if followed, the batch requests section documented a config option that no longer exists, and several config option names no longer exist under their documented names. All corrections were verified against the 0.46.7.0 Alaveteli source (`config/general.yml-example`, `config/routes.rb`, `app/models/`, `app/views/admin*`, `app/views/alaveteli_pro/`, and `doc/CHANGES.md`).

## Implementation notes

- Changes are deliberately conservative: no prose restyling beyond the rewritten batch requests and upgrade sections, Jekyll front matter and Liquid tags untouched, English pages only.
- The `server.md` config anchors (`#secret_key_base`, `#recaptcha_site_key`, `#recaptcha_secret_key`) assume the customising/config page documents these settings under those names.
- The Capistrano upgrade recommendation was removed (rather than reworded) because Capistrano support was dropped on `develop` (commit `29e015934`, "This method of deploying Alaveteli won't work") and the installing-section deploy page is being reduced to a removal note in a parallel change, so nothing here links to Capistrano workflows any more.
- Not changed, flagged for maintainers:
  - `redaction.md` says basic censor rule text replacement is always case sensitive, which is correct for 0.46.7.0 (`CensorRule#apply_to_text` does a literal `gsub`). A `case_sensitive` toggle (and diacritic-insensitive matching) exists on `develop` but is unreleased, so the page will need updating at the next release.
  - `adwords.md` documents Google Ads/Grants, not Alaveteli; nothing there is verifiable against the source tree so it was left untouched (it still uses the retired "Adwords" branding throughout).
  - Historical/anecdotal WhatDoTheyKnow content (2010 support statistics, mail import error rates) left as-is since it is presented as historical example.

## Screenshots

N/A

## Notes to reviewer

The `notes` CSV column removal and the batch requests rewrite are the most substantive changes — see mysociety/alaveteli#7217 / commit `2a2c1da84` for the CSV change, and the 0.41.1.0 entry in `doc/CHANGES.md` ("Remove legacy batch request interface") for batch requests. Everything else is label/name/link corrections verified directly against the 0.46.7.0 source. The example CSV now uses `home_page` in place of `notes`; the double-quote escaping guidance remains as prose bullets above the example.

## AI disclosure

This change was prepared with the assistance of AI coding agents (opencode, coordinated via Orca orchestration), working under human direction. All edits were verified by the agents against the Alaveteli source at the 0.46.7.0 release tag. Please review with the usual scrutiny and flag anything that reads as unverified.

---

Have you updated the changelog? [skip changelog] - documentation-site-only change.

### DIFF
--- a/docs/running/admin_manual.md
+++ b/docs/running/admin_manual.md
@@ -409,12 +409,12 @@ its *url_title*.
 <div class="attention-box info">
   <p><strong>How to find a request's <em>id</em> or <em>url_title</em></strong></p>
   <p>
-    A request's <em>id</em> is the number after <code>/show/</code> in the
-    admin interface's URL when you are looking at that request.
-    For example, if the URL is <code>/admin/request/show/118</code>, then the
+    A request's <em>id</em> is the number after <code>/admin/requests/</code>
+    in the admin interface's URL when you are looking at that request.
+    For example, if the URL is <code>/admin/requests/118</code>, then the
     <em>id</em> is <code>118</code>. Similarly, if you know that you want to see the
     admin interface's page for the request with id <code>118</code>, you know it will
-    be <code>/admin/request/show/118</code>.
+    be <code>/admin/requests/118</code>.
   </p>
   <p>
     A request's <em>url_title</em> is the part after <code>/request/</code>
--- a/docs/running/admin_manual.md
+++ b/docs/running/admin_manual.md
@@ -377,14 +377,14 @@ your attention. Click on any one of them to see the details.
 
 <div class="attention-box helpful-hint">
   If the message does not belong to any request, you can delete it instead.
-  Simply click on the <strong>Destroy Message</strong> button instead of
+  Simply click on the <strong>Destroy message</strong> button instead of
   redelivering it.
 </div>
 
 When you inspect a message, you may see a guess made by Alaveteli as to which
 request the message belongs to. Check this request. If the guess is right
 &mdash; the incoming email really is a response to that request &mdash;
-the request's *title_url* will already be in the input box: click the
+the request's *url_title* will already be in the input box: click the
 **Redeliver to another request** button.
 
 If there is not a guess, or Alaveteli's guess is wrong, look at the  `To:`
@@ -464,7 +464,7 @@ effectively dead.
 To add an email address to the spam address list you need to copy it from an
 incoming message and paste it into the spam addresss list. The easiest way to
 do this is to click on **Summary** at the top of any admin page, and then click
-on **Put misdelivered responses with the right requests** to see the contents
+on **Put misdelivered responses with the right request** to see the contents
 of the holding pen.
 
 <div class="attention-box info">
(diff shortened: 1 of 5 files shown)


garethrees (approved):
This is great, thanks!

PR #9460, opened 2026-08-11
Show body version history

### LINKED ISSUE #7997: Show PublicBody version history when handling PublicBodyChangeRequest
Almost every time I do one of these (for an existing authority) I open the body admin show page to see the version history to e.g. check that the proposed email isn't one we've already tried, or if we're using the current email for a specific campaigning reason.

Could just extract the history code to a partial and render under the form like so:

Screenshot 2023-11-06 at 17 04 19

laurentS:
## Relevant issue(s)

Fixes #7997

## What does this do?

This is a simple refactor of the existing template for reuse in the edit view. The template itself is not modified.

## Implementation notes

The history is a bit dense when shown on the edit page, as width is more limited. I added an alternating background colour to make it somewhat easier to read.

## Screenshots
image

## Notes to reviewer

<hr>

### DIFF
--- a/app/views/admin_public_body/edit.html.erb
+++ b/app/views/admin_public_body/edit.html.erb
@@ -41,6 +41,9 @@
         (this is permanent!)
       <% end %>
     </div>
+    <hr>
+    <h2>History</h2>
+    <%= render partial: 'edit_history', locals: {versions: @public_body.versions.order(version: :desc)} %>
 
   </div>
   <div class="span4">
--- a/app/views/admin_public_body/show.html.erb
+++ b/app/views/admin_public_body/show.html.erb
@@ -69,48 +69,7 @@
 
 <h2>History</h2>
 
-<% @versions.each_with_index do |version, i| %>
-  <div class="row">
-    <div class="span2">
-      <b>
-        <%= "Version #{ version.version }" %>
-      </b>
-    </div>
-
-    <div class="span4">
-      <%= version.updated_at.to_fs(:db) %>
-      (<%= time_ago_in_words(version.updated_at) %> ago)
-    </div>
-
-    <div class="span6">
-      <p>
-        “<%= h(version.last_edit_comment) %>”
-
-        <% if version.editor %>
-          – <%= link_to version.editor.name, admin_user_path(version.editor) %>
-        <% else %>
-          – <%= version.last_edit_editor %>
-        <% end %>
-      </p>
-
-      <% if i == @versions.length - 1 %>
-        <ul>
-          <li>This is the first version.</li>
-        </ul>
-      <% else %>
-        <ul>
-          <% version.compare(@versions[i+1]) do |change| %>
-            <li>
-              <%= change[:name] %> was changed from
-              <code><%= change.fetch(:from, '-') %></code> to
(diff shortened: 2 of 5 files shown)


app/views/admin_public_body/edit.html.erb
```diff
@@ -41,6 +41,10 @@
         (this is permanent!)
       <% end %>
     </div>
+    <hr>
+    <h2>History</h2>
+    <%= render :partial => 'edit_history', :locals => {:versions => @public_body.versions.order(version: :desc)} %>
```
garethrees: Would err towards the more modern ruby syntax for hashes by default:

```diff
- <%= render :partial => 'edit_history', :locals => {:versions => @public_body.versions.order(version: :desc)} %>
+ <%= render partial: 'edit_history', locals: { versions: @public_body.versions.order(version: :desc)} %>
```

app/views/admin_public_body/edit.html.erb
```diff
@@ -52,3 +56,4 @@
     <%= render :partial => 'tag_help' %>
   </div>
 </div>
+
```
garethrees: Unnecessary extra whitespace

doc/CHANGES.md
```diff
@@ -205,7 +206,8 @@
 The following templates have been changed. Please use `script/reconcile-theme`
 to update overrides in your theme to match the new templates.
 
-    None yet
+    app/views/admin_public_body/edit.html.erb
+    app/views/admin_public_body/show.html.erb
```
garethrees: We tend to generate these at release time so no need to do so on a per-PR level

app/views/admin_public_body/show.html.erb
```diff
@@ -69,48 +69,7 @@
...
-              <code><%= change.fetch(:from, '-') %></code> to
-              <code><%= change.fetch(:to, '-') %></code>
-            </li>
-          <% end %>
-        </ul>
-      <% end %>
-    </div>
-  </div>
-<% end %>
+<%= render :partial => 'edit_history', :locals => {:versions => @public_body.versions.order(version: :desc)} %>
```
garethrees: Same note RE: hash syntax above

garethrees (approved):
Looks great! Thanks!

Left some _very_ minor nitpicks, but would be totally happy to merge as is.

Also, for smaller one-commit features we tend to squash the changelog update into the main commit, but no big deal at all.

app/views/admin_public_body/edit.html.erb
laurentS: Ok, I went with the old syntax for consistency with the rest of the file, but I prefer the new one, personally :)
Updated.

laurentS:
> for smaller one-commit features we tend to squash the changelog update into the main commit, but no big deal at all.

This is mostly a mark of me forgetting to update it :sweat_smile:

Updated as per your comments above.

app/views/admin_public_body/edit.html.erb
garethrees: Either is fine, but I'd go for new style by default unless it creates hassle (e.g. if you're changing a large existing hash, and  the rest of the commit is quite large, then sometimes its more sensible to just add the new entry in the old style).

PR #9461, opened 2026-08-11
Allow admins to hide a request without notifying the user

### LINKED ISSUE #8092: Allow admins to use the request hiding interface without sending a message to the user
At WDTK, in cases where admins are hiding an old request, we do not always think it is appropriate to inform a user that their request has been hidden. This is for a variety of reasons.

However, it is useful to be able to use the hiding interface on the main request administration page, particularly given the prominence reason it sets.

Therefore, it would be ideal if an option could be selected to not automatically send a message to the user. The default could change depending on the date the request was made or of the last action (ideally, from my perspective, the last action the user personally took).

laurentS:
## Relevant issue(s)

Fixes #8092 

## What does this do?

This adds an extra button on the /admin/requests/<nnnn> page that allows hiding a request with or without notifying the user by email.

## Why was this needed?

See issue above.

## Screenshots
image

## Notes to reviewer

<hr>

### DIFF
--- a/app/controllers/admin_request_controller.rb
+++ b/app/controllers/admin_request_controller.rb
@@ -190,16 +190,21 @@ class AdminRequestController < AdminController
       @info_request.save!
 
       if ! @info_request.is_external?
-        ContactMailer.from_admin_message(
-          @info_request.user.name,
-          @info_request.user.email,
-          subject,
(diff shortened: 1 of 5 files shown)


garethrees (approved):
Wonderful thanks!

PR #9465, opened 2026-08-13
Warn users that their latest request might not be listed

laurentS:
## Relevant issue(s)

N/A

## What does this do?

This PR adds a note on a user's page to warn them that recent requests may not yet be listed.

This is only shown to the user themselves, not others, and only if they created a request within the last hour.

## Why was this needed?

Various support emails show that users get confused when they post a request, and try to see it on `/user/<myself>`. As xapian might take a few minutes to update, they think it wasn't saved, leading to actions like resending the request (which bothers public bodies) or emailing support. Hopefully this note helps reduce these occurrences.

## Implementation notes

This will not be needed anymore once we switch over to the new search engine.

## Screenshots

image

## Notes to reviewer

<hr>

### DIFF
--- a/app/controllers/user_controller.rb
+++ b/app/controllers/user_controller.rb
@@ -43,6 +43,7 @@ class UserController < ApplicationController
     set_show_requests if @show_requests
 
     @private_requests = []
+    @has_recent_requests = false
 
     if @is_you
       private_requests =
@@ -71,6 +72,15 @@ class UserController < ApplicationController
       @track_things_grouped = @track_things.group_by(&:track_type)
       # Requests you need to describe
       @undescribed_requests = @display_user.get_undescribed_requests
+
+      # very recent requests might not show up immediately on a user's page
+      # leading to unneeded actions like resends, or support emails. Warn them.
+      @has_recent_requests = 
+        @display_user.
+          info_requests.
+          visible_to_requester.
+          where(created_at: ..1.hour.ago).
+          exists?
     end
 
     respond_to do |format|
--- a/app/views/user/show/_show_requests.html.erb
+++ b/app/views/user/show/_show_requests.html.erb
@@ -40,6 +40,13 @@
           <% if @is_you %>
             <%= _('You have made no Freedom of Information requests using ' \
                   'this site.') %>
+          </p>
+          <% if @has_recent_requests %>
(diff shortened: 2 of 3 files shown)


app/controllers/user_controller.rb
```diff
@@ -71,6 +72,15 @@ def show
...
       # Requests you need to describe
       @undescribed_requests = @display_user.get_undescribed_requests
+
+      # very recent requests might not show up immediately on a user's page
+      # leading to unneeded actions like resends, or support emails. Warn them.
+      @has_recent_requests = 
+        @display_user.
+          info_requests.
+          visible_to_requester.
+          where('created_at > ?', Time.zone.now - 1.hour).
```
garethrees: Nicer version:

```rb
where(created_at: ..1.hour.ago).
```

app/views/user/show/_show_requests.html.erb
```diff
@@ -69,6 +76,13 @@
         <%= @page_desc %>
       </h2>
 
+      <% if @has_recent_requests %>
```
garethrees, line 79: I'm sure there's a good reason but quicker to ask – why do we need this twice in this file?

garethrees (approved):
Looks good in principle – one query

app/controllers/user_controller.rb
laurentS: that's nicer indeed :)

app/views/user/show/_show_requests.html.erb
laurentS: I'm sure there's a way to refactor that better. The page is structured in a way that splits the view content between "there are requests" and "there are none" but then nests "@is_you" (or not) under each, hence the duplication. (I think we'd have to flip around the `if`s to make the structure less redundant)

Given that this page is probably going to be overhauled substantially when we move to the new search engine, I thought I'd not try to modify the structure too much now, and instead do a proper cleanup in a few weeks.

garethrees: Ah gotcha. Yep totally agree with the simple approach for now but just wanted to ask as it wasn't obvious from a scan.

garethrees:
Thanks for pushing the query improvement – will merge when green

Issue #9467, opened 2026-08-14
Detect disposable and relay email domains in `UserSpamScorer`


benrfairless:
Following on from a conversation on openaustralia/righttoknow#1049, where @garethrees suggested moving WhatDoTheyKnow's UserCheck.com integration into core rather than each site reimplementing it at theme level.

## The problem

Core's only email-domain defence is static lists: `DEFAULT_SUSPICIOUS_DOMAINS`, `DEFAULT_SPAM_DOMAINS` and `DEFAULT_SPAM_TLDS` in `lib/user_spam_scorer.rb`, overridable per site via `config/user_spam_scorer.yml`. Between them they cover a few dozen domains, and disposable-mail providers add domains faster than a hand-maintained list can track them.

This came up on #1473, which was closed on the basis that "we'll never win. These services are constantly cycling domains … the only way to do this would be to whitelist major email providers". That holds for a static list. A live lookup against a service that tracks these domains is the thing that changes it, and it doesn't require whitelisting anyone.

## What already exists

WhatDoTheyKnow solved this at theme level:

- mysociety/whatdotheyknow-theme#2020: `lib/user_check_integration.rb`, a `UserCheck` module wrapping `GET https://api.usercheck.com/domain/{domain}` with a 5 second timeout and a 28 day `Rails.cache` entry per domain, registering scoring methods through the API added in #8753.
- mysociety/whatdotheyknow-theme#2023: the relay domain check.
- mysociety/fixmystreet#5508: a separate UserCheck integration in FixMyStreet.

The checks, with WDTK's weights:

| Check | Score |
| --- | --- |
| `email_domain_is_disposable` | 20 |
| `email_domain_is_relay` | 10 |
| `email_domain_invalid_mx` | 5 |

## Proposal

Move that into core roughly as-is, with the three changes @garethrees suggested:

1. **Enable on API key presence.** If `USERCHECK_API_KEY` is set the checks register, otherwise they don't. The API does answer unauthenticated requests, but treating the key as the switch keeps the implementation simple and makes the opt-in explicit.
2. **Make the scores configurable** via `config/user_spam_scorer.yml`, so sites can weigh these signals for their own context. This needs a small change to `UserSpamScorer`. See below.
3. **Leave out** WDTK's `content_limit` override for accounts tagged `disposable_email`, which was a temporary measure specific to that rollout.

### Configurable scores needs a change to `UserSpamScorer`

[...]

PR #9468, opened 2026-08-14
Detect disposable and relay email domains in UserSpamScorer

benrfairless:
## Relevant issue(s)

mysociety/alaveteli#9467

## What does this do?

Adds an opt-in integration with [UserCheck](https://www.usercheck.com) to `UserSpamScorer`. When `USERCHECK_API_KEY` is set in `config/general.yml`, three extra scoring methods register, each backed by a live lookup of the email **domain** of the user being scored:

| Check | Default score |
| --- | --- |
| `email_domain_is_disposable` | 20 |
| `email_domain_is_relay` | 10 |
| `email_domain_invalid_mx` | 5 |

When the key is blank nothing registers and behaviour is unchanged.

It also makes `score_mappings` genuinely configurable: scores passed to `register_custom_scoring_method` now act as defaults which `config/user_spam_scorer.yml` can override, and String keys from the YAML config are symbolised so they hit the custom method registry.

## Why was this needed?

Core's only email-domain defence is static lists (`DEFAULT_SUSPICIOUS_DOMAINS`, `DEFAULT_SPAM_DOMAINS`, `DEFAULT_SPAM_TLDS`), and disposable-mail providers add domains faster than a hand-maintained list can track. WhatDoTheyKnow solved this at theme level (mysociety/whatdotheyknow-theme#2020, mysociety/whatdotheyknow-theme#2023); this upstreams it so each site doesn't reimplement it, per @garethrees' suggestion on openaustralia/righttoknow#1049.

## Implementation notes

- **Privacy contract** (documented on the module): only the email domain is ever sent to the API, never the local part. Successful results are cached per domain for 28 days, so an already-seen domain generates no further requests. This keeps third-party disclosure down to "someone signed up with an address at this domain", which matters for deployments with cross-border transfer rules (APP 8, GDPR Chapter V).
- **Score mapping precedence**: `#initialize` previously merged registered custom scores in *after* configured mappings, so registration always won. Registered scores are now reverse-merged as defaults.
- **String keys from YAML**: previously a configured custom method score would miss the Symbol-keyed registry and fall through to `send`, raising `NoMethodError`. Keys are now symbolised.
- **Unknown mappings**: a mapping whose method is neither registered nor an instance method is now skipped with a `Rails.logger.warn` instead of raising, so a site that configures scores for the UserCheck checks doesn't crash scoring in environments where the key is unset.
- **Admin debug page**: `sensitive_key_patterns` extended with `API_KEY` so the key renders as `[FILTERED]`.
- Deliberately leaves out WDTK's `content_limit` override for accounts tagged `disposable_email`, which was a temporary rollout measure (per the issue).

## Screenshots

N/A

## Notes to reviewer

- **Silent failure trade-off**: API timeouts/errors are logged and score nothing, so an outage degrades to current behaviour — but a site can't easily tell the checks have stopped running. This mirrors the WDTK theme implementation; happy to add more visibility if you'd prefer.
- This implements the issue as filed; there's no maintainer sign-off on the shape yet, so treat the design choices above as proposals.
- A companion documentation PR against `gh-pages` is being opened and will be cross-linked here.
- Test evidence (run via docker compose): `bundle exec rspec spec/lib/user_check_spec.rb spec/lib/user_spam_scorer_spec.rb spec/lib/configuration_spec.rb` → 146 examples, 0 failures; `spec/controllers/user_controller_spec.rb`, `spec/controllers/users/sessions_controller_spec.rb`, `spec/controllers/user_profile/about_me_controller_spec.rb` → all green.

<hr>

Have you updated the changelog? Yes — see `doc/CHANGES.md` under `# develop`.

### DIFF
--- a/config/general.yml-example
+++ b/config/general.yml-example
@@ -1023,6 +1023,30 @@ BLOCK_SPAM_SIGNINS: false
 # ---
 BLOCK_SPAM_REQUESTS: false
 
+# API key for UserCheck (https://www.usercheck.com), used to detect email
+# domains that are disposable, are relay/forwarding services, or have invalid
+# MX records. When set, the user spam scorer gains three extra checks with
+# live lookups against the UserCheck API; when blank, the checks are disabled
+# and only the static domain lists apply.
+#
+# Only the email domain is ever sent to the API, never the full address.
+# Results are cached for 28 days per domain, so a domain that has already
+# been checked generates no further requests. API failures are logged and
+# score nothing, so an outage degrades to the static list behaviour.
+#
+# The default scores (email_domain_is_disposable: 20, email_domain_is_relay:
+# 10, email_domain_invalid_mx: 5) can be tuned via score_mappings in
+# config/user_spam_scorer.yml.
+#
+# USERCHECK_API_KEY - String (default: '')
+#
+# Example:
+#
+#   USERCHECK_API_KEY: 'xxxxxxxxxxxxxxxxxxxxxxxx'
+#
+# ---
+USERCHECK_API_KEY: ''
+
 # Restrict IP addresses from some countries from performing some actions.
 # If set, requests from IP addresses in the countries specified will be
 # prevented from making new requests on the site.
--- a/config/initializers/user_spam_scorer.rb
+++ b/config/initializers/user_spam_scorer.rb
@@ -1,4 +1,5 @@
 require 'ipaddr'
+require 'user_check'
 require 'user_spam_scorer'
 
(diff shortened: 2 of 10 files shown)


benrfairless:
Companion documentation PR for alaveteli.org: mysociety/alaveteli#9469 (against `gh-pages`; marked not to merge until this lands).

garethrees:
Just quickly saying thanks so much for working on this. We'll get it reviewed as soon as we can. We're in the last phase of a funded project around improving the search so that's taking most of our attention but we'll try to get to this in a gap.

PR #9470, opened 2026-08-14
Allow admins to download batch requests

### LINKED ISSUE #5955: Admins should be able to download zips of batches
https://www.whatdotheyknow.com/info_request_batch/330

Zip download by an admin on a public batch gives “The web page at … might be temporarily down or it may have moved permanently to a new web address.ERR_INVALID_RESPONSE”.

Can work around this by logging in as the author of the batch.

benrfairless:
Fixes #5955. Downstream report with full investigation: openaustralia/righttoknow#1043.

## What's the problem?

The Downloads box on a batch request page is shown to batch owners **and admins** (`InfoRequestBatch#is_owning_user?`), but the `:download` ability and `AlaveteliPro::BatchDownloadsController` only permit **pro batch owners**:

- An admin clicking ZIP gets `ERR_INVALID_RESPONSE` — the controller scopes the lookup to `current_user.info_request_batches`, raises `RecordNotFound`, and `render_exception` answers the `.zip` format with a zero-byte body still labelled `application/zip`.
- A non-pro batch owner is bounced to the homepage by `pro_user_authenticated?`.

## What's the fix?

Per the discussion in #5955, batch download becomes a pro and admin feature (individual requests within a batch remain downloadable by anyone):

- **Ability**: `:download` now also allows anyone who `can? :admin` the batch — admins for public batches, pro admins for embargoed ones. Pro batch owners are unchanged; non-pro owners still can't bulk-export correspondence.
- **Controller**: skip `pro_user_authenticated?` (same approach as `AlaveteliPro::EmbargoesController` for admin-usable actions), look the batch up directly rather than through `current_user`, and rely on `authorize! :download`. Unauthenticated users are still asked to log in.
- **View**: the Downloads box now renders only when `can? :download`, so non-pro owners and (for embargoed batches) non-pro admins stop seeing links that never worked.

One behaviour change worth noting: a signed-in user requesting someone else's batch download now gets `PermissionDenied` (403) rather than a 404. Batches themselves are public objects, so this doesn't leak anything.

Out of scope: `render_exception` still answers non-HTML formats with a bare `head @status`, which is what turned the original 404 into an opaque browser error — that's worth a separate fix (see the downstream issue for details).

## Testing

Updated `spec/controllers/alaveteli_pro/batch_downloads_controller_spec.rb`:

- admin (non-owner) can download ZIP/CSV of a public batch
- admin cannot download an embargoed batch; pro admin can
- non-pro owner is denied
- pro user requesting someone else's batch now expects `CanCan::AccessDenied` instead of `RecordNotFound`

`spec/controllers/alaveteli_pro/batch_downloads_controller_spec.rb` (23 examples), `spec/models/ability_spec.rb` (264) and `spec/controllers/info_request_batch_controller_spec.rb` (16) all pass.

---

Written with AI assistance (Claude Code, model `anthropic.claude-fable-5`). All changes reviewed and specs run locally.

### DIFF
--- a/app/controllers/alaveteli_pro/batch_downloads_controller.rb
+++ b/app/controllers/alaveteli_pro/batch_downloads_controller.rb
@@ -6,6 +6,8 @@ class AlaveteliPro::BatchDownloadsController < AlaveteliPro::BaseController
   include ActionController::Live
 
   skip_before_action :html_response
+  skip_before_action :pro_user_authenticated?
+  before_action :user_authenticated?
 
   def show
     authorize! :download, info_request_batch
@@ -21,8 +23,17 @@ class AlaveteliPro::BatchDownloadsController < AlaveteliPro::BaseController
 
   private
 
+  def user_authenticated?
+    return if authenticated?
+
+    ask_to_login(
+      web: _('To download batch requests'),
+      email: _('Then you can download batch requests')
+    )
+  end
+
   def info_request_batch
-    @info_request_batch ||= current_user.info_request_batches.
+    @info_request_batch ||= InfoRequestBatch.
       find(params[:info_request_batch_id])
   end
 
--- a/app/models/ability.rb
+++ b/app/models/ability.rb
@@ -125,7 +125,8 @@ class Ability
 
     # Downloading batch requests
     can :download, InfoRequestBatch do |batch_request|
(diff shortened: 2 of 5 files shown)


garethrees (approved):
Looks great. Will merge assuming the specs pass (forgot to approve the run before reviewing locally).

PR #9474, opened 2026-08-17
WIP: Search through InfoRequest children

laurentS:
## Relevant issue(s)

Relates to #8814
Follows PR #9439 (this PR includes the commits in it, we should merge that one first then rebase here)

## What does this do?

This makes 3 models searchable:
- `OutgoingMessage`
- `IncomingMessage`
- `FoiAttachment`

It also sets up admin UI to actually search through those models, and links that with `InfoRequest` search.

### Notes on performance of reindexing

see also #9334 

All benchmarks below done on a random-ish sample of real files. I run `rbspy` during all the benchmarks, so there's some minor overhead.

#### OutgoingMessage

```
Benchmark.bm do |x|
  x.report { OutgoingMessage.where(id: 55_000..).limit(1000).map(&:reindex) }
end
       user     system      total        real
  23.023317   1.784569  24.807886 ( 31.737047)
```

This works out to about 30ms per `OutgoingMessage`:

- 10k messages: 5 minutes
- 2.1M messages: ~17 hours

Quick rbspy on the above:
- 29% of time is spent calculating the diff of unredacted vs redacted content (CPU bound)
- 30% is spent figuring out OutgoingMessage.body (including 24% on InfoRequest.applicable_censor_rules) (IO bound on db)
- total time spent in InfoRequest.applicable_censor_rules is 33% of total, about 5-6hrs for WDTK (IO bound on db again)
- about 30% of the total time is spent waiting on the db (some of that time is included in the above, for instance to list out censor rules)

So we have a mix of CPU+IO bound work here.

Quick win: models/info_request:1465 in applicable_censor_rules, fetch all censor rules for an `InfoRequest` in one single SQL query. This will be called something like 12M times to reindex all of WDTK. Every 1ms we save here is about 3.5hrs of indexing time saved (probably more given network travel time between rails <-> postgres is slower than on dev machine).

More work: make the `reindex_all` method `async`, so that ruby can do CPU work while waiting for the db. Async code is very painful to debug, I'd only go there if the sync approach is unacceptable. Simpler option is to run N processes, each of them in charge a chunk of objects (eg. split the ID range between them).

==> the indexing for `OutgoingMessage` was extracted to #9504 

#### IncomingMessage

The same benchmark as above for `IncomingMessage` run twice takes:
- 18.5s for 100 messages, ie. 185ms / message (recent ones, mostly on disk)
- 30s for 100 older messages, ie. 300ms / message (older ones, mostly on cloud storage)

Again, with `rbspy`, the breakdown is roughly:
- 11% of time spent fetching from storage for recent requests (mostly on disk), 43% for older requests (mostly on cloud storage)
- 22-30% of time is spent extracting text content (ie. forking to whatever child process)
- db upserts are only about 4% of time spent

Rough estimates (on disk -- on cloud):
- 10k messages: 30 -- 50 minutes
- 3.5M messages (WDTK): 4.3 days -- 12.1 days (to be compared to the 2.8 days listed in [my other comment](https://github.com/mysociety/alaveteli/issues/9334#issuecomment-5266285934))

Given that the vast majority of files are on the cloud storage, we'll be much nearer the higher bound.

#### FoiAttachment

Same benchmark again, with older/cloud files and newer/disk files:
- older/cloud: 153s (2.5 min) for 100 attachments, ie: 1.5s / attachment.

Rough estimates:
- 10k attachments: 4.2 hrs
- 6M attachments: 104 days :skull: 

Profiling:
- diffing is again ~33% of total time
- fetching files from cloud is only about 21% of total time.
- extracting text content is expensive: 49% of total time, it looks like spreadsheets are taking a lot more time than other files (~80% of extraction although they seem to represent only 10-15% of the files sampled)
- there is redundant text extraction. `FoiAttachment::body_to_text` will pull the redacted file from storage and run it through the text extraction on each call. We then do the same for the unmasked body, meaning we have 3 file downloads and 3 calls to the text extractor.

time saving options:
- deduplicate (or cache) text extraction
- censor rules don't necessarily work too well on PDF, so it might be better to pull the unredacted content once, extract text then apply censor rules to the text and diff the result. This results in a single download from cloud, and a single text extraction, but the resulting diff may be different (probably more accurate). This would not work for attachments where the original version has been removed.
- diffing redacted vs unredacted content is great for the search index size (and search speed), but it's costing us about 1 month (!!) of CPU time to index WDTK. Can we find an alternative implementation that's (a lot) faster? (see EDITS below)
- maybe worth investigating spreadsheet content extraction a bit further, as it's likely going to be very painful on WDTK, given the amount of spreadsheets there compared to my sample.

#### EDITS

- `FoiAttachment` now skips certains file types where we can't extract any content yet (images, videos)
- a super basic diff algorithm is used for now: if the strings are equal, the diff is empty, otherwise it's the entire string. While far from ideal, it's super cheap to run until we find a better option.

### Takeaways

Will iterate on this PR, but clearly, we can't spend 4 months reindexing content!

Running multiple processes will help, as the DB is unlikely to be the bottle neck for a while.

## Notes to reviewer

<hr>


garethrees:
I think it would be worth breaking `OutgoingMessage` into its own PR. That will allow us to deploy and reindex _relatively_ quickly since its the simpler of them to start testing the new search over larger free text content.

laurentS:
> I think it would be worth breaking `OutgoingMessage` into its own PR. That will allow us to deploy and reindex _relatively_ quickly since its the simpler of them to start testing the new search over larger free text content.

Done in #9504

There are lots of test failures which seem to boil down to the following:
The `after_commit` call to `reindex` forces the `IncomingMessage` to cache its body, which in turn breaks a number of tests (particularly around censor rules) which implicitly relied on the cached fields to be `nil`. This can be solved by calling `perform_enqueued_jobs` after creating a censor rule.
Alternatively, we can prevent `IncomingMessage` from reindexing if its main body content is not cached.

This second solution feels less intrusive as it prevents breaking the app's logic in a variety of places. This is what I've implemented here, based on conversation with @garethrees today.

garethrees:
@laurentS the query you asked for:

```
foi=> select content_type, count(*) from foi_attachments group by content_type order by count desc;
                               content_type                                |  count
---------------------------------------------------------------------------+---------
 text/html                                                                 | 2533533
 application/pdf                                                           |  992101
 text/plain                                                                |  984882
 image/png                                                                 |  591898
 image/jpeg                                                                |  368320
 application/vnd.ms-word                                                   |  175806
 application/vnd.openxmlformats-officedocument.wordprocessingml.document   |  137460
 application/vnd.openxmlformats-officedocument.spreadsheetml.sheet         |   97048
 image/gif                                                                 |   65698
 application/vnd.ms-excel                                                  |   27749
 message/delivery-status                                                   |   11130
 application/zip                                                           |    9060
 text/csv                                                                  |    7515
 application/octet-stream                                                  |    3137
 application/rtf                                                           |    2611
 application/vnd.openxmlformats-officedocument.presentationml.presentation |    2460
 text/rfc822-headers                                                       |    2385
 application/vnd.oasis.opendocument.text                                   |    2170
 audio/mpeg                                                                |    1952
 image/tiff                                                                |    1777
 text/rtf                                                                  |    1595
 application/gzip                                                          |    1477
 image/bmp                                                                 |    1474
 application/vnd.ms-powerpoint                                             |    1302
 application/x-ms-wmz                                                      |    1169
 image/jpg                                                                 |     947
 application/x-gzip                                                        |     748
 image/x-citrix-jpeg                                                       |     706
 application/vnd.ms-excel.sheet.macroenabled.12                            |     477
 application/oleobject                                                     |     397
 text/css                                                                  |     281
 application/csv                                                           |     204
 application/vnd.ms-excel.sheet.binary.macroenabled.12                     |     176
 text/x-vcard                                                              |     175
 application/vnd.openxmlformats-officedocument.wordprocessingml.template   |     160
 application/vnd.ms-word.document.macroenabled.12                          |     158
 application/x-microsoft-rpmsg-message                                     |     143
 application/x-mspublisher                                                 |     140
 text/xml                                                                  |     138
 image/x-citrix-gif                                                        |     135
 text/x-pdf                                                                |     128
 application/x-7z-compressed                                               |     125
 application/x-dbf                                                         |     105
 application/x-empty                                                       |     103
 document/pdf                                                              |     101
 audio/wav                                                                 |     100
 application/x-pdf                                                         |     100
 application/vnd.oasis.opendocument.spreadsheet                            |      95
 application/vnd.visio                                                     |      93
 application/vnd.ms-office                                                 |      87
 application/xml                                                           |      71
 image/svg+xml                                                             |      66
 application/octet                                                         |      65
 application/pgp-signature                                                 |      49
 image/pjpeg                                                               |      49
 application/x-shapefile                                                   |      45
 application/softgrid-pdf                                                  |      43
 application/unknown                                                       |      42
 image/x-png                                                               |      41
 application/softgrid-png                                                  |      30
 video/mp4                                                                 |      24
 text/x-c                                                                  |      22
 application/vnd.ms-officetheme                                            |      21
 application/vnd.ms-publisher                                              |      21
 application/x-download                                                    |      20
 application/json                                                          |      19
 application/x-compressed                                                  |      19
 application/x-rar                                                         |      19
 video/x-flv                                                               |      18
 application/vnd.oasis.opendocument.presentation                           |      18
 image/xbitmap                                                             |      18
 application/pkcs7-signature                                               |      17
 image/svg                                                                 |      14
 message/rfc822-headers                                                    |      14
 application/x-msexcel                                                     |      13
 application/vnd.openxmlformats                                            |      13
 application/vnd.openxmlformats-officedocument.spreadsheetml.template      |      13
 application/x-sqlite3                                                     |      12
 application/vnd.ms-powerpoint.presentation.macroenabled.12                |      12
 application/msword.document                                               |       9
 application/softgrid-jpg                                                  |       9
 application/binary                                                        |       9
 application/vnd.ms-word.template.macroenabled.12                          |       8
 text/x-fortran                                                            |       8
 application/vnd.ms-xpsdocument                                            |       7
 message/xdelivery-status                                                  |       7
 application/vnd.ms-project                                                |       7
 application/jpg                                                           |       6
 message/global                                                            |       6
 image/btl                                                                 |       6
 text/x-pascal                                                             |       6
 application/x-rpt                                                         |       6
 application/applefile                                                     |       5
 application/vnd.sun.xml.writer                                            |       4
 text/x-mail                                                               |       4
 application/val                                                           |       4
 image/webp                                                                |       4
 text/calendar                                                             |       4
 application/ms-word                                                       |       4
 application/vnd.adobe.pdfxml                                              |       3
 image/pdf                                                                 |       3
 image/vnd.dwg                                                             |       3
 application/x-msword                                                      |       3
 text/x-asm                                                                |       3
 application/x-wine-extension-ini                                          |       3
 audio/mid                                                                 |       3
 audio/mp3                                                                 |       3
 application/vnd.openxmlformats-officedocument.presentationml.template     |       3
 message/global-delivery-status                                            |       3
 application/vnd.openxmlformats-officedocument.presentationml.slideshow    |       3
 application/wordperfect5.1                                                |       2
 application/x-geopackage                                                  |       2
 application/x-pkcs7-signature                                             |       2
 application/vnd.ms-visio.viewer                                           |       2
 audio/mp4                                                                 |       2
 audio/x-ms-wma                                                            |       2
 audio/x-wav                                                               |       2
 application/vnd.lotus-wordpro                                             |       2
 image/omr                                                                 |       2
 application/postscript                                                    |       2
 image/vnd.ms-modi                                                         |       2
 image/whatdotheyknow.com                                                  |       2
 application/png                                                           |       2
 application/pdf.a520491b_3bf7_494d_8855_7fac2c6c0608                      |       2
 message/rfc822-stubbed                                                    |       2
 application/ms-excel                                                      |       2
 application/ics                                                           |       2
 text/x-diff                                                               |       2
 application/doc                                                           |       2
 text/x-uri                                                                |       2
 video/quicktime                                                           |       2
 image/001                                                                 |       1
 application/vnd.ms-excel.template.macroenabled.12                         |       1
 application/vnd.ms-powerpoint.slideshow.macroenabled.12                   |       1
 application/x-vnd.oasis.opendocument.spreadsheet                          |       1
 application/msoutlook                                                     |       1
 application/x-dosexec                                                     |       1
 application/mixed                                                         |       1
 application/x-viewer                                                      |       1
 text/tab-separated-values                                                 |       1
 application/x-tex                                                         |       1
 application/winhelp                                                       |       1
 application/vnd.openxmlformats-officedocument.wordprocessing              |       1
 application/x-sharing-metadata-xml                                        |       1
 application/vnd.ms-visio.drawing                                          |       1
 application/adobe                                                         |       1
 application/pgp-keys                                                      |       1
 application/eml                                                           |       1
 message/disposition-notification                                          |       1
 application/softgrid-gif                                                  |       1
 application/vnd.lotus-approach                                            |       1
 application/vnd.ms-cab-compressed                                         |       1
 application/vnd.openxmlformats-officedocument.spreadsheetml.sh            |       1
(153 rows)
```

laurentS:
The commits above should allow indexing without breaking the existing tests, but some of the fixes might imply that those specific points could break outside tests without mocking.

We still need to find a way to decouple reindexing from saving objects, as this causes lots of annoying side effects for `IncomingMessage` and `FoiAttachment`.

gbp:
I've been working on bring the request search over to the frontend, and I think we're going to clash on one point. Worth agreeing on it before either lands.

## The problem

Both branches need to search a request's content and get back requests. On this branch the admin page searches `search_documents` directly and gets one row per matching document, so a request shows up once for each message or attachment that matched. On mine the request listings need one `InfoRequest` per match, ranked by its best hit and paginated.

To group hits by request, each searchable type reaches its request by a different route. `InfoRequest` is the request, `OutgoingMessage` and `IncomingMessage` have `info_request_id`, and `FoiAttachment` goes through `incoming_message`. 

Your `SearchDocument::NotEmbargoedQuery` here already uses four `LEFT JOIN`s to get back to the request. A grouped request search would need the same four joins again.

## The fix

Store the top of the tree on each document. I've started work that adds `root_type` and `root_id` to `search_documents`, polymorphic like `searchable`. A record with no parent is its own root, so `InfoRequest`, `PublicBody` and `User` point at themselves. A nested model names its root in `searchable`:

```ruby
searchable index: { ... }, root: :info_request
```

That takes a belongs_to, so the in-database reindex still works. We will also need to support ` root: { type:, id: }` column pair for polymorphic parents like Citation.

With root columns in place:

- Grouping by request is `GROUP BY root_id WHERE root_type = 'InfoRequest'`, done inside the query before the limit, so pages are stable. SearchDocument.request_search on my branch does this and returns a chainable InfoRequest relation.
- NotEmbargoedQuery becomes one join from root_id to embargoes.

This PR would need `root: :info_request` on IncomingMessage, and a `has_one :info_request, through: :incoming_message` on FoiAttachment to name as its root.

Thoughts?

garethrees:
Sounds like a reasonable approach to me, and flexible such that we could do similar with a different tree structure if we added one elsewhere in the app.

gbp:
@laurentS is there a reason for why you've gone the `SearchDocument::NotEmbargoedQuery` approach instead of preventing the embargoed content from being indexed in the first place?

laurentS:
@gbp I think your proposal makes sense. Only side effects I can think of would be a bit more storage space (but the increase should be tiny compared to the rest of that table), and a tiny overhead due to extra indices during reindexing, but again negligible.

In terms of usage, being able to list requests with a pointer to the specific item that matched would be very nice, and actually make it easier in the future to build per page/per subitem search results that don't swamp the user with tons of tiny results.

To make this happen, since you're already working on it, I suggest you update the `SearchDocument` table in a small PR that you can merge quickly, then I can rebase this PR on top of that and adjust code as needed.

> @laurentS is there a reason for why you've gone the `SearchDocument::NotEmbargoedQuery` approach instead of preventing the embargoed content from being indexed in the first place?

We do want pro admins to be able to find embargoed content, for instance someone's name when they've sent a GDPR removal request.

gbp:
> We do want pro admins to be able to find embargoed content, for instance someone's name when they've sent a GDPR removal request.

Wouldn't that be stored in the admin index then? I guess your original approach was to just store the diffs in the admin which is why you have it in the public index.

I'm just concerned there is a chance private embargoed/requester only content is exposed.

laurentS:
> Wouldn't that be stored in the admin index then? I guess your original approach was to just store the diffs in the admin which is why you have it in the public index.
> 
> I'm just concerned there is a chance private embargoed/requester only content is exposed.

I think we have different use cases:
- we want pro_admins to find any content, including embargoed and censored stuff.
- we want a pro user to be able to find their own requests (and subitems)

So I did things this way so that:
- uncensored content is always in the `admin_index` (whether we store only the diff or the full uncensored content). The `admin_index` should never be accessible to non-admin users.
- an embargoed request is in the public index (in its censored version) so that non-admins can find it. But we do need an extra layer of checks to restrict who can actually find them outside of admin pages. I guess for pro users, we will need to add an extra join on `User` to check ownership of embargoed requests, and your `root` column will help with that too.

I hope I didn't miss anything here.

gbp:
> * we want a pro user to be able to find their own requests (and subitems)

Ah yes. I had forgotten this use case where they would be searching for their own embargoed requests.

app/models/foi_attachment.rb
```diff
@@ -44,6 +44,7 @@ class FoiAttachment < ApplicationRecord
   include Lockable
   include Maskable
   include Replaceable
+  include Searchable
```
gbp: To fix the CI failures I think we need to move this include after `has_one_attached :file, service: :attachments` so the file upload callback runs before `after_commit :reindex`

gbp:
@laurentS FYI - I merged #9551 to fix the hanging issue so saw yesterday - It'll error in the future instead. And merged #9552 to fix the raw PDF/ZIP dumps in the spec output on the CI runs here.

spec/factories/foi_attachments.rb
```diff
@@ -115,6 +118,9 @@
       content_type { 'application/pdf' }
       filename { 'interesting.pdf' }
       body { load_file_fixture('interesting.pdf') }
+      after(:build) do |attachment|
+        attachment.define_singleton_method(:is_indexable?) { false }
```
laurentS: @gbp I had to add this line for the "factories" test you added in https://github.com/mysociety/alaveteli/pull/9552/changes#diff-208b9a865cb5357ccae3240cdeec91de2011ed512170ef7196782226974059e3R2569 to keep passing. My understanding is the after_commit hook forces a file fetch, and overwrites the raw email patching you did the in the factory in that PR.
I don't like disabling the reindexing here, because I'm worried that we're not testing the actual behaviour, so we have a bug looming. On the other hand, the factory is already creating an object that should not exist "in real life", so maybe I worry about something that doesn't actually happen. Happy to have your thoughts on this specific point.

app/models/foi_attachment.rb
```diff
@@ -69,6 +72,58 @@ class FoiAttachment < ApplicationRecord
...
+  searchable(
+    index: {
+      ".body_for_indexing": "A"
+    },
+    admin_index: {
+      "filename": "A",
+      ".unredacted_diff_for_admin_indexing": "A",
+      "prominence_reason": "A",
+      "replaced_reason": "A"
+    }
```
gbp:
```suggestion
    },
    root: :info_request
```

app/models/incoming_message.rb
```diff
@@ -85,6 +86,30 @@ class IncomingMessage < ApplicationRecord
...
+      # We index the body of the email here instead of on FoiAttachment,
+      # as most people think of it as part of the message rather than as
+      # an attachment.
+      ".get_body_for_indexing": "A",
+      prominence_reason: "D"
+    },
+    admin_index: {
+      from_email: "D",
+      ".main_body_text_uncensored_for_indexing": "A"
+    }
```
gbp:
```suggestion
    },
    root: :info_request
```

app/models/foi_attachment.rb
laurentS: :facepalm:

spec/factories/foi_attachments.rb
gbp: We really don't want to disable the reindexing. I'm unclear why these stubs are needed but if it's a blocker revert 9552 and we can revisit that next week. We do tho want to avoid the unnecessary "Ignoring unparsable header" warnings in the test output which 9552 fixed.

Ultimately I think the reindex in the after commit should move that processing into background job which I would hope would fix this.

PR #9475, opened 2026-08-18
Drop support for PostgreSQL 13

gbp:
PostgreSQL 13 reached end of life in November 2025, so move to version 15, the release shipped with Debian Bookworm.

CI used to pull `fixmystreet/postgres`, which has no tag past 13.5. All that image adds to the official one is an en_GB locale, the same thing our own `docker/Dockerfile-postgres` does, so CI now builds and runs that instead. Its `createdb.sh` sets the databases up, replacing the psql step.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -55,20 +55,7 @@ jobs:
       fail-fast: false
       matrix:
         include:
-        - { ruby: '3.4', postgres: 13.5 }
-
-    services:
-      postgres:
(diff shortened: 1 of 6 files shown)


gbp:
@garethrees might need to rebuild your docker db image - do mind checking these changes?

garethrees:
Yep will take a look 👍

garethrees (commented):
Rebuilt docker container and all works nicely. Now running pg 15 🎉

Issue #9304, opened 2026-06-04
Test new search engine results quality and performance


laurentS:
This is a followup ticket to #9239.

Some observations about ranking of results when searching for public bodies:

- the final ranking is calculated by summing the rankings in 3 different searches: admin tsvectors, "public" tsvectors, and a sql `LIKE` search. Each of these is weighted equally. The `LIKE` search is unable to provide any meaningful ranking at this point (as it is a binary "match/no match" search), so its results negate the outputs of the tsvector searches, leading to almost random results. I see 2 options here:
  1. assign different weights to the 3 sub queries, so that `LIKE` results are ranked lower than others, but this might make it hard to actually surface results that would only match there.
  2. when `exact_mode` is on, disable the other subqueries, and do a `LIKE` search only. This would make regular admin search faster (as we wouldn't run any `LIKE` query), and we provide a checkbox for admins to choose `exact match` when they need it, which hopefully wouldn't be too often.
- tsvector ranking is calculated in postgres by reading the `content_tsv` data for each match, which could be several thousands of records for generic search queries. This can be IO-heavy. Considering that users are probably not going to page through hundreds of pages of results, maybe we should limit results to 1-2k (which is already a lot of results to look at for a human), yielding faster responses, and assume users will modify their search query if they don't find what they want.

garethrees:
Some quick notes from our call today where we focused on the question of how do we test that the new search results are good quality and have good enough performance.

The latter should be relatively easy to test, but the former is not. We have a large amount of content and our Xapian baseline isn't 

In general we need to do some research on how others test search results. We'll all (@garethrees, @gbp, @laurentS) do a bit of research and report back with some resources and a bit of a summary of what we should try.

The ideas we came up with on the call were:

* Generate a load of sample data/fixtures so that we can test at scale (rather than the handful of fixtures we currently have)
* Parse actual searches from logs; rerun these searches using old/new searches and and compare result sets
  * If result sets are broadly similar then we can assert that we've at least matched the existing baseline
  * If result sets have a wide variance then that is a sign that we should investigate manually to determine whether we've improved/worsened
* We could do a similar test where we compare against google/bing results
* We could dump the database to Elasticsearch and compare against that; in general we didn't feel great about this approach as there'd be a high setup cost and we're aiming to make the postgres search work well; not re-compare options
* We could A/B test with users; our existing [feature flag library](https://github.com/flippercloud/flipper) supports basic AB testing; 
[...]

PR #9487, opened 2026-08-24
Add PG search admin stats

gbp:
## Relevant issue(s)

https://github.com/mysociety/alaveteli/issues/9304

## What does this do?

Add PG search admin stats allow us to measure and explain PG queries from the Alaveteli admin where we've introduced the new search.

<hr>

[skip changelog]


gbp:
merged eba8ae014f9a91f12016341fd3b62daaafb11335

PR #9492, opened 2026-08-25
Update tagging widget as a Stimulus controller

gbp:
## What does this do?

Update tagging widget as a Stimulus controller

## Why was this needed?

The widget shipped in the shared admin bundle, so it loaded on every admin page and fetched the tag list whether or not there was a field to use it. It constructed markup by hand without sanitization leading to possible XSS attacks.

A partial holds the markup now, with a template for one tag. The controller clones that template and puts the tag name in as a text node, so a stored tag is never parsed as markup again.

A datalist offers the tags already in use, less the ones on the record. Space ends a tag, Tab fills the input in from the list and ends the tag once there is nothing left to fill in, and backspace in an empty box takes the last one off. Enter is read on keyup after the default action where the datalist value is selected.

Text left sitting in the box would be dropped on save, so it validates the form until it is either made into a tag or cleared.

The cross is a character rather than an image, and the colours move to the stylesheet.

<hr>

[skip changelog]


laurentS (approved):

gbp:
Merged in f4caa1563c

PR #9494, opened 2026-08-26
Drop random rotation on pro marketing highlight

gbp:
The rotation is worked out when assets are precompiled. Each build picks -1deg or 0deg, so the asset digest hash changes and the user's browser cache gets thrown away.

<hr>

[skip changelog]


lucascumsille (approved):

lucascumsille:
It looks good =)

PR #9504, opened 2026-08-27
Make OutgoingMessage searchable

laurentS:
## Relevant issue(s)

#8814 

## What does this do?

- makes `OutgoingMessage` searchable, so that we can start indexing for testing on actual data/use cases
- adds a generic content diffing method for `searchable`s (which will be reused by other models)
- adds a `start` argument to `reindex_all` to allow running indexing in batches

## Implementation notes

See notes in #9474 about indexing performance. There is still room for improvement in `applicable_censor_rules` but it's left out of this PR.

There is no code yet to search through outgoing messages from the UI, this will be addressed separately while indexing is happening.

## Notes to reviewer

Skipping changelog as per other search related PRs, just too many small changes :)

<hr>

### DIFF
--- a/app/models/concerns/searchable.rb
+++ b/app/models/concerns/searchable.rb
@@ -257,7 +257,7 @@ module Searchable
     # Reindex all instances of a model.
     # This would normally not be run beyond the initial indexing of a
     # pre-existing database.
-    def reindex_all(batch_size: 1000)
+    def reindex_all(start_id: nil, batch_count: nil)
       options = search_options
       return if options.nil?
 
@@ -272,7 +272,13 @@ module Searchable
 
       start = Time.zone.now
       count = 0
-      indexable.find_each(batch_size: batch_size) do |record|
+      if start_id.nil?
+        items_to_index = indexable
+      else
+        items_to_index = indexable.where(id: start_id..(start_id + batch_count))
+      end
+      items_to_index.find_each() do |record|
+        puts(record.id)
         record.reindex
         count += 1
       end
--- a/app/models/outgoing_message.rb
+++ b/app/models/outgoing_message.rb
@@ -29,6 +29,7 @@ class OutgoingMessage < ApplicationRecord
   include Rails.application.routes.url_helpers
   include LinkToHelper
   include Taggable
+  include Searchable
 
   include OutgoingMessage::DeliveryStatus
 
@@ -82,6 +83,17 @@ class OutgoingMessage < ApplicationRecord
 
   scope :followup, -> { where(message_type: 'followup') }
   scope :is_searchable, -> { where(prominence: 'normal') }
+
+  searchable(
+    index: {
+      ".get_text_for_indexing": "A",
+      ".safe_from_name": "D"
+    },
+    admin_index: {
+      from_name: "B",
+      prominence_reason: "D",
+    }
+  )
 
   def self.default_salutation(public_body)
     _("Dear {{public_body_name}},", public_body_name: public_body.name)


app/models/outgoing_message.rb
```diff
@@ -83,6 +84,25 @@ class OutgoingMessage < ApplicationRecord
   scope :followup, -> { where(message_type: 'followup') }
   scope :is_searchable, -> { where(prominence: 'normal') }
 
+  searchable(
+    index: {
+      prominence_reason: "D",
```
garethrees: We were going to pop this in the admin index only

app/models/outgoing_message.rb
```diff
@@ -83,6 +84,25 @@ class OutgoingMessage < ApplicationRecord
...
+      ".get_text_for_indexing": "A",
+      ".safe_from_name": "D"
+    },
+    admin_index: {
+      ".diff_of_body_redactions": "A",
+      from_name: "D"
```
garethrees: Would make this a B in the context of admin as we're quite likely to be searching by name?

garethrees (commented):
minor comments but otherwise lgtm

app/models/outgoing_message.rb
laurentS: missed that, sorry. fixed.

laurentS:
@garethrees I updated by code, because the initial version wasn't actually working :/
It looks like the `start` keyword was used by something else, so it wasn't actually passing the expected value, but something completely different. I've renamed them to something conflict-safe (hopefully).

garethrees:
@gbp can you take a look at this; feel free to make any fixes necessary to get merged and an initial indexing run kicked off. Thanks!

app/models/concerns/searchable.rb
```diff
@@ -272,7 +321,13 @@ def reindex_all(batch_size: 1000)
 
       start = Time.zone.now
       count = 0
-      indexable.find_each(batch_size: batch_size) do |record|
+      if start_id.nil? && end_id.nil?
```
gbp: end_id is undefined

app/models/concerns/searchable.rb
```diff
@@ -9,6 +9,9 @@
 # @example
 #   User.where.not(banned_at: nil).search_scope('alice').order(:created_at)
 #
+
+require 'diff/lcs'
```
gbp: This will break production deploy. Won't be in the gem bundle as this gem is dev/test only. Need to add to the Gemfile

gbp (changes requested):

app/models/outgoing_message.rb
```diff
@@ -83,6 +84,25 @@ class OutgoingMessage < ApplicationRecord
...
+      ".diff_of_body_redactions": "A",
+      from_name: "B",
+      prominence_reason: "D",
+    }
+  )
+
+  def diff_of_body_redactions
+    diff_unredacted_and_redacted_content(
+      body,
+      get_text_for_indexing
```
gbp: aren't both of these redacted already, should we use raw_body?

app/models/outgoing_message.rb
```diff
@@ -83,6 +84,25 @@ class OutgoingMessage < ApplicationRecord
...
+    },
+    admin_index: {
+      ".diff_of_body_redactions": "A",
+      from_name: "B",
+      prominence_reason: "D",
+    }
+  )
+
+  def diff_of_body_redactions
+    diff_unredacted_and_redacted_content(
```
gbp: Trying to workout what `diff_unredacted_and_redacted_content` actually does: take for example this test which passed. The returned value of `bp.co` means what exactly? How is that any use?

```ruby
  describe '#diff_of_body_redactions' do
    subject { message.diff_of_body_redactions }

    let(:message) { FactoryBot.build(:initial_request, body: body) }

    context 'when an email address is masked for indexing' do
      let(:body) { 'Write to bob@example.com please.' }
      it { is_expected.to eq('bp.co') }
    end
  end
```

app/models/concerns/searchable.rb
```diff
@@ -272,7 +321,13 @@ def reindex_all(batch_size: 1000)
...
       start = Time.zone.now
       count = 0
-      indexable.find_each(batch_size: batch_size) do |record|
+      if start_id.nil? && end_id.nil?
+        items_to_index = indexable
+      else
+        items_to_index = indexable.where(id: start_id..(start_id + batch_count))
+      end
+      items_to_index.find_each() do |record|
+        puts(record.id)
```
gbp:
```suggestion
```

gbp:
In order to get this deployed I've dropped the diffing as it doesn't seem to work as expected.

app/models/outgoing_message.rb
laurentS: what's the censor rule applied to the message? This does look very weird indeed.

garethrees: I think @gbp was saying that the [default email address text mask](https://github.com/mysociety/alaveteli/blob/develop/lib/alaveteli_text_masker.rb#L10-L13) would have matched "`bob@example.com`" and replaced it with "`[email address]`" in the above example, without needing a specific `CensorRule`.

laurentS: Just for the record, I am looking into this just now. The `diff/lcs` algorithm tries to be smart about things, but it backfires when I try to make sense of it in my subsequent code. I wrote it based on an example where the diff identified characters changed, but somehow the example you picked decides to consider that the first `b` was replaced with `[`, but then it thinks `ob@` was removed, then it matches the `e` in `exampl` with the one in `email`, and so on.
Why it does this is unclear. At this point, considering that the diffing is mainly an optimisation rather than a must-have, I'll continue without it.

PR #9505, opened 2026-08-27
[#8823] Track when a CensorRule redacts something

### LINKED ISSUE #8823: Admin UI showing which where redactions are applied and redactions apply to what

garethrees:
## Relevant issue(s)

Fixes #8823

## What does this do?

Record a CensorRule::Redaction row per (rule, redactable, attribute) whenever a rule's application to text or binary content actually changes it, so we can later show which records a given rule has affected.

## Why was this needed?

When many censor rules _could_ apply to something, its not easy to determine which ones did, if they did at all.

This is confusing for admins, and also when making erasure decisions in the code (e.g. in https://github.com/mysociety/alaveteli/pull/9138)

## Implementation notes

I was originally going to store the diff of what was removed by the redaction so that we could use this as a basis for admin search, but that got complicated so for now just adding some basic tracking.

## Screenshots

Showing which redactions are actually doing something on the request page:

Screenshot 2026-08-27 at 16 40 46

Clicking through shows which child records the redaction has been applied to:

Screenshot 2026-08-27 at 17 05 01


app/models/censor_rule/recordable.rb
```diff
@@ -0,0 +1,40 @@
+# Record when a CensorRule has actually been applied to a Redactable
+module CensorRule::Recordable
+  extend ActiveSupport::Concern
+
+  PATTERN_COLUMNS = %w[text regexp case_sensitive ignore_diacritics].freeze
```
garethrees, line 5: couldn't think of a better name – its the attributes where, if changed, may result in different content being redacted so we'd reprocess (but e.g. changing the last edit comment wouldn't make a difference to the redacted content)

app/models/outgoing_message.rb
```diff
@@ -203,6 +203,7 @@ def body(options = {})
       info_request.try(:applicable_censor_rules) or []
     end
 
+    # FIXME: We appear not to be applying text masks here
```
garethrees, line 209: Not sure why we aren't applying masks here. I haven't looked in to why as it's a bit outside the scope of the PR and this one is a bit more consequential. May revisit before merge but think we can get by without doing so.

app/models/censor_rule/redaction.rb
```diff
@@ -0,0 +1,41 @@
...
+#  created_at         :datetime         not null
+#  updated_at         :datetime         not null
+#
+class CensorRule::Redaction < ApplicationRecord
+  belongs_to :censor_rule
+  belongs_to :redactable, polymorphic: true
+
+  validates_presence_of :redacted_attribute
+
+  scope :for_request, ->(info_request) {
```
garethrees, line 19: Might be a simpler way of writing this – please do suggest improvements!

app/models/censor_rule/recordable.rb
```diff
@@ -0,0 +1,40 @@
+# Record when a CensorRule has actually been applied to a Redactable
+module CensorRule::Recordable
```
garethrees, line 2: Not sure this is the best name either, but didn't want to clash with `Redactable` – the concern for including in records that can be redacted.

garethrees:
I haven’t fully tested it since making a couple of small tweaks, adding some tests and rebasing but it should be okay to start taking a look through.

app/models/censor_rule/redaction.rb
laurentS: Could you reuse the `applicable_censor_rules` improvements that were just merged and find corresponding redactions via `CensorRule.id`, something like this?
```ruby
where(censor_rule_id: [info_request.applicable_censor_rules.map(&:id)])
```

app/models/censor_rule/redaction.rb
```diff
@@ -0,0 +1,41 @@
...
+#  redactable_id      :bigint           not null
+#  redacted_attribute :string           not null
+#  created_at         :datetime         not null
+#  updated_at         :datetime         not null
+#
+class CensorRule::Redaction < ApplicationRecord
+  belongs_to :censor_rule
+  belongs_to :redactable, polymorphic: true
+
+  validates_presence_of :redacted_attribute
```
laurentS, line 17: Is it worth validating against a list of acceptable values?

app/models/censor_rule/recordable.rb
laurentS: Maybe just add this comment in the code?

laurentS (commented):
A couple of thoughts after playing with this a bit:
- I have 5 global rules in my test dataset. They are shown on each `InfoRequest` admin page although they're not applied (we should update them to be more specific but that's besides the point). It could be nice to split up the list of censor rules into "Used / not used" or maybe sort them by number of actual redactions, to make it easier to see what's actually used.
- opening an old request that has 3 InfoRequest scoped rules + 5 global ones on it, the admin page first showed no redactions at all, although the request scoped ones are actually applied. After clicking "edit" on a rule and cancelling that, the count was updated, but on a different rule. I'm not sure what caused this update, but right now, the numbers shown to me are wrong. Is there going to be some rake task to update all these values? The screenshot below shows the current state. The 2 circled counts are wrong, I think they should be at least 2 (incoming messages) for both (the request page shows rules correctly applied), while the "3" looks correct. Let me know how I can help with a repro.

Image

garethrees:
> split up the list of censor rules into "Used / not used" or maybe sort them by number of actual redactions…

Interesting yeah. I don't think we should sort by this as the current sort order helps you understand the application order. I think hiding non-used rules with a little note that you can click to reveal them would be great. If I can add it trivally with what's available in bootstrap I will, but might have to come back to this as a later improvement. The colour-coding of the count badges was intended to make this as easy as possible at a quick-ish glance.

> Is there going to be some rake task to update all these values?

Yeah, we'll need to expire the requests so that they're re-applied _and_ tracked. The edit action causes the [`:expire_requests` callback](https://github.com/mysociety/alaveteli/blob/8823-censor-rule-application-tracking-v3/app/models/censor_rule/expiry.rb#L6) to fire, which is why the counts get correctly updated.

 I thought I'd mentioned we'd need a data migration task somewhere but obviously not, sorry! We'll need something like `CensorRule.find_each(&:expire_requests)`. I'll need to check, but I think that only clears the caches and then the next public load will repopulate them. For this it would be more ideal for the task to also do the repopulation so that the counts are correct even if you go direct to the admin UI for an affected request.

app/models/censor_rule/redaction.rb
garethrees: I guess we could hard code known attributes where censor rules are applied, but there are likely some themes that would need updating where they've added new attributes (e.g. `User#id_number` or something). The more ideal solution is to get https://github.com/mysociety/alaveteli/pull/9347 over the line and dynamically pull the values defined in the `redactable` calls. I'll ponder this, but my initial lean is to keep it as is for this PR.

app/models/censor_rule/redaction.rb
garethrees: Hm so that just gets me, essentially, all censor rules that _may_ apply to the request. We still need to find ones that _have_ been applied, which needs knowledge of the redactable. Pushed a much cleaner implementation of this in 4545bac1a.

laurentS:
> Interesting yeah. I don't think we should sort by this as the current sort order helps you understand the application order.

Just noting you had a sort by `created_at` somewhere, which would be different from the application order. If you reuse `applicable_censor_rules`, you should get the correct order though.

> I thought I'd mentioned we'd need a data migration task somewhere but obviously not, sorry! We'll need something like `CensorRule.find_each(&:expire_requests)`.

This would trigger a search engine reindex, so it's probably worth thinking of the workload this would cause, as you'd save/reindex the request, all messages and attachments. I don't remember how many censor rules you have on wdtk, but the impact may not be trivial.

> For this it would be more ideal for the task to also do the repopulation so that the counts are correct even if you go direct to the admin UI for an affected request.

I would agree with that.

garethrees:
> This would trigger a search engine reindex… the impact may not be trivial

Yeah, this is something to consider. Just been looking at a data migration task and well, its a lot of requests that we'd need to expire:

```rb
ids = CensorRule.find_each.flat_map { |rule| rule.censorable_requests.ids }.uniq
infoRequest.where(id: ids).count
# => 84980
```

garethrees:
> Just noting you had a sort by `created_at` somewhere, which would be different from the application order. If you reuse `applicable_censor_rules`, you should get the correct order though.

Yeah we already use this when [rendering the list of censor rules](https://github.com/mysociety/alaveteli/blob/8823-censor-rule-application-tracking-v3/app/views/admin_request/show.html.erb#L407-L409). The [`created_at` ordering](https://github.com/mysociety/alaveteli/blob/8823-censor-rule-application-tracking-v3/app/controllers/admin/censor_rules/redactions_controller.rb#L13) is for the page that lists the content that's been actually redacted by a specific rule (second screenshot in PR body).

app/models/outgoing_message.rb
garethrees: We call `remove_privacy_sensitive_things!` in `get_body_for_html_display`, but that only redacts emails – not e.g. the mobile number mask. In any case I think this is a different issue.

garethrees:
@laurentS I couldn't reproduce the issue after discussion on the call, but I have now pushed up a backfill script. After updating your local branch (note there have been force-pushes), I'd try the following:

```sh
# Enable the feature
bin/rails runner "AlaveteliFeatures.backend.enable(:redaction_tracking)"

# Clear all redaction tracking (check the output is 0 – i.e. no tracked redactions)
bin/rails runner "CensorRule::Redaction.delete_all ; puts CensorRule::Redaction.count"

# Run the backfill script
bin/rails temp:record_censor_rule_redactions
```

lib/tasks/temp.rake
```diff
@@ -115,6 +115,8 @@ namespace :temp do
         attempt(errors, attachment) do
           attachment.redacted_filename
           attachment.body
+          attachment.default_body
+          attachment.body_to_text
```
garethrees: Actually these don't do anything since the methods themselves don't call `apply_masks` – this happens at the callsite. Need to pass them through the masking pipeline explicitly to mirror the callsite (yuck!)

```
    attachment.apply_masks(attachment.default_body,
                           attachment.content_type,
                           redacted_attribute: :default_body)
```

app/models/info_request.rb
```diff
@@ -584,7 +584,7 @@ def from_name
   def safe_from_name
     return external_user_name if is_external?
 
-    apply_censor_rules_to_text(from_name)
+    apply_masks(from_name)
```
gbp: Need to be careful when from_name is `nil`. Calling `apply_masks` will raise NoMethodError

app/models/outgoing_message.rb
```diff
@@ -148,7 +148,7 @@ def from_name
   def safe_from_name
     return info_request.external_user_name if info_request.is_external?
 
-    info_request.apply_censor_rules_to_text(from_name)
+    apply_masks(from_name)
```
gbp: Need to be careful when from_name is nil. Calling apply_masks will raise NoMethodError

app/models/foi_attachment/maskable.rb
```diff
@@ -39,6 +35,28 @@ def mask_siblings
     end
   end
 
+  def apply_masks(text, content_type = 'text/plain', redacted_attribute:)
+    info_request.apply_masks(
+      text, content_type,
+      redactable: self, redacted_attribute: redacted_attribute
+    )
+  end
```
gbp:
```suggestion
  def apply_masks(text, content_type = 'text/plain', redacted_attribute:, redactable: self)
    info_request.apply_masks(
      text, content_type,
      redactable: redactable, redacted_attribute: redacted_attribute
    )
  end
```

app/models/incoming_message.rb
```diff
@@ -132,6 +133,13 @@ def update_request
     info_request.update_last_public_response_at
   end
 
+  def apply_masks(text, content_type = 'text/plain', redacted_attribute:, redactable: self)
+    info_request.apply_masks(
+      text, content_type,
+      redactable: redactable, redacted_attribute: redacted_attribute
+    )
+  end
+
```
gbp, line 142: Remove #apply_masks as it is included in Redactable

app/models/foi_attachment.rb
```diff
@@ -172,7 +173,7 @@ def redacted_filename
     return filename unless info_request
     return filename if locked? && !locking?
 
-    info_request.apply_censor_rules_to_text(filename)
+    apply_masks(filename, redacted_attribute: :filename)
```
gbp, line 176: Nb. This will attachment URLs

app/models/censor_rule/recordable.rb
```diff
@@ -0,0 +1,47 @@
...
+
+    attributes = {
+      redactable: redactable,
+      redacted_attribute: redacted_attribute
+    }
+
+    if before == after
+      redactions.where(attributes).delete_all
+    else
+      redactions.create_or_find_by!(attributes)
```
gbp, line 37: Could this might be more efficient?

```suggestion
    scope = redactions.where(attributes)

    if before == after
      scope.delete_all if scope.exists?
    else
      redactions.create_or_find_by!(attributes) unless scope.exists?
```

app/controllers/attachments_controller.rb
```diff
@@ -47,8 +47,6 @@ def show_as_html
       }
     )
 
-    html = @incoming_message.apply_masks(html, response.media_type)
-
```
gbp, line 51: This switches from binary to plain/text masking. Is that expected? Do we want to do that change here?

Previously it depended on `response.media_type` which I believe was always nil

lib/tasks/temp.rake
```diff
@@ -1,3 +1,5 @@
+include AlaveteliFeatures::Helpers
```
gbp, line 1: This leak into Object. Remove and change line 55 to suggestion

lib/tasks/temp.rake
```diff
@@ -41,6 +43,105 @@ namespace :temp do
...
+  desc 'Record redactions made to existing content by current CensorRules'
+  task record_censor_rule_redactions: :environment do
+    # CensorRule::Redaction rows are only written as a rule is applied, so
+    # content redacted before the tracking existed has nothing recorded
+    # against it. Expire each affected request to drop the cached (already
+    # redacted) content, then read each redactable attribute back so the
+    # masking pipeline runs again and records what it removes.
+    #
+    # Safe to re-run and can be picked up from a given request with START_AT.
+    unless feature_enabled?(:redaction_tracking)
```
gbp, line 55:
```suggestion
    unless AlaveteliFeatures.backend.enabled?(:redaction_tracking)
```

app/views/admin/censor_rules/_list.html.erb
```diff
@@ -33,6 +44,22 @@
             <% end %>
           <% end %>
 
+          <% if feature_enabled?(:redaction_tracking) && defined?(info_request) %>
+            <td>
+              <% count = censor_rule.redactions.for_request(info_request).count %>
```
gbp: N+1 here. Should move outside the `censor_rules.each` block.

Maybe as: `redaction_counts = CensorRule::Redaction.for_request(info_request).group(:censor_rule_id).count`

Could also extract out of the `feature_enabled?(:redaction_tracking) && defined?(info_request)` calls too

app/models/outgoing_message.rb
```diff
@@ -212,7 +215,10 @@ def body(options = {})
       info_request.try(:applicable_censor_rules) or []
     end
 
-    censor_rules.reduce(text) { |t, rule| rule.apply_to_text(t) }
+    # FIXME: We appear not to be applying text masks here
```
gbp: FIXME. Left on purpose?

app/models/incoming_message.rb
```diff
@@ -114,7 +113,9 @@ def redeliverable?
   #
   # Returns a String
   def safe_from_name
-    info_request.apply_censor_rules_to_text(from_name) if from_name
+    return unless from_name
+
+    apply_masks(from_name, redacted_attribute: :from_name)
```
gbp, line 118: Need to be careful when from_name is nil. Calling apply_masks will raise NoMethodError

app/models/info_request.rb
```diff
@@ -584,7 +585,7 @@ def from_name
   def safe_from_name
     return external_user_name if is_external?
 
-    apply_censor_rules_to_text(from_name)
+    apply_masks(from_name, redactable: self, redacted_attribute: :from_name)
```
gbp, line 588: Need to be careful when from_name is nil. Calling apply_masks will raise NoMethodError

gbp (changes requested):

app/models/censor_rule/recordable.rb
garethrees: In what way were you thinking? Wouldn't it mean two database statements rather than one in each case?

app/models/foi_attachment.rb
garethrees: Think there's a missing word here – "affect"? This should just track the redaction that was already happening.

app/models/incoming_message.rb
garethrees: Ah yes. Need to pull in https://github.com/mysociety/alaveteli/pull/9347/changes/6587a1fdc6260d9a26da5a82f966e817bef51950.

app/models/foi_attachment.rb
garethrees: Oh sorry I see this was more related to the change to `apply_masks`. Hmm will think about this

app/models/info_request.rb
garethrees: Need to pull in 6587a1f.

app/models/incoming_message.rb
garethrees: Did you mean a different concern? It's not defined in [`Redactable`](https://github.com/mysociety/alaveteli/blob/8823-censor-rule-application-tracking-v3/app/models/concerns/redactable.rb)

gbp: Oh Sorry I thought Maskable was a concern and not just scoped to FoiAttachment::Maskable. Maybe it should be?

app/models/censor_rule/recordable.rb
gbp: Every time there isn't a change we would do a delete_all, I think a single exists is going to be faster.

app/models/foi_attachment.rb
gbp: Yes - `s/will attachment/will change attachment/`

app/models/incoming_message.rb
garethrees: Yeah, would be good to unify these a bit, but might leave that for another day as there's some overlap with https://github.com/mysociety/alaveteli/pull/9347. I think once we get these PRs merged we can look at improving the general structure.

app/views/admin/censor_rules/_list.html.erb
garethrees: f6bf5d2

app/controllers/attachments_controller.rb
garethrees: Hmm, well:

* The HTML version of the attachment is derived from the already-masked body
* The masks are being applied to HTML content here, so `text/html` seemed like it made sense

I'd thought the only material difference would be that instead of redactions appearing as multiple `x` chars, we'd get the actual replacement string – however I now realise that binary masks don't actually apply the full set of masks. I think I'll update this to force binary masking with a note to review/fix later, as I _think_ this change should be okay

app/models/censor_rule/recordable.rb
garethrees: Updated to this approach

PR #9515, opened 2026-09-03
Improve new search performance again

gbp:
## Relevant issue(s)

#9304. Follows on from #9489. Part of #8813.

## What does this do?

Makes the new admin search fast again, and orders results by how well they match rather than by nothing at all.

Listings with sort buttons offer "Relevance" as one of the choices while the new engine is searching. The ones without buttons come back best first.

## Why was this needed?

#9489 swapped `DISTINCT` for matching on the IDs, which gave PostgreSQL a semi-join. With the page `LIMIT` on top it answered that by scanning the whole table and running the search again for every row it looked at. The bodies listing took about 30 seconds for "cardiff" on production.

The same change removed an ordering we didn't know we had. `DISTINCT` over every column sorts by every column, starting with `id`, so results came back oldest first and looked ranked. Nothing replaced it, so the score the search works out was thrown away and rows came back in whatever order the plan produced. Paging wasn't stable either, as a record could turn up on two pages or on none.

<hr>

[skip changelog]

### DIFF
--- a/app/models/search_document.rb
+++ b/app/models/search_document.rb
@@ -205,29 +205,32 @@ class SearchDocument < ApplicationRecord
       SearchDocument.where("sd_id IN (SELECT s.sd_id FROM (#{sql[:query]}) s)",
 sql[:values])
     else
-      # De-duplicate records that match through several translations or
-      # sections (the join yields one row per matching search_document),
-      # keeping the relation chainable, countable and paginatable.
+      # A record can match through several translations or sections, so
+      # group the documents to one row per record, scored by the best of
+      # them. The relation stays chainable, countable and paginatable.
       #
       # Join the ids rather than matching them with IN. A semi-join lets
       # PostgreSQL scan the whole table and re-run the search for every
       # row it looks at.
       matching_ids = SearchDocument.
-        select(:searchable_id).
+        select(:searchable_id, "max(search_results.score) AS score").
         where(searchable_type: model.to_s).
         joins(
           "JOIN search_results " \
           "ON search_results.sd_id = search_documents.sd_id"
         ).
-        distinct
+        group(:searchable_id)
+
+      record_id = "#{relation.quoted_table_name}." \
+                  "#{relation.quoted_primary_key}"
 
       relation.
         with(search_results: Arel.sql(sql[:query], **sql[:values])).
         joins(
           "JOIN (#{matching_ids.to_sql}) search_matches " \
-          "ON search_matches.searchable_id = " \
-          "#{relation.quoted_table_name}.#{relation.quoted_primary_key}"
-        )
+          "ON search_matches.searchable_id = #{record_id}"
+        ).
+        order(Arel.sql("search_matches.score DESC, #{record_id}"))
     end
   end
 end
--- a/app/models/search_document.rb
+++ b/app/models/search_document.rb
@@ -206,21 +206,28 @@ class SearchDocument < ApplicationRecord
 sql[:values])
     else
       # De-duplicate records that match through several translations or
-      # sections (the join yields one row per matching search_document).
-      # Matching on the ids keeps the relation chainable, countable and
-      # paginatable, and leaves PostgreSQL free to fetch just the rows
-      # that matched.
+      # sections (the join yields one row per matching search_document),
+      # keeping the relation chainable, countable and paginatable.
+      #
+      # Join the ids rather than matching them with IN. A semi-join lets
+      # PostgreSQL scan the whole table and re-run the search for every
+      # row it looks at.
       matching_ids = SearchDocument.
         select(:searchable_id).
         where(searchable_type: model.to_s).
         joins(
           "JOIN search_results " \
           "ON search_results.sd_id = search_documents.sd_id"
-        )
+        ).
+        distinct
 
       relation.
         with(search_results: Arel.sql(sql[:query], **sql[:values])).
-        where(id: matching_ids)
+        joins(
+          "JOIN (#{matching_ids.to_sql}) search_matches " \
+          "ON search_matches.searchable_id = " \
+          "#{relation.quoted_table_name}.#{relation.quoted_primary_key}"
+        )
     end
   end
 end
--- a/app/models/search_document.rb
+++ b/app/models/search_document.rb
@@ -75,6 +75,8 @@ class SearchDocument < ApplicationRecord
       search_queries << <<~SQL.chomp
           SELECT
               sd_id,
+              searchable_type,
+              searchable_id,
               rank() OVER (
                 ORDER BY ts_rank_cd(admin_content_tsv, websearch_to_tsquery(:language, unaccent(:query))) DESC
               ) AS rank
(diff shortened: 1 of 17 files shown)


laurentS:
Quick comparison of search speeds between this PR and #9516 using a copy of madada's data (using search terms that actually match something):

|search term   |    hits   | time (ms)  |  9516 (ms)  | notes         |
|--------------|-----------|------------|-------------|---------------|
| mairie       |  ~33_674  |  320-340   | 220-280     | limit: 50000  |
| cardiff      |  0        |  3-5       |  3-5        |               |
| paris        |  194      |  5.4-5.9   |  4.5-5.1    |               |
| bergerac     |  15       |  3-6       |  3-4.5      |               |
| saint        |  5403     |  45-100    |  50-120     |               |

All tests above with (except the one for `mairie` where I increased the limit to get all results):
```
PublicBody.search_scope(<search_term>, backend: :postgresql, admin_mode: true, limit: 10000)
```

Repeating the same search benefits from postgresql caching, so timings tend
to improve when repeating the search a few times. Changing search terms on every
call leads to search times that seem to stay in the ranges above.
There seems to be a slight advantage to the other version on bigger result sets, apparently because it does 1 less join, and postgres seems smart enough to be able to reuse the ordering on `search_results` directly in the output. But overall, both solutions are essentially the same in terms of speed.

Given that the test suite is failing in the other PR because of the tweaks I made in the ActiveRecords methods, I'm going to close that one, and focus here. If we face more performance issues as result sets grow, we can always revisit.

gbp:
@laurentS I agree let's use this branch as a basis for this work. I have actually been looking at your commits are bringing them over to here. Just running the fully test suite for a final check I haven't broken anything. I think using materialize makes complete sense and reckon I have bought that over without breaking anything.

I've cherry picked split up your commit to the changes are a bit more logical - will push shortly.

app/models/search_document.rb
```diff
@@ -205,22 +205,32 @@ def self.hybrid_search(query,
       SearchDocument.where("sd_id IN (SELECT s.sd_id FROM (#{sql[:query]}) s)",
```
laurentS: I think we need to apply the same fixes as below here:

```suggestion
      SearchDocument.
        with(search_results: Arel.sql(sql[:query], **sql[:values])).
        joins(
          "JOIN search_results " \
          "ON search_results.sd_id = search_documents.sd_id"
        ).
        order(Arel.sql("search_results.score DESC"))

```

app/models/search_document.rb
```diff
@@ -112,9 +116,11 @@ def self.hybrid_search_internal(
     end
 
     # all searches use the FTS ts_vectors
-    search_queries << <<~SQL.chomp
+    search_queries << <<~SQL.squish
```
laurentS: I actually quite like the `chomp` version, because it makes requests easier to copy/paste for debugging/explaining (the single long line breaks when pasting from a terminal into `psql`) and just to read as well.
Could we do something like:

```ruby
# config/initializers/alaveteli.rb (or another file)
String.class_eval do
  if Rails.env.development?
    def cleanup
      self.chomp
    end
  else
    def cleanup
      self.squish
    end
  end
end
```
and then apply `<<~SQL.cleanup` instead? (it looks like we don't actually need the first 3, just applying the squish on the ones below this comment seems to be enough)

lib/core_ext/active_record_materialized_cte.rb
```diff
@@ -0,0 +1,22 @@
+##
```
laurentS, line 1: this feels a lot more elegant than what I did ✨

spec/models/public_body_spec.rb
```diff
@@ -2134,7 +2133,6 @@ def set_default_attributes(public_body)
             :public_body,
             name: "Ministère de l'Intérieur"
           )
-          body.reindex
           expect(PublicBody.newsearch(
```
laurentS, line 2136: Should we switch to `search_scope` for the tests? It should be the same result, but it would make sense to test the code path we're actually using.

spec/models/search_document_spec.rb
```diff
@@ -107,6 +107,37 @@
...
+      results = User.newsearch('Otter', admin_mode: true)
+
+      expect(results.to_a).to eq([strong, weak])
+    end
+  end
+
+  context 'searching every model at once' do
+    it 'returns the documents best match first' do
+      FactoryBot.create(:user, name: 'Ferret Watcher')
+      strong = FactoryBot.create(:user, name: 'Ferret Ferret Ferret')
```
laurentS, line 133: Shouldn't this line be some other model, eg. `PublicBody` ?

laurentS (approved):

gbp:
Here are some benchmarks for queries which 1. aren't paginated 2. and those that are. Shows that without pagination it would be fast and we wouldn't had noticed the slowness:

| term | develop<br>`09cbd2ab23` | join fix<br>`d325927177` | HEAD<br>`9fb4f22a21` | HEAD no MATERIALIZED<br>`9fb4f22a21` |
|---|---|---|---|---|
| council | 49.2 / **7052.8** | 52.2 / 41.6 | 43.3 / 31.8 | 43.0 / 35.1 |
| cardiff | 4.2 / **1360.5** | 11.2 / 12.3 | 4.6 / 6.5 | 8.8 / 4.9 |
| health | 22.2 / **5205.6** | 29.8 / 19.5 | 21.9 / 9.7 | 20.4 / 10.2 |
| nhs | 20.5 / **8392.6** | 29.8 / 19.3 | 20.5 / 13.7 | 19.8 / 10.6 |
| school | 68.8 / **>60000** | 77.9 / 67.1 | 70.3 / 58.3 | 68.3 / 56.6 |
| department | 5.4 / **3698.0** | 12.1 / 12.7 | 5.1 / 4.2 | 5.4 / 4.2 |

- **develop** — parent of the join fix. `WHERE id IN (…)`, plain CTE.
- **join fix** — branch before @laurentS's commits. Joined derived table, plain CTE.
- **HEAD** — two materialized CTEs, no join back to `search_documents`.
- **HEAD no MATERIALIZED** — HEAD's exact SQL with ` AS MATERIALIZED ` replaced by ` AS `.

gbp:
Also shows `MATERIALIZED` can be slightly slower but it gives us insurance that the query is faster just in case the PostgreSQL planner decides to do something strange!

app/models/search_document.rb
gbp: Ah this is where our personal preference differ. I prefer the consistency with other queries

spec/models/public_body_spec.rb
gbp: Yes we should but lets drop `newsearch` in a different branch

spec/models/search_document_spec.rb
gbp: Yeah. We could although I think we already have tests for search multiple models. Using the same model allows us to be sure the ranking is the same.

app/models/search_document.rb
gbp: I'll drop this change for now

PR #9516, opened 2026-09-04
Speed up search query

laurentS:
## Relevant issue(s)
https://github.com/mysociety/alaveteli/issues/9304#issuecomment-5506792378
(I now see that @gbp proposed a fix for the same issue yesterday :disappointed: So this does more or less the same work as #9515 but in a different way.)

I will try to compare both approaches in more detail to see what is faster

## What does this do?

The change relies mainly on patching 2 internal rails methods based on an open PR so that we can `MATERIALIZE` the search CTE. This in turn prevents postgres from trying to optimise the query, which led to the search subquery being run several thousand times for a single search,
causing the slow search speeds observed.

This also removes one of the 2 joins we were using (by fetching some extra columns in the subqueries), which helps save a bit more execution time.

Ordering of results is also fixed.

On some local testing:
- searching for `cardiff` in public bodies was taking over a minute, now 12-15ms
- even searching for "mairie" in my dataset (which should return > 35k results) completes in ~600ms

This also works with other models, and even across multiple models (which will be handy for searching through requests and their children elements).

## Why was this needed?

#9489 fixed one slowness problem by adding a different one. This PR hopefully fixes both :crossed_fingers: 

## Implementation notes

This takes some of the code in https://github.com/rails/rails/pull/54322 which means fiddling a bit with rails' internals. I've gone with it under the assumption that the code is good (as several people seem to have reviewed it or even deployed it to prod, and I think I understand what it does :wink:), and the PR would eventually be merged upstream so we could drop the change locally.

## Notes to reviewer

<hr>

### DIFF
--- a/app/models/search_document.rb
+++ b/app/models/search_document.rb
@@ -75,6 +75,8 @@ class SearchDocument < ApplicationRecord
       search_queries << <<~SQL.chomp
           SELECT
               sd_id,
+              searchable_type,
+              searchable_id,
               rank() OVER (
                 ORDER BY ts_rank_cd(admin_content_tsv, websearch_to_tsquery(:language, unaccent(:query))) DESC
               ) AS rank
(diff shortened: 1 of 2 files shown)


laurentS:
Closing in favour of #9515

PR #9527, opened 2026-09-07
Add Japanese (ja) translation

katsushi2441:
Adds `locale/ja/app.po` (1,467 / 1,553 strings). Placeholders (`{{name}}`), HTML tags, entities and URLs were machine-checked to match each `msgid`; no fuzzy entries. Checked on the Docker dev environment with `AVAILABLE_LOCALES: 'ja en'` and `DEFAULT_LOCALE: 'ja'`. Terminology follows the Japanese FOI system (request=請求, public authority=行政機関, successful=開示, refused=不開示, partially successful=一部開示). The remaining ~86 strings are help texts with nested `<strong>` markup that I'd rather leave to a native reviewer than risk broken markup.

I understand Transifex is the preferred route for translations; if you'd rather I upload there, tell me and I'll do that. Screenshots: https://github.com/katsushi2441/alaveteli-jp/tree/jp/docs/screenshots

### DIFF
--- a/locale/ja/app.po
+++ b/locale/ja/app.po
@@ -1,0 +1,5992 @@
+# SOME DESCRIPTIVE TITLE.
+# Copyright (C) YEAR THE PACKAGE'S COPYRIGHT HOLDER
+# This file is distributed under the same license as the PACKAGE package.
+# FIRST AUTHOR <EMAIL@ADDRESS>, YEAR.
+#
+msgid ""
+msgstr ""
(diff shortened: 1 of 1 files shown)


garethrees:
Hi @katsushi2441, yes, please add strings via transifex. We have some guidance on this at https://alaveteli.org/docs/customising/translation/. It hasn't been updated for a while but should be roughly correct.

Also, do let us know if you launch an Alaveteli in Japan: https://alaveteli.org/community/

PR #9528, opened 2026-09-07
Improve the search query to save one join

laurentS:
## Relevant issue(s)

Adds a commit to #9515 

## What does this do?

This removes the need for an intermediate join using `search_documents` to get back `sd_id` which we already had from the previous query.

It also seems to save one sort in postgres, as the database seems smart enough to reuse the ordered materialized `search_results` and just fetch the extra columns it needs from the required model.

In terms of ranking, this version uses the sum of all scores instead of keeping the max for a given object. So if an attachment were to match on multiple pages (when we implement a `SearchDocument` for each page), we would assign it the sum of the score for each page, rather than just the highest score of any of its pages. When joining on `FoiAttachment` for example, we would end up with a single result for one attachment, pointing at the first page that matched, but a score representing the entire attachment. (this is a bit theoretical as we don't implement this yet, so difficult to test!)

Visual explain before/after show the 2 bits of work (highlighted in red) that are not needed anymore. The other parts of the query are almost exactly the same, except we return two extra columns in the CTE scan (sd_id and rank_sum).

### before
https://explain.dalibo.com/plan/a991b2adah93ef15
image

### after
https://explain.dalibo.com/plan/5b98ab117296cb00
image

## Implementation notes

This is a PR on top of the linked PR to make it easier to review separately.

## Notes to reviewer

<hr>

### DIFF
--- a/app/models/search_document.rb
+++ b/app/models/search_document.rb
@@ -133,15 +133,22 @@ class SearchDocument < ApplicationRecord
         LIMIT #{limit * limit_ratio}
       SQL
 
+    # when the subqueries return multiple search_documents for a single
+    # searchable, we sum their scores under the assumption that if the
+    # record matched multiple times, it is probably more relevant.
+    # Keeping min(sd_id) allows us to group on the searchable to prevent
+    # duplicates, while keeping a valid sd_id if we need to join on it
+    # in a later query, without having to go through the search_documents
+    # table one more time.
     sql = <<~SQL.squish
       SELECT
-        searches.sd_id,
         searches.searchable_type,
         searches.searchable_id,
+        min(searches.sd_id) as sd_id,
         sum(searches.rank) AS rank_sum,
         sum(rrf_score(searches.rank)) AS score
       FROM ((#{search_queries.join(") UNION ALL (")})) searches
-      GROUP BY searches.sd_id, searches.searchable_type, searches.searchable_id
+      GROUP BY searches.searchable_type, searches.searchable_id
       ORDER BY score DESC
       LIMIT #{limit}
     SQL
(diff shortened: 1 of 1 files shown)


app/models/search_document.rb
```diff
@@ -228,23 +235,14 @@ def self.hybrid_search(query,
...
         joins(
-          "JOIN search_records " \
-          "ON search_records.searchable_type = '#{model}' " \
-          "AND search_records.searchable_id = #{record_id}"
+          "JOIN search_results " \
+          "ON search_results.searchable_type = '#{model}' " \
+          "AND search_results.searchable_id = #{record_id}"
         ).
-        order(Arel.sql("search_records.score DESC, #{record_id}"))
+        order(Arel.sql("search_results.score DESC"))
```
gbp, line 245: I do think we need `record_id` here to ensure we have consistent ordering in case search result scores are equal to another result.

gbp (commented):
I don't think summing scores is the correct thing to do. Take for example searchable with more than one locale, or one (say for an attachment with content over multiple search documents) using sum over max would mean the searchable is more likely to ranked higher.

Also `min(sd_id)` picks the oldest search document and not necessarily the best match.

The speed improvement doesn't really seem worth it:

| term | HEAD<br>`9fb4f22a21` | + one CTE<br>`8a51633213` |
|---|---|---|
| council | 43.3 / 31.8 | 41.5 / 32.8 |
| cardiff | 4.6 / 6.5 | 3.7 / 3.2 |
| health | 21.9 / 9.7 | 19.6 / 8.7 |
| nhs | 20.5 / 13.7 | 19.4 / 8.3 |
| school | 70.3 / 58.3 | 68.1 / 57.9 |
| department | 5.1 / 4.2 | 4.7 / 3.4 |

app/models/search_document.rb
laurentS: Yeah, this is a good point. And in that case, the plan will include the final sort step, unless we push this second sort order down the subqueries.

laurentS:
> I don't think summing scores is the correct thing to do. Take for example searchable with more than one locale, or one (say for an attachment with content over multiple search documents) using sum over max would mean the searchable is more likely to ranked higher.
> 
> Also `min(sd_id)` picks the oldest search document and not necessarily the best match.

In my mind, the `SearchDocument` object is a search abstraction, we don't really show it to the user. But indeed, if we pick the `min(sd_id)` of multiple pages of a `FoiAttachment`, we would show the same attachment as if we picked `sd_id of the best matching page`, but if we end up showing an excerpt, then yeah, this falls apart.

As for sum vs max, I guess the question is do we want to prioritize a doc that has one very good page over another that has many ok pages. Thinking about it in abstraction made sense in my mind, but it's not so clear now...

I'll close this PR for now.

Issue #8810, opened 2025-08-15
Disentangle Xapian from alert generation


garethrees:
request thread / places where xapian is used where it would be better to just get the records from the database associations

garethrees:
Meant to document here that this is more about email alerts. The request thread itself is based off the events, pulled out through SQL. The email alerts are also based off the events, but they're pulled out via Xapian.

gbp:
🤖 I audited production before starting, to check the feature justifies the work and to find where Xapian is genuinely load-bearing. IP addresses and saved search text are withheld as they contain personal data.

**Headline: 93% of alert volume can move to the database without waiting for the PostgreSQL search work.**

- **200,928 tracks** across **63,248 users**; **126,375 (63%)** belong to accounts that would actually be emailed (`should_be_emailed?`)
- **5,412 users** received a digest in the last month
- **8,860 new tracks** in the last 12 months — declining gently from a ~2013–2021 peak, but alive
- **2.65M alert rows** in one month
- Feeds: **9,315 hits/day, ≤295 distinct IPs**, ~2.3% of all site GETs. Polling pattern (~4½ hourly, ~6 feeds per IP) looks like real RSS readers

### Five of six track types don't need search at all

`TrackThing#track_query` is a Xapian query string, but for five of the six types it's a pure structural filter that maps onto associations we already have:

| `track_type` | `track_query` | Equivalent |
|---|---|---|
| `request_updates` | `request:<url_title>` | `info_request.info_request_events` |
| `all_new_requests` | `variety:sent` | `InfoRequestEvent.sent_events` |
| `all_successful_requests` | `variety:response (status:successful OR …)` | `response_events.where(calculated_state: …)` |
| `public_body_updates` | `requested_from:<url_name>` | `public_body.info_requests` → events |
| `user_updates` | `requested_by:… OR commented_by:…` | tracked user's requests + comments |
| `search_query` | free text + filters | needs a backend — but see below |

Three don't even need the query parsed: `info_request_id`, `public_body_id` and `tracked_user_id` are already columns on `track_things`.

### Subscription count and alert volume run in opposite directions

| Type | Alerts/month | Share | Live tracks | Per track |
|---|---:|---:|---:|---:|
| `search_query` | 2,263,971 | **85.3%** | 3,908 | 579 |
| `public_body_updates` | 270,545 | 10.2% | 8,367 | 32 |
| `all_successful_requests` | 84,752 | 3.2% | 29 | 2,922 |
[...]

PR #9530, opened 2026-09-10
Disentangle Xapian from Track Things alerts

gbp:
## Relevant issue(s)

Most of #8810

## What does this do?

Adds a strategy per track type, so five of the six can find their events in the database instead of Xapian.

Alerts, feeds, the wall and the river all go through `TrackThing#matches`, which returns events and the words to highlight.

## Why was this needed?

Alerts blocks the Postgres search work. They have to keep working whichever backend is in place, so they need to stop depending on one.

Saved searches are the exception. They carry free text, so they still need a search index and keep using it.

## Implementation notes

The new `database_backed_alerts` flag picks the database strategies. It is a runtime flag, off by default, so it can be turned on and off from a console while the alert daemon is running.

<hr>

[skip changelog]

### DIFF
--- a/app/controllers/track_controller.rb
+++ b/app/controllers/track_controller.rb
@@ -171,7 +171,7 @@ class TrackController < ApplicationController
 
   def atom_feed_internal
     search_results = perform_track_search(@track_thing)
-    @events = search_results.results.map { |result| result[:model] }
+    @events = search_results.events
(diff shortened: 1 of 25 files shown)


gbp:
Merged e8f9d60bb798f13945b8642de2fff3dad666a27b

